PR #82, opened 2016-05-27
Ft/common req auth

jmunoznaranjo:
This PR copies S3's auth library into Arsenal so Vault can access it in order to perform query signature verification.

- Core code, tests and helper files are copied.
- package.json deps are updated
- An api is exposed so Vault can overwrite the core authentication code with its own. However the old "vault.js" is left as default.

Related:
Vault: https://github.com/scality/Vault/pull/359
vaultclient: https://github.com/scality/vaultclient/pull/72


lib/auth/v2/getCanonicalizedAmzHeaders.js
```diff
@@ -0,0 +1,44 @@
+'use strict'; // eslint-disable-line strict
+
+function getCanonicalizedAmzHeaders(headers) {
+    /*
+    Iterate through headers and pull any headers that are x-amz headers.
+    Need to include 'x-amz-date' here even though AWS docs
+    ambiguous on this.
+    */
+    const amzHeaders = Object.keys(headers)
+        .filter(val => val.substr(0, 6) === 'x-amz-')
```
jmunoznaranjo: These lines are not affected by this PR, so I'd rather preserve the original code and make the minimal changes possible in order to avoid possible side effects. Does that sound good? ;)

lib/auth/v2/getCanonicalizedAmzHeaders.js
```diff
@@ -0,0 +1,44 @@
...
+    */
+
+    // If there are no amz headers, just return an empty string
+    if (amzHeaders.length === 0) {
+        return '';
+    }
+
+
+    // Sort the amz headers by key (first item in tuple)
+    amzHeaders.sort((a, b) => {
```
jmunoznaranjo: Same as my previous comment, even if an enhancement was possible that's not the purpose of this PR ;)

lib/auth/v2/headerAuthCheck.js
```diff
@@ -0,0 +1,74 @@
...
+    log.trace('signature from request', { signatureFromRequest });
+    const stringToSign = constructStringToSign(request, log);
+    log.trace('constructed string to sign', { stringToSign });
+    const algo = algoCheck(signatureFromRequest.length);
+    log.trace('algo for calculating signature', { algo });
+    if (algo === undefined) {
+        return callback(errors.InvalidArgument);
+    }
+    return vault.authenticateV2Request(accessKey,
+        signatureFromRequest, stringToSign, algo, log, (err, authInfo) => {
```
jmunoznaranjo: Not for this PR ;)

lib/auth/v2/queryAuthCheck.js
```diff
@@ -0,0 +1,68 @@
...
+    const stringToSign = constructStringToSign(request, log);
+    log.trace('constructed string to sign', { stringToSign });
+    const algo = algoCheck(signatureFromRequest.length);
+    log.trace('algo for calculating signature', { algo });
+    if (algo === undefined) {
+        return callback(errors.InvalidArgument);
+    }
+    return vault.authenticateV2Request(accessKey,
+        signatureFromRequest, stringToSign, algo, log, (err, authInfo) => {
+            if (err) {
```
jmunoznaranjo: Same :stuck_out_tongue_closed_eyes:

lib/auth/in_memory/vault.json
```diff
@@ -0,0 +1,79 @@
+{
```
jmunoznaranjo: I understand your point. My reason for moving the whole backend is that  we need the in_memory backend and vault.js for testing the whole thing.

I've set vault.js as the default authentication handler, so S3 can just import Arsenal and use it for signature generation. Looks transparent and easy to me.

lib/auth/auth.js.Design.md
```diff
@@ -0,0 +1,95 @@
...
+```js
+router.get("/getObjectRoute", function(request, response){
+    checkAuth(request, function(err, accessKey){
+        if(err){
+            // Return an error response
+        }
+        // Continue on with request
+    })
+})
+```
```
jmunoznaranjo: Fine for me, I'll remove it here. We can recover it from the commit history, and it's still in S3.

lib/auth/auth.js
```diff
@@ -0,0 +1,56 @@
...
+    // Check whether signature is in header
+    if (authHeader) {
+        log.trace('authorization header', { authHeader });
+        // TODO: Check for security token header to
+        // handle temporary security credentials
+        if (authHeader.startsWith('AWS ')) {
+            log.trace('authenticating request with auth v2 using headers');
+            authV2.headerAuthCheck.check(request, log, cb);
+        } else if (authHeader.startsWith('AWS4')) {
+            log.debug('authenticating request with Auth V4 using headers');
```
AntoninCoulibaly: `'authenticating request with auth V4 using headers'`
to be compliant with the other one

lib/auth/auth.js
```diff
@@ -0,0 +1,56 @@
...
+        } else {
+            log.warn('missing authorization security header');
+            return cb(errors.MissingSecurityHeader);
+        }
+    } else if (request.query.Signature) {
+        // Check whether signature is in query string
+        log.trace('authenticating request with auth v2 using query string');
+        authV2.queryAuthCheck.check(request, log, cb);
+    } else if (request.query['X-Amz-Algorithm']) {
+        log.debug('authenticating request with Auth v4 using query string');
```
AntoninCoulibaly: same as above

lib/auth/auth.js
jmunoznaranjo: As in previous comments, this is part of the original code from S3. I don't want to make this kind of changes in this PR with the aim of minimizing the diff. Let's leave it as it is for now ;)

lib/auth/auth.js
jmunoznaranjo: same ;)

jmunoznaranjo:
history squashed, please review

lib/constants.js
```diff
@@ -0,0 +1,54 @@
+'use strict'; // eslint-disable-line strict
+
+module.exports = {
```
LaurenSpiegel: Do you need all of these?  I would just keep what you need.

LaurenSpiegel:
Other than slimming down on constants.js, 👍

lib/constants.js
jmunoznaranjo: done ;)

lib/auth/auth.js
```diff
@@ -53,4 +58,49 @@ auth.doAuth = (request, log, cb) => {
     return undefined;
 };
 
+auth.generateV4Headers =
+    (request, data, accessKey, secretKeyValue) => {
+        Object.assign(request, { headers: {} });
+        const amzDate = convertUTCtoISO8601(Date.now());
```
AntoninCoulibaly: why don't you call it timestamp directly ?

lib/auth/v4/validateInputs.js
```diff
@@ -39,7 +39,7 @@ function validateCredentials(credentials, timestamp, log) {
             { scopeDate, timestampDate });
         return false;
     }
-    if (service !== 's3') {
+    if (service !== 's3' && service !== 'iam') {
         log.warn('service in credentials is not s3', { service });
```
AntoninCoulibaly: 'service in credentials is not s3 or aws'

AntoninCoulibaly:
just 2 little comments :+1:

ThibaultRiviere:
:+1:

PR #87, opened 2016-06-02
Move extensions to Arsenal

vrancurel:
S3 metadata file backend uses the same extensions than metadata to process delimiter and MPU queries.

At some point in time, Metadata would need to load extensions from here with an associative array (no need to dynamically load those files).

### DIFF
--- a/lib/extension/list.extension.js
+++ b/lib/extension/list.extension.js
@@ -1,0 +1,40 @@
+'use strict'; // eslint-disable-line strict
+/**
+ *  Class of an extension doing the simple listing
+ */
+class List {
+    /**
+     *  Constructor
+     *  Set the logger and the res
+     *  @param {Object} parameters - The parameters you sent to DBD
+     *  @param {RequestLogger} logger - The logger of the request
+     *  @return {undefined}
+     */
+    constructor(parameters, logger) {
+        this.logger = logger;
+        this.res = [];
+    }
+
+    /**
+     *  Function apply on each element
+     *  Just add it to the array
+     *  @param {Object} elem - The data from the database
+     *  @return {Boolean} - True = continue the stream
+     */
+    filter(elem) {
+        this.res.push(elem);
+        return true;
+    }
+
+    /**
+     *  Function returning the result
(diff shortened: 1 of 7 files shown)


vrancurel:
I want to keep the extensions files 'AS IS' because they are moved from Metadata, until Metadata uses them, just in case we need to manually import those files again.

rahulreddy:
👍

RachedBenMustapha:
As long as it's just a copy we don't need to post-critique the code 👍

lib/extension/list.extension.js
```diff
@@ -0,0 +1,40 @@
...
+     *  @return {Boolean} - True = continue the stream
+     */
+    filter(elem) {
+        this.res.push(elem);
+        return true;
+    }
+
+    /**
+     *  Function returning the result
+     *  @return {Array} - The listed elements
```
MathieuCassagne: We do not know, you can store strings, numbers, objects .., it's an array of elem whatever what elem is.

lib/extension/list.extension.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+class List {
+    /**
+     *  Constructor
+     *  Set the logger and the res
+     *  @param {Object} parameters - The parameters you sent to DBD
+     *  @param {RequestLogger} logger - The logger of the request
+     *  @return {undefined}
+     */
+    constructor(parameters, logger) {
```
MathieuCassagne: Yep, the extension have to follow a constructor that takes parameters and logger.

lib/extension/list.extension.js
```diff
@@ -0,0 +1,40 @@
...
+        this.res = [];
+    }
+
+    /**
+     *  Function apply on each element
+     *  Just add it to the array
+     *  @param {Object} elem - The data from the database
+     *  @return {Boolean} - True = continue the stream
+     */
+    filter(elem) {
```
MathieuCassagne: Yes, but the filter function is the function called on each element given by the readStream on the database it's the same function for ALL extensions, it's a convention.

PR #88, opened 2016-06-08
[Clustering](ft): Add clustering

alexandre-merle:
Add a clustering, simple to use, with an internal
watchdog

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,3 +46,24 @@ console.log(errors.AccessDenied);
 //    AccessDenied: true }
 
 ```
+
+## Clustering
+
+The clustering class can be used to set up a cluster of worker. The class will
+create at least 1 worker, will log any worker event (started, exited).
+The class has also a watchdog who will restart workers until the stop() method
+was called.
+
+### Usage
+
+```
+import { Clustering } from 'arsenal';
+
+const clusters = new Clustering(clusterSize, logger);
+clusters.start(i => {
+    // Put here the logic of every worker.
+    // 'i' is the index of the worker
+});
+```
+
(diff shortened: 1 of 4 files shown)


alexandre-merle:
Updated :) @MathieuCassagne Did you see any feature you will need in DBD ? @LaurenSpiegel same question :)

README.md
```diff
@@ -46,3 +46,24 @@ console.log(errors.AccessDenied);
 //    AccessDenied: true }
 
 ```
+
+## Clustering
+
+The clustering class can be used to set up a cluster of worker. The class will
```
LaurenSpiegel: of workers

README.md
```diff
@@ -46,3 +46,24 @@ console.log(errors.AccessDenied);
...
+create at least 1 worker, will log any worker event (started, exited).
+The class has also a watchdog who will restart workers until the stop() method
+was called.
```
LaurenSpiegel: is called

lib/Clustering.js
```diff
@@ -0,0 +1,149 @@
...
+     * Constructor
+     *
+     * @param {number} size Cluster size
+     * @param {Logger} logger Logger object
+     * @return {Clustering} itself
+     */
+    constructor(size, logger) {
+        this._size = size;
+        if (size < 1) {
+            throw new Error('Cluster size must superior or equal to 1');
```
LaurenSpiegel: must be greater than or equal to

LaurenSpiegel:
We're not listening to the on "disconnect" event here.  I need some convincing that this is as safe as the setInterval approach in S3's current server.

PR #97, opened 2016-06-13
Fix/listing limit

MathieuCassagne:
#### Extension(fix): Default value:
- Add 'checkLlimit' into tools.js to check the maxKeys of all extensions
  in one function.
- The default value is now set in the extension itself.
- ListExtension: Add a limit on max elementes returned.
- ListExtension, DelimiterExtension, ListMPUExtension: Check if reached
  maxKeys at start of fiter function to deal with the case of
  maxKeys <= 0

#### Extension(tests): Add unit tests:
- Added unit test for DelimiterExtension, ListExtension and checkLimit
  function.

#### Add missing extension to index.js

### DIFF
--- a/lib/extension/listMPU.extension.js
+++ b/lib/extension/listMPU.extension.js
@@ -1,9 +1,13 @@
 'use strict'; // eslint-disable-line strict
 
+const checkLimit = require('./tools').checkLimit;
+const DEFAULT_MAX_KEYS = 1000;
+
 function numberDefault(num, defaultNum) {
     const parsedNum = Number.parseInt(num, 10);
     return Number.isNaN(parsedNum) ? defaultNum : parsedNum;
 }
+
 /**
  *  Class for the ListMultipartUploads extension
  */
@@ -24,7 +28,7 @@ class ListMultipartUploads {
         this.prefixLength = 0;
         this.queryPrefixLength = numberDefault(params.queryPrefixLength, 0);
         this.keys = 0;
-        this.maxKeys = numberDefault(params.maxKeys, 1000);
+        this.maxKeys = checkLimit(params.maxKeys, DEFAULT_MAX_KEYS);
         this.delimiter = params.delimiter;
         this.splitter = params.splitter;
         this.logger = logger;
@@ -80,6 +84,12 @@ class ListMultipartUploads {
      *  @return {Boolean} - True: Continue, False: Stop
      */
     filter(obj) {
+        // Check first in case of maxkeys = 0
+        if (this.keys >= this.maxKeys) {
+            // In cases of maxKeys <= 0 => IsTruncated = false
+            this.IsTruncated = this.maxKeys > 0;
+            return false;
+        }
         const key = obj.key;
         const value = obj.value;
         if (this.delimiter) {
@@ -96,10 +106,6 @@ class ListMultipartUploads {
             }
         } else {
             this.addUpload(value);
-        }
-        if (this.keys === this.maxKeys) {
-            this.IsTruncated = true;
-            return false;
         }
         return true;
     }
--- a/tests/unit/extension/delimiter.js
+++ b/tests/unit/extension/delimiter.js
@@ -1,0 +1,175 @@
+'use strict'; // eslint-disable-line strict
+
+const assert = require('assert');
+const Delimiter =
+    require('../../../lib/extension/delimiter.extension').Delimiter;
+const Werelogs = require('werelogs');
+const logger = new Werelogs('listTest');
+const performListing = require('../../utils/performListing');
+
+class Test {
+    constructor(name, input, output, filter) {
+        this.name = name;
+        this.input = input;
+        this.output = output;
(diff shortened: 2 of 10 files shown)


tests/unit/extension/delimiter.js
```diff
@@ -0,0 +1,175 @@
...
+        this.output = output;
+        this.filter = filter || this._defaultFilter;
+    }
+
+    _defaultFilter() {
+        return true;
+    }
+}
+
+describe('Delimiter extenstion', () => {
```
DavidPineauScality: extens~~t~~ion

tests/unit/extension/tools.js
```diff
@@ -0,0 +1,24 @@
...
+
+describe('checkLimit function', () => {
+    const tests = [
+        { input: [ 1000, 2500 ], output: 1000 },
+        { input: [ 5000, 2500 ], output: 2500 },
+        { input: [ 'lala', 2500 ], output: 2500 },
+        { input: [ 15000, 0 ], output: 15000 },
+        { input: [ 'lala', 0 ], output: 0 },
+        { input: [ 2000, undefined ], output: 2000 },
+        { input: [ 'lala', undefined ], output: undefined },
```
DavidPineauScality: what about cases such as 0,0 ?
I see that you test with an undefined or 0-defined limit, but as it is known as an invalid value for our implementation, shouldn't we ensure that this cant be done, or at least that the result yielded is valid for our system ?

tests/unit/extension/listMpuExtension.js
```diff
@@ -2,10 +2,11 @@
 
 const assert = require('assert');
 const ListMultipartUploads =
-    require('../../lib/extension/listMPU.extension').
+    require('../../../lib/extension/listMPU.extension').
     ListMultipartUploads;
 const werelogs = require('werelogs');
 const logger = new werelogs('listMpuTest');
+const performListing = require('./performListing');
```
DavidPineauScality: this file might be more fit for a path such as `tests/utils/`

tests/unit/extension/list.js
```diff
@@ -0,0 +1,41 @@
+'use strict'; // eslint-disable-line strict
+
+const assert = require('assert');
+const List = require('../../../lib/extension/list.extension').List;
+const Werelogs = require('werelogs');
+const logger = new Werelogs('listTest');
+const performListing = require('./performListing');
```
DavidPineauScality: in tests/utils ? :)

tests/unit/extension/delimiter.js
```diff
@@ -0,0 +1,175 @@
+'use strict'; // eslint-disable-line strict
+
+const assert = require('assert');
+const Delimiter =
+    require('../../../lib/extension/delimiter.extension').Delimiter;
+const Werelogs = require('werelogs');
+const logger = new Werelogs('listTest');
+const performListing = require('./performListing');
```
DavidPineauScality: `tests/utils` too

lib/extension/listMPU.extension.js
```diff
@@ -1,9 +1,14 @@
 'use strict'; // eslint-disable-line strict
 
+const checkLimit = require('./tools').checkLimit;
+const DEFAULT_MAX_KEYS = 1000;
```
DavidPineauScality: I see that both Delimiter and ListMultipartUpload define a default of 1000, while the basic listing defines a default of 10000.

Shouldn't we ensure that the default is the same for all ? Is there a risk doing so ?

DavidPineauScality:
A few comments, nothing really major, but still a few important cleanups (at least to me).
Mostly good :)
Thanks for the added tests, by the way !

tests/unit/extension/tools.js
MathieuCassagne: Nice catch :) The test is added :)

tests/unit/extension/listMpuExtension.js
MathieuCassagne: Yep, I'll move it :)

lib/extension/listMPU.extension.js
MathieuCassagne: The limit of 1000 comes from the AWS documentation, it's an hard limit.
But, the simple list extension is not intended to be used by S3 to do an object listing, bucketClient enforce the use of delimiter or listMPU extension, the limit of 10000 is here to avoid to use more RAM than the process can use.

DavidPineauScality:
Good for me now :+1:

rahulreddy:
Do we have to do end-to-end as well as run functional tests on S3 with this PR since the file backend in S3 uses this extension?

DavidPineauScality:
well end-to-end is not using the file backend, so I do not see that as a necessity. But running S3's functional tests again would be useful with this PR.

rahulreddy:
MetaData is using this extension as well, I would run both S3 and MetaData's functional tests to make sure.
👍 I am ok with the code

rahulreddy:
MetaData and S3 tests are passing.

PR #98, opened 2016-06-22
FT: Use data && make check synchronous

alexandre-merle:
- Replace request.query by an object data to
  pass to the function
- Allow using another service than s3 in the
  signingKey
- Split the check and the backend call for the
  authentification

### DIFF
--- a/lib/auth/auth.js
+++ b/lib/auth/auth.js
@@ -30,23 +30,23 @@ auth.check = (request, log, awsService, data) => {
             log.trace('authenticating request with auth v2 using headers');
             return authV2.headerAuthCheck.check(request, log, data);
         } else if (authHeader.startsWith('AWS4')) {
-            log.debug('authenticating request with Auth V4 using headers');
+            log.trace('authenticating request with Auth V4 using headers');
             return authV4.headerAuthCheck.check(request, log, awsService);
         }
-        log.warn('missing authorization security header');
+        log.debug('missing authorization security header');
         return { err: errors.MissingSecurityHeader };
     } else if (data.Signature) {
         // Check whether signature is in query string
         log.trace('authenticating request with auth v2 using query string');
         return authV2.queryAuthCheck.check(request, log, data);
     } else if (data['X-Amz-Algorithm']) {
-        log.debug('authenticating request with Auth v4 using query string');
+        log.trace('authenticating request with Auth v4 using query string');
         return authV4.queryAuthCheck.check(request, log, data);
     }
     // If no auth information is provided in request, then
     // user is part of 'All Users Group' so send back this
     // group as the canonicalID
-    log.trace('No authentication provided. User identified as public');
+    log.debug('No authentication provided. User identified as public');
     const authInfo = new AuthInfo({ canonicalID: constants.publicId });
     return { err: null, data: authInfo };
 };
--- a/lib/auth/auth.js
+++ b/lib/auth/auth.js
@@ -10,18 +10,15 @@ const constructStringToSignV4 = require('./v4/constructStringToSign');
 const convertUTCtoISO8601 = require('./v4/timeUtils').convertUTCtoISO8601;
 const crypto = require('crypto');
 const vaultUtilities = require('./in_memory/vaultUtilities');
-
+let vault = null;
 const auth = {};
 
 auth.setAuthHandler = handler => {
-    authV2.headerAuthCheck.setAuthHandler(handler);
-    authV2.queryAuthCheck.setAuthHandler(handler);
-    authV4.headerAuthCheck.setAuthHandler(handler);
-    authV4.queryAuthCheck.setAuthHandler(handler);
+    vault = handler;
     return auth;
 };
 
-auth.doAuth = (request, log, cb, awsService, data) => {
+auth.check = (request, log, awsService, data) => {
     log.debug('running auth checks', { method: 'auth' });
     const authHeader = request.headers.authorization;
     // Check whether signature is in header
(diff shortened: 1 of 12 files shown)


DavidPineauScality:
There are some parts of this PR that I'm going to remove very soon.
Would it be possible to rework it so that it only introduces the useful bits ? I'll comment on the unnecessary bits.

DavidPineauScality:
Note that the plan to rework the API is described in #102

alexandre-merle:
Updated :) with an synchronous api :)

alexandre-merle:
end to end running http://ci.ironmann.io/gh/scality/Integration/1835

alexandre-merle:
see http://ci.ironmann.io/gh/scality/Integration/1837

lib/auth/auth.js
```diff
@@ -10,19 +10,15 @@ const constructStringToSignV4 = require('./v4/constructStringToSign');
 const convertUTCtoISO8601 = require('./v4/timeUtils').convertUTCtoISO8601;
 const crypto = require('crypto');
 const vaultUtilities = require('./in_memory/vaultUtilities');
-
-
+let vault = require('./vault');
```
LaurenSpiegel: Why require vault -- let's just set handler.

lib/auth/auth.js
```diff
@@ -32,41 +28,64 @@ auth.doAuth = (request, log, cb, awsService) => {
...
         } else if (authHeader.startsWith('AWS4')) {
             log.debug('authenticating request with Auth V4 using headers');
-            authV4.headerAuthCheck.check(request, log, cb, awsService);
-        } else {
-            log.warn('missing authorization security header');
-            return cb(errors.MissingSecurityHeader);
+            return authV4.headerAuthCheck.check(request, log, awsService);
         }
-    } else if (request.query.Signature) {
+        log.warn('missing authorization security header');
```
rahulreddy: This should be trace

lib/auth/auth.js
```diff
@@ -32,41 +28,64 @@ auth.doAuth = (request, log, cb, awsService) => {
...
+                return cb(null, authInfo);
+            });
+    } else if (res.version === 4) {
+        res.data.log = log;
+        return vault.authenticateV4Request(res.data, cb, awsService);
+    } else if (res.data instanceof AuthInfo) {
+        return cb(null, res.data);
     }
-    return undefined;
+    log.error('Cannot found authentification method', {
```
rahulreddy: This could be re-worded to `authentication method not found`

lib/auth/auth.js
```diff
@@ -32,41 +28,64 @@ auth.doAuth = (request, log, cb, awsService) => {
...
+    }
+    // If no auth information is provided in request, then
+    // user is part of 'All Users Group' so send back this
+    // group as the canonicalID
+    log.trace('No authentication provided. User identified as public');
+    const authInfo = new AuthInfo({ canonicalID: constants.publicId });
+    return { err: null, data: authInfo };
+};
+
+auth.doAuth = (request, log, cb, awsService, data) => {
```
LaurenSpiegel: This function is temporary, I assume?

lib/auth/auth.js
```diff
@@ -32,41 +28,64 @@ auth.doAuth = (request, log, cb, awsService) => {
...
-    (request, data, accessKey, secretKeyValue) => {
+    (request, data, accessKey, secretKeyValue, awsService) => {
         Object.assign(request, { headers: {} });
         const amzDate = convertUTCtoISO8601(Date.now());
         // get date without time
         const scopeDate = amzDate.slice(0, amzDate.indexOf('T'));
         const signedHeaders = 'host;x-amz-date;x-amz-content-sha256';
         const region = 'us-east-1';
-        const service = 'iam';
+        const service = awsService || 'iam';
```
LaurenSpiegel: I think this is against our style rules.

alexandre-merle: It's not :) and it's more efficient for v8 in that way :) but i can change it :)

tests/unit/auth/v4/headerAuthCheck.js
```diff
@@ -200,39 +187,19 @@ describe('v4 headerAuthCheck', () => {
...
-            done();
-        });
+        const res = headerAuthCheck(alteredRequest, log);
+        clock.uninstall();
+        assert.ifError(res.err);
+        done();
     });
 
-    it('should successfully authenticate', done => {
+    it('should successfully return information', done => {
```
LaurenSpiegel: Do you still need the done?
It seems you are just testing that the auth version is returned.  Is that all we should be testing here?  If so, let's change the should statement to -- it should return the auth version type

tests/unit/auth/v4/queryAuthCheck.js
```diff
@@ -171,23 +154,20 @@ describe('v4 queryAuthCheck', () => {
...
-            done();
-        });
+        const res = queryAuthCheck(alteredRequest, log, alteredRequest.query);
+        clock.uninstall();
+        assert.deepStrictEqual(res.err, errors.InvalidArgument);
+        done();
     });
 
-    it('should successfully authenticate', done => {
+    it('should successfully return informations', done => {
```
LaurenSpiegel: Same comment as above.

tests/unit/auth/v4/headerAuthCheck.js
alexandre-merle: Yes, in fact checking there is no error is enought, but we still check the good work of this function, not really relate to the version

lib/auth/auth.js
alexandre-merle: yes :)

alexandre-merle:
Updated :)

rahulreddy:
👍

alexandre-merle:
end to end running http://ci.ironmann.io/gh/scality/Integration/1842

LaurenSpiegel:
👍

alexandre-merle:
end to end passed, dropme removed :)

PR #104, opened 2016-06-24
Forward-port of rel/1.1 into master

DavidPineauScality:


DavidPineauScality:
Merging on behalf of @rahulreddy since there was no conflict, and the diff is mostly a fix that was already tested.

PR #105, opened 2016-06-28
Fix: Compute canonicalRequest for post request

alexandre-merle:
- Aws cli don't use payload to iam request, so the payload is compute with an empty string
- POST request compute the body into the payload even is the payload is missing

Should be merge with https://github.com/scality/Vault/pull/392 and https://github.com/scality/vaultclient/pull/83

### DIFF
--- a/lib/auth/auth.js
+++ b/lib/auth/auth.js
@@ -31,7 +31,7 @@ auth.check = (request, log, awsService, data) => {
             return authV2.headerAuthCheck.check(request, log, data);
         } else if (authHeader.startsWith('AWS4')) {
             log.trace('authenticating request with Auth V4 using headers');
-            return authV4.headerAuthCheck.check(request, log, awsService);
+            return authV4.headerAuthCheck.check(request, log, data, awsService);
         }
         log.debug('missing authorization security header');
         return { err: errors.MissingSecurityHeader };
@@ -89,15 +89,26 @@ auth.generateV4Headers =
         const credentialScope =
             `${scopeDate}/${region}/${service}/aws4_request`;
         const timestamp = amzDate;
-        const query = request.query || {};
+        const query = request.query || data;
         const algorithm = 'AWS4-HMAC-SHA256';
 
-        const payload = !!data ? data : '';
+        let payload = '';
+        if (request.method === 'POST') {
+            const formattedData = [];
+            Object.keys(data).forEach(key => {
+                const formattedKey = encodeURIComponent(key);
+                const formattedValue = encodeURIComponent(data[key]);
+                if (formattedData.length > 0) {
+                    formattedData.push('&');
+                }
+                formattedData.push(`${formattedKey}=${formattedValue}`);
+            });
+            payload = formattedData.join('');
+        }
         const payloadChecksum = crypto.createHash('sha256').update(payload)
                                       .digest('hex');
         const path = request.path;
         Object.assign(request, { path: '/' });
-
         request.setHeader('host', request._headers.host);
         request.setHeader('x-amz-date', amzDate);
         request.setHeader('x-amz-content-sha256', payloadChecksum);
--- a/lib/auth/auth.js
+++ b/lib/auth/auth.js
@@ -1,7 +1,7 @@
 'use strict'; // eslint-disable-line strict
 
 const errors = require('../errors');
-
+const queryString = require('querystring');
 const AuthInfo = require('./AuthInfo');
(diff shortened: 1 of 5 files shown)


lib/auth/v4/createCanonicalRequest.js
```diff
@@ -16,13 +17,35 @@ function createCanonicalRequest(params) {
...
+                .digest('hex').toLowerCase();
+        }
+    }
 
     const canonicalURI = !!pResource ? awsURIencode(pResource, false) : '/';
 
     // canonical query string
     let canonicalQueryStr = '';
-    if (pQuery) {
+    if (pQuery && !(service === 'iam' && pHttpVerb === 'POST')) {
```
MathieuCassagne: You do not like ```||``` ? :D

jmunoznaranjo:
:+1:

lib/auth/auth.js
```diff
@@ -89,15 +89,26 @@ auth.generateV4Headers =
...
         const timestamp = amzDate;
-        const query = request.query || {};
+        const query = request.query || data;
         const algorithm = 'AWS4-HMAC-SHA256';
 
-        const payload = !!data ? data : '';
+        let payload = '';
+        if (request.method === 'POST') {
+            const formattedData = [];
+            Object.keys(data).forEach(key => {
```
rahulreddy: You can use node's `querystring` module to generate the payload string.

lib/auth/v4/createCanonicalRequest.js
```diff
@@ -16,13 +17,35 @@ function createCanonicalRequest(params) {
...
+
+    let payloadChecksum = params.payloadChecksum;
+
+    if (!payloadChecksum) {
+        if (pHttpVerb === 'GET') {
+            payloadChecksum = crypto.createHash('sha256').update('')
+                .digest('hex').toLowerCase();
+        } else if (pHttpVerb === 'POST') {
+            const formattedData = [];
+            Object.keys(pQuery).forEach(key => {
```
rahulreddy: Same. Please use node's `QueryString` module.

lib/auth/auth.js
```diff
@@ -31,7 +31,7 @@ auth.check = (request, log, awsService, data) => {
             return authV2.headerAuthCheck.check(request, log, data);
         } else if (authHeader.startsWith('AWS4')) {
             log.trace('authenticating request with Auth V4 using headers');
-            return authV4.headerAuthCheck.check(request, log, awsService);
+            return authV4.headerAuthCheck.check(request, log, data, awsService);
```
LaurenSpiegel: Will you please add a jsdoc for auth.check above?  It will make it clearer what the difference is between request and data.

lib/auth/auth.js
```diff
@@ -89,15 +89,26 @@ auth.generateV4Headers =
         const credentialScope =
             `${scopeDate}/${region}/${service}/aws4_request`;
         const timestamp = amzDate;
-        const query = request.query || {};
+        const query = request.query || data;
```
LaurenSpiegel: Is it possible to have a request.query on the IAM requests?  Would it be clearer/safer to say that if the service is 'iam', use data as query?

lib/auth/v4/createCanonicalRequest.js
```diff
@@ -16,13 +17,35 @@ function createCanonicalRequest(params) {
...
     const pHeaders = params.pHeaders;
     const pSignedHeaders = params.pSignedHeaders;
-    const payloadChecksum = params.payloadChecksum;
+    const service = params.service;
+
+    let payloadChecksum = params.payloadChecksum;
+
+    if (!payloadChecksum) {
+        if (pHttpVerb === 'GET') {
+            payloadChecksum = crypto.createHash('sha256').update('')
```
LaurenSpiegel: Can't you just set the hash for an empty string without calculating it each time?

lib/auth/v4/headerAuthCheck.js
```diff
@@ -14,9 +14,11 @@ const headerAuthCheck = {};
  * V4 header auth check
  * @param {object} request - HTTP request object
  * @param {object} log - logging object
+ * @param {object} data - Data of request
```
LaurenSpiegel: Please be more specific here so we know how this differs from the request parameter above.

alexandre-merle:
Updated :)

LaurenSpiegel:
👍

rahulreddy:
👍

PR #106, opened 2016-06-30
FT: User Policy validator

rahulreddy:
Validates user policies against the JSON schema per AWS' documentation
 and validator implementation.

### DIFF
--- a/errors/arsenalErrors.json
+++ b/errors/arsenalErrors.json
@@ -293,13 +293,13 @@
     "description": "The request signature we calculated does not match the signature you provided."
   },
   "_comment" : {
-    "note" : "This is an AWS S3 specific error. We are opting to use the more general 'ServiceUnavailable' error used throughout AWS (IAM/EC2) to have uniformity of error messages even though we are potentially compromising S3 compatibility.",    
+    "note" : "This is an AWS S3 specific error. We are opting to use the more general 'ServiceUnavailable' error used throughout AWS (IAM/EC2) to have uniformity of error messages even though we are potentially compromising S3 compatibility.",
     "ServiceUnavailable": {
       "code": 503,
       "description": "Reduce your request rate."
     }
   },
-  "ServiceUnavailable": { 
+  "ServiceUnavailable": {
     "code": 503,
     "description": "The request has failed due to a temporary failure of the server."
   },
@@ -418,6 +418,18 @@
   "AccountNotFound": {
     "code": 500,
     "description": "No account was found in Vault, please contact your system administrator."
+  },
+  "ValidationError": {
+    "code": 400,
+    "description": "The specified value is invalid."
+  },
+  "MalformedPolicyDocument": {
+     "code": 400,
+     "description": "Syntax errors in policy."
+  },
+  "InvalidInput": {
+     "code": 400,
+     "description": "The request was rejected because an invalid or out-of-range value was supplied for an input parameter."
   },
   "_comment": "-------------- Special non-AWS S3 errors --------------",
   "MPUinProgress": {
@@ -550,7 +562,7 @@
   "SecretKeyDoesNotExist": {
     "description": "secret key does not exist",
     "code": 5030
-  }, 
+  },
   "InvalidRegion": {
     "description": "Region was not provided or is not recognized by the system",
     "code": 5031
@@ -583,15 +595,15 @@
   "BadUrl": {
     "description": "url not ok",
     "code": 5038
-  }, 
+  },
   "BadClientIdList": {
     "description": "client id list not ok'",
     "code": 5039
-  }, 
+  },
   "BadThumbprintList": {
     "description": "thumbprint list not ok'",
     "code": 5040
-  }, 
+  },
   "BadObject": {
     "description": "Object not ok'",
     "code": 5041
@@ -600,12 +612,12 @@
   "BadRole": {
     "description": "role not ok",
     "code": 5042
-  }, 
+  },
   "_comment": "#### SamlpErrors ####",
   "BadSamlp": {
     "description": "samlp not ok",
     "code": 5043
-  }, 
+  },
   "BadMetadataDocument": {
     "description": "metadata document not ok",
     "code": 5044
--- a/index.js
+++ b/index.js
@@ -18,5 +18,7 @@ module.exports = {
     },
     policies: {
         evaluators: require('./lib/policyEvaluator/evaluator.js'),
+        validateUserPolicy: require('./lib/policy/policyValidator')
+            .validateUserPolicy,
     },
 };
--- a/lib/policy/policyValidator.js
+++ b/lib/policy/policyValidator.js
@@ -1,0 +1,117 @@
+'use strict'; // eslint-disable-line strict
+
+const Ajv = require('ajv');
+const userPolicySchema = require('./userPolicySchema');
+const errors = require('../errors');
+
+const ajValidate = new Ajv({ allErrors: true });
+// compiles schema to functions and caches them for all cases
+const userPolicyValidate = ajValidate.compile(userPolicySchema);
+
+const errDict = {
+    required: {
+        Version: 'Policy document must be version 2012-10-17 or greater.',
+        Action: 'Policy statement must contain actions.',
+    },
+    pattern: {
+        Action: 'Actions/Conditions must be prefaced by a vendor,' +
+            ' e.g., iam, sdb, ec2, etc.',
+        Resource: 'Resource must be in ARN format or "*".',
+    },
+    minItems: {
+        Resource: 'Policy statement must contain resources.',
+    },
+};
+
+// parse ajv errors and return early with the first relevant error
+function _parseErrors(ajvErrors) {
+    // deep copy is needed as we have to assign custom error description
+    const parsedErr = Object.assign({}, errors.MalformedPolicyDocument);
+    parsedErr.description = 'Syntax errors in policy.';
+    ajvErrors.some(err => {
+        const resource = err.dataPath;
+        const field = err.params ? err.params.missingProperty : undefined;
+        const errType = err.keyword;
+        if (errType === 'type' && (resource === '.Statement' ||
+            resource === '.Statement.Resource' ||
+            resource === '.Statement.NotResource')) {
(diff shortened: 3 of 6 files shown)


rahulreddy:
I have a few more unit tests to add, but I want to get some reviews going meanwhile.

lib/policy/policyValidator.js
```diff
@@ -0,0 +1,108 @@
...
+                parsedErr = errors.PolicyInvalidVersion;
+            }
+        } else {
+            parsedErr = errors.InvalidPolicyDocument;
+        }
+        return parsedErr instanceof Error;
+    });
+    return parsedErr;
+}
+
```
LaurenSpiegel: clever 😄

lib/policy/policyValidator.js
```diff
@@ -0,0 +1,108 @@
...
+            return { error: errors.PolicyInvalidJSON, valid: false };
+        }
+        userPolicyValidate(parseRes);
+        if (userPolicyValidate.errors) {
+            return { error: _parseErrors(userPolicyValidate.errors),
+                valid: false };
+        }
+        return { error: null, valid: true };
+    }
+    // todo: add support for resource policies
```
LaurenSpiegel: Sore thumb!

lib/policy/userPolicySchema.json
```diff
@@ -0,0 +1,401 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "type": "object",
+    "title": "AWS Policy schema.",
+    "description": "This schema describes user policy per AWS policy grammar rules",
```
LaurenSpiegel: a user policy

lib/policy/userPolicySchema.json
```diff
@@ -0,0 +1,401 @@
...
+    "definitions": {
+        "actionItem": {
+            "type": "string",
+            "pattern": "^[^*:]+:([^:])+|^\\*{1}$"
+        },
+        "resourceItem": {
+            "type": "string",
+            "pattern": "^\\*|arn:aws(:(\\*{1}|[a-z0-9\\*\\-]{2,})*){2}:.*$"
```
LaurenSpiegel: Need 5 ":"  Please confirm this works

lib/policy/userPolicySchema.json
```diff
@@ -0,0 +1,401 @@
...
+                },
+                "DateGreaterThanEquals": {
+                    "type": "object"
+                },
+                "Bool": {
+                    "type": "object"
+                },
+                "BinaryEquals": {
+                    "type": "object"
+                },
```
LaurenSpiegel: need BinaryNotEquals too.  I think it's missing from their docs but it's in the policy generator.

LaurenSpiegel:
A few comments, otherwise 👍

lib/policy/userPolicySchema.json
rahulreddy: You are right, my wildcard check was masking it. I have updated the regex to check specifically for 5 occurrences of `:` and added test asserting the behavior.

lib/policy/userPolicySchema.json
rahulreddy: Added `BinaryNotEquals` and `BinaryNotEqualsIfExists`

lib/policy/policyValidator.js
```diff
@@ -0,0 +1,108 @@
...
+    });
+    return parsedErr;
+}
+
+
+// parse JSON safely without throwing an exception
+function _safeJSONParse(s) {
+    let res;
+    try {
+        res = JSON.parse(s);
```
alexandre-merle: you can return res here :)

alexandre-merle:
LGTM :+1:

lib/policy/policyValidator.js
```diff
@@ -0,0 +1,106 @@
...
+        } else if (err.keyword === 'type') {
+            // skip if it's Statement as it does not have enough
+            // error context
+            if (resource === 'Version') {
+                parsedErr = errors.PolicyInvalidVersion;
+            }
+        } else {
+            parsedErr = errors.InvalidPolicyDocument;
+        }
+        return parsedErr instanceof Error;
```
michael-zapata: So, in some cases you send a `bool` rather than an `Error`?

package.json
```diff
@@ -14,6 +14,7 @@
   },
   "homepage": "https://github.com/scality/Arsenal#readme",
   "dependencies": {
+    "ajv": "^4.1.3",
```
michael-zapata: Version `4.1.5` is out. Could we use a fixed version to promote build reproducibility and avoid random breakage?

lib/policy/policyValidator.js
rahulreddy: The main function returns either `undefined` or `Error`. The test function for Array.some always returns `bool` which would only exit the Array.some not the main function.

package.json
rahulreddy: Upgraded the dependency to 4.1.5 and uses fixed version.

lib/policy/policyValidator.js
```diff
@@ -0,0 +1,106 @@
...
+            } else if (field === 'Action') {
+                parsedErr = errors.MissingPolicyAction;
+            } else if (field === 'Effect') {
+                parsedErr = errors.MissingPolicyEffect;
+            } else if (field === 'Resource') {
+                parsedErr = errors.MissingPolicyResource;
+            } else {
+                parsedErr = errors.InvalidPolicyDocument;
+            }
+        } else if (err.keyword === 'minItems' && resource === 'Statement') {
```
michael-zapata: `resource` is only used here, we shouldn't compute it every time.

rahulreddy: I didn't understand your comment, I need resource to evaluate the else if block and it changes with every err item in the list. Also, all it's doing is changing `.Statement` => `Statement`

errors/arsenalErrors.json
```diff
@@ -419,6 +419,18 @@
     "code": 500,
     "description": "No account was found in Vault, please contact your system administrator."
   },
+  "ValidationError": {
+    "code": 400,
+    "description": "The specified value is invalid"
```
RachedBenMustapha: Missing dot here?

index.js
```diff
@@ -18,5 +18,7 @@ module.exports = {
     },
     policies: {
         evaluators: require('./lib/policyEvaluator/evaluator.js'),
```
RachedBenMustapha: We probably don't need the `.js` here.

LaurenSpiegel:
👍

lib/policy/policyValidator.js
```diff
@@ -0,0 +1,117 @@
...
+            ' e.g., iam, sdb, ec2, etc.',
+        Resource: 'Resource must be in ARN format or "*".',
+    },
+    minItems: {
+        Resource: 'Policy statement must contain resources.',
+    },
+};
+
+// parse ajv errors and build list of erros
+function _parseErrors(ajvErrors) {
```
RachedBenMustapha: It seems we are not building a list here but returning the first usable error (with the `some` shortcut), the comment could be updated.

index.js
rahulreddy: It's not related to this PR.

lib/policy/policyValidator.js
```diff
@@ -0,0 +1,117 @@
...
+// validates policy using the validation schema
+function _validatePolicy(type, policy) {
+    if (type === 'user') {
+        const parseRes = _safeJSONParse(policy);
+        if (parseRes instanceof Error) {
+            return { error: Object.assign({}, errors.MalformedPolicyDocument),
+                valid: false };
+        }
+        userPolicyValidate(parseRes);
+        if (userPolicyValidate.errors) {
```
RachedBenMustapha: Their API feels strange, having one global function that monkey-patches itself. Seems OK here but feels so unsafe 😭

lib/policy/userPolicySchema.json
```diff
@@ -0,0 +1,407 @@
...
+    "title": "AWS Policy schema.",
+    "description": "This schema describes a user policy per AWS policy grammar rules",
+    "definitions": {
+        "actionItem": {
+            "type": "string",
+            "pattern": "^[^*:]+:([^:])+|^\\*{1}$"
+        },
+        "resourceItem": {
+            "type": "string",
+            "pattern": "^\\*|arn:aws(:(\\*{1}|[a-z0-9\\*\\-]{2,})*?){3}:((?!\\$\\{\\}).)*?$"
```
RachedBenMustapha: I trust you have validated this as my eyes are bleeding 😆

index.js
RachedBenMustapha: Oh yes sorry I missed the non-green background :)

RachedBenMustapha:
Minor comment on a comment, otherwise 👍

lib/policy/policyValidator.js
rahulreddy: I know...it's very strange.

lib/policy/userPolicySchema.json
rahulreddy: Yes 😁 . I have unit tests confirming it's valid and invalid cases.

lib/policy/policyValidator.js
rahulreddy: True, this function has changed a lot since it was initially written. I have upgraded the description reflecting the current usage.

lib/policy/policyValidator.js
```diff
@@ -8,40 +8,50 @@ const ajValidate = new Ajv({ allErrors: true });
...
+    required: {
+        Version: 'Policy document must be version 2012-10-17 or greater.',
+        Action: 'Policy statement must contain actions.',
+    },
+    pattern: {
+        Action: 'Actions/Conditions must be prefaced by a vendor,' +
+            ' e.g., iam, sdb, ec2, etc.',
+        Resource: 'Resource must be in ARN format or "*".',
+    },
+    minItems: {
```
alexandre-merle: where minItems it's used ? :)

rahulreddy: Good catch, I forgot to add it. I have added the condition and a test for it.

alexandre-merle:
Good to me :+1:

PR #107, opened 2016-07-01
Implement policy evaluator

LaurenSpiegel:
Implements and tests a user policy evaluator.  Can be extended to handle resource policies.

### DIFF
--- a/lib/policyEvaluator/evaluator.js
+++ b/lib/policyEvaluator/evaluator.js
@@ -1,0 +1,283 @@
+'use strict'; // eslint-disable-line strict
+
+const substituteVariables = require('./utils/variables.js');
+const handleWildcards = require('./utils/wildcards.js').handleWildcards;
+const conditions = require('./utils/conditions.js');
+const findConditionKey = conditions.findConditionKey;
+const convertConditionOperator = conditions.convertConditionOperator;
+const checkArnMatch = require('./utils/checkArnMatch.js');
+
+const evaluators = {};
+
+
+/**
+ * Check whether resource in policy statement applies to request resource
+ * @param {object} requestContext - info about request
+ * @param {string | [string]} statementResource - Resource(s) impacted
+ * by policy statement
+ * @param {object} log - logger
+ * @return {boolean} true if applicable, false if not
+ */
+function isResourceApplicable(requestContext, statementResource, log) {
+    const resource = requestContext.resource;
+    if (!Array.isArray(statementResource)) {
+        // eslint-disable-next-line no-param-reassign
+        statementResource = [statementResource];
+    }
+    // ARN format:
+    // arn:partition:service:region:namespace:relative-id
+    const requestResourceArr = resource.split(':');
+    // Pull just the relative id because there is no restriction that it
+    // does not contain ":"
+    const requestRelativeId = requestResourceArr.slice(5).join(':');
+    for (let i = 0; i < statementResource.length; i ++) {
+        // Handle variables (must handle BEFORE wildcards)
+        const policyResource =
+            substituteVariables(statementResource[i], requestContext);
+        // Handle wildcards
+        const arnSegmentsMatch =
+            checkArnMatch(policyResource, requestRelativeId,
+                requestResourceArr, true);
+        if (arnSegmentsMatch) {
+            log.trace('policy resource is applicable to request',
+            { requestResource: resource, policyResource });
+            return true;
+        }
+        continue;
+    }
+    log.trace('no policy resource is applicable to request',
+        { requestResource: resource });
+    // If no match found, no resource is applicable
+    return false;
+}
+
+/**
+ * Check whether action in policy statement applies to request
+ * @param {Object} requestAction - Type of client request
+ * @param {string | [string]} statementAction - Action(s) impacted
+ * by policy statement
+ * @param {Object} log - logger
+ * @return {boolean} true if applicable, false if not
+ */
+function isActionApplicable(requestAction, statementAction, log) {
+    if (!Array.isArray(statementAction)) {
+        // eslint-disable-next-line no-param-reassign
+        statementAction = [statementAction];
+    }
+    const length = statementAction.length;
+    for (let i = 0; i < length; i ++) {
+        // No variables in actions so no need to handle
+        const regExStrOfStatementAction =
+            handleWildcards(statementAction[i]);
+        const actualRegEx = new RegExp(regExStrOfStatementAction);
+        if (actualRegEx.test(requestAction)) {
+            log.trace('policy action is applicable to request action', {
+                requestAction, policyAction: statementAction[i],
+            });
+            return true;
+        }
+    }
+    log.trace('no action in policy applicable to request action',
+        { requestAction });
+    // If no match found, return false
+    return false;
+}
+
+/**
+ * Check whether request meets policy conditions
+ * @param {Object} requestContext - info about request
+ * @param {Object} statementCondition - Condition statement from policy
+ * @param {Object} log - logger
+ * @return {boolean} true if meet conditions, false if not
+ */
+function meetConditions(requestContext, statementCondition, log) {
+    // The Condition portion of a policy is an object with different
+    // operators as keys
+    const operators = Object.keys(statementCondition);
+    const length = operators.length;
+    for (let i = 0; i < length; i ++) {
+        const operator = operators[i];
+        const hasIfExistsCondition = operator.endsWith('IfExists');
+        // If has "IfExists" added to operator name, find operator name
+        // without "IfExists"
+        const bareOperator = hasIfExistsCondition ? operator.slice(0, -8) :
+            operator;
+        const operatorsWithVariables = ['StringEquals', 'StringNotEquals',
+            'StringEqualsIgnoreCase', 'StringNotEqualsIgnoreCase',
+            'StringLike', 'StringNotLike', 'ArnEquals', 'ArnNotEquals',
+            'ArnLike', 'ArnNotLike'];
+        const operatorCanHaveVariables =
+            operatorsWithVariables.indexOf(bareOperator) > -1;
+        // Loop through conditions with the same operator
+        // Note: this should be the actual operator name, not the bareOperator
+        const conditionsWithSameOperator = statementCondition[operator];
+        const conditionKeys = Object.keys(conditionsWithSameOperator);
+        const conditionKeysLength = conditionKeys.length;
+        for (let j = 0; j < conditionKeysLength;
+            j ++) {
+            const key = conditionKeys[j];
+            let value = conditionsWithSameOperator[key];
+            if (!Array.isArray(value)) {
+                value = [value];
+            }
+            // Handle variables
+            if (operatorCanHaveVariables) {
+                value = value.map(item =>
+                    substituteVariables(item, requestContext));
+            }
+            // Pull key using requestContext
+            // TODO: If applicable to S3, handle policy set operations
+            // where a keyBasedOnRequestContext returns multiple values and
+            // condition has "ForAnyValue" or "ForAllValues".
+            // (see http://docs.aws.amazon.com/IAM/latest/UserGuide/
+            // reference_policies_multi-value-conditions.html)
+            const keyBasedOnRequestContext =
+                findConditionKey(key, requestContext);
+            // Handle IfExists
+            if (keyBasedOnRequestContext === undefined
+                && hasIfExistsCondition) {
+                log.trace('satisfies condition due to IfExists operator');
+                continue;
+            }
+            // If no IfExists qualifier, the key does not exist and the
+            // condition operator is not Null, the
+            // condition is not met so return false.
+            if (keyBasedOnRequestContext === undefined
+                && bareOperator !== 'Null') {
+                log.trace('condition not satisfied due to ' +
+                'missing info', { operator,
+                    conditionKey: key, policyValue: value });
+                return false;
+            }
+            // Transalate operator into function using bareOperator
+            const operatorFunction = convertConditionOperator(bareOperator);
+            // Note: Wildcards are handled in the comparison operator function
+            // itself since StringLike, StringNotLike, ArnLike and ArnNotLike
+            // are the only operators where wildcards are allowed
+            if (!operatorFunction(keyBasedOnRequestContext, value)) {
+                log.trace('did not satisfy condition', { operator: bareOperator,
+                        keyBasedOnRequestContext, policyValue: value });
+                return false;
+            }
+        }
+    }
+    return true;
+}
+
+/**
+ * Evaluate whether a request is permitted under a policy.
+ * @param {object} requestContext - Info necessary to evaluate permission
+ * See http://docs.aws.amazon.com/IAM/latest/UserGuide/
+ * reference_policies_evaluation-logic.html#policy-eval-reqcontext
+ * @param {object} requestContext.resource - arn of requested resource
+ * @param {string} requestContext.action - Action type of request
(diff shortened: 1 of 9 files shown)


lib/policyEvaluator/evaluator.js
```diff
@@ -0,0 +1,257 @@
...
+ * @param {string | [string]} statementAction - Action(s) impacted
+ * by policy statement
+ * @return {boolean} true if applicable, false if not
+ */
+function isActionApplicable(requestAction, statementAction) {
+    if (!Array.isArray(statementAction)) {
+        // eslint-disable-next-line no-param-reassign
+        statementAction = [statementAction];
+    }
+    for (let i = 0; i < statementAction.length; i ++) {
```
michael-zapata: This one can be a `reduce()`.

lib/policyEvaluator/evaluator.js
```diff
@@ -0,0 +1,257 @@
...
+ * @param {object} requestContext - Info necessary to evaluate permission
+ * See http://docs.aws.amazon.com/IAM/latest/UserGuide/
+ * reference_policies_evaluation-logic.html#policy-eval-reqcontext
+ * @param {[object]} allPolicies - all applicable IAM or resource policies
+ * @return {string} Allow if permitted, Deny if not permitted.
+ * Default is to Deny. Deny overrides an Allow
+ */
+evaluators.evaluateAllPolicies = (requestContext, allPolicies) => {
+    let verdict = 'Deny';
+    for (let i = 0; i < allPolicies.length; i++) {
```
michael-zapata: This one can also be a `reduce()`

lib/policyEvaluator/utils/conditions.js
```diff
@@ -0,0 +1,407 @@
...
+/**
+ * findConditionKey finds the value of a condition key based on requestContext
+ * @param {string} key - condition key name
+ * @param {object} requestContext - info sent with request
+ * @return {string} condition key value
+ */
+conditions.findConditionKey = (key, requestContext) => {
+    // TODO: Consider combining with findVariable function if no benefit
+    // to keeping separate
+    const map = new Map();
```
michael-zapata: Instead of doing all those separate `.set()` you could use the `new Map(iterator)`

lib/policyEvaluator/utils/conditions.js
```diff
@@ -0,0 +1,407 @@
...
+    // or security issue with object assignment
+    const operatorMap = {
+        StringEquals: function strictEqual(key, value) {
+            return value.some(item => {
+                const swtichedOutChars = convertSpecialChars(item);
+                return swtichedOutChars === key;
+            });
+        },
+        StringNotEquals: function strictNotEquals(key, value) {
+            // eslint-disable-next-line new-cap
```
michael-zapata: Why not fix the warning instead?

tests/unit/policyEvaluator.js
```diff
@@ -0,0 +1,1006 @@
...
+
+function check(requestContext, policy, expected) {
+    const result = evaluatePolicy(requestContext, policy);
+    assert.deepStrictEqual(result, expected);
+}
+let policy;
+let requestContext;
+
+describe('policyEvaluator', () => {
+    describe('evaluate a single policy', () => {
```
michael-zapata: Please avoid test duplication.

lib/policyEvaluator/evaluator.js
alexandre-merle: For performance, i'm to keep that way, but store the length before to avoid object access :)

lib/policyEvaluator/utils/conditions.js
LaurenSpiegel: Is there a performance difference?

lib/policyEvaluator/utils/conditions.js
LaurenSpiegel: I wanted to match the names that AWS uses.

lib/policyEvaluator/evaluator.js
```diff
@@ -0,0 +1,257 @@
...
+    if (!Array.isArray(statementResource)) {
+        // eslint-disable-next-line no-param-reassign
+        statementResource = [statementResource];
+    }
+    // ARN format:
+    // arn:partition:service:region:namespace:relative-id
+    const requestResourceArr = resource.split(':');
+    // Pull just the relative id because there is no restriction that it
+    // does not contain ":"
+    const requestRelativeId = requestResourceArr.slice(5).join(':');
```
alexandre-merle: I think a better way to compute the relative id is to take the length of the 5 first part and then do a substring, because here, we create an array without the first 5 five part, and then we create a string with the remaining parts :)

lib/policyEvaluator/evaluator.js
```diff
@@ -0,0 +1,257 @@
...
+/**
+ * Check whether request meets policy conditions
+ * @param {Object} requestContext - info about request
+ * @param {Object} statementCondition - Condition statement from policy
+ * @return {boolean} true if meet conditions, false if not
+ */
+function meetConditions(requestContext, statementCondition) {
+    // The Condition portion of a policy is an object with different
+    // operators as keys
+    for (let i = 0; i < Object.keys(statementCondition).length; i ++) {
```
alexandre-merle: May be store the keys and the length before :)

lib/policyEvaluator/evaluator.js
```diff
@@ -0,0 +1,257 @@
...
+        const operatorsWithVariables = ['StringEquals', 'StringNotEquals',
+            'StringEqualsIgnoreCase', 'StringNotEqualsIgnoreCase',
+            'StringLike', 'StringNotLike', 'ArnEquals', 'ArnNotEquals',
+            'ArnLike', 'ArnNotLike'];
+        const operatorCanHaveVariables =
+            operatorsWithVariables.indexOf(bareOperator) > -1;
+        // Loop through conditions with the same operator
+        // Note: this should be the actual operator name, not the bareOperator
+        const conditionsWithSameOperator = statementCondition[operator];
+        for (let j = 0; j < Object.keys(conditionsWithSameOperator).length;
```
alexandre-merle: Same here :)

lib/policyEvaluator/evaluator.js
```diff
@@ -0,0 +1,257 @@
...
+    // Pull just the relative id because there is no restriction that it
+    // does not contain ":"
+    const requestRelativeId = requestResourceArr.slice(5).join(':');
+    for (let i = 0; i < statementResource.length; i ++) {
+        // Handle variables (must handle BEFORE wildcards)
+        const policyResource =
+            substituteVariables(statementResource[i], requestContext);
+        // Handle wildcards
+        const arnSegmentsMatch =
+            checkArnMatch(policyResource, requestRelativeId,
```
alexandre-merle: You can pass the relativeid here, to avoir recompute it in checkArnMatch :)

lib/policyEvaluator/evaluator.js
LaurenSpiegel: I'd rather return early if there is any deny.

lib/policyEvaluator/evaluator.js
LaurenSpiegel: I use the array in checkArnMatch.

lib/policyEvaluator/evaluator.js
LaurenSpiegel: The length of this list is probably at most 5.

lib/policyEvaluator/utils/conditions.js
```diff
@@ -0,0 +1,407 @@
...
+        },
+        BinaryNotEquals: function binaryNotEquals(key, value) {
+            // eslint-disable-next-line new-cap
+            return !operatorMap.BinaryEquals(key, value);
+        },
+        IpAddress: function ipAddress(key, value) {
+            return value.some(item => {
+                let inRange = false;
+                try {
+                    inRange = ip.cidrSubnet(item).contains(key);
```
alexandre-merle: you can just return here :)

lib/policyEvaluator/utils/variables.js
```diff
@@ -0,0 +1,103 @@
...
+ */
+function substituteVariables(string, requestContext) {
+    const arr = string.split('');
+    let startOfVariable = arr.indexOf('$');
+    while (startOfVariable > -1) {
+        if (arr[startOfVariable + 1] !== '{') {
+            startOfVariable = arr.indexOf('$', startOfVariable + 1);
+            continue;
+        }
+        const end = arr.indexOf('}', startOfVariable + 1);
```
alexandre-merle: +2 :)

tests/unit/policyEvaluator.js
```diff
@@ -0,0 +1,1006 @@
...
+                    policy.Statement.Resource = '*';
+                    policy.Statement.Condition = {
+                        StringNotLike: { 's3:prefix': [
+                            'home/${aws:username}/?/*',
+                            'home/',
+                        ] } };
+                    requestContext.action = 's3:ListBucket';
+                    requestContext.requesterInfo.username = 'Pete';
+                    // pattern is home/${aws:username}/?/*
+                    requestContext.query.prefix = 'home/Pete/a/something';
```
alexandre-merle: you don't match the pattern here ?

tests/unit/policyEvaluator.js
LaurenSpiegel: Tests refactored.

GiorgioRegni:
Love reading the comments in the code, very precise and concise!

lib/policyEvaluator/evaluator.js
rahulreddy: It's not the length of the Object but the call itself is very expensive. You should cache that in a variable.

lib/policyEvaluator/evaluator.js
```diff
@@ -0,0 +1,286 @@
...
+            // where a keyBasedOnRequestContext returns multiple values and
+            // condition has "ForAnyValue" or "ForAllValues".
+            // (see http://docs.aws.amazon.com/IAM/latest/UserGuide/
+            // reference_policies_multi-value-conditions.html)
+            const keyBasedOnRequestContext =
+                findConditionKey(key, requestContext);
+            // Handle IfExists
+            if (keyBasedOnRequestContext === undefined
+                && hasIfExistsCondition) {
+                log.trace('request does not contain info to satisfy condition' +
```
rahulreddy: We should aim for short, succinct log messages, long messages are less readable in Kibana and log files.

lib/policyEvaluator/utils/conditions.js
```diff
@@ -0,0 +1,407 @@
...
+ * convertToEpochTime checks whether epoch or ISO time and converts to epoch
+ * if necessary
+ * @param {string | array} time - value or values to be converted
+ * @return {string | array} converted value or values
+ */
+function convertToEpochTime(time) {
+    function convertSingle(item) {
+        // If ISO time
+        if (item.indexOf(':') > -1) {
+            return new Date(item).valueOf();
```
rahulreddy: This line would return a number and not a string. To be consistent, we should always return number/array of numbers since this function is used along with `NumericEquals` and others. So, if it's not ISO and its Epoch it needs to be converted to a number here as well.

lib/policyEvaluator/utils/conditions.js
```diff
@@ -0,0 +1,407 @@
...
+                const wildRegEx = new RegExp(wildItem);
+                return wildRegEx.test(key);
+            });
+        },
+        StringNotLike: function stringNotLike(key, value) {
+            // eslint-disable-next-line new-cap
+            return !operatorMap.StringLike(key, value);
+        },
+        NumericEquals: function numericEquals(key, value) {
+            return value.some(item => item === key);
```
rahulreddy: As other numeric functions are doing, you would have to convert the key and value to numbers if they are not. It's so that you can avoid comparisons of (string, number) or (number, string)

lib/policyEvaluator/utils/conditions.js
```diff
@@ -0,0 +1,407 @@
...
+        },
+        NumericNotEquals: function numericNotEquals(key, value) {
+            // eslint-disable-next-line new-cap
+            return !operatorMap.NumericEquals(key, value);
+        },
+        NumericLessThan: function lessThan(key, value) {
+            const numberKey = Number.parseInt(key, 10);
```
rahulreddy: Do you mind extracting this into a function
```javascript
const numberItem = Number.parseInt(item, 10);
if (Number.isNaN(numberItem)) {
    return false;
}
```
and re-use the function everywhere?

lib/policyEvaluator/utils/conditions.js
```diff
@@ -0,0 +1,407 @@
...
+            return value.some(item => item === base64Key);
+        },
+        BinaryNotEquals: function binaryNotEquals(key, value) {
+            // eslint-disable-next-line new-cap
+            return !operatorMap.BinaryEquals(key, value);
+        },
+        IpAddress: function ipAddress(key, value) {
+            return value.some(item => {
+                let inRange = false;
+                try {
```
rahulreddy: For V8 performance reasons, can you extract the try...catch into an isolated function?

lib/policyEvaluator/utils/wildcards.js
```diff
@@ -0,0 +1,62 @@
...
+ * @param {string} string - any input string
+ * @return {string} converted string
+ */
+wildcards.handleWildcards = string => {
+    // Replace all '*' with '.*' (allow any combo of letters)
+    // and all '?' with '.{1}' (allow for any one character)
+    // If *, ? or $ are enclosed in ${}, keep literal *, ?, or $
+    function characterMap(char) {
+        const map = {
+            '\\*': '.*',
```
rahulreddy: I would need a second opinion on this but I think should we convert all greedy regex checks to lazy. It would be eating a ton of CPU cycles if not.
So `*` becomes `*?`, unless you are matching for a `*` character of course.

tests/unit/policyEvaluator.js
```diff
@@ -0,0 +1,930 @@
...
+                        'arn:aws:s3:::notVeryPrivate/object';
+                    check(requestContext,
+                        samples['Multi-Statement Policy'],
+                        'Allow');
+                });
+
+            it('should be neutral under multi-statement ' +
+                'policy if request action not in policy',
+                () => {
+                    requestContext.action = 's3:DeleteObject';
```
rahulreddy: Is there a benefit to changing the action and resource in each of these tests?

rahulreddy:
We need some tests for the Date conditions using Unix epoch times as well.

lib/policyEvaluator/utils/conditions.js
LaurenSpiegel: Will not actually reduce the lines of code.

tests/unit/policyEvaluator.js
LaurenSpiegel: I tried to change them where they made sense.  For instance, checking if an action is not in a policy, I had to change the action.  Also, if conditions did not make sense for an action, I changed it (such as having a condition with maxKeys, etc. with a listing action).

LaurenSpiegel:
Comments addressed

lib/policyEvaluator/utils/conditions.js
rahulreddy: Sad, I overlooked that return part.

rahulreddy:
👍

alexandre-merle:
LGTM :+1:

lib/policyEvaluator/utils/conditions.js
michael-zapata: Between a single allocation and multiple ones?

PR #109, opened 2016-07-08
Ft/policy auth

LaurenSpiegel:
1) Creates a RequestContext class to be used in S3 and Vault to gather the information needed to evaluate IAM policies.

2) Pulls information from the authentication code to add to a RequestContext instance.

@DavidPineauScality, I had to mess with the authentication code a bit...

### DIFF
--- a/lib/auth/auth.js
+++ b/lib/auth/auth.js
@@ -41,7 +41,8 @@ auth.check = (request, log, awsService, data) => {
             return authV2.headerAuthCheck.check(request, log, data);
         } else if (authHeader.startsWith('AWS4')) {
             log.trace('authenticating request with Auth V4 using headers');
-            return authV4.headerAuthCheck.check(request, log, data, awsService);
+            return authV4.headerAuthCheck.check(request, log, data,
+                awsService);
         }
         log.debug('missing authorization security header');
         return { err: errors.MissingSecurityHeader };
@@ -61,14 +62,17 @@ auth.check = (request, log, awsService, data) => {
     return { err: null, data: authInfo };
 };
 
-auth.doAuth = (request, log, cb, awsService, data) => {
-    const res = auth.check(request, log, awsService, data);
+auth.doAuth = (request, log, cb, awsService, requestContext) => {
+    const res = auth.check(request, log, awsService, request.query);
+    requestContext.setAuthType(res.authType);
+    requestContext.setSignatureVersion(res.signatureVersion);
+    requestContext.setSignatureAge(res.signatureAge);
     if (res.err) {
         return cb(res.err);
     } else if (res.version === 2) {
         return vault.authenticateV2Request(res.data.accessKey,
             res.data.signatureFromRequest,
-            res.data.stringToSign, res.data.algo, log,
+            res.data.stringToSign, res.data.algo, requestContext, log,
             (err, authInfo) => {
                 if (err) {
                     return cb(err);
@@ -77,7 +81,8 @@ auth.doAuth = (request, log, cb, awsService, data) => {
             });
     } else if (res.version === 4) {
         res.data.log = log;
-        return vault.authenticateV4Request(res.data, cb, awsService);
+        return vault.authenticateV4Request(res.data, requestContext,
+            cb, awsService);
     } else if (res.data instanceof AuthInfo) {
         return cb(null, res.data);
     }
--- a/lib/policyEvaluator/RequestContext.js
+++ b/lib/policyEvaluator/RequestContext.js
@@ -1,0 +1,417 @@
+'use strict'; // eslint-disable-line strict
+
+
+// http://docs.aws.amazon.com/IAM/latest/UserGuide/list_s3.html
+// For MPU actions:
+// http://docs.aws.amazon.com/AmazonS3/latest/dev/mpuAndPermissions.html
+// For bucket head and object head:
+// http://docs.aws.amazon.com/AmazonS3/latest/dev/
+// using-with-s3-actions.html
+const _actionMap = {
+    bucketDelete: 's3:DeleteBucket',
+    bucketGet: 's3:ListBucket',
+    bucketGetACL: 's3:GetBucketAcl',
+    bucketHead: 's3:ListBucket',
+    bucketPut: 's3:CreateBucket',
+    bucketPutACL: 's3:PutBucketAcl',
+    completeMultipartUpload: 's3:PutObject',
+    initiateMultipartUpload: 's3:PutObject',
+    listMultipartUploads: 's3:ListBucketMultipartUploads',
+    listParts: 's3:ListMultipartUploadParts',
+    multipartDelete: 's3:AbortMultipartUpload',
+    objectDelete: 's3:DeleteObject',
+    objectGet: 's3:GetObject',
+    objectGetACL: 's3:GetObjectAcl',
+    objectHead: 's3:GetObject',
+    objectPut: 's3:PutObject',
+    objectPutACL: 's3:PutObjectAcl',
+    objectPutPart: 's3:PutObject',
+    serviceGet: 's3:ListAllMyBuckets',
+};
+
+
+function _findAction(service, method) {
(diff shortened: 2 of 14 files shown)


lib/auth/auth.js
```diff
@@ -38,20 +38,33 @@ auth.check = (request, log, awsService, data) => {
         // handle temporary security credentials
         if (authHeader.startsWith('AWS ')) {
             log.trace('authenticating request with auth v2 using headers');
-            return authV2.headerAuthCheck.check(request, log, data);
+            const res = authV2.headerAuthCheck.check(request, log, data);
+            res.authType = 'REST-HEADER';
```
alexandre-merle: I think those information can be into headerAuthCheck :) Same for the three others :)

lib/policyEvaluator/RequestContext.js
```diff
@@ -0,0 +1,400 @@
...
+
+function _findAction(service, method) {
+    if (service === 's3') {
+        // http://docs.aws.amazon.com/IAM/latest/UserGuide/list_s3.html
+        // For MPU actions:
+        // http://docs.aws.amazon.com/AmazonS3/latest/dev/mpuAndPermissions.html
+        // For bucket head and object head:
+        // http://docs.aws.amazon.com/AmazonS3/latest/dev/
+        // using-with-s3-actions.html
+        const map = {
```
alexandre-merle: I think this map will be better outside the function :)

lib/policyEvaluator/RequestContext.js
```diff
@@ -0,0 +1,400 @@
...
+        return JSON.stringify(requestInfo);
+    }
+
+    /**
+     * deSerialize the JSON string
+     * @param {string} stringRequest - the stringified requestContext
+     * @return {object} - parsed string
+     */
+    static deSerialize(stringRequest) {
+        const obj = JSON.parse(stringRequest);
```
alexandre-merle: May be perform try catch here, and handle an potential error :)

lib/policyEvaluator/RequestContext.js
```diff
@@ -0,0 +1,400 @@
...
+            obj.awsService, obj.locationConstraint, obj.requesterInfo,
+            obj.signatureVersion, obj.authType, obj.signatureAge);
+    }
+
+    /**
+    * Get the request action
+    * @return {string} action
+    */
+    getAction() {
+        return _findAction(this._awsService, this._apiMethod);
```
alexandre-merle: For performance consideration, create another variable to store the result of the _findAction in the constructor :)
nb: same for getRessource()

alexandre-merle:
Add a dropme commit to set the package version to 1.1.0-policies, to avoid any suprise in the end to end :)

lib/auth/auth.js
rahulreddy: Yeah, I agree. @LaurenSpiegel If you want to keep the `headerAuthCheck.check` pure from this logic, you can make a utility function in the same file which deals with this logic.

lib/policyEvaluator/RequestContext.js
```diff
@@ -0,0 +1,417 @@
...
+    objectPutPart: 's3:PutObject',
+    serviceGet: 's3:ListAllMyBuckets',
+};
+
+
+function _findAction(service, method) {
+    if (service === 's3') {
+        return _actionMap[method];
+    }
+    return undefined;
```
rahulreddy: `return;` is sufficient

lib/policyEvaluator/RequestContext.js
```diff
@@ -0,0 +1,417 @@
...
+function _buildArn(service, bucket, object) {
+    if (service === 's3') {
+        if (bucket && object) {
+            return `arn:aws:s3:::${bucket}/${object}`;
+        } else if (bucket) {
+            return `arn:aws:s3:::${bucket}`;
+        }
+        return 'arn:aws:s3:::';
+    }
+    return undefined;
```
rahulreddy: `return;` is sufficient

lib/policyEvaluator/RequestContext.js
```diff
@@ -0,0 +1,417 @@
...
+
+class RequestContext {
+    constructor(headers, query, bucketName, objectName,
+        requesterIp, isEncrypted, apiMethod,
+        awsService, locationConstraint, requesterInfo,
+        signatureVersion, authType, signatureAge) {
+        this._headers = headers;
+        this._query = query;
+        this._requesterIp = requesterIp;
+        this._sslEnabled = isEncrypted;
```
rahulreddy: `isEncrypted` may be confusing, can we name the param something on the lines of `sslEnabled`?

lib/policyEvaluator/RequestContext.js
```diff
@@ -0,0 +1,417 @@
...
+        this._apiMethod = apiMethod;
+        this._awsService = awsService;
+        this._bucket = bucketName;
+        this._object = objectName;
+        this._locationConstraint = locationConstraint;
+        // Not implemented
+        this._multiFactorAuthPresent = undefined;
```
rahulreddy: a `null` would make more sense than `undefined` for properties that are not set. This has zero impact on logic but conventional code uses it that way, so it's easy for someone new looking at the code.

lib/policyEvaluator/utils/variables.js
```diff
@@ -27,43 +27,44 @@ function findVariable(variable, requestContext) {
...
     map.set('aws:SecureTransport',
-        requestContext.sslEnabled ? 'true' : 'false');
+        requestContext.getSslEnabled() ? 'true' : 'false');
     // aws:SourceIp is requester's IP address, for use with IP address
     // conditions
-    map.set('aws:SourceIp', requestContext.requesterIp);
+    map.set('aws:SourceIp', requestContext.getRequesterIp());
     // aws:UserAgent is information about the requester's client application
-    map.set('aws:UserAgent', requestContext.headers['user-agent']);
+    map.set('aws:UserAgent', requestContext.getHeaders()['user-agent']);
```
rahulreddy: Can you assign requestContext calls to a variable and then use them here?

tests/unit/auth/v2/errorHandling.js
```diff
@@ -26,10 +28,13 @@ describe('Error handling in checkAuth', () => {
             url: '/bucket',
             query: {},
         };
+        const requestContext = new RequestContext(request.headers,
```
rahulreddy: I think we can do some cleanup of code duplication while we are here. It would be ideal to move this code into a check function and just pass only the required stuff.

tests/unit/policyEvaluator.js
```diff
@@ -58,9 +53,9 @@ describe('policyEvaluator', () => {
             it('should grant access under multi-statement ' +
                 'policy if action and resource match those in policy',
                 () => {
-                    requestContext.action = 's3:GetObject';
-                    requestContext.resource =
-                        'arn:aws:s3:::notVeryPrivate/object';
+                    requestContext._apiMethod = 'objectGet';
```
rahulreddy: I would recommended a function for this as well. It would be easier to change any method names or add new ones in one place in the future.

lib/policyEvaluator/RequestContext.js
LaurenSpiegel: linter complains 😞

lib/policyEvaluator/RequestContext.js
LaurenSpiegel: same

rahulreddy: Oh yeah, the linter. We (me & linter) are never on the same page.

tests/unit/auth/v2/errorHandling.js
LaurenSpiegel: Out of scope here and David is refactoring this code anywayl

LaurenSpiegel:
comments addressed

rahulreddy:
👍

alexandre-merle:
LGTM :+1:

PR #115, opened 2016-07-16
Handle space for all clients in post request auth

LaurenSpiegel:
This change is needed in the signature calculation for the go sdk to work.  Instead of adding to the "if" clause, I have removed the "if" check.  Without it, the signature works for AWS-CLI, the go sdk and the node sdk.


LaurenSpiegel:
end to end started from vault passed: https://github.com/scality/Vault/pull/433
http://ci.ironmann.io/gh/scality/Integration/2066

rahulreddy:
👍

rahulreddy:
Can you make this PR against rel/1.1, I will handle the fwdport.

alexandre-merle:
LGTM :+1:

PR #118, opened 2016-07-27
add versioning listing algorithm

ttvi:
and extend a bit existing listing algorithm for listing latest versions.

### DIFF
--- a/lib/algos/list/delimiter.js
+++ b/lib/algos/list/delimiter.js
@@ -3,7 +3,9 @@
 const checkLimit = require('./tools').checkLimit;
 const DEFAULT_MAX_KEYS = 1000;
 /**
- *  Class for the delimiter extension
+ *  Base class for the delimiter extension. This is here for the legacy reason.
+ *  In future versions, metadata should only do delimiting and it's S3 the part
+ *  to extract the desired properties of each entry from the received list.
  */
 class Delimiter {
     /**
--- a/lib/algos/list/delimiterMaster.js
+++ b/lib/algos/list/delimiterMaster.js
@@ -1,0 +1,84 @@
+'use strict'; // eslint-disable-line strict
+
+const Delimiter = require('./delimiter').Delimiter;
+const VSUtils = require('../../versioning/utils').VersioningUtils;
+
+/**
+ * Extended delimiter class for versioning.
+ */
+class DelimiterMaster extends Delimiter {
+    /**
+     * Overriding the base function to extract the versionId of the entry.
+     *
+     * @param {string} key - the key of the entry
+     * @param {object} value - the value of the entry
+     * @return {undefined}
+     */
+    addContents(key, value) {
+        this.Contents.push({
+            key,
+            value: {
+                Size: value['content-length'],
+                ETag: value['content-md5'],
+                LastModified: value['last-modified'],
+                // <versioning>
+                VersionId: VSUtils.getts(value),
+                // </versioning>
+                Owner: {
+                    DisplayName: value['owner-display-name'],
+                    ID: value['owner-id'],
+                },
+                StorageClass: value['x-amz-storage-class'],
+                Initiated: value.initiated,
+                Initiator: value.initiator,
+                EventualStorageBucket: value.eventualStorageBucket,
+                partLocations: value.partLocations,
+                creationDate: value.creationDate,
+            },
+        });
+        this.NextMarker = key;
+        ++this.keys;
+    }
+
+    /**
+     * Overriding the filter function that formats the
+     * listing results based on the listing algorithm.
+     *
+     * @param {object} obj - metadata entry in the form of { key, value }
+     * @return {boolean} - continue filtering or return the formatted list
+     */
+    filter(obj) {
+        // Check first in case of maxkeys <= 0
+        if (this.keys >= this.maxKeys) {
+            // In cases of maxKeys <= 0 => IsTruncated = false
+            this.IsTruncated = this.maxKeys > 0;
+            return false;
+        }
+        // <versioning>
+        const value = VSUtils.decodeVersion(obj.value);
+        // ignore it if the master version is a delete marker
+        if (VSUtils.isDeleteMarker(value)) {
+            return true;
+        }
+        // use the original object name for delimitering to work correctly
+        const key = VSUtils.getObjectNameFromMasterKey(obj.key);
+        // </versioning>
+        if (this.delimiter) {
+            const commonPrefixIndex =
+                key.indexOf(this.delimiter, this.searchStart);
+            if (commonPrefixIndex === -1) {
+                this.addContents(key, value);
+            } else {
+                this.addCommonPrefix(
+                    key.substring(0, commonPrefixIndex + this.delimLen));
+            }
+        } else {
+            this.addContents(key, value);
+        }
+        return true;
+    }
+}
+
+module.exports = {
+    DelimiterMaster,
+};
--- a/lib/algos/list/delimiterVersions.js
+++ b/lib/algos/list/delimiterVersions.js
@@ -1,0 +1,106 @@
+'use strict'; // eslint-disable-line strict
+
+const Delimiter = require('./delimiter').Delimiter;
+const VSUtils = require('../../versioning/utils').VersioningUtils;
+
+/**
+ * Extended delimiter class for versioning.
+ */
+class DelimiterVersions extends Delimiter {
+    /**
+     *  Constructor of the extension
+     *  Init and check parameters
+     *  @param {Object} parameters - parameters sent to DBD
+     *  @param {RequestLogger} logger - werelogs request logger
+     *  @param {object} latestVersions - latest versions of some keys
+     *  @return {undefined}
+     */
+    constructor(parameters, logger, latestVersions) {
+        super(parameters, logger);
+        this.NextVersionMarker = undefined; // next version marker
+        this.latestVersions = undefined; // final list of the latest versions
+        this._latestVersions = latestVersions; // reserved for caching
+    }
+
+    /**
+     * Overriding the base function to not process the metadata entry here,
+     * leaving the job of extracting object's attributes to S3.
+     *
+     * @param {string} key - key of the entry
+     * @param {string} value - value of the entry
+     * @return {undefined}
+     */
+    addContents(key, value) {
+        const components =
+            VSUtils.getObjectNameAndVersionIdFromVersionKey(key);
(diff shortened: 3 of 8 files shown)


lib/algos/list/delimiter.js
```diff
@@ -3,7 +3,9 @@
 const checkLimit = require('./tools').checkLimit;
 const DEFAULT_MAX_KEYS = 1000;
 /**
- *  Class for the delimiter extension
+ *  Base class for the delimiter extension. This is here for the legacy reason.
+ *  In future versions, metadata should only do delimiting and it's S3 the part
```
rahulreddy: 👍

rahulreddy:
I would recommend extracting `DelimiterVersions` class into it's own file, it would make it easier to refer to the code in the future.

lib/algos/list/delimiter.js
```diff
@@ -133,6 +141,62 @@ class Delimiter {
...
+/**
+ * Extended delimiter class for versioning.
+ */
+class DelimiterVersions extends Delimiter {
+    /**
+     *  Constructor of the extension
+     *  Init and check parameters
+     *  @param {Object} parameters - parameters sent to DBD
+     *  @param {RequestLogger} logger - werelogs request logger
+     *  @param {object} latests - latest versions of some keys
```
rahulreddy: Can you call rename `latests` to `latestVersions` instead? It would be nice to be more descriptive and meaningful.

tests/unit/algos/list/delimiterVersions.js
```diff
@@ -0,0 +1,172 @@
...
+        { key: '/notes/spring/1.txt', value },
+        { key: '/notes/spring/2.txt', value },
+        { key: '/notes/spring/march/1.txt', value },
+        { key: '/notes/summer/1.txt', value },
+        { key: '/notes/summer/2.txt', value },
+        { key: '/notes/summer/august/1.txt', value },
+        { key: '/notes/year.txt', value },
+        { key: '/Pâtisserie=中文-español-English', value },
+    ];
+    const tests = [
```
michael-zapata: Nice!

lib/versioning/utils.js
```diff
@@ -0,0 +1,36 @@
...
+        const indexOfVersionId = versionKey.lastIndexOf(SEP);
+        return versionKey.slice(VER.length, indexOfVersionId);
+    },
+    getObjectNameAndVersionIdFromVersionKey: versionKey => {
+        const indexOfVersionId = versionKey.lastIndexOf(SEP);
+        const objectName = versionKey.slice(VER.length, indexOfVersionId);
+        const versionId = versionKey.slice(indexOfVersionId + 1);
+        return { objectName, versionId };
+    },
+    decodeMasterVersion: masterVersion => JSON.parse(masterVersion),
```
MathieuCassagne: You can write it like:
```
decodeMasterVersion: JSON.parse,
```
A bit smaller, but not too important :)

lib/versioning/utils.js
```diff
@@ -0,0 +1,36 @@
...
+    // future extensions, such as attatching external information to the string
+    // value from request instead of having to parse and stringify the value
+    // Example: "prefixlen_{prefix}{value_from_request}"
+    encodeVersion: version => JSON.stringify(version),
+    decodeVersion: version => JSON.parse(version),
+    // using these functions to manipulate the attributes of the value also
+    // help us from restricting ourselves to hardcoding arbitrary attributes
+    setattr: (version, attr, value) => version[attr] = value,
+    setts: (version, ts) => version[VSConst.ATTR_VERSION_ID] = ts,
+    getts: version => version[VSConst.ATTR_VERSION_ID],
```
LaurenSpiegel: This naming is confusing.  It implies timestamp but has the version id.

ttvi: We used the term "timestamp" for unique identifier that increases monotonically and that happened to be used for "versionId". Though they are the same and "ts" is shorter but we can change it to "versionId" without any issue.

lib/algos/list/delimiterVersions.js
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Extended delimiter class for versioning.
+ */
+class DelimiterVersions extends Delimiter {
+    /**
+     *  Constructor of the extension
+     *  Init and check parameters
+     *  @param {Object} parameters - parameters sent to DBD
+     *  @param {RequestLogger} logger - werelogs request logger
+     *  @param {object} latestVersions - latest versions of some keys
```
LaurenSpiegel: where does this come from?  what determines this?

ttvi: From metadata, as an extension cannot do listing by itself and listing is required to know the latest versions of a collection of objects (listing the master keys).

ThibaultRiviere:
:+1:

tests/unit/algos/list/delimiterMaster.js
```diff
@@ -0,0 +1,179 @@
+'use strict'; // eslint-disable-line strict
+
+const assert = require('assert');
+const DelimiterMaster =
+    require('../../../../lib/algos/list/delimiterMaster').DelimiterMaster;
```
michael-zapata: You could always require the index file instead, that would allow you to test from a API consumer perspective.

lib/algos/list/delimiterMaster.js
```diff
@@ -0,0 +1,67 @@
...
+                EventualStorageBucket: value.eventualStorageBucket,
+                partLocations: value.partLocations,
+                creationDate: value.creationDate,
+            },
+        });
+        this.NextMarker = key;
+        ++this.keys;
+    }
+
+    filter(obj) {
```
michael-zapata: I'd welcome a bit of documentation here.

tests/unit/algos/list/delimiterMaster.js
ttvi: The unit tests failed as I changed it to `require('../../../../index').algorithms.list.DelimiterMaster`. Can you shed some light here?

lib/algos/list/delimiterMaster.js
ttvi: This is overriding the function of the base class. Some comment for the next commit should not be troublesome anyway.

tests/unit/algos/list/delimiterMaster.js
```diff
@@ -2,12 +2,12 @@
 
 const assert = require('assert');
 const DelimiterMaster =
-    require('../../../../lib/algos/list/delimiterMaster').DelimiterMaster;
+    require('../../../../index').algorithms.list.DelimiterMaster;
```
ttvi: Modifying these lines doesn't make the delimiter unit tests fail but the policy unit tests instead; see [here](http://ci.ironmann.io/gh/scality/Arsenal/829).

michael-zapata: I found the issue, it actually exposed a bug that'd happen to the API consumer, because of mixed imports.

lib/versioning/utils.js
```diff
@@ -0,0 +1,36 @@
...
+const VERLEN = VSConst.VERSIONING_AFFIXES.OTHER_VERSION_PREFIX.length;
+const SEP = VSConst.VERSIONING_AFFIXES.SEPARATOR;
+
+module.exports.VersioningUtils = {
+    // using explicit encode and decode functions for version object to enable
+    // future extensions, such as attatching external information to the string
+    // value from request instead of having to parse and stringify the value
+    // Example: "prefixlen_{prefix}{value_from_request}"
+    encodeVersion: version => JSON.stringify(version),
+    decodeVersion: version => JSON.parse(version),
```
LaurenSpiegel: wrap in try/catch here?

lib/versioning/utils.js
```diff
@@ -0,0 +1,36 @@
...
+    // using these functions to manipulate the attributes of the value also
+    // help us from restricting ourselves to hardcoding arbitrary attributes
+    setattr: (version, attr, value) => version[attr] = value,
```
LaurenSpiegel: this is to set any attribute on a specific version?  Should we have a similar getter?

lib/versioning/utils.js
LaurenSpiegel: I think setVersionId and getVersionId would be clearer.

lib/versioning/utils.js
```diff
@@ -0,0 +1,36 @@
+const VSConst = require('./constants').VersioningConstants;
+
+const MSTLEN = VSConst.VERSIONING_AFFIXES.MASTER_VERSION_PREFIX.length;
```
LaurenSpiegel: masterVersionPrefixLength

lib/versioning/utils.js
```diff
@@ -0,0 +1,36 @@
...
+const VERLEN = VSConst.VERSIONING_AFFIXES.OTHER_VERSION_PREFIX.length;
```
LaurenSpiegel: otherVersionPrefixLength

lib/versioning/utils.js
ttvi: Will update. But personally I think crashing should be the intended result, rather than covering it and messing with the errors later. After all, it should be an internal error if there is anything wrong with the format here, in such case we should restart rather than trying to make more errors.

ttvi:
updated

lib/versioning/utils.js
```diff
@@ -0,0 +1,63 @@
...
+    // value from request instead of having to parse and stringify the value
+    // Example: "prefixlen_{prefix}{value_from_request}"
+    encodeVersion: version => JSON.stringify(version),
+    decodeVersion: version => {
+        try {
+            return JSON.parse(version);
+        } catch (err) {
+            // using undefined here because JSON.parse("null") === null
+            return undefined;
+            // what to do with the error
```
LaurenSpiegel: I agree that we can't just do nothing about the error.  If this is being used in such a way that throwing makes more sense, then leave that.  Or we can do a safeJsonParse function like the following that logs an error and returns an Internal Error back to the client.

```javascript
export function safeJsonParse(request, toParse) {
    try {
        return JSON.parse(toParse);
    } catch (error) {
        log.error('failed to parse json');
        return errors.InternalError;
    }
}
```

ttvi:
updated

LaurenSpiegel:
👍

ttvi:
@michael-zapata @rahulreddy (whoever in charge) please merge.

rahulreddy:
@ttvi Sorry I missed this.

ttvi:
Need the branch for demo.

PR #120, opened 2016-07-27
Remove reassignment of path.

LaurenSpiegel:
If this function is used for a request that has a path, it won't work since the function is currently changing the path to '/'


LaurenSpiegel:
End to end passed http://ci.ironmann.io/gh/scality/Integration/2277

ploki:
:+1:

rahulreddy:
👍

alexandre-merle:
:+1:

PR #123, opened 2016-07-28
Fix/check headers

LaurenSpiegel:
adds a check to make sure all signed headers are included in list of signed headers


ploki:
:+1:

LaurenSpiegel:
End to end passed: http://ci.ironmann.io/gh/scality/Integration/2283

PR #135, opened 2016-08-08
Handles mapped ipv4 request ip addresses.

### LINKED ISSUE #133: Policy evaluation: IPV4 mapped as IPV6 addresses need to be converted
Since nodejs server listen's to ipv4 and ipv6 addresses on the same socket, `net.Socket.remoteAddress` gives us an ipv4 address in ipv6 format like `::ffff:127.0.0.1`. This needs to be converted to ipv4 if the policy evaluation condition is specified in ipv4. 
https://github.com/whitequark/ipaddr.js seems to be a good module to accomplish that replacing the current `node-ip` module.

LaurenSpiegel:
This closes #133.

### DIFF
--- a/lib/policyEvaluator/utils/conditions.js
+++ b/lib/policyEvaluator/utils/conditions.js
@@ -188,13 +188,17 @@ function convertToEpochTime(time) {
 
 
 /**
- * checkIPinRange checks whether a given ip address is in an ip address
- * range
- * @param {string} item - ip address range
+ * checkIPinRangeOrMatch checks whether a given ip address is in an ip address
+ * range or matches the given ip address
+ * @param {string} item - ip address range or ip address
  * @param {object} key - parsed ip address
  * @return {boolean} true if in range, false if not
  */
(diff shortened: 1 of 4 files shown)


lib/policyEvaluator/utils/conditions.js
```diff
@@ -190,15 +190,27 @@ function convertToEpochTime(time) {
...
+        // not ipv4 so try ipv6
+            range = ipAddr.IPv6.parseCIDR(item);
+        } catch (err) {
+            // range is not valid ipv4 or ipv6
+            return false;
+        }
+    }
     try {
-        inRange = ip.cidrSubnet(item).contains(key);
+        inRange = key.match(range);
```
alexandre-merle: just return ```key.match(range)``` here :) that way you don't need the inRange variable :)

alexandre-merle:
Minor comment, LGTM :+1:

nicolas2bert:
👍

PR #143, opened 2016-08-15
FIX negation operator without request context

nicolas2bert:
This fix allows user to make an S3request not related to the negation condition operator.

### DIFF
--- a/lib/policyEvaluator/evaluator.js
+++ b/lib/policyEvaluator/evaluator.js
@@ -136,14 +136,14 @@ function meetConditions(requestContext, statementCondition, log) {
             // Handle IfExists and negation operators
             const operatorsWithNegation = ['StringNotEquals',
             'StringNotEqualsIgnoreCase', 'StringNotLike', 'ArnNotEquals',
-            'ArnNotLike'];
+            'ArnNotLike', 'NumericNotEquals'];
             const isNegationOperator =
                 operatorsWithNegation.indexOf(bareOperator) > -1;
             if ((keyBasedOnRequestContext === undefined ||
                 keyBasedOnRequestContext === null) &&
                 (hasIfExistsCondition || isNegationOperator)) {
                 log.trace('satisfies condition due to IfExists operator or ' +
-                'negation operator');
+                'negation operator', { method: 'evaluators.evaluatePolicy' });
                 continue;
             }
             // If no IfExists qualifier, the key does not exist and the
--- a/lib/policyEvaluator/evaluator.js
+++ b/lib/policyEvaluator/evaluator.js
@@ -133,11 +133,17 @@ function meetConditions(requestContext, statementCondition, log) {
             // reference_policies_multi-value-conditions.html)
             const keyBasedOnRequestContext =
                 findConditionKey(key, requestContext);
-            // Handle IfExists
+            // Handle IfExists and negation operators
+            const operatorsWithNegation = ['StringNotEquals',
+            'StringNotEqualsIgnoreCase', 'StringNotLike', 'ArnNotEquals',
+            'ArnNotLike'];
+            const isNegationOperator =
+                operatorsWithNegation.indexOf(bareOperator) > -1;
             if ((keyBasedOnRequestContext === undefined ||
                 keyBasedOnRequestContext === null) &&
-                hasIfExistsCondition) {
-                log.trace('satisfies condition due to IfExists operator');
+                (hasIfExistsCondition || isNegationOperator)) {
+                log.trace('satisfies condition due to IfExists operator or ' +
+                'negation operator');
                 continue;
             }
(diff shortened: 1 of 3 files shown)


LaurenSpiegel:
Looks good.  Please add a couple unit tests and check whether the number not equals operators work the same way.

nicolas2bert:
Just checked, it is working the same 👍

nicolas2bert:
@LaurenSpiegel just added some unit tests

tests/unit/policyEvaluator.js
```diff
@@ -332,6 +332,24 @@ describe('policyEvaluator', () => {
                         check(requestContext, rcModifiers, policy, 'Allow');
                     });
 
+            it('should allow access for StringNotLike condition with ' +
+            'variables and wildcards: without any prefix in the query',
```
LaurenSpiegel: it should allow access for StringNotLike condition if condition parameter is completely missing from request

tests/unit/policyEvaluator.js
```diff
@@ -387,6 +405,20 @@ describe('policyEvaluator', () => {
                     check(requestContext, rcModifiers, policy, 'Allow');
                 });
 
+            it('should allow access for StringNotEquals condition if ' +
```
LaurenSpiegel: use similar description as above

tests/unit/policyEvaluator.js
```diff
@@ -443,6 +475,15 @@ describe('policyEvaluator', () => {
                     };
                     check(requestContext, rcModifiers, policy, 'Allow');
                 });
+            it('should allow access for StringNotEqualsIgnoreCase condition ' +
```
LaurenSpiegel: same

LaurenSpiegel:
what about the numericNotEquals operators?

nicolas2bert:
Updated it

LaurenSpiegel:
what about DateNotEquals, BinaryNotEquals and NotIpAddress?

LaurenSpiegel:
👍

lib/policyEvaluator/evaluator.js
```diff
@@ -133,11 +133,17 @@ function meetConditions(requestContext, statementCondition, log) {
...
+            'ArnNotLike', 'NumericNotEquals'];
+            const isNegationOperator =
+                operatorsWithNegation.indexOf(bareOperator) > -1;
             if ((keyBasedOnRequestContext === undefined ||
                 keyBasedOnRequestContext === null) &&
-                hasIfExistsCondition) {
-                log.trace('satisfies condition due to IfExists operator');
+                (hasIfExistsCondition || isNegationOperator)) {
+                log.trace('satisfies condition due to IfExists operator or ' +
+                'negation operator');
```
rahulreddy: one last comment, can you add the method attribute to the log? In this case, method should be
`evaluators.evaluatePolicy`

tests/unit/policyEvaluator.js
```diff
@@ -332,6 +332,24 @@ describe('policyEvaluator', () => {
                         check(requestContext, rcModifiers, policy, 'Allow');
                     });
 
+            it('should allow access for StringNotLike condition if condition' +
+            ' parameter is completely missing from request',
+                    () => {
```
michael-zapata: Small nit: indent feels off.

lib/policyEvaluator/evaluator.js
```diff
@@ -133,11 +133,17 @@ function meetConditions(requestContext, statementCondition, log) {
             // reference_policies_multi-value-conditions.html)
             const keyBasedOnRequestContext =
                 findConditionKey(key, requestContext);
-            // Handle IfExists
+            // Handle IfExists and negation operators
+            const operatorsWithNegation = ['StringNotEquals',
```
michael-zapata: Is there a list of available operators somewhere? That would be great to have. Also, why construct the array every time instead of having it available globally once and for all?

tests/unit/policyEvaluator.js
nicolas2bert: 👍

lib/policyEvaluator/evaluator.js
nicolas2bert: More functional are coming in Integration.
I did not find one, so I run locally my functional test against Amazon S3 to see how these different operators behaved.
I moved both array outside the two loops.

michael-zapata: They are still constructed every time we call the function, that's better than every iteration, but it's still an allocation we'll do on every function call.

nicolas2bert: I changed it.

DavidPineauScality:
Merge on behalf of @rahulreddy at the request of @nicolas2bert

PR #146, opened 2016-08-26
Change v2 query auth window to 1 hour 1 second

LaurenSpiegel:
This is a sort of forward port of https://github.com/scality/S3/pull/169
Since auth moved to arsenal after rel/1.1, the changes need to be made here and the tests added to s3 separately.

Once this is merged here and into master, we can forward port the s3 tests in scality/S3#169 from rel/1.1 into s3 rel/6.2.0-beta5 and master.

### DIFF
--- a/lib/auth/v2/queryAuthCheck.js
+++ b/lib/auth/v2/queryAuthCheck.js
@@ -4,7 +4,6 @@ const errors = require('../../errors');
 
 const algoCheck = require('./algoCheck');
 const constructStringToSign = require('./constructStringToSign');
-const checkRequestExpiry = require('./checkRequestExpiry');
 
 const queryAuthCheck = {};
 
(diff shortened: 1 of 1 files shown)


rahulreddy:
👍

DavidPineauScality:
Since no test is validating this specifically (and I understand the complexity of testing a one-hour timeout...), there is a high risk of regression when I'll finally be able to come back with the auth refactor. Any way to mitigate this risk ?

LaurenSpiegel:
@DavidPineauScality these will fail if the time is shortened -- https://github.com/scality/S3/pull/169/files.  I also have a test there that tests an expires (with a shorter timeframe). I need this merged before doing the forward port in s3.

DavidPineauScality:
I'm guessing the not-anymore-imported file is still used for V4 (otherwise please remove it).

:+1: on the change itself, as it looks relatively straight-forward :)

PR #150, opened 2016-08-30
Fix compat for query auth

LaurenSpiegel:
This fixes https://github.com/scality/Arsenal/issues/151.

I will open a PR in S3 with tests tomorrow.

### DIFF
--- a/lib/auth/v2/constructStringToSign.js
+++ b/lib/auth/v2/constructStringToSign.js
@@ -35,9 +35,10 @@ function constructStringToSign(request, data, log) {
     s3cmd includes x-amz-date in amzHeaders rather
     than here in stringToSign so we have replicated that.
     */
-    const date = headers.date ? headers.date : query.Expires;
+    const date = query.Expires ? query.Expires : headers.date;
+    const combinedQueryHeaders = Object.assign(headers, query);
     stringToSign += (date ? `${date}\n` : '\n')
-        + getCanonicalizedAmzHeaders(headers)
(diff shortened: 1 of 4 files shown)


DavidPineauScality:
Can't this be tested in Arsenal first at least ? Tests should always (to the best possible) accompany the code itself (unit tests), before testing its integration (integration tests).

lib/auth/v2/constructStringToSign.js
```diff
@@ -35,9 +35,10 @@ function constructStringToSign(request, data, log) {
     s3cmd includes x-amz-date in amzHeaders rather
     than here in stringToSign so we have replicated that.
     */
-    const date = headers.date ? headers.date : query.Expires;
+    const date = query.Expires ? query.Expires : headers.date;
+    const combinedQueryHeaders = Object.assign(headers, query);
```
alexandre-merle: for performance, it's may be better to use a for loop :)

alexandre-merle:
One small comment, otherwise LGTM :+1:

LaurenSpiegel:
didn't kill vault: http://ci.ironmann.io/gh/scality/Integration/2836
s3 still works too: http://ci.ironmann.io/gh/scality/S3/804

lib/auth/v2/constructStringToSign.js
rahulreddy: Object.assign isn't bad by default, it just sucks if the object we are cloning is nested. Since `query` is just a single-level object, it does no more than forEach on the properties.

rahulreddy:
👍

PR #153, opened 2016-09-06
Ft/utapi auth

LaurenSpiegel:
These changes are needed to enable authorization checks for utapi requests.

### DIFF
--- a/lib/policyEvaluator/RequestContext.js
+++ b/lib/policyEvaluator/RequestContext.js
@@ -35,17 +35,32 @@ function _findAction(service, method) {
     if (service === 's3') {
         return _actionMap[method];
     }
+    if (service === 'utapi') {
+        // currently only method is ListMetrics
+        return `utapi:${method}`;
+    }
     return undefined;
 }
 
-function _buildArn(service, bucket, object) {
+function _buildArn(service, generalResource, specificResource) {
+      // arn:partition:service:region:account-id:resourcetype/resource
     if (service === 's3') {
-        if (bucket && object) {
-            return `arn:aws:s3:::${bucket}/${object}`;
-        } else if (bucket) {
-            return `arn:aws:s3:::${bucket}`;
+        // arn:aws:s3:::bucket/object
+        // General resource is bucketName
+        if (generalResource && specificResource) {
+            return `arn:aws:s3:::${generalResource}/${specificResource}`;
+        } else if (generalResource) {
+            return `arn:aws:s3:::${generalResource}`;
         }
         return 'arn:aws:s3:::';
+    }
+    if (service === 'utapi') {
+        // arn:scality:utapi:::resourcetype/resource
+        // (possible resource types are buckets, accounts or users)
+        if (specificResource) {
+            return `arn:scality:utapi:::${generalResource}/${specificResource}`;
+        }
+        return `arn:scality:utapi:::${generalResource}/`;
     }
     return undefined;
 }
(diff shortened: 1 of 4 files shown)


rahulreddy:
👍

nicolas2bert:
👍

alexandre-merle:
LGTM :+1:

AntoninCoulibaly:
:+1:

lib/policyEvaluator/RequestContext.js
```diff
@@ -35,27 +35,44 @@ function _findAction(service, method) {
     if (service === 's3') {
         return _actionMap[method];
     }
+    if (service === 'utapi') {
```
michael-zapata: `else if` would make more sense. Or a `switch` statement.

lib/policyEvaluator/RequestContext.js
```diff
@@ -35,27 +35,44 @@ function _findAction(service, method) {
...
+        // arn:aws:s3:::bucket/object
+        // General resource is bucketName
+        if (generalResource && specificResource) {
+            return `arn:aws:s3:::${generalResource}/${specificResource}`;
+        } else if (generalResource) {
+            return `arn:aws:s3:::${generalResource}`;
         }
         return 'arn:aws:s3:::';
     }
+    if (service === 'utapi') {
```
michael-zapata: Same here regarding `else`.

lib/policyEvaluator/RequestContext.js
```diff
@@ -35,27 +35,44 @@ function _findAction(service, method) {
...
+        // arn:scality:utapi:::resourcetype/resource
+        // (possible resource types are buckets, accounts or users)
+        if (specificResource) {
+            return `arn:scality:utapi:::${generalResource}/${specificResource}`;
```
michael-zapata: This is plain duplication if we look a few lines above. Can we get rid of that?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "arsenal",
-  "version": "1.1.0",
+  "version": "1.1.0-utapi",
```
michael-zapata: Is this really wanted?

tests/unit/policyEvaluator.js
```diff
@@ -1112,6 +1112,24 @@ describe('policyEvaluator', () => {
     });
 });
 
+describe('policyEvaluator for utapi', () => {
```
michael-zapata: Only two tests for utapi?

lib/policyEvaluator/RequestContext.js
LaurenSpiegel: This will be expanding at some point so I want to be clear which service is being used.  When we get beyond 3 items, we can organize with an object or map structure.

lib/policyEvaluator/RequestContext.js
LaurenSpiegel: same

lib/policyEvaluator/RequestContext.js
LaurenSpiegel: I want to keep the different services separate at least for now.

package.json
LaurenSpiegel: it's in a drop me commit to make sure integration was picking up the right version.

tests/unit/policyEvaluator.js
LaurenSpiegel: the wildcards, variables, conditions and operators are not different for utapi so no need to replicate the prior tests.

PR #158, opened 2016-09-13
Enable batch policy evaluation

LaurenSpiegel:
To facilitate:
https://github.com/scality/S3/pull/204
https://github.com/scality/Vault/pull/592

### DIFF
--- a/lib/auth/auth.js
+++ b/lib/auth/auth.js
@@ -62,26 +62,23 @@ auth.check = (request, log, awsService, data) => {
     return { err: null, data: authInfo };
 };
 
-auth.doAuth = (request, log, cb, awsService, requestContext) => {
+auth.doAuth = (request, log, cb, awsService, requestContexts) => {
     const res = auth.check(request, log, awsService, request.query);
-    requestContext.setAuthType(res.authType);
-    requestContext.setSignatureVersion(res.signatureVersion);
-    requestContext.setSignatureAge(res.signatureAge);
+    requestContexts.forEach(requestContext => {
+        requestContext.setAuthType(res.authType);
+        requestContext.setSignatureVersion(res.signatureVersion);
+        requestContext.setSignatureAge(res.signatureAge);
+    });
     if (res.err) {
         return cb(res.err);
     } else if (res.version === 2) {
         return vault.authenticateV2Request(res.data.accessKey,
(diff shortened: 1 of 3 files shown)


tests/unit/auth/v2/publicAccess.js
```diff
@@ -22,10 +22,10 @@ describe('Public Access', () => {
             url: '/bucket',
             query: {},
         };
-        const requestContext = new RequestContext(request.headers,
+        const requestContext = [new RequestContext(request.headers,
```
michael-zapata: Since you have tests for single policies, why not add a few tests for batches?

lib/auth/auth.js
```diff
@@ -62,26 +62,23 @@ auth.check = (request, log, awsService, data) => {
     return { err: null, data: authInfo };
 };
 
-auth.doAuth = (request, log, cb, awsService, requestContext) => {
+auth.doAuth = (request, log, cb, awsService, requestContexts) => {
     const res = auth.check(request, log, awsService, request.query);
-    requestContext.setAuthType(res.authType);
-    requestContext.setSignatureVersion(res.signatureVersion);
-    requestContext.setSignatureAge(res.signatureAge);
+    requestContexts.forEach((requestContext) => {
```
michael-zapata: Please avoid adding lint warnings.

lib/auth/auth.js
```diff
@@ -62,26 +62,23 @@ auth.check = (request, log, awsService, data) => {
...
         return cb(res.err);
     } else if (res.version === 2) {
         return vault.authenticateV2Request(res.data.accessKey,
             res.data.signatureFromRequest,
-            res.data.stringToSign, res.data.algo, requestContext, log,
-            (err, authInfo) => {
-                if (err) {
-                    return cb(err);
-                }
-                return cb(null, authInfo);
```
michael-zapata: Removing this means you will send back `undefined` instead of `null`, could you ensure there's no breakage anywhere due to this subtle change? (S3 tests come to mind).

lib/auth/auth.js
LaurenSpiegel: fixed

lib/auth/auth.js
LaurenSpiegel: tests still pass.

PR #161, opened 2016-09-22
Remove warnings and set warnings as hard errors

GuillaumeGomez:


lib/shuffle.js
```diff
@@ -113,9 +114,11 @@ function randomRange(min, max) {
 module.exports = function shuffle(array) {
```
michael-zapata: Since we return the array, can we not mutate the original one? (That'll require an end-to-end though, so that can wait for later)

GuillaumeGomez: This isn't the point of this PR so I'll open another one to do this.

GuillaumeGomez:
cc @rahulreddy

Issue #163, opened 2016-09-22
Setting empty signed header value to undefined (v4 header auth)


LaurenSpiegel:
Since an empty string is interpreted as false, just checking whether the header value exists results in missing a signed header that has an empty string as its value.

rahulreddy:
Fixed by #164

PR #164, opened 2016-09-22
FIX: Handle empty signed headers

LaurenSpiegel:
Since an empty string is interpreted as false, just checking whether the header value existed was resulting in missing a signed header that had an empty string as its value.

Fixes https://github.com/scality/Arsenal/issues/163


LaurenSpiegel:
@ironman-machine please

michael-zapata:
Shouldn't this be also backported in `rel/1.1`? cc @vrancurel

LaurenSpiegel:
This issue was not introduced into rel/1.1.

PR #165, opened 2016-09-26
fix(listing): improve edge cases

ghost:


rahulreddy:
@michael-zapata Can you make the fix against rel/6.2? You can open it for reviews if you think the PR is ready.

michael-zapata:
As the tag is implying, it is still very much WIP, the edge-cases we're trying to solve are still not working as intended.

rahulreddy:
@michael-zapata Thanks

michael-zapata:
Blocked by #183.

michael-zapata:
Retargeting `master as this is expected for the `6.3` release. Some more work will be needed in Metadata to get it completely right. But I'd welcome @ploki's insight on the tests that I am adding to see if this is the intended behaviour.

michael-zapata:
E2E succeeded: http://ci.ironmann.io/gh/scality/Integration/6485#build-parameters

S3 will need updating with some tests to assert more strictly during E2E, since current tests are not assessing the functionality as well as arsenal's unit testing.

lib/algos/list/delimiter.js
```diff
@@ -1,52 +1,69 @@
...
-     *  Constructor of the extension
-     *  Init and check parameters
-     *  @param {Object} parameters - parameters sent to DBD
-     *  @param {RequestLogger} logger - werelogs request logger
-     *  @return {undefined}
+     * Create a new Delimiter instance
+     * @constructor
+     * @param {Object} parameters           - listing parameters
+     * @param {String} parameters.delimiter - delimiter per amazon format
+     * @param {String} parameters.prefix    - prefix per amazon format
```
LaurenSpiegel: missing parameters.gt

lib/algos/list/delimiter.js
```diff
@@ -70,67 +88,72 @@ class Delimiter {
...
-            // In cases of maxKeys <= 0 => IsTruncated = false
+            // In cases of maxKeys <= 0 -> IsTruncated = false
             this.IsTruncated = this.maxKeys > 0;
             return false;
         }
-        const key = obj.key;
-        const value = obj.value;
-        if (this.prefix && !key.startsWith(this.prefix)) {
+        if (this.prefix && !key.startsWith(this.prefix)
+                && key.startsWith(this.NextMarker)) {
```
LaurenSpiegel: what if no nextMarker?

lib/algos/list/delimiter.js
```diff
@@ -1,52 +1,70 @@
...
+function getCommonPrefix(key, delimiter, delimiterIndex) {
+    return key.substring(0, delimiterIndex + delimiter.length);
+}
+
+/**
+ * Handle object listing with parameters
+ *
+ * @prop {String[]} CommonPrefixes     - 'folders' defined by the delimiter
+ * @prop {String[]} Contents           - 'files' to list
+ * @prop {Boolean} IsTruncated         - marker per amazon format
```
LaurenSpiegel: description is wrong

lib/algos/list/delimiter.js
michael-zapata: That evaluates to false then, which means we go through the other checks as usual.

LaurenSpiegel: what if there was no delimiter?  just because you have the proper prefix and your key starts with the key of the prior key does not mean you should skip adding this key to the contents.

michael-zapata: Which highlights even more the lack of testing outside of the happy path then. Changing to `key === this.NextMarker` should do the trick.

rahulreddy:
Since we don't want to do this in a rush, this PR will not be merged for the release. We want to fully test against AWS, add any needed tests in S3 (so that they can run against AWS), add @ploki 's tests and then merge this PR accompanying those changes.

ThibaultRiviere (approved):

ploki:
@ploki's tests will be added in S3. Hence, I see no objection to merge this branch

LaurenSpiegel (approved):

rahulreddy (approved):
@ploki Can you squash the commits and rebase the branch? We can run end-to-end after that?

PR #169, opened 2016-10-07
FIX: werelogs in dependencies of http server

alexandre-merle:
Werelogs is actually require by the http server
of arsenal, it should not, since its not a prod
dependencies.
This PR fix that

### DIFF
--- a/lib/network/http/server.js
+++ b/lib/network/http/server.js
@@ -3,7 +3,6 @@
 const http = require('http');
 const https = require('https');
 const assert = require('assert');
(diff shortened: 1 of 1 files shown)


alexandre-merle:
fix validation running at http://ci.ironmann.io/gh/scality/Integration/4983

rahulreddy:
I was on the fence of running an end to end while merging and did not do that. I should have done that.

rahulreddy:
Fix validated at http://ci.ironmann.io/gh/scality/Integration/4987

PR #172, opened 2016-10-12
Compat: Check for invalid Date header

nicolas2bert:
Fixes https://github.com/scality/Arsenal/issues/179

When empty/invalid/unreadable/none Date header : return AccessDenied error
**On [ceph/s3-tests](https://github.com/ceph/s3-tests), these changes fix**:

_on object with V2 auth:_
- `FAIL: s3tests.functional.test_headers.test_object_create_bad_date_empty_aws2`
- `FAIL: s3tests.functional.test_headers.test_object_create_bad_date_invalid_aws2`
- `FAIL: s3tests.functional.test_headers.test_object_create_bad_date_unreadable_aws2`
- `FAIL: s3tests.functional.test_headers.test_object_create_bad_date_none_aws2`

_on bucket with V2 auth:_
- `FAIL: s3tests.functional.test_headers.test_bucket_create_bad_date_empty_aws2`
- `FAIL: s3tests.functional.test_headers.test_bucket_create_bad_date_invalid_aws2`
- `FAIL: s3tests.functional.test_headers.test_bucket_create_bad_date_unreadable_aws2`
- `FAIL: s3tests.functional.test_headers.test_bucket_create_bad_date_none_aws2`

_on object with V4 auth:_
- `FAIL: s3tests.functional.test_headers.test_object_create_bad_amz_date_empty_aws4`
- `FAIL: s3tests.functional.test_headers.test_object_create_bad_amz_date_invalid_aws4`
- `FAIL: s3tests.functional.test_headers.test_object_create_bad_amz_date_unreadable_aws4`
- `FAIL: s3tests.functional.test_headers.test_object_create_bad_amz_date_none_aws4`

_on bucket with V4 auth:_
- `FAIL: s3tests.functional.test_headers.test_bucket_create_bad_amz_date_empty_aws4`
- `FAIL: s3tests.functional.test_headers.test_bucket_create_bad_amz_date_invalid_aws4`
- `FAIL: s3tests.functional.test_headers.test_bucket_create_bad_amz_date_unreadable_aws4`
- `FAIL: s3tests.functional.test_headers.test_bucket_create_bad_amz_date_none_aws4`

### DIFF
--- a/lib/auth/v2/headerAuthCheck.js
+++ b/lib/auth/v2/headerAuthCheck.js
@@ -14,8 +14,11 @@ function check(request, log, data) {
         headers['x-amz-date'] : headers.date;
     timestamp = Date.parse(timestamp);
     if (!timestamp) {
-        log.debug('missing security header: invalid date/timestamp');
-        return { err: errors.MissingSecurityHeader };
+        log.debug('missing or invalid date header',
+        { method: 'auth/v2/headerAuthCheck.check' });
+        return { err: errors.AccessDenied.
+          customizeDescription('Authentication requires a valid Date or ' +
+          'x-amz-date header') };
     }
 
     const timeout = checkRequestExpiry(timestamp, log);
--- a/lib/auth/v4/headerAuthCheck.js
+++ b/lib/auth/v4/headerAuthCheck.js
@@ -38,7 +38,7 @@ function check(request, log, data, awsService) {
         log.trace('requesting streaming v4 auth');
         if (request.method !== 'PUT') {
             log.debug('streaming v4 auth for put only',
-                { method: 'v4HeaderAuthCheck' });
+                { method: 'auth/v4/headerAuthCheck.check' });
             return { err: errors.InvalidArgument };
         }
         if (!request.headers['x-amz-decoded-content-length']) {
@@ -70,8 +70,11 @@ function check(request, log, data, awsService) {
         timestamp = convertUTCtoISO8601(request.headers.date);
     }
     if (!timestamp) {
-        log.debug('missing date header');
-        return { err: errors.MissingSecurityHeader };
+        log.debug('missing or invalid date header',
+          { method: 'auth/v4/headerAuthCheck.check' });
+        return { err: errors.AccessDenied.
+          customizeDescription('Authentication requires a valid Date or ' +
+          'x-amz-date header') };
     }
 
     if (!validateCredentials(credentialsArr, timestamp, log)) {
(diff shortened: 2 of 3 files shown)


lib/auth/v2/headerAuthCheck.js
```diff
@@ -15,7 +15,7 @@ function check(request, log, data) {
     timestamp = Date.parse(timestamp);
     if (!timestamp) {
         log.debug('missing security header: invalid date/timestamp');
```
michael-zapata: This debug is not aptly named then.

lib/auth/v4/headerAuthCheck.js
```diff
@@ -71,7 +71,7 @@ function check(request, log, data, awsService) {
     }
     if (!timestamp) {
         log.debug('missing date header');
```
michael-zapata: Same here.

lib/auth/v2/headerAuthCheck.js
```diff
@@ -15,7 +15,7 @@ function check(request, log, data) {
...
-        return { err: errors.MissingSecurityHeader };
+        return { err: errors.AccessDenied };
```
rahulreddy: I would recommend adding a custom error description accompanying the error something like
```js
errors.customizeDescription('Missing Date header');
```

MathieuCassagne: I do not know if this error can happen often, but when using ```customizeDescription()``` from ArsenalError we're creating a new ArsenalError and not using the same object like the "natives" ArsenalErrors. Can this lead to some loss of performances here ?

rahulreddy: I don't think there is a loss of performance by just creating a new Error instance, but the request has already errored out at this point, so we don't have to worry about it. And we can't re-use the available instance to customize the error as it will update the error description in global scope.
Even if this doesn't happen often it's going to help that one time `AccessDenied` is sent to the client and the error is too vague to interpret what actually happened.

nicolas2bert:
@michael-zapata 
Here is the way I am doing it:
Let s start with header-tests compability,
1. Enumerate all the header-tests which fail on S3server and pass on AWS
2. For each test which fails : fix it and look if other tests status has changed (failed or passed) due to this fix.
3. If no other ceph tests has failed (and was passing before) because of my changes : Enumerate ceph test that has been fixed
4. If other ceph tests have failed, try to figure out what is wrong until no ceph tests fail.

lib/auth/v2/headerAuthCheck.js
michael-zapata: I'm sure Amazon has a special format for that, we might as well use that, right?

nicolas2bert: Amazon return: `AWS authentication requires a valid Date or x-amz-date header`

PR #174, opened 2016-10-13
ft: utility to check if IP matches a cidr range

rahulreddy:
This PR includes the following commits

- refactor: move ipcheck methods for reuse
- ft: utility to check if IP matches a cidr range

### DIFF
--- a/lib/ipCheck.js
+++ b/lib/ipCheck.js
@@ -1,0 +1,59 @@
+'use strict'; // eslint-disable-line strict
+
+const ipaddr = require('ipaddr.js');
+
+/**
+ * checkIPinRangeOrMatch checks whether a given ip address is in an ip address
+ * range or matches the given ip address
+ * @param {string} cidr - ip address range or ip address
+ * @param {object} ip - parsed ip address
+ * @return {boolean} true if in range, false if not
+ */
+function checkIPinRangeOrMatch(cidr, ip) {
+    // If there is an exact match of the ip address, no need to check ranges
+    if (ip.toString() === cidr) {
+        return true;
+    }
+    let range;
+
+    try {
+        range = ipaddr.IPv4.parseCIDR(cidr);
+    } catch (err) {
+        try {
+        // not ipv4 so try ipv6
+            range = ipaddr.IPv6.parseCIDR(cidr);
+        } catch (err) {
+            // range is not valid ipv4 or ipv6
+            return false;
+        }
+    }
+    try {
+        return ip.match(range);
+    } catch (err) {
+        return false;
+    }
+}
+
+/**
+* Parse IP address into object representation
+* @param {string} ip - IPV4/IPV6/IPV4-mapped IPV6 address
+* @return {object} parsedIp - Object representation of parsed IP
+*/
+function parseIp(ip) {
+    if (ipaddr.IPv4.isValid(ip)) {
+        return ipaddr.parse(ip);
+    }
+    if (ipaddr.IPv6.isValid(ip)) {
+        // .process pares v6 as v6 and parses mapped v4 addresses as IPv4
+        //  (unmapped)
+        return ipaddr.process(ip);
+    }
+    // not valid ip address according to module, so return empty object
+    // which will obviously not match a given condition range
+    return {};
+}
+
+module.exports = {
+    checkIPinRangeOrMatch,
+    parseIp,
+};
--- a/lib/ipCheck.js
+++ b/lib/ipCheck.js
@@ -44,16 +44,40 @@ function parseIp(ip) {
         return ipaddr.parse(ip);
     }
     if (ipaddr.IPv6.isValid(ip)) {
-        // .process pares v6 as v6 and parses mapped v4 addresses as IPv4
-        //  (unmapped)
+        // also parses IPv6 mapped IPv4 addresses into IPv4 representation
         return ipaddr.process(ip);
     }
     // not valid ip address according to module, so return empty object
-    // which will obviously not match a given condition range
+    // which will obviously not match a range of ip addresses that the parsedIp
+    // is being tested against
     return {};
+}
+
+
+/**
+* Checks if an IP adress matches a given list of CIDR ranges
+* @param {string[]} cidrList - List of CIDR ranges
+* @param {string} ip - IP address
+* @return {boolean} - true if there is match or false for no match
+*/
+function ipMatchCidrList(cidrList, ip) {
+    const parsedIp = parseIp(ip);
+    return cidrList.some(item => {
+        let cidr;
+        // patch the cidr if range is not specified
+        if (item.indexOf('/') === -1) {
+            if (item.startsWith('127.')) {
+                cidr = `${item}/8`;
+            } else if (ipaddr.IPv4.isValid(item)) {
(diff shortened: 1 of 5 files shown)


tests/unit/ipCheck.js
```diff
@@ -0,0 +1,76 @@
...
+        () => parseValidIpCheck('192.168.1.1', ipaddr.IPv4));
+
+    it('should parse IPv6 address',
+        () => parseValidIpCheck('2001:cdba::3257:9652', ipaddr.IPv6));
+
+    it('should parse IPv4 mapped IPv6 address',
+        // ::ffff:c0a8:101 mapped for 192.168.1.1
+        () => parseValidIpCheck('::ffff:c0a8:101', ipaddr.IPv4));
+
+    ['192.168', 'INVALID', '', null, undefined].forEach(item => {
```
RachedBenMustapha: While not RFC'd (I think), `192.168` will parse correctly about everywhere to `192.0.0.168` (description [here](https://linux.die.net/man/3/inet_aton)).

RachedBenMustapha: While not RFC'd (I think), 192.168 will parse correctly about everywhere to 192.0.0.168 (description here).

To have something invalid, I would suggest 260.384.2.1

lib/ipCheck.js
```diff
@@ -0,0 +1,84 @@
...
+function ipMatchCidrList(cidrList, ip) {
+    const parsedIp = parseIp(ip);
+    return cidrList.some(item => {
+        let cidr = item;
+        // patch the cidr if range is not specified only if cidr is one if
+        // a) loopback b) IPv4 and not 10.x.x.x
+        if (cidr.indexOf('/') === -1) {
+            if (cidr.startsWith('127.')) {
+                cidr += '/8';
+            } else if (!cidr.startsWith('10.') && ipaddr.IPv4.isValid(cidr)) {
```
RachedBenMustapha: RFC1918 ranges do not have `/24` masks - could you just patch with `/32` to not make it a range? It's pretty dangerous to turn, for example, `1.2.3.4` into the `1.2.3.X` range.

tests/unit/ipCheck.js
rahulreddy: Interesting, that's good to know. I made the change to 260.384.2.1

rahulreddy:
Yup, I just fixed the reference.

lib/ipCheck.js
```diff
@@ -0,0 +1,83 @@
+'use strict'; // eslint-disable-line strict
+
+const ipaddr = require('ipaddr.js');
+
+/**
+ * checkIPinRangeOrMatch checks whether a given ip address is in an ip address
+ * range or matches the given ip address
+ * @param {string} cidr - ip address range or ip address
+ * @param {object} ip - parsed ip address
```
michael-zapata: What exactly is the format of this object? Any properties to document?

lib/ipCheck.js
```diff
@@ -0,0 +1,83 @@
...
+/**
+* Checks if an IP adress matches a given list of CIDR ranges
+* @param {string[]} cidrList - List of CIDR ranges
+* @param {string} ip - IP address
+* @return {boolean} - true if there is match or false for no match
+*/
+function ipMatchCidrList(cidrList, ip) {
+    const parsedIp = parseIp(ip);
+    return cidrList.some(item => {
+        let cidr = item;
```
michael-zapata: Why not use `item` directly then use a template string to avoid additional copies/refs?

lib/ipCheck.js
```diff
@@ -0,0 +1,83 @@
...
+* Parse IP address into object representation
+* @param {string} ip - IPV4/IPV6/IPV4-mapped IPV6 address
+* @return {object} parsedIp - Object representation of parsed IP
+*/
+function parseIp(ip) {
+    if (ipaddr.IPv4.isValid(ip)) {
+        return ipaddr.parse(ip);
+    }
+    if (ipaddr.IPv6.isValid(ip)) {
+        // .process pares v6 as v6 and parses mapped v4 addresses as IPv4
```
michael-zapata: I think you meant `parses v6`. Also, I don't really get the comment, is this changing the internal representation of the object?

lib/ipCheck.js
michael-zapata: What I meant was:
```es6
    const parsedIp = parseIp(ip);
    return cidrList.some(item => {
        let cidr;
        // patch the cidr if range is not specified only if cidr is one if
        // a) loopback b) IPv4 and not 10.x.x.x
        if (item.indexOf('/') === -1) {
            if (item.startsWith('127.')) {
                cidr = `${item}/8`;
            } else if (!item.startsWith('10.') && ipaddr.IPv4.isValid(item)) {
                cidr = `${item}/24`;
            }
        }
        return checkIPinRangeOrMatch(cidr || item, parsedIp);
});
```

tests/unit/ipCheck.js
```diff
@@ -0,0 +1,77 @@
...
+    });
+});
+
+describe('Check IP matches a list of CIDR ranges', () => {
+    it('should match IP in a valid range',
+        () => cidrListMatchCheck(['192.168.1.0/24', '192.168.100.14/24',
+            '2001:db8::'], '192.168.100.1', true));
+
+    [
+        [['127.0.0.1'], '127.0.0.2'],
```
michael-zapata: Isn't there a parse error there? I'm surprised this isn't picked up by the CI.

rahulreddy: There is no parse error. It's an array of arguments, first argument is an array, second one is a string.

rahulreddy: The tests didn't run after that as there was a `.only` on the last test. I removed the `.only`, I don't see any parse error.

lib/ipCheck.js
```diff
@@ -0,0 +1,83 @@
...
+* Checks if an IP adress matches a given list of CIDR ranges
+* @param {string[]} cidrList - List of CIDR ranges
+* @param {string} ip - IP address
+* @return {boolean} - true if there is match or false for no match
+*/
+function ipMatchCidrList(cidrList, ip) {
+    const parsedIp = parseIp(ip);
+    return cidrList.some(item => {
+        let cidr;
+        // patch the cidr if range is not specified only if cidr is one if
```
LaurenSpiegel: patch the cidr if range is not specified **and** only if cidr is one **of**

lib/ipCheck.js
```diff
@@ -0,0 +1,83 @@
...
+function parseIp(ip) {
+    if (ipaddr.IPv4.isValid(ip)) {
+        return ipaddr.parse(ip);
+    }
+    if (ipaddr.IPv6.isValid(ip)) {
+        // also parses IPv6 mapped IPv4 addresses into IPv4 representation
+        return ipaddr.process(ip);
+    }
+    // not valid ip address according to module, so return empty object
+    // which will obviously not match a given condition range
```
LaurenSpiegel: which will obviously not match a range of ip addresses that the parsedIp is being tested against

(need to generalize the comment since no longer just for policy eval)

rahulreddy:
@ironman-machine ...

PR #184, opened 2016-10-21
Ft/batch delete

LaurenSpiegel:
This is needed for the multi-object delete API in S3.

### DIFF
--- a/lib/auth/auth.js
+++ b/lib/auth/auth.js
@@ -100,7 +100,9 @@ function extractParams(request, log, awsService, data) {
  * @param {object} log - Logger object
  * @param {function} cb - the callback
  * @param {string} awsService - Aws service related
- * @param {RequestContext[]} requestContexts - array of RequestContext
+ * @param {RequestContext[] | null} requestContexts - array of RequestContext
+ * or null if no requestContexts to be sent to Vault (for instance,
+ * in multi-object delete request)
  * @return {undefined}
  */
 function doAuth(request, log, cb, awsService, requestContexts) {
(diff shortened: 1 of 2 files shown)


lib/auth/auth.js
```diff
@@ -111,11 +113,14 @@ function doAuth(request, log, cb, awsService, requestContexts) {
         return cb(null, res.params);
     }
 
-    requestContexts.forEach(requestContext => {
-        requestContext.setAuthType(res.params.data.authType);
-        requestContext.setSignatureVersion(res.params.data.signatureVersion);
-        requestContext.setSignatureAge(res.params.data.signatureAge);
-    });
+    if (requestContexts) {
```
DavidPineauScality: So I take it that the batch delete only tries to authenticate, and not check that policies allow the user to delete ?
If not, could you clarify the cases in some sort of doc/comment ?

lib/auth/auth.js
```diff
@@ -100,7 +100,9 @@ function extractParams(request, log, awsService, data) {
  * @param {object} log - Logger object
  * @param {function} cb - the callback
  * @param {string} awsService - Aws service related
- * @param {RequestContext[]} requestContexts - array of RequestContext
+ * @param {RequestContext[] | null} requestContexts - array of
```
DavidPineauScality: Why allow a null value when we could simply provide an empty array and not modify the code ? I do not have the proper vision to understand the impact of this.

lib/auth/auth.js
LaurenSpiegel: For batch delete, we authenticate first without checking policies.  This means that we don't have to parse the xml sent with the request before authenticating.  Once we have authenticated successfully, we send a request to vault to just authorize against the policies if the request is from a user.

lib/auth/auth.js
LaurenSpiegel: Sending an empty array results in validation problems in vault since we are using the querystring to send authentication request info to vault.  
> querystring.stringify({rc: []})
''

Also, for chunk authentication requests in streaming v4 auth, we are also sending vault null for the request context.  So, by sending null here we limit the options vault has to deal with --either null or an array of requestContext instances (rather than null, an array of requestContext instances or an empty string key once querystring.parse does its magic).

LaurenSpiegel:
@ironman-machine good with this?

LaurenSpiegel:
end to end passed: http://ci.ironmann.io/gh/scality/Integration/5995

Issue #188, opened 2016-11-02
authv4 sort query params


LaurenSpiegel:
In creating the canonical request, we are not sorting query params that have a mix of capital and lowercase letters correctly.

LaurenSpiegel:
Fixed by #189

PR #189, opened 2016-11-02
BF: Sort query params for v4 auth.

LaurenSpiegel:
Before this change we were only taking into account
capital letters that appeared at the beginning of a query
param key.  With this change we take into account
capital letters throughout the key.

Fixes https://github.com/scality/Arsenal/issues/188


LaurenSpiegel:
@ironman-machine go for it

LaurenSpiegel:
end to end passed

PR #191, opened 2016-11-04
FIX: Signing key binary

alexandre-merle:
The binary digest for the signing key make
vaultclient under node v6 to generate wrong signature.
This pull request fix that.


alexandre-merle:
@ironman-machine let's go

alexandre-merle:
@michael-zapata the digest is not utf8 :)

michael-zapata:
Ok! In this case, I'm perfectly fine with no explicit digest form.

alexandre-merle:
@michael-zapata in fact, when not specified, the digest is return directly as a buffer :) which is what we want here :)

rahulreddy:
As this is a compatibility fix for node 4 -> node 6, we removed the bug label. So this doesn't warrant  a test.

PR #194, opened 2016-11-05
FT: Iam policies

alexandre-merle:
Implement iam policies


alexandre-merle:
@LaurenSpiegel yes, when i will implement policies for policies call :)

alexandre-merle:
@ironman-machine go alone

PR #195, opened 2016-11-09
BF: Fetches the correct value for `StorageClass`

bennettbuchanan:
* Related to https://github.com/scality/S3/pull/345


bennettbuchanan:
@ironman-machine

rahulreddy:
@ironman-machine lets run this before merging..

PR #196, opened 2016-11-22
FT: Add PutBucketWebsite to RequestContext map

electrachong:
Part of S3 Bucket Website API implementation


rahulreddy:
@ironman-machine

PR #197, opened 2016-11-29
COMPAT: Always return delimiter

bennettbuchanan:
Currently, when using the S3 file-backend, if a `Delimiter` parameter is
defined and it does not result in a `CommonPrefixes` array that is
greater than 0, `Delimiter` is set to ''. This causes the XML response
to omit the `Delimeter` key. For compatibility with AWS, this commit
fixes the `Delimiter` response to include any `Delimiter` string given 
as a parameter or return a `Delimiter` that is `undefined`.

See the associated fix in S3: https://github.com/scality/S3/pull/435

### DIFF
--- a/lib/algos/list/MPU.js
+++ b/lib/algos/list/MPU.js
@@ -115,7 +115,6 @@ class MultipartUploads {
      *  @return {Object} - The result.
      */
     result() {
-        const delimiter = this.CommonPrefixes.length > 0 ? this.delimiter : '';
         return {
             CommonPrefixes: this.CommonPrefixes,
             Uploads: this.Uploads,
@@ -123,7 +122,7 @@ class MultipartUploads {
             NextKeyMarker: this.NextKeyMarker,
             MaxKeys: this.maxKeys,
             NextUploadIdMarker: this.NextUploadIdMarker,
-            Delimiter: delimiter,
+            Delimiter: this.delimiter,
         };
     }
 }


bennettbuchanan:
@ironman-machine

DavidPineauScality:
Would end-to-end tests not validate that ? I'm unsure, but I agree that for such a change, we need to ensure nothing breaks for metadata.

bennettbuchanan:
Yes, last time a similarly small change was made (see https://github.com/scality/Arsenal/pull/195), it passed end-to-end but then caused Metadata tests to fail when it was merged.

DavidPineauScality:
After offline discussion with @ThibaultRiviere, What we need to do is the following:
 - One MD branch starting from rel/6.2.1 with updated dependency to this PR's branch
 - One MD branch starting from master with updated dependency to this PR's branch
Check unit/functional tests status for both

This needs to be done at least for any listing change in Arsenal.

bennettbuchanan:
Thanks @DavidPineauScality, I will create those two branches. I suppose I should open pull requests and run end-to-end tests with both. I will add tags that they are just for review.

rahulreddy:
@DavidPineauScality Curious as to why we want to test this with Metadata#rel/6.2.1 since this is targeting only master and we don't expect to bring this change to rel/6.2.1 as this is a compat change.

DavidPineauScality:
@rahulreddy As I said in the edit, that's because both branches rely on Arsenal#master due to breaking changes. As such, if we ever make a unit/functional test run in MD and it breaks the tests, this will be painful if noticed later on. This does not have any impact on Release builds and CI runs, since the dependency patcher alleviates this by enforcing the use of the non-breaking-change versions while not running the broken tests (unit and functional)

rahulreddy:
@bennettbuchanan Please reduce the commit message to 50 chars. and then we will need one more approval to get this merged.

bennettbuchanan:
This commit's change is passing unit/ft tests on both master and rel/6.2.1 branch in MetaData.

bennettbuchanan:
@ironman-machine 
SCALITY_METADATA_BRANCH=dev/COMPAT/Arsenal_rel/6.2.1_branch_compat_check

bennettbuchanan:
@ironman-machine 
SCALITY_METADATA_BRANCH=dev/COMPAT/Arsenal_master_branch_compat_check

bennettbuchanan:
@ironman-machine check MD master

bennettbuchanan:
@ironman-machine check MD rel/6.2.1 after POC commit
SCALITY_METADATA_BRANCH=rel/6.2.1

bennettbuchanan:
After converstation with @alexandre-merle, dropped POC commit since ft/unit tests were already passing in MD

PR #199, opened 2016-12-01
COMPAT: Date before epoch

### LINKED ISSUE #198: COMPAT: return AccessDenied if Date before epochTime (01/01/1970)
On **ceph/s3-tests**, these changes fix:

`FAIL: s3tests.functional.test_headers.test_object_create_bad_date_before_epoch_aws2`
`FAIL: s3tests.functional.test_headers.test_bucket_create_bad_date_before_epoch_aws2`

nicolas2bert:
FIXES #198

### DIFF
--- a/lib/auth/v2/checkRequestExpiry.js
+++ b/lib/auth/v2/checkRequestExpiry.js
@@ -1,24 +1,36 @@
 'use strict'; // eslint-disable-line strict
+const errors = require('../../errors');
+
+const epochTime = new Date('1970-01-01').getTime();
 
 function checkRequestExpiry(timestamp, log) {
+    // If timestamp is before epochTime, the request is invalid and return
+    // errors.AccessDenied
+    if (timestamp < epochTime) {
+        log.debug('request time is invalid', { timestamp });
+        return errors.AccessDenied;
+    }
     // If timestamp is not within 15 minutes of current time, or if
     // timestamp is more than 15 minutes in the future, the request
-    // has expired and return true
+    // has expired and return errors.RequestTimeTooSkewed
     const currentTime = Date.now();
     log.trace('request timestamp', { requestTimestamp: timestamp });
     log.trace('current timestamp', { currentTimestamp: currentTime });
+
     const fifteenMinutes = (15 * 60 * 1000);
     if (currentTime - timestamp > fifteenMinutes) {
         log.trace('request timestamp is not within 15 minutes of current time');
-        return true;
+        log.debug('request time too skewed', { timestamp });
+        return errors.RequestTimeTooSkewed;
     }
 
     if (currentTime + fifteenMinutes < timestamp) {
         log.trace('request timestamp is more than 15 minutes into future');
-        return true;
+        log.debug('request time too skewed', { timestamp });
+        return errors.RequestTimeTooSkewed;
     }
 
-    return false;
+    return undefined;
 }
 
 module.exports = checkRequestExpiry;
--- a/lib/auth/v2/headerAuthCheck.js
+++ b/lib/auth/v2/headerAuthCheck.js
@@ -21,10 +21,9 @@ function check(request, log, data) {
           'x-amz-date header') };
     }
 
-    const timeout = checkRequestExpiry(timestamp, log);
-    if (timeout) {
-        log.debug('request time too skewed', { timestamp });
-        return { err: errors.RequestTimeTooSkewed };
+    const err = checkRequestExpiry(timestamp, log);
+    if (err) {
+        return { err };
     }
     // Authorization Header should be
     // in the format of 'AWS AccessKey:Signature'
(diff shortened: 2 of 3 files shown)


lib/auth/v2/checkRequestExpiry.js
```diff
@@ -1,24 +1,36 @@
 'use strict'; // eslint-disable-line strict
+const errors = require('../../errors');
 
 function checkRequestExpiry(timestamp, log) {
+    const epochTime = new Date('1970-01-01').getTime();
+    // If timestamp is before epochTime, the request is invalid and return
+    // errors.AccessDenied
+    if (timestamp < epochTime) {
+        log.trace('request timestamp is before epochTime');
```
DavidPineauScality: we don't need two log entries here, one should suffice.

lib/auth/v2/checkRequestExpiry.js
```diff
@@ -1,24 +1,36 @@
...
 
     if (currentTime + fifteenMinutes < timestamp) {
         log.trace('request timestamp is more than 15 minutes into future');
-        return true;
+        log.debug('request time too skewed', { timestamp });
+        return errors.RequestTimeTooSkewed;
     }
 
-    return false;
+    return {};
```
DavidPineauScality: Should we not use "undefined" here, to imply that there's no error, rather than an empty object ?
this looks like counter-intuitive way, that may go against usual "error" management expectations, and lead to errors where the empty object is seen as an error that does not actually say anything,

lib/auth/v2/headerAuthCheck.js
```diff
@@ -21,10 +21,9 @@ function check(request, log, data) {
           'x-amz-date header') };
     }
 
-    const timeout = checkRequestExpiry(timestamp, log);
-    if (timeout) {
-        log.debug('request time too skewed', { timestamp });
-        return { err: errors.RequestTimeTooSkewed };
+    const timeoutResult = checkRequestExpiry(timestamp, log);
```
DavidPineauScality: following my previous comment, this could be changed into:
```
const err = checkRequestExpiry(timestamp, log);
if (err) {
    return { err } ; // Uses the shorthand syntax for the wrapping (not necessary)
}
```

tests/unit/auth/v2/checkRequestExpiry.js
```diff
@@ -5,25 +5,36 @@ const assert = require('assert');
...
+        assert.deepEqual(timeoutResult, errors.RequestTimeTooSkewed);
     });
 
-    it('should return false if the date in the header is ' +
+    it('should return no error if the date in the header is ' +
        'within 15 minutes of current time', () => {
         const timestamp = new Date();
-        assert(!checkRequestExpiry(timestamp, log));
+        const timeoutResult = checkRequestExpiry(timestamp, log);
+        assert(timeoutResult, {});
```
DavidPineauScality: to be changed if you follow my previous comments

lib/auth/v2/checkRequestExpiry.js
nicolas2bert: Got it! Thanks for the advice.

nicolas2bert:
@ironman-machine ok

lib/auth/v2/checkRequestExpiry.js
```diff
@@ -1,24 +1,35 @@
 'use strict'; // eslint-disable-line strict
+const errors = require('../../errors');
 
 function checkRequestExpiry(timestamp, log) {
+    const epochTime = new Date('1970-01-01').getTime();
```
michael-zapata: We can have this as a global instead, meaning we don't create it every time we go in this function.

tests/unit/auth/v2/checkRequestExpiry.js
```diff
@@ -5,25 +5,36 @@ const assert = require('assert');
...
+        const timeoutResult = checkRequestExpiry(timestamp, log);
+        assert.deepEqual(timeoutResult, errors.AccessDenied);
+    });
+
+    it('should return RequestTimeTooSkewed error if the date in the ' +
+       'header is more than 15 minutes old', () => {
         const timestamp = new Date(Date.now() - 16 * 60000);
-        assert(checkRequestExpiry(timestamp, log));
+        const timeoutResult = checkRequestExpiry(timestamp, log);
+        assert.deepEqual(timeoutResult, errors.RequestTimeTooSkewed);
```
michael-zapata: `deepStrictEqual()` would be better.

lib/auth/v2/checkRequestExpiry.js
nicolas2bert: Agree.

tests/unit/auth/v2/checkRequestExpiry.js
nicolas2bert: Why not.

nicolas2bert:
@ironman-machine ok2

nicolas2bert:
@ironman-machine ok3

nicolas2bert:
@ironman-machine ok4

PR #200, opened 2016-12-10
FT: Add GetBucketWebsite to RequestContext map

electrachong:
Part of S3 getBucketWebsite API implementation


electrachong:
@ironman-machine

PR #202, opened 2016-12-13
FT: Add DeleteBucketWebsite to RequestContext map

electrachong:
Part of S3 deleteBucketWebsite API implementation


electrachong:
thanks Dora
@ironman-machine

PR #205, opened 2016-12-16
Auth/translate headers

LaurenSpiegel:


LaurenSpiegel:
@ironman-machine go

LaurenSpiegel:
@ironman-machine check it now

PR #207, opened 2016-12-19
BF: Sort query params for v4 auth.

rahulreddy:
Before this change we were only taking into account
capital letters that appeared at the beginning of a query
param key.  With this change we take into account
capital letters throughout the key.


LaurenSpiegel:
why do no changes show up?

rahulreddy:
I think there was a conflict when the forward ports were being done and it was resolved not including this commit's changes. When I cherry-picked, it thinks the commit existed before and hence no changes are required. I am gonna try forward porting to 6.2.5 to 6.3 instead since I see the change in 6.2.5

PR #217, opened 2017-01-26
DEV: getBucketLocation

nicolas2bert:


rahulreddy:
End to end http://ci.ironmann.io/gh/scality/Integration/9584

PR #218, opened 2017-01-26
fix: auth v4 canonical request resource

### LINKED ISSUE #177: auth v4 canonical request
Query parameters for POST request should not infer in the payloadChecksum, like this :

```
POST /?Param1=value1 HTTP/1.1
Host:example.amazonaws.com
X-Amz-Date:20150830T123600Z
```

should generate :

```
POST
/
Param1=value1
host:example.amazonaws.com
x-amz-date:20150830T123600Z

host;x-amz-date
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
```

when we generate this payload `9095672bbd1f56dfc5b65f3e153adc8731a4a654192329106275f4c7b24d0b6e`

rahulreddy:
fixes #177

### DIFF
--- a/lib/auth/v4/constructStringToSign.js
+++ b/lib/auth/v4/constructStringToSign.js
@@ -1,6 +1,7 @@
 'use strict'; // eslint-disable-line strict
 
 const crypto = require('crypto');
+const url = require('url');
 
 const createCanonicalRequest = require('./createCanonicalRequest');
 
@@ -20,7 +21,7 @@ function constructStringToSign(params) {
 
     const canonicalReqResult = createCanonicalRequest({
         pHttpVerb: request.method,
-        pResource: request.path,
+        pResource: url.parse(request.path).pathname,
         pQuery: query,
         pHeaders: request.headers,
         pSignedHeaders: signedHeaders,


rahulreddy:
@ironman-machine try

rahulreddy:
We would want to test this with Go SDK to ensure we are not breaking compatibility and a test asserting the fix.

DavidPineauScality:
Is the Go SDK specific in that regard ?

DavidPineauScality:
My bad, got the explanation offline :)

rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine try

bennettbuchanan:
With https://github.com/scality/utapi/pull/92, these changes enable compatibility with the Go SDK.

rahulreddy:
Pasting from docs for reference

S3 V4 Auth
> CanonicalURI is the URI-encoded version of the absolute path component of the URI—everything starting with the "/" that follows the domain name and up to the end of the string or to the question mark character ('?') if you have query string parameters.

http://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html

General V4 Auth
> The canonical URI is the URI-encoded version of the absolute path component of the URI, which is everything in the URI from the HTTP host to the question mark character ("?") that begins the query string parameters (if any).

http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html

rahulreddy:
Favoring a fix in Utapi as UT is sending path when Arsenal method is requesting a pathname. I will make a PR to add a comment for reference.

PR #222, opened 2017-02-03
FT: Node v6

alexandre-merle:
switch to node v6

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,9 @@
 {
   "name": "arsenal",
-  "version": "1.1.0",
+  "engines": {
+    "node": "6.9.5"
+  },
+  "version": "7.0.0",
   "description": "Common utilities for the S3 project components",
   "main": "index.js",
   "repository": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -14,7 +17,8 @@
   },
   "homepage": "https://github.com/scality/Arsenal#readme",
   "dependencies": {
-    "ajv": "4.1.5",
+    "ajv": "^4.11.2",
+    "eslint-plugin-react": "^4.3.0",
```
rahulreddy: Why do we need to add this plugin?

alexandre-merle: in npm 3, the peer dependencies are not automatically installed anymore, so we need this plugin :)

package.json
```diff
@@ -14,7 +17,8 @@
   },
   "homepage": "https://github.com/scality/Arsenal#readme",
   "dependencies": {
-    "ajv": "4.1.5",
+    "ajv": "^4.11.2",
```
rahulreddy: Use specific version instead of `^`

alexandre-merle:
@rahulreddy updated, the other depedencies will be pinned when forward port of the 6.4 pull request

package.json
```diff
@@ -16,6 +19,7 @@
   "dependencies": {
     "ajv": "4.1.5",
     "ipaddr.js": "1.2.0",
+    "eslint-plugin-react": "^4.3.0",
```
rahulreddy: Isn't this a devDependency though?

alexandre-merle: right :)

rahulreddy:
@alexandre-merle good to go?

Issue #219, opened 2017-01-26
Policy evaluation, action should be case insensitive


alexandre-merle:
During policy evaluation, we do not manage lowercase action in policies : 
```
The name must match an action that is supported by the service. The prefix and the action name are case insensitive. For example, iam:ListAccessKeys is the same as IAM:listaccesskeys
```

PR #223, opened 2017-02-15
FIX: Case insensitive action

alexandre-merle:
Check with case insensitive the policy action

Fixes https://github.com/scality/Arsenal/issues/219

### DIFF
--- a/lib/policyEvaluator/evaluator.js
+++ b/lib/policyEvaluator/evaluator.js
@@ -77,7 +77,7 @@ function isActionApplicable(requestAction, statementAction, log) {
         // No variables in actions so no need to handle
         const regExStrOfStatementAction =
             handleWildcards(statementAction[i]);
-        const actualRegEx = new RegExp(regExStrOfStatementAction);
+        const actualRegEx = new RegExp(regExStrOfStatementAction, 'i');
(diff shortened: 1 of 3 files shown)


alexandre-merle:
@ironman-machine try

rahulreddy:
Has this behavior been tested against AWS IAM?

rahulreddy:
>You specify a value using a namespace that identifies a service (iam, ec2 sqs, sns, s3, etc.) followed by the name of the action to allow or deny. The name must match an action that is supported by the service. The prefix and the action name are case insensitive. For example, iam:ListAccessKeys is the same as IAM:listaccesskeys. The following examples show Action elements for different services.

Nevermind, I see it in the docs http://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements.html#Action

alexandre-merle:
@rahulreddy yes, i saw that in the doc too, that's why i made the pull request :)

alexandre-merle:
@ironman-machine try

alexandre-merle:
End to end here http://ci.ironmann.io/gh/scality/Integration/9605

DavidPineauScality:
Merging on behalf of the maintainer

PR #225, opened 2017-02-21
FT: Fix deps

alexandre-merle:
Fix dependencies


DavidPineauScality:
@ironman-machine

DavidPineauScality:
Merging, as E2E passed, and this is dependency version fixing only

PR #228, opened 2017-02-27
update with new versioning

ttvi:
What changed:
- update listing algorithms `delimiterMaster` and `delimiterVersions` to reflect new versioning in metadata (with metadata's functional tests to verify)
- remove unused constants
- change versioning tools to version class
- update unit tests


DavidPineauScality:
What  about S3 functional tests ? The listing algorithm is used for the file backend.

ttvi:
There should be no impact. The listing algorithms `delimiterMaster` and `delimiterVersions` are specific to versioning only.

ThibaultRiviere (approved):

lib/algos/list/Extension.js
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @param {object} entry - a listing entry from metadata
+     *                         expected format: { key, value }
+     * @return {number} - result of filtering the entry:
+     *                    > 0: entry is accepted and included in the result
+     *                    = 0: entry is accepted but not included (skipping)
+     *                    < 0: entry is not accepted, listing should finish
+     */
+    filter(entry) {
+        return entry ? 0 : 0;
```
MathieuCassagne: You always return 0 ?

lib/algos/list/MPU.js
```diff
@@ -3,6 +3,11 @@
 const checkLimit = require('./tools').checkLimit;
 const DEFAULT_MAX_KEYS = 1000;
 
+function inc(str) {
```
MathieuCassagne: This function is present in 3 files, can you make it into a separate file ?

lib/algos/list/basic.js
```diff
@@ -15,16 +17,34 @@ class List {
      *  @return {undefined}
      */
     constructor(parameters, logger) {
-        this.logger = logger;
+        super(parameters, logger);
         this.res = [];
         if (parameters) {
-            this.maxKeys = checkLimit(parameters.maxKeys, DEFAULT_MAX_KEYS);
+            const maxKeys = parameters.maxKeys;
```
MathieuCassagne: Why do  this in two steps ?

lib/algos/list/Extension.js
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @return {string} - the insight; a common prefix or a master key
+     */
+    skipping() {
+        return '';
+    }
+
+    /**
+     * Get the listing resutls. Format depends on derivatives' specific logic.
+     * @return {Array} - The listed elements
```
MathieuCassagne: this.res is always undefined.

lib/algos/list/basic.js
```diff
@@ -34,11 +54,11 @@ class List {
     filter(elem) {
```
MathieuCassagne: You return a Number now.

lib/algos/list/delimiter.js
```diff
@@ -94,30 +127,12 @@ class Delimiter {
      */
     addContents(key, value) {
```
MathieuCassagne: Same about the return.

lib/algos/list/basic.js
```diff
@@ -1,12 +1,14 @@
...
+const Extension = require('./Extension').default;
+
 const checkLimit = require('./tools').checkLimit;
 const DEFAULT_MAX_KEYS = 10000;
 
 /**
  *  Class of an extension doing the simple listing
  */
-class List {
+class List extends Extension {
```
rahulreddy: `extends` currently performs poor in pure ES6 implementation. You can choose to use the class directly instead or introduce base class in a PR later when performance regression is fixed in node 6.

rahulreddy (changes requested):

lib/algos/list/basic.js
DavidPineauScality: isn't that part of the "class" tests ? They report "same" as I see it, or I may not be looking at the right row ?

lib/versioning/VersionID.js
```diff
@@ -0,0 +1,136 @@
...
+ * in that millisecond, and the identifier of the local site (which could be
+ * datacenter, region, or server depending on the notion of geographics). This
+ * function is stateful which means it keeps some values in the memory and the
+ * next call depends on the previous call.
+ *
+ * @param {string} info - the additional info to ensure uniqueness if desired
+ * @return {string} - the formated versionId string
+ */
+function generateVersionId(info) {
+    // Need to wait for the millisecond slot got "flushed". We wait for
```
DavidPineauScality: I'm not sure to understand this bit. I get that you wait for one millisecond, but not why ?
Is it because you need to ensure that it starts fresh on a "new" millisecond, due to the inner workings  relying on millisecond + sequence number ? Are you relying on the fact that Node.JS computation is monothreaded + this mechanism, to make sure that any two instantiations of this class cannot start on the same millisecond ?
If yes, please fix doc to reflect this.

lib/versioning/VersionID.js
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/**
+ * Encrypt a versionId to add a validation layer before processing
+ * bucket versioning related requests.
+ *
+ * @param {string} str - the versionId to encrypt
+ * @return {string} - the encrypted versionId
+ */
+function encrypt(str) {
```
DavidPineauScality: the encrypt/decrypt part is not used for MD.
It should have gone to another PR, depending on this one.

No matter it's too late, but please keep this in mind, as such huge reviews are a hassle for everyone of us.

tests/unit/algos/list/delimiterMaster.js
```diff
@@ -1,179 +0,0 @@
-'use strict'; // eslint-disable-line strict
```
DavidPineauScality: Why have no tests for this one ?

ghost (changes requested):
Lots of good things in there, + a few questions about doc, explanations/clarifications, and design improvement suggestions

lib/algos/list/delimiter.js
```diff
@@ -94,30 +127,12 @@ class Delimiter {
...
-                },
-                StorageClass: tmp['x-amz-storage-class'],
-                Initiated: tmp.initiated,
-                Initiator: tmp.initiator,
-                EventualStorageBucket: tmp.eventualStorageBucket,
-                partLocations: tmp.partLocations,
-                creationDate: tmp.creationDate,
-            },
-        });
+        this.Contents.push({ key, value });
```
ttvi: To make S3 to do the computation instead of the leader node. S3 needs to do the parsing https://github.com/scality/S3/pull/651/commits/44157881f57ddcd55e0db5b4e0aab0b8bf7c247b#diff-d1fea59d8501a89ee1ce02464ecdcd16; no metadata impact.
To make them consistent across extensions (w.r.t Basic and Versions originally).

lib/algos/list/delimiterMaster.js
```diff
@@ -1,84 +1,110 @@
...
-     * @return {boolean} - continue filtering or return the formatted list
+     *  Filter to apply on each iteration, based on:
+     *  - prefix
+     *  - delimiter
+     *  - maxKeys
+     *  The marker is being handled directly by levelDB
+     *  @param {Object} obj       - The key and value of the element
+     *  @param {String} obj.key   - The key of the element
+     *  @param {String} obj.value - The value of the element
+     *  @return {Boolean}         - indicates if iteration should continue
```
ttvi: I only do the declaration in the base class (Extension.js). The derivatives don't need to describe again unless they do different things. Nevertheless I don't mind to specify it.

lib/algos/list/delimiterVersions.js
```diff
@@ -1,106 +1,175 @@
...
-            Contents: this.Contents,
-            // <versioning>
-            LatestVersions: this.latestVersions,
-            // </versioning>
+            Versions: this.Contents,
             IsTruncated: this.IsTruncated,
-            NextMarker: this.IsTruncated ? this.NextMarker : undefined,
-            NextVersionMarker: this.IsTruncated ?
-                this.NextVersionMarker : undefined,
+            NextKeyMarker: this.IsTruncated ? this.NextMarker : undefined,
```
ttvi: To match AWS specifications.

lib/versioning/Version.js
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /**
+     * [MetaData Internal] Check if a version is a place holder for deletion.
+     *
+     * @param {string} value - version to check
+     * @return {boolean} - whether this is a PHD version
+     */
+    static isPHD(value) {
+        // check the length of the value before parsing
```
ttvi: That's the description for the previous version. Will be removed.

lib/versioning/VersionID.js
ttvi: Epochs are at millisecond level. Wait one millisecond to make sure the first epoch when this module started is always the latest, regardless of the sequence number.

tests/unit/algos/list/delimiterMaster.js
ttvi: It has been merged with the tests for `delimiter`.

lib/versioning/VersionID.js
ttvi: Arsenal is in the middle to provide common knowledge to both S3 and MetaData. Since it's related to versioning, it belongs to this specific PR.

lib/algos/list/Extension.js
ttvi: This is the base class.

lib/algos/list/Extension.js
```diff
@@ -0,0 +1,72 @@
...
+        return {};
+    }
+
+    /**
+     * This function receives a data entry from metadata and decides if it will
+     * include the entry in the listing result or not.
+     *
+     * @param {object} entry - a listing entry from metadata
+     *                         expected format: { key, value }
+     * @return {number} - result of filtering the entry:
```
ttvi: I originally defined the results as positive and negative so that we can declare arbitrary values for specific instructions later. But declaring it now is also OK, why not.

lib/algos/list/basic.js
ttvi: Thanks. Updated.

lib/algos/list/delimiter.js
ttvi: Updated.

lib/versioning/Version.js
```diff
@@ -0,0 +1,168 @@
...
+        // all PHD versions should have the same length
+        if (value.indexOf('isPHD') < 0) {
+            return false;
+        }
+        // parse the value if it has the length of a PHD version
+        try {
+            return Version.from(value).isPHDVersion();
+        } catch (exception) { // eslint-disable-line strict
+            return false; // nice, Vault
```
ttvi: Vault is sending plain text data not in JSON format. Vault should never contain "isPHD" in the data it sends, but if it does, we have this covered.

lib/algos/list/basic.js
ttvi: Is this still an issue here considering the time nodejs spends to do "super" compared to the listing process itself? The number 4x is significant but in the context of listing I do not really see that as an issue.
Definitely that needs to be improved though. What is your suggestion here: removing `super` and duplicating the code or is there any cool approach you can think of?

rahulreddy: I don't think it's an issue. The benchmark compares to the ES5 implementation, so it's 4x slower than the ES5 implementation but fast enough for us since the code would spend nowhere near time number of times making as many super calls as the benchmark.

lib/algos/list/MPU.js
```diff
@@ -34,6 +40,28 @@ class MultipartUploads {
         this.logger = logger;
     }
 
+    genMDParams() {
+        const params = {};
+        if (this.params.keyMarker) {
+            params.gt = `overview${this.params.splitter}` +
```
ttvi: Let's keep it as-is.

lib/algos/list/basic.js
ttvi: Let's keep it as-is. @rahulreddy let me know if you want to open an issue for this.

lib/algos/list/Extension.js
ttvi: Updated.

lib/algos/list/Extension.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * Provides the insight into why filter is skipping an entry. This could be
+     * because it is skipping a range of delimited keys or a range of specific
+     * version when doing master version listing.
+     *
+     * @return {string} - the insight; a common prefix or a master key
+     */
+    skipping() {
```
ttvi: Checked.

lib/algos/list/delimiterMaster.js
```diff
@@ -1,84 +1,110 @@
...
+     * Delimiter listing of master versions.
+     * @param {Object} parameters           - listing parameters
+     * @param {String} parameters.delimiter - delimiter per amazon format
+     * @param {String} parameters.prefix    - prefix per amazon format
+     * @param {String} parameters.marker    - marker per amazon format
+     * @param {Number} parameters.maxKeys   - number of keys to list
+     */
+    constructor(parameters) {
+        super(parameters);
+        this.PHD = undefined;
```
ttvi: Renamed.

lib/algos/list/basic.js
```diff
@@ -25,20 +27,38 @@ class List {
         this.keys = 0;
     }
 
+    genMDParams() {
```
ThibaultRiviere: genMDParams will allocate an object and remove useless property.
But leveldb accept and ignore the undefined property.
This is not an issue but we should be able to avoid this step.

lib/algos/list/delimiterMaster.js
DavidPineauScality: well, it's fine if that's the reason, but the generable JSDOC won't show it outright. minor anyways

lib/algos/list/delimiterVersions.js
DavidPineauScality: Okay, gotcha

lib/versioning/Version.js
DavidPineauScality: ok, nice defensive coding then :)

tests/unit/algos/list/delimiterMaster.js
DavidPineauScality: oh ok

ghost (approved):

DavidPineauScality:
LGTM now, Good work o/

ThibaultRiviere (approved):

lib/versioning/VersionID.js
```diff
@@ -0,0 +1,136 @@
...
+const LENGTH_SQ = 6; // position in ms slot
+const LENGTH_ST = 5; // site identifier
+
+// empty string template for the variables in a versionId
+const TEMPLATE_TS = new Array(LENGTH_TS + 1).join('0');
+const TEMPLATE_SQ = new Array(LENGTH_SQ + 1).join('0');
+const TEMPLATE_ST = new Array(LENGTH_ST + 1).join(' ');
+
+// site identifier, like PARIS, TOKYO; will be trimmed if exceeding max length
+const SITE_ID = `${process.env.SITE_ID}${TEMPLATE_ST}`.slice(0, LENGTH_ST);
```
lamphamsy: Could `process.env.SITE_ID` be undefined? If yes, should it be `process.env.SITE_ID || ${a_default_name}`?

lib/versioning/VersionID.js
```diff
@@ -0,0 +1,136 @@
...
+ * @return {string} - the encrypted versionId
+ */
+function encrypt(str) {
+    const cipher = crypto.createCipher(
+            VID_CRYPTO_ALGORITHM, VID_CRYPTO_PASSWORD);
+    let encrypted = cipher.update(str,
+            VID_CRYPTO_ENCODING_TXT, VID_CRYPTO_ENCODING);
+    encrypted += cipher.final(VID_CRYPTO_ENCODING);
+    let length = encrypted.length;
+    while (length && encrypted.charAt(length - 1) === '=') {
```
lamphamsy: Why '=' must be removed?

lib/versioning/VersionID.js
ttvi: No. SITE_ID is a new required variable to be available to both S3 and MetaData https://github.com/scality/Federation/pull/937.

lib/versioning/VersionID.js
ttvi: To beautify the base64 versionId string without impacting the correctness.

ttvi:
Integration with versioning OK: http://ci.ironmann.io/gh/scality/Integration/11082.
Trying integration without versioning: http://ci.ironmann.io/gh/scality/Integration/11089.

ttvi:
Ah, I just realized that this has completely changed the format of listing so this branch of Arsenal cannot be tested against the non-versioning branch of the other repos. @LaurenSpiegel @rahulreddy please merge this all together with S3's https://github.com/scality/S3/pull/651 and MetaData's https://github.com/scality/MetaData/pull/1009.

lib/algos/list/basic.js
ttvi: Weird.

lib/versioning/VersionID.js
jonathan-gramain: A regexp could be used to make things slightly clearer than a loop IMO, say, this should be equivalent (considering that base64 reserves `=` only for the trailer):
```
encrypted = encrypted.replace(/=/g,'')
```

lib/versioning/VersionID.js
```diff
@@ -0,0 +1,136 @@
...
+
+// site identifier, like PARIS, TOKYO; will be trimmed if exceeding max length
+const SITE_ID = `${process.env.SITE_ID}${TEMPLATE_ST}`.slice(0, LENGTH_ST);
+
+// constants for max epoch and max sequential number in the same epoch
+const MAX_TS = Math.pow(10, LENGTH_TS) - 1; // good until 16 Nov 5138
+const MAX_SQ = Math.pow(10, LENGTH_SQ) - 1; // good for 1 billion ops
+
+// the earliest versionId, used for versions before versioning
+const VID_INF = `${TEMPLATE_TS}${MAX_TS}`.slice(-LENGTH_TS) +
```
jonathan-gramain: You could consider using helper functions to do the proper padding.

tests/unit/versioning/VersionID.js
```diff
@@ -0,0 +1,50 @@
+const VID = require('../../../lib/versioning/VersionID.js');
+const assert = require('assert');
+
+function randkey(length = 15) {
```
ttvi: This is the comment I like; it touches the right issues + having convincing reasoning for its suggestions.

ttvi: Ah, about the answer, it's there to represent the approximated length of a versionId string. I will move it to the caller.

lib/versioning/VersionID.js
ttvi: Yes it was not carefully crafted. Some other alternatives can be used like a simple `indexOf`, but regex seems concise and is a good solution.

lib/versioning/VersionID.js
```diff
@@ -0,0 +1,136 @@
...
+const TEMPLATE_TS = new Array(LENGTH_TS + 1).join('0');
+const TEMPLATE_SQ = new Array(LENGTH_SQ + 1).join('0');
+const TEMPLATE_ST = new Array(LENGTH_ST + 1).join(' ');
+
+// site identifier, like PARIS, TOKYO; will be trimmed if exceeding max length
+const SITE_ID = `${process.env.SITE_ID}${TEMPLATE_ST}`.slice(0, LENGTH_ST);
+
+// constants for max epoch and max sequential number in the same epoch
+const MAX_TS = Math.pow(10, LENGTH_TS) - 1; // good until 16 Nov 5138
+const MAX_SQ = Math.pow(10, LENGTH_SQ) - 1; // good for 1 billion ops
```
ttvi: Epochs are in milliseconds.

ttvi: The comment does not give enough context due to max-len restriction.

lib/versioning/Version.js
```diff
@@ -0,0 +1,168 @@
...
+        return new Version(value ? JSON.parse(value) : undefined);
+    }
+
+    /**
+     * [MetaData Internal] Check if a version is a place holder for deletion.
+     *
+     * @param {string} value - version to check
+     * @return {boolean} - whether this is a PHD version
+     */
+    static isPHD(value) {
```
ttvi: It is metadata internal as clearly state. I have a more detail description here https://github.com/scality/MetaData/pull/1009/files#diff-a6a06b67d14df0938c340ab3d479523aR15.

lib/algos/list/tools.js
```diff
@@ -1,3 +1,9 @@
+// constants for extensions
+const SKIP_NONE = undefined; // to be inline with the values of NextMarker
```
ttvi: ignored

lib/algos/list/delimiterVersions.js
```diff
@@ -1,106 +1,165 @@
...
+                return this.addCommonPrefix(key, delimiterIndex);
+            }
+        }
+        return this.addContents({ key, value: obj.value, versionId });
+    }
+
+    skipping() {
+        if (this.NextMarker) {
+            const index = this.NextMarker.lastIndexOf(this.delimiter);
+            if (index === this.NextMarker.length) {
```
ttvi: Why?

lib/algos/list/delimiterVersions.js
```diff
@@ -1,106 +1,165 @@
...
-            value,
-        });
-        this.NextMarker = objectName;
-        this.NextVersionMarker = versionId;
-        ++this.keys;
-        // only include the latest versions of the keys in the resulting list
-        // this is not actually used now, it's reserved for caching in future
-        if (this._latestVersions) {
-            this.latestVersions[objectName] = this._latestVersions[objectName];
+    addContents(obj) {
```
ttvi: True.

ttvi: But because this is a short function and we would need to check the returned value from the base implementation to add `this.NextVersionIdMarker = obj.versionId;` so I think it's clearer to do it this way while being as concise.

lib/algos/list/delimiterMaster.js
```diff
@@ -1,84 +1,112 @@
...
+    }
+
+    skipping() {
+        if (this.NextMarker) {
+            // next marker:
+            // - foo/ : skipping foo/
+            // - foo  : skipping foo.
+            const index = this.NextMarker.lastIndexOf(this.delimiter);
+            return this.NextMarker +
+                (index === this.NextMarker.length ? '' : VID_SEP);
```
ttvi: We can use that but then we need to switch the order of the results. The purpose is to use `VID_SEP` as a sub-delimiter in the case we need to skip a range of versions of the same object (`this.NextMarker` is not delimited by the main delimiter). I have the comments above stating that clearly.

lib/algos/list/delimiterMaster.js
```diff
@@ -1,84 +1,112 @@
...
-                key.indexOf(this.delimiter, this.searchStart);
-            if (commonPrefixIndex === -1) {
-                this.addContents(key, value);
-            } else {
-                this.addCommonPrefix(
-                    key.substring(0, commonPrefixIndex + this.delimLen));
+            // check if the key has the delimiter
+            const baseIndex = this.prefix ? this.prefix.length : 0;
+            const delimiterIndex = key.indexOf(this.delimiter, baseIndex);
+            if (delimiterIndex < 0) {
```
ttvi: True.

lib/algos/list/delimiterMaster.js
```diff
@@ -1,84 +1,112 @@
...
+     *  @param {String} obj.value - The value of the element
+     *  @return {number}         - indicates if iteration should continue
      */
     filter(obj) {
-        // Check first in case of maxkeys <= 0
-        if (this.keys >= this.maxKeys) {
-            // In cases of maxKeys <= 0 => IsTruncated = false
-            this.IsTruncated = this.maxKeys > 0;
-            return false;
+        let key = obj.key;
```
ttvi: No. the derivatives of `Delimiter` are filtering the super-set of what `Delimiter` does.

lib/algos/list/delimiterMaster.js
```diff
@@ -1,84 +1,112 @@
...
+
+    /**
+     *  Add a (key, value) tuple to the listing
+     *  Set the NextMarker to the current key
+     *  Increment the keys counter
+     *  @param {String} key   - The key to add
+     *  @param {String} value - The value of the key
+     *  @return {number}      - indicates if iteration should continue
      */
     addContents(key, value) {
```
ttvi: Seems to be true.

lib/algos/list/MPU.js
```diff
@@ -34,6 +35,28 @@ class MultipartUploads {
...
+                `${this.params.keyMarker}${this.params.splitter}`;
+            if (this.params.uploadIdMarker) {
+                params.gt += `${this.params.uploadIdMarker}`;
+            }
+            // advance so that lower bound does not include the supplied
+            // markers
+            params.gt = inc(params.gt);
```
ttvi: I've got them all from the old code https://github.com/scality/MetaData/pull/1009/files#diff-b596dc6bd6d51b111a52033729463bd1L189. Maybe @LaurenSpiegel knows what it does.

lib/algos/list/Extension.js
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @param {object} entry - a listing entry from metadata
+     *                         expected format: { key, value }
+     * @return {number} - result of filtering the entry:
+     *                    > 0: entry is accepted and included in the result
+     *                    = 0: entry is accepted but not included (skipping)
+     *                    < 0: entry is not accepted, listing should finish
+     */
+    filter(entry) {
+        return entry ? FILTER_SKIP : FILTER_SKIP;
```
ttvi: I agree this does not look nice. As this is an interface, I wanted to give some specific description about each function can so that we can get the general idea of the derivatives. About the double `FILTER_SKIP`s there, they exist for some purposes: first, to give an example on what to return (matching what described in JSDoc), and second, to avoid the linting warning of not using the input value `entry` (while I still want to keep the `entry` there).

lib/algos/list/MPU.js
```diff
@@ -34,6 +35,28 @@ class MultipartUploads {
         this.logger = logger;
     }
 
+    genMDParams() {
+        const params = {};
+        if (this.params.keyMarker) {
+            params.gt = `overview${this.params.splitter}` +
+                `${this.params.keyMarker}${this.params.splitter}`;
+            if (this.params.uploadIdMarker) {
+                params.gt += `${this.params.uploadIdMarker}`;
```
ttvi: I don't know if `this.params.uploadIdMarker` could be a number. As with the `param.gt`, I copied it from the old code.

lib/algos/list/Extension.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * Provides the insight into why filter is skipping an entry. This could be
+     * because it is skipping a range of delimited keys or a range of specific
+     * version when doing master version listing.
+     *
+     * @return {string} - the insight: a common prefix or a master key,
+     *                                 or undefined if there is no insight
```
ttvi: updated

lib/algos/list/MPU.js
```diff
@@ -81,14 +104,14 @@ class MultipartUploads {
...
-     *  @return {Boolean} - True: Continue, False: Stop
+     *  @return {number} - > 0: Continue, < 0: Stop
      */
     filter(obj) {
         // Check first in case of maxkeys = 0
         if (this.keys >= this.maxKeys) {
             // In cases of maxKeys <= 0 => IsTruncated = false
             this.IsTruncated = this.maxKeys > 0;
-            return false;
+            return -1;
```
ttvi: used

lib/algos/list/basic.js
```diff
@@ -25,20 +27,38 @@ class List {
...
+        });
+        return params;
+    }
+
     /**
      *  Function apply on each element
      *  Just add it to the array
      *  @param {Object} elem - The data from the database
-     *  @return {Boolean} - True = continue the stream
+     *  @return {Boolean} - > 0 : continue listing
```
ttvi: updated

lib/algos/list/delimiter.js
```diff
@@ -36,26 +39,29 @@ function getCommonPrefix(key, delimiter, delimiterIndex) {
...
      * Create a new Delimiter instance
      * @constructor
      * @param {Object} parameters           - listing parameters
      * @param {String} parameters.delimiter - delimiter per amazon format
-     * @param {String} parameters.start     - prefix per amazon format
-     * @param {String} [parameters.gt]      - NextMarker per amazon format
-     * @param {Number} [parameters.maxKeys] - number of keys to list
+     * @param {String} parameters.prefix    - prefix per amazon format
+     * @param {String} parameters.marker    - marker per amazon format
+     * @param {Number} parameters.maxKeys   - number of keys to list
```
ttvi: documented

lib/algos/list/delimiterMaster.js
ttvi: removed

lib/algos/list/delimiterMaster.js
ttvi: upgraded

lib/versioning/VersionID.js
ttvi: ok

lib/versioning/VersionID.js
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * @param {Number} span - time to wait in nanoseconds (1/1000000 millisecond)
+ * @return {Undefined} - nothing
+ */
+function wait(span) {
+    function getspan(diff) {
+        return diff[0] * 1e9 + diff[1];
+    }
+    const start = process.hrtime();
+    while (getspan(process.hrtime(start)) < span);
```
ttvi: good

tests/unit/versioning/VersionID.js
ttvi: moved

lib/algos/list/MPU.js
jonathan-gramain: I think the `+` operator on a string automatically converts its second argument into a string.

lib/algos/list/delimiterMaster.js
jonathan-gramain: More pragmatically, my understanding of `lastIndexOf` is that it returns either an index number in `0..length-1`, or -1 if not found, so here I don't think `index === this.NextMarker.length` can ever be true, which would mean that you always return `this.nextMarker + VID_SEP`.

lib/versioning/Version.js
jonathan-gramain: IMO, internal functions need to be self-documented as much as possible, especially because internal tends to have more intricate complexity, if not for the API user since they're internal, but for a programmer who may have to dig into it later. It's good that the meaning of PHD is explained somewhere, but I think it would be even better to have a more descriptive name for the function name itself as well.

lib/algos/list/delimiterVersions.js
jonathan-gramain: Fine for me

lib/algos/list/delimiterVersions.js
jonathan-gramain: see https://github.com/scality/Arsenal/pull/228#discussion_r109256720

lib/algos/list/delimiterMaster.js
ttvi: Yes. I've committed some crime on this PR.

lib/algos/list/delimiterVersions.js
ttvi: Yes. I've committed some crime on this PR.

PR #229, opened 2017-03-08
S3C-35 Implement a new communication channel to use functionality of level-sublevel remotely

jonathan-gramain:
It allows the client to do the normal levelDB operations (put, get,
delete, list keys via createReadStream) as well as creating new
sublevels. (Note that this is a purely virtual operation, nothing is
created initially but the client gets a handle to manipulate the new
sublevel).

It's built on top of socket.io which provides messaging abstraction
and helps maintaining a reliable channel (detects connection failures,
attempts to reconnect etc.). It also allows to use several namespaces
and "rooms" on the same communication channel, which may be useful
later.

There is an additional timeout for each operation, which triggers an
error after 5 seconds without an answer by default.

Unit tests added: ping, basic put and get, listing of keys and
rewrite, deletes, timeout test.

### DIFF
--- a/index.js
+++ b/index.js
@@ -40,5 +40,12 @@ module.exports = {
         http: {
             server: require('./lib/network/http/server'),
         },
+        level: require('./lib/network/level-net'),
+    },
+    storage: {
+        metadata: {
+            server: require('./lib/storage/metadata/file/server'),
+            client: require('./lib/storage/metadata/file/client'),
+        },
     },
 };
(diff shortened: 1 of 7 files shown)


jonathan-gramain:
Please hold-on on this PR for now.

I did a quick smoke test on performance, a simple ls of directories that needs to fetch around a thousand entries is about 3 times slower with this branch compared to master. Also, tcpdump shows about 50x (!) more packets exchanged every time I list the bucket, though when I look at the debug logs I don't see significantly more payload generated than the current implementation. There may be something wrong with the use of socket.io, maybe not, but I need to investigate a bit first, and possibly revert back to one of the previous approaches I tried before if there's no solution to this performance issue with socket.io.

jonathan-gramain:
I suspect this is mostly due to the ping-pong between client and server for every listing entry returned.

jonathan-gramain:
Will create a new PR with custom implementation on stream for performance later on, this one is obsolete.

PR #230, opened 2017-03-14
S3C-35 New communication channel for remote sublevel

jonathan-gramain:
It allows the client to do the normal levelDB operations (put, get,
delete, list keys via createReadStream) as well as creating new
sublevels. (Note that this is a purely virtual operation, nothing is
created initially but the client gets a handle to manipulate the new
sublevel).

It's built on top of socket.io which provides messaging abstraction
and helps maintaining a reliable channel (detects connection failures,
attempts to reconnect etc.). It also allows to use several namespaces
and "rooms" on the same communication channel, which may be useful
later.

There is an additional timeout for each operation, which triggers an
error after 5 seconds without an answer by default.

A custom object stream implementation has been added in order to list
keys more efficiently than the third-party "socket.io-stream" module
which does a round-trip for every entry. This implementation both
gathers written objects in a single packet, and can pipeline multiple
packets to the server while waiting for acks (max 5 by default, should
be configurable). This should help a lot when the remote latency is
high.

Note that levelDB createReadStream() becomes asynchronous, the user
must pass a callback and the stream is returned as a callback
argument.

Unit tests added:

 - ping
 - basic CRUD test
 - sublevels (separation of namespace, nesting)
 - listing of keys (to the end and aborted by the client) + parallel rewrites
 - parallel random reads
 - parallel deletes
 - command timeout

### DIFF
--- a/lib/network/level-net/index.js
+++ b/lib/network/level-net/index.js
@@ -1,0 +1,343 @@
+'use strict'; // eslint-disable-line
+
+const http = require('http');
+const io = require('socket.io');
+const ioClient = require('socket.io-client');
+const sioStream = require('./sio-stream');
+const async = require('async');
+const assert = require('assert');
+
+const flattenError = require('./utils').flattenError;
+const reconstructError = require('./utils').reconstructError;
+
+const DEFAULT_CALL_TIMEOUT_MS = 5000;
+
+class SocketIOConnection {
+
+    /**
+     * @constructor
+     * @param {String} url URL of the socket.io namespace,
+     *   e.g. 'http://localhost:9990/metadata'
+     * @param {Object} params constructor additional params
+     * @param {Logger} params.logger logger object
+     * @param {Number} [params.streamMaxPendingAck] max number of
+     *   in-flight output stream packets sent to the server without an ack
+     *   received yet
+     * @param {Number} [params.streamAckTimeoutMs] timeout for receiving
+     *   an ack after an output stream packet is sent to the server
+     */
+    constructor(url, params) {
+        assert(params.logger);
+        this.logger = params.logger;
+        this.socket = ioClient(url);
+        this.socketStreams = sioStream.createSocket(
+            this.socket,
+            params.logger,
+            params.streamMaxPendingAck,
+            params.streamAckTimeoutMs);
+        this.socket.on('error', err => {
+            this.logger.warn('connectivity error to storage daemon',
+                             { error: err });
+            return undefined;
+        });
+    }
+
+    /**
+     * @brief internal RPC implementation w/o timeout
+     *
+     * @param {String} remoteCall name of the remote function to call
+     * @param {Array} args list of arguments to the remote function
+     * @param {function} cb callback called when done
+     * @return {undefined}
+     */
+    _call(remoteCall, args, cb) {
+        const wrapCb = (err, data) => {
+            cb(reconstructError(err),
+               this.socketStreams.decodeStreams(data));
+        };
+        this.logger.debug('remote call', { remoteCall, args });
+        this.socket.emit('call', remoteCall,
+                         this.socketStreams.encodeStreams(args), wrapCb);
+        return undefined;
+    }
+
+    /**
+     * @brief call a remote function named @a remoteCall, with
+     * arguments @a args and callback @a cb
+     *
+     * @a cb is called when the remote function returns an ack, or
+     * when the timeout set by @a timeoutMs expires, whichever comes
+     * first. When an ack is received, the callback gets the arguments
+     * sent by the remote function in the ack response. In the case of
+     * timeout, it's passed a single Error argument with the code:
+     * 'ETIMEDOUT' property, and a self-described string in the 'info'
+     * property.
+     *
+     * @param {String} remoteCall name of the remote function to call
+     * @param {Array} args list of arguments to the remote function
+     * @param {function} cb callback called when done or timeout
+     * @param {Number} timeoutMs timeout in milliseconds
+     * @return {undefined}
+     */
+    callTimeout(remoteCall, args, cb, timeoutMs = DEFAULT_CALL_TIMEOUT_MS) {
+        if (typeof cb !== 'function') {
+            throw new Error(`argument cb=${cb} is not a callback`);
+        }
+        async.timeout(this._call.bind(this), timeoutMs,
+                      `operation ${remoteCall} timed out`)(remoteCall,
+                                                           args, cb);
+        return undefined;
+    }
+}
+
+const dbAsyncCommandNames = ['put', 'get', 'del', 'batch'];
+const dbSyncCommandNames = ['createReadStream'];
+
+const miscSyncCommands = {
+    ping: function ping() {
+        return 'pong';
+    },
+};
+
+const miscAsyncCommands = {
+    pingAsync: function pingAsync(cb) {
+        setImmediate(() => cb(null, 'pong'));
+    },
+    slowCommand: function slowCommand(cb) {
+        setTimeout(() => cb(null, 'ok'), 2000);
+    },
+};
+
+/**
+ * @brief get a LevelDB client object that proxies requests to a
+ * remote DB server through socket.io events
+ *
+ * NOTE: synchronous calls on the server-side API (i.e those which
+ * take no callback argument) become asynchronous on the client, take
+ * one additional parameter (the callback), then:
+ *
+ * - if it throws, the error is passed as callback's first argument,
+ *   otherwise null is passed
+ * - the return value is passed as callback's second argument (unless
+ *   an error occurred).
+ *
+ * The only DB call in this case currently is createReadStream().
+ *
+ * @param {Object} options options object
+ * @param {Logger} options.logger logger object
+ * @param {String} [options.baseNs=] custom namespace to use
+ * @param {Number} [options.callTimeoutMs] timeout for remote calls
+ * @param {Number} [options.streamMaxPendingAck] max number of
+ *   in-flight output stream packets sent to the server without an ack
+ *   received yet
+ * @param {Number} [options.streamAckTimeoutMs] timeout for receiving
+ *   an ack after an output stream packet is sent to the server
+ * @return {Object} a LevelDB proxy object that can be called with
+ * leveldb API functions directly, or sub-leveled with openSub()
+ */
+module.exports.client = function LevelClient(options) {
+    assert(options.logger);
+
+    const dbClient = {
+        path: [],
+        baseNs: options.baseNs || '',
+        logger: options.logger,
+        callTimeoutMs: options.callTimeoutMs,
+    };
+    assert(dbClient.baseNs === '' || dbClient.baseNs.startsWith('/'));
+
+    dbClient.getCallTimeout = function getCallTimeout() {
+        return this.callTimeoutMs;
+    };
+    dbClient.setCallTimeout = function setCallTimeout(newTimeoutMs) {
+        this.callTimeoutMs = newTimeoutMs;
+    };
+
+    dbClient.connect = function connect(host, port) {
+        this.mdClient = new SocketIOConnection(
+            `http://${host}:${port}${this.baseNs}/metadata`, options);
+        this.miscClient = new SocketIOConnection(
+            `http://${host}:${port}${this.baseNs}/misc`, options);
+    };
+    for (const remoteCall of dbSyncCommandNames.concat(dbAsyncCommandNames)) {
+        dbClient[remoteCall] = function onCall(...rpcArgs) {
+            const cb = rpcArgs.pop();
+            const args = { rpcArgs, subLevel: this.path };
+            this.mdClient.callTimeout(remoteCall, args, cb,
+                                      this.callTimeoutMs);
+        };
+    }
+    for (const remoteCall of Object.keys(miscSyncCommands).concat(
+        Object.keys(miscAsyncCommands))) {
+        dbClient[remoteCall] = function onCall(...rpcArgs) {
+            const cb = rpcArgs.pop();
+            const args = { rpcArgs };
+            this.miscClient.callTimeout(remoteCall, args, cb,
+                                        this.callTimeoutMs);
+        };
+    }
+
+    /**
+     * @brief return a handle to a sublevel database
+     *
+     * @note this function has no side-effect on the db, it just
+     * returns a handle properly configured to access the sublevel db
+     * from the client.
+     *
+     * @param {String} subName name of sublevel
+     * @return {Object} a handle to the sublevel database that has the
+     * same API as its parent
+     */
+    dbClient.openSub = function openSub(subName) {
+        const subLevel = {};
+        Object.assign(subLevel, this);
+        // maintain path as a list of nested sublevels
+        subLevel.path = subLevel.path.slice();
+        subLevel.path.push(subName);
+        return subLevel;
+    };
+
+    return dbClient;
+};
+
+/**
+ * @brief create a server object that serves remote LevelDB requests
+ * through socket.io events
+ *
+ * @param {Object} db The root LevelDB database object to expose to
+ * remote clients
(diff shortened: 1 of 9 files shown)


lib/network/level-net/utils.js
```diff
@@ -0,0 +1,30 @@
...
+        return err;
+    }
+    const serializedErr = Object.assign({}, err);
+
+    serializedErr.message = err.message;
+    for (const k in err) {
+        if (!(k in serializedErr)) {
+            serializedErr[k] = err[k];
+        }
+    }
```
LaurenSpiegel: why is this needed in addition to object assign?  also, is there a need to copy the err?

lib/storage/data/file/backend.js
```diff
@@ -0,0 +1,113 @@
+import { errors, stringHash } from 'arsenal';
```
LaurenSpiegel: seems this file should be moved over in a separate PR in which you create a dataFile server

lib/storage/metadata/file/client.js
```diff
@@ -0,0 +1,56 @@
...
+            .newRequestLogger();
+    }
+
+    /**
+     * Open the remote metadata database (backed by leveldb)
+     *
+     * @return {Object} handle to the remote database
+     */
+    openDB() {
+        this.client = levelNet.client({ baseNs: '/MDFile',
```
LaurenSpiegel: seems there should be a constants file that has '/MDFile' to be used by both client and server (and other constants).

lib/network/level-net/sio-stream.js
```diff
@@ -0,0 +1,372 @@
...
+            // eslint-disable-next-line
+            for (const k in arg) {
+                encodedObj[k] = this.encodeStreams(arg[k]);
+            }
+            return encodedObj;
+        }
+        return arg;
+    }
+
+    decodeStreams(arg) {
```
LaurenSpiegel: please add jsdoc

lib/network/level-net/sio-stream.js
```diff
@@ -0,0 +1,372 @@
...
+                });
+            }
+            if (isWriteStream) {
+                transportStream.pipe(arg);
+            }
+            return encodedStream;
+        }
+        if (typeof(arg) === 'object') {
+            let encodedObj;
+            if (Array.isArray(arg)) {
```
LaurenSpiegel: would this ever be an array?

lib/network/level-net/sio-stream.js
```diff
@@ -0,0 +1,372 @@
...
+                encodedObj = [];
+            } else {
+                encodedObj = {};
+            }
+            // user objects are simple flat objects and we want to
+            // copy all their properties
+            // eslint-disable-next-line
+            for (const k in arg) {
+                encodedObj[k] = this.encodeStreams(arg[k]);
```
LaurenSpiegel: note that a for in loop will not preserve order for an array.  is that needed?

lib/network/level-net/sio-stream.js
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    destroy() {
+        debug('destroy called', this.streamId);
+        this._destroyed = true;
+        this.pause();
+        this.removeAllListeners('data');
+        this.removeAllListeners('end');
+        this._readState = {
+            pushBuffer: [],
```
LaurenSpiegel: will this really clean out the array?

lib/storage/metadata/file/server.js
```diff
@@ -0,0 +1,67 @@
...
+    startServer() {
+        const rootDB = level(this.metadataPath + ROOT_DB);
+        this.db = sublevel(rootDB);
+        this.logger.info('starting metadata file backend server');
+        /* We start a server that will serve the sublevel capable
+           rootDB to clients */
+        const server = levelNet.createServer(
+            this.db, { streamMaxPendingAck: this.streamMaxPendingAck,
+                       streamAckTimeoutMs: this.streamAckTimeoutMs });
+        server.initMetadataService('/MDFile');
```
jonathan-gramain: Good idea, even define it in a new shared file (constants.js?) would make sense since the same string shall be used by client and server.

lib/network/level-net/index.js
```diff
@@ -0,0 +1,283 @@
...
+     * returns a handle properly configured to access the sublevel db
+     * from the client.
+     *
+     * @param {String} subName name of sublevel
+     * @return {Object} a handle to the sublevel database that has the
+     * same API as its parent
+     */
+    dbClient.openSub = function openSub(subName) {
+        const subLevel = {};
+        for (const prop in this) {
```
jonathan-gramain: ask the linter ;)

lib/network/level-net/index.js
```diff
@@ -0,0 +1,283 @@
...
+            if (!(prop in subLevel)) {
+                subLevel[prop] = this[prop];
+            }
```
jonathan-gramain: You're right

lib/network/level-net/index.js
```diff
@@ -0,0 +1,283 @@
...
+    const ioServer = io(httpServer);
+
+    ioServer.initMetadataService = function initMetadataService(baseNs = '') {
+        assert(baseNs === '' || baseNs.startsWith('/'));
+
+        const mdSock = this.of(`${baseNs}/metadata`);
+        mdSock.on('connection', conn => {
+            function openSub(path) {
+                let subDb = db;
+                path.forEach(pathItem => {
```
jonathan-gramain: yes

lib/network/level-net/index.js
```diff
@@ -0,0 +1,283 @@
...
+                    subDb = subDb.sublevel(pathItem);
+                });
+                return subDb;
```
jonathan-gramain: This does a recursive lookup of a sublevel, going through every subpath found in 'path' in order. That is why we return the latest sublevel (the deepest). I actually renamed the function lookupSubLevel().

lib/network/level-net/index.js
```diff
@@ -0,0 +1,283 @@
...
+            }
+
+            const mdStreams = sioStream(conn,
+                                        options.streamMaxPendingAck,
+                                        options.streamAckTimeoutMs);
+            conn.on('error', () => {});
```
jonathan-gramain: We could log something here though in the current code index does not have a werelogs logging context, and it shall be logged in an upper level. We have to have a listener for errors though otherwise it's going to throw.

jonathan-gramain: I guess to turn all debug() calls into log.debug it will be necessary to have a logging context :)

lib/network/level-net/index.js
```diff
@@ -0,0 +1,283 @@
...
+                if (dbAsyncCommandNames.includes(remoteCall)) {
+                    try {
+                        const path = args.pop();
+                        args.push((err, data) => {
+                            cb(serializeError(err),
+                               mdStreams.encodeStreams(data));
+                        });
+
+                        const subDb = openSub(path);
+                        subDb[remoteCall].apply(subDb, args);
```
jonathan-gramain: I don't think that subDb[remoteCall] may throw (I hope not, it ought to call the cb instead). I don't think that openSub throws either (via `db.sublevel()`) but I'm not 100% sure. Though if it does, I think there's something going seriously wrong so we could let the exception bubble up and crash the process. (so the try/catch may be removed here). I don't have a strong opinion on keeping or removing it here.

lib/network/level-net/sio-stream.js
jonathan-gramain: What do you mean more precisely, .e.g in case we keep a reference to _readState? Normally not it should be safe (and normally having an empty pushBuffer is not essential because it should not be used anymore at this stage, but it's cleaner to clean it I think)

LaurenSpiegel: please ignore.

lib/network/level-net/sio-stream.js
```diff
@@ -0,0 +1,372 @@
...
+
+        this.socket.on('stream-end', (payload, cb) => {
+            const { streamId } = payload;
+            debug('received stream-end event streamId', streamId);
+            const stream = this.remoteStreams[streamId];
+            if (!stream) {
+                debug('no remote stream registered with id', streamId);
+                return;
+            }
+            stream._onend();
```
jonathan-gramain: I reworked the stream event handling to make it simpler and more predictable (and correct, hopefully!)

lib/network/level-net/sio-stream.js
```diff
@@ -0,0 +1,372 @@
...
+                writable: isWriteStream,
+            };
+            debug('encoded stream:', encodedStream);
+            let transportStream;
+            if (isReadStream && isWriteStream) {
+                transportStream = new SIODuplexStream(this, streamId,
+                                                      this.maxPendingAck,
+                                                      this.ackTimeoutMs);
+            } else if (isReadStream) {
+                transportStream = new SIOOutputStream(this, streamId,
```
jonathan-gramain: I removed the duplex streams since a. they are not used and b. they are probably broken.

I think we can keep the support for both readable and writable streams though, as it only adds a few lines to encode and decode functions as the rest is fully symmetric (events can be emitted or listened to by either side the same way) and I can imagine use cases where writable streams on the client could be useful (possibly write gathering) otherwise I think they do no harm being unused (but you may disagree).

lib/network/level-net/sio-stream.js
jonathan-gramain: This is custom application parameters, so it may want to pass an array as a function's argument (e.g. the "future" db.batch() I think is using an array to pass the list of operations).

lib/network/level-net/sio-stream.js
jonathan-gramain: It will preserve the order, because the for loop will iterate over array "keys" 0, 1, ...

lib/network/level-net/utils.js
jonathan-gramain: Some properties are not copied by Object.assign (don't ask me why exactly, probably related to computed properties). Anyway, the loop is able to catch those properties, e.g. not doing it fails a few tests like in CRUD test when checking `err.notFound`.

jonathan-gramain: I think we want to copy a new error since it's going to be a separate serialized object.

lib/storage/data/file/backend.js
jonathan-gramain: Right, this is a mistake.

lib/network/level-net/sio-stream.js
jonathan-gramain: This page explains well the pros and cons of using `for..in` on arrays:
http://stackoverflow.com/questions/500504/why-is-using-for-in-with-array-iteration-a-bad-idea

It's not a matter of order in our case, because we're writing to the same "property name" on the output array, but it may possibly write other internal properties that are not part of the array's data. So overall it can be better to use a normal `for` loop iterating over array's length. (the other consequence is that all undefined elements will be iterated too, but that's fine for this use case I think).

lib/network/level-net/index.js
```diff
@@ -0,0 +1,289 @@
...
+     * @brief internal RPC implementation w/o timeout
+     *
+     * @param {String} remoteCall name of the remote function to call
+     * @param {Array} args list of arguments to the remote function
+     * @param {function} cb callback called when done
+     * @return {undefined}
+     */
+    _call(remoteCall, args, cb) {
+        if (typeof cb !== 'function') {
+            return cb(new Error(`argument cb=${cb} is not a callback`));
```
jonathan-gramain: But this is gonna throw! (just kidding, yes you're perfectly right, thanks for spotting it :))

lib/network/level-net/index.js
LaurenSpiegel: would you mind adding a comment or jsdoc for the function

lib/network/level-net/sio-stream.js
```diff
@@ -0,0 +1,372 @@
+'use strict'; // eslint-disable-line
+
+const uuid = require('uuid');
+const stream = require('stream');
+const debug = require('debug')('sio-stream');
```
LaurenSpiegel: not won over by werelogs? 😄

lib/network/level-net/utils.js
```diff
@@ -0,0 +1,30 @@
+'use strict'; // eslint-disable-line
+
+module.exports.serializeError = function serializeError(err) {
```
LaurenSpiegel: perhaps rename (flattenError?) and jsdocs

lib/network/level-net/utils.js
```diff
@@ -0,0 +1,30 @@
...
+    serializedErr.message = err.message;
+    for (const k in err) {
+        if (!(k in serializedErr)) {
+            serializedErr[k] = err[k];
+        }
+    }
+    return serializedErr;
+};
+
+module.exports.deserializeError = function deserializeError(err) {
```
LaurenSpiegel: perhaps constructErrObj?  please add jsdoc

jonathan-gramain: done

lib/network/level-net/utils.js
jonathan-gramain: done

lib/network/level-net/index.js
jonathan-gramain: done

lib/network/level-net/utils.js
```diff
@@ -0,0 +1,50 @@
...
+ * @return {Error} a reconstructed Error object inheriting @a err
+ *   attributes
+ */
+module.exports.reconstructError = function reconstructError(err) {
+    if (!err) {
+        return err;
+    }
+    const reconstructedErr = new Error(err.message);
+
+    for (const k in err) {
```
rahulreddy: This is super minor but I think it's good to stick to `Object.keys(obj).forEach` pattern of looping through object properties as this pattern performs slow when the object is in hashtable mode(lots property modifications, deletions etc).

lib/network/level-net/index.js
```diff
@@ -0,0 +1,299 @@
...
+ * @param {Object} options.logger logger object
+ * @param {Number} [options.streamMaxPendingAck] max number of
+ *   in-flight output stream packets sent to the server without an ack
+ *   received yet
+ * @param {Number} [options.streamAckTimeoutMs] timeout for receiving
+ *   an ack after an output stream packet is sent to the server
+ * @return {Object} a server object, not yet listening on a TCP port
+ * (you must call listen(port) on the returned object)
+ */
+module.exports.createServer = function LevelServer(db, options = {}) {
```
rahulreddy: `options = {}` would not work since options.logger is a required property.

rahulreddy:
@ironman-machine try

lib/network/level-net/index.js
```diff
@@ -0,0 +1,343 @@
...
+            throw new Error(`argument cb=${cb} is not a callback`);
+        }
+        async.timeout(this._call.bind(this), timeoutMs,
+                      `operation ${remoteCall} timed out`)(remoteCall,
+                                                           args, cb);
+        return undefined;
+    }
+}
+
+const dbAsyncCommandNames = ['put', 'get', 'del', 'batch'];
```
ttvi: I know it has been merged, but I think even with a small number of keys it's still a good practice here to use an object (O(1)) instead of an array (O(n)) as a dictionary.

lib/network/level-net/index.js
```diff
@@ -0,0 +1,343 @@
...
+             * the main db handle.
+             *
+             * @param {String []} path path to the sublevel, as a
+             * piecewise array of sub-levels
+             * @return {Object} the handle to the sublevel
+             */
+            function lookupSubLevel(path) {
+                let subDb = db;
+                path.forEach(pathItem => {
+                    subDb = subDb.sublevel(pathItem);
```
ttvi: This sounds like you have a tree of sub-level dbs that you have to search for each using its absolute path. If what I understood is correct then do you think it's necessary to map a flat structure (buckets) to this hierarchical structure (even though it looks like there are only two levels: namespace and bucket)?

PR #234, opened 2017-03-21
S3C-35 init metadata storage in arsenal

jonathan-gramain:
Follow-up of https://github.com/scality/Arsenal/pull/230

Note: base branch is ft/S3C-35-metadataServer in order to see the changes on top, but this shall be merged to master.

### DIFF
--- a/lib/storage/metadata/file/server.js
+++ b/lib/storage/metadata/file/server.js
@@ -1,8 +1,13 @@
 'use strict'; // eslint-disable-line
 
+const fs = require('fs');
+const os = require('os');
 const assert = require('assert');
+const uuid = require('uuid');
+
 const Logger = require('werelogs').Logger;
 const constants = require('../../../constants');
+const storageUtils = require('../../utils');
 
 const level = require('level');
 const levelNet = require('../../../network/level-net');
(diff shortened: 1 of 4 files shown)


lib/storage/metadata/file/server.js
```diff
@@ -50,12 +55,52 @@ class MetadataFileServer {
...
     /**
      * Start the metadata server and listen to incoming connections
      *
      * @return {undefined}
      */
     startServer() {
+        fs.accessSync(this.metadataPath, fs.F_OK | fs.R_OK | fs.W_OK);
+
+        const warning =
+                  'WARNING: Synchronization directory updates are not ' +
```
rahulreddy: You can use template strings for multiline strings

lib/storage/metadata/file/server.js
```diff
@@ -50,12 +55,52 @@ class MetadataFileServer {
...
+                  'supported on this platform. Newly written data could ' +
+                  'be lost if your system crashes before the operating ' +
+                  'system is able to write directory updates.';
+        if (os.type() === 'Linux' && os.endianness() === 'LE') {
+            try {
+                storageUtils.setDirSyncFlag(this.metadataPath);
+            } catch (err) {
+                this.logger.warn(warning, { error: err.stack });
```
rahulreddy: error: err.message, errorStack: err.stack

lib/storage/utils.js
```diff
@@ -0,0 +1,29 @@
+'use strict'; // eslint-disable-line
+
+const fs = require('fs');
+const assert = require('assert');
+
+module.exports.setDirSyncFlag = function setDirSyncFlag(path) {
+    // may throw if ioctl is not available
+    const ioctl = require('ioctl');
```
rahulreddy: You need to add ioctl to package.json

lib/storage/metadata/file/server.js
jonathan-gramain: I'm not convinced this will make it more readable: AFAIK, all spaces present in a template string in the code file are part of the string, so I would have to align the second line to the left margin, which makes indentation confusing. Otherwise I can concatenate two template strings, but then it's not much different than using plain strings.

Please correct me if I'm wrong.

lib/storage/utils.js
jonathan-gramain: I did this only in my data PR in-progress, not here, thanks for pointing it.

lib/storage/metadata/file/server.js
jonathan-gramain: I just copy and paste this I think, though it's definitely a mistake

lib/storage/metadata/file/server.js
rahulreddy: Oh nvm, since this was meant for a logger message.

lib/storage/metadata/file/server.js
jonathan-gramain: fixed

lib/storage/utils.js
jonathan-gramain: done

rahulreddy:
Please redo the commit message to be under 50 characters.

PR #235, opened 2017-03-23
S3C-158 REST interface for datafile backend

jonathan-gramain:
This is how we will be able to do the data storage in a separate
storage daemon, running on another container or host. For now a data
REST server will be spawned locally by the S3 server at startup (in S3
PR).

There are actually three parts: the REST client, the REST server, and
the DataFileStore which handles the storage logic. The DataFileStore
implementation comes from the original data/file implementation in S3
server.

The REST API uses a service base path named /DataFile, and does roughly:

 - a PUT directly on /DataFile URL creates a new object file and
   returns its new random hex-encoded key through its URL in a
   Location response header, along with a '201 Created' response
   code. The REST client extracts the key from this URL and returns it
   in the callback.

 - GET and DELETE on the URL returned in the 'Location' header shall
   do their duty, though the file backend appends the known key to the
   base path to recreate the URL.

### DIFF
--- a/lib/jsutil.js
+++ b/lib/jsutil.js
@@ -1,0 +1,32 @@
+'use strict'; // eslint-disable-line
+
+const debug = require('util').debuglog('jsutil');
+
+// JavaScript utility functions
+
+/**
+ * force @a func to be called only once, even if actually called
+ * multiple times. The cached result of the first call is then
+ * returned (if any).
+ *
+ * @note underscore.js provides this functionality but not worth
+ * adding a new dependency for such a small use case.
+ *
+ * @param {function} func function to call at most once
+
+ * @return {function} a callable wrapper mirroring @a func but
+ * only calls @a func at first invocation.
+ */
+module.exports.once = function once(func) {
+    const state = { called: false, res: undefined };
+    return function wrapper(...args) {
+        if (!state.called) {
+            state.called = true;
+            state.res = func.apply(func, args);
+        } else {
+            debug('function already called:', func,
+                  'returning cached result:', state.res);
+        }
+        return state.res;
+    };
+};
--- a/lib/jsutil.js
+++ b/lib/jsutil.js
@@ -5,7 +5,7 @@ const debug = require('util').debuglog('jsutil');
 // JavaScript utility functions
 
 /**
- * force @a func to be called only once, even if actually called
+ * force <tt>func</tt> to be called only once, even if actually called
  * multiple times. The cached result of the first call is then
  * returned (if any).
  *
@@ -14,8 +14,8 @@ const debug = require('util').debuglog('jsutil');
  *
  * @param {function} func function to call at most once
 
- * @return {function} a callable wrapper mirroring @a func but
- * only calls @a func at first invocation.
+ * @return {function} a callable wrapper mirroring <tt>func</tt> but
+ * only calls <tt>func</tt> at first invocation.
  */
 module.exports.once = function once(func) {
     const state = { called: false, res: undefined };
--- a/lib/network/http/utils.js
+++ b/lib/network/http/utils.js
@@ -1,0 +1,109 @@
+'use strict'; // eslint-disable-line
+
+const errors = require('../../errors');
+
+/**
+ * Parse the Range header into an object
+ *
+ * @param {String} rangeHeader - The 'Range' header value
+
+ * @return {Object} object containing a range specification, with
+ * either of:
+ * - start and end attributes: a fully specified range request
+ * - a single start attribute: no end is specified in the range request
+ * - a suffix attribute: suffix range request
+ * - an error attribute of type errors.InvalidArgument if the range
+ *     syntax is invalid
+ */
+function parseRangeSpec(rangeHeader) {
+    const rangeMatch = /^bytes=([0-9]+)?-([0-9]+)?$/.exec(rangeHeader);
+    if (rangeMatch) {
+        const rangeValues = rangeMatch.slice(1, 3);
+        if (rangeValues[0] === undefined) {
+            if (rangeValues[1] !== undefined) {
+                return { suffix: Number.parseInt(rangeValues[1], 10) };
+            }
+        } else {
+            const rangeSpec = { start: Number.parseInt(rangeValues[0], 10) };
+            if (rangeValues[1] === undefined) {
+                return rangeSpec;
+            }
+            rangeSpec.end = Number.parseInt(rangeValues[1], 10);
+            if (rangeSpec.start <= rangeSpec.end) {
+                return rangeSpec;
+            }
+        }
+    }
+    return { error: errors.InvalidArgument };
+}
+
+/**
+ * Convert a range specification as given by parseRangeSpec() into a
+ * fully specified absolute byte range
+ *
+ * @param {Number []} rangeSpec - Parsed range specification as returned
+ *   by parseRangeSpec()
+ * @param {Number} objectSize - Total byte size of the whole object
+
+ * @return {Object} object containing either:
+ * - a 'range' attribute which is a fully specified byte range [start,
+       end], as the inclusive absolute byte range to request from the
+       object
+ * - or no attribute if the requested range is a valid range request
+       for a whole empty object (non-zero suffix range)
+ * - or an 'error' attribute of type errors.InvalidRange if the
+ *     requested range is out of object's boundaries.
+ */
+function getByteRangeFromSpec(rangeSpec, objectSize) {
+    if (rangeSpec.suffix !== undefined) {
+        if (rangeSpec.suffix === 0) {
+            // 0-byte suffix is always invalid (even on empty objects)
+            return { error: errors.InvalidRange };
+        }
+        if (objectSize === 0) {
+            // any other suffix range on an empty object returns the
+            // full object (0 bytes)
+            return {};
+        }
+        return { range: [Math.max(objectSize - rangeSpec.suffix, 0),
+                         objectSize - 1] };
+    }
+    if (rangeSpec.start < objectSize) {
+        // test is false if end is undefined
+        return { range: [rangeSpec.start,
+                         (rangeSpec.end < objectSize ?
+                          rangeSpec.end : objectSize - 1)] };
+    }
+    return { error: errors.InvalidRange };
+}
+
+/**
+ * Convenience function that combines parseRangeSpec() and
+ * getByteRangeFromSpec()
+ *
+ * @param {String} rangeHeader - The 'Range' header value
+ * @param {Number} objectSize - Total byte size of the whole object
+
+ * @return {Object} object containing either:
+ * - a 'range' attribute which is a fully specified byte range [start,
+ *     end], as the inclusive absolute byte range to request from the
+ *     object
+ * - or no attribute if the requested range is either syntactically
+ *     incorrect or is a valid range request for an empty object
+ *     (non-zero suffix range)
+ * - or an 'error' attribute instead of type errors.InvalidRange if
+ *     the requested range is out of object's boundaries.
+ */
+function parseRange(rangeHeader, objectSize) {
+    const rangeSpec = parseRangeSpec(rangeHeader);
+    if (rangeSpec.error) {
+        // invalid range syntax is silently ignored in HTTP spec,
+        // hence returns the whole object
+        return {};
+    }
+    return getByteRangeFromSpec(rangeSpec, objectSize);
+}
+
+module.exports = { parseRangeSpec,
+                   getByteRangeFromSpec,
+                   parseRange };
--- a/lib/network/rest/RESTClient.js
+++ b/lib/network/rest/RESTClient.js
@@ -1,0 +1,255 @@
+'use strict'; // eslint-disable-line
+
+const assert = require('assert');
+const http = require('http');
+
+const Logger = require('werelogs').Logger;
+const constants = require('../../constants');
+const utils = require('./utils');
+const errors = require('../../errors');
+
+function setRequestUids(reqHeaders, reqUids) {
+    // inhibit 'assignment to property of function parameter' -
+    // this is what we want
+    // eslint-disable-next-line
+    reqHeaders['X-Scal-Request-Uids'] = reqUids;
+}
+
+function setRange(reqHeaders, range) {
+    const rangeStart = range[0] !== undefined ? range[0].toString() : '';
+    const rangeEnd = range[1] !== undefined ? range[1].toString() : '';
+    // inhibit 'assignment to property of function parameter' -
+    // this is what we want
+    // eslint-disable-next-line
+    reqHeaders['Range'] = `bytes=${rangeStart}-${rangeEnd}`;
+}
+
+function setContentType(reqHeaders, contentType) {
+    // inhibit 'assignment to property of function parameter' -
+    // this is what we want
+    // eslint-disable-next-line
+    reqHeaders['Content-Type'] = contentType;
+}
+
+function setContentLength(reqHeaders, size) {
+    // inhibit 'assignment to property of function parameter' -
+    // this is what we want
+    // eslint-disable-next-line
+    reqHeaders['Content-Length'] = size.toString();
+}
+
+function makeErrorFromHTTPResponse(response) {
+    const rawBody = response.read();
+    const body = (rawBody !== null ? rawBody.toString() : '');
+    let error;
+    try {
+        const fields = JSON.parse(body);
+        error = errors[fields.errorType]
+            .customizeDescription(fields.errorMessage);
+    } catch (err) {
+        error = new Error(body);
+    }
+    // error is always a newly created object, so we can modify its
+    // properties
+    error.remote = true;
+    return error;
+}
+
+
+/**
+ * @class
+ * @classdesc REST Client interface
+ *
+ * The API is usable when the object is constructed.
+ */
+class RESTClient {
+    /**
+     * Interface to the data file server
+     * @constructor
+     * @param {Object} params - Contains the basic configuration.
+     * @param {String} params.host - hostname or ip address of the
+     *   RESTServer instance
+     * @param {Number} params.port - port number that the RESTServer
+     *   instance listens to
+     * @param {Logger} [params.log] - logging configuration
+     */
+    constructor(params) {
+        assert(params.host);
+        assert(params.port);
+
+        this.host = params.host;
+        this.port = params.port;
+        this.setupLogging(params.log);
+        this.httpAgent = new http.Agent({ keepAlive: true });
+    }
+
+    setupLogging(config) {
+        let options = undefined;
+        if (config !== undefined) {
+            options = {
+                level: config.logLevel,
+                dump: config.dumpLevel,
+            };
+        }
+        this.logging = new Logger('DataFileRESTClient', options);
+    }
+
+    createLogger(reqUids) {
+        return reqUids ?
+            this.logging.newRequestLoggerFromSerializedUids(reqUids) :
+            this.logging.newRequestLogger();
+    }
+
+    doRequest(method, headers, key, log, responseCb) {
+        const reqHeaders = headers || {};
+        const urlKey = key || '';
+        const reqParams = {
+            hostname: this.host,
+            port: this.port,
+            method,
+            path: `${constants.dataFileURL}/${urlKey}`,
+            headers: reqHeaders,
+            agent: this.httpAgent,
+        };
+        log.debug(`about to send ${method} request`, {
+            hostname: reqParams.hostname,
+            port: reqParams.port,
+            path: reqParams.path,
+            headers: reqParams.headers });
+        const request = http.request(reqParams, responseCb);
+
+        // disable nagle algorithm
+        request.setNoDelay(true);
+        return request;
+    }
+
+    /**
+     * This sends a PUT request to the the REST server
+     * @param {http.IncomingMessage} stream - Request with the data to send
+     * @param {string} stream.contentHash - hash of the data to send
+     * @param {integer} size - size
+     * @param {string} reqUids - The serialized request ids
+     * @param {RESTClient~putCallback} callback - callback
+     * @returns {undefined}
+     */
+    put(stream, size, reqUids, callback) {
+        const log = this.createLogger(reqUids);
+        const headers = {};
+        setRequestUids(headers, reqUids);
+        setContentType(headers, 'application/octet-stream');
+        setContentLength(headers, size);
+
+        const request = this.doRequest('PUT', headers, null, log, response => {
+            response.once('readable', () => {
+                // expects '201 Created'
+                if (response.statusCode !== 201) {
+                    return callback(makeErrorFromHTTPResponse(response));
+                }
+                // retrieve the key from the Location response header
+                // containing the complete URL to the object, like
+                // /DataFile/abcdef.
+                const location = response.headers.location;
+                if (location === undefined) {
+                    return callback(new Error(
+                        'missing Location header in the response'));
+                }
+                const locationInfo = utils.explodePath(location);
+                if (!locationInfo) {
+                    return callback(new Error(
+                        `bad Location response header: ${location}`));
+                }
+                return callback(null, locationInfo.key);
+            });
+        }).on('finish', () => {
+            log.debug('finished sending PUT data to the REST server', {
+                component: 'RESTClient',
+                method: 'put',
+                contentLength: size,
+            });
+        }).on('error', callback);
+
+        stream.pipe(request);
+        stream.on('error', err => {
+            log.error('error from readable stream', {
+                error: err,
+                method: 'put',
+                component: 'RESTClient',
+            });
+            request.end();
+        });
+    }
+
+    /**
+     * send a GET request to the REST server
+     * @param {String} key - The key associated to the value
+     * @param { Number [] | Undefined} range - range (if any) a
+     *   [start, end] inclusive range specification, as defined in
+     *   HTTP/1.1 RFC.
+     * @param {String} reqUids - The serialized request ids
+     * @param {RESTClient~getCallback} callback - callback
+     * @returns {undefined}
+     */
+    get(key, range, reqUids, callback) {
+        const log = this.createLogger(reqUids);
+        const headers = {};
+        setRequestUids(headers, reqUids);
+        if (range) {
+            setRange(headers, range);
+        }
+        const request = this.doRequest('GET', headers, key, log, response => {
+            response.once('readable', () => {
+                if (response.statusCode !== 200 &&
+                    response.statusCode !== 206) {
+                    return callback(makeErrorFromHTTPResponse(response));
+                }
+                return callback(null, response);
+            });
+        }).on('error', callback);
+
+        request.end();
+    }
+
+    /**
+     * send a DELETE request to the REST server
+     * @param {String} key - The key associated to the values
+     * @param {String} reqUids - The serialized request ids
+     * @param {RESTClient~deleteCallback} callback - callback
+     * @returns {undefined}
+     */
+    delete(key, reqUids, callback) {
+        const log = this.createLogger(reqUids);
+        const headers = {};
+        setRequestUids(headers, reqUids);
+
+        const request = this.doRequest(
+            'DELETE', headers, key, log, response => {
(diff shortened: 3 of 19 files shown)


jonathan-gramain:
Hold on: this PR introduces a concurrency issue that I have to investigate.

jonathan-gramain:
looks good now.

lib/storage/data/file/store.js
```diff
@@ -0,0 +1,274 @@
...
+     * @param {stream.Readable} dataStream - input stream providing the
+     *   object data
+     * @param {Number} size - Total byte size of the data to put
+     * @param {Logger} log - logging object
+     * @param {DataFileStore~putCallback} callback - called when done
+     * @return {undefined}
+     */
+    put(dataStream, size, log, callback) {
+        const key = crypto.pseudoRandomBytes(20).toString('hex');
+        const filePath = this.getFilePath(key);
```
ttvi: Do you think we should consider other alternatives, such as leveldb, since pure file backend may be more prone to the limitation of the number of inodes?

package.json
```diff
@@ -32,6 +32,9 @@
   "optionalDependencies": {
       "ioctl": "2.0.0"
   },
+  "optionalDependencies": {
+    "ioctl": "^2.0.0"
```
ttvi: You are dupplicating `optionalDependencies` here?

lib/network/rest/server.js
```diff
@@ -0,0 +1,332 @@
...
+        const reqUids = req.headers['x-scal-request-uids'];
+        const log = this.createLogger(reqUids);
+        log.debug('request received', { method: req.method,
+                                        url: req.url });
+        const reqMethods = {
+            PUT: this._onPut.bind(this),
+            GET: this._onGet.bind(this),
+            DELETE: this._onDelete.bind(this),
+        };
+        if (req.method in reqMethods) {
```
ttvi: Using `in` may not be the only option. For this kind of fixed corpus, you can also use `switch ... case` so that you don't have to create the object `reqMethods` every single request.
The `bind` is not needed here neither btw.

lib/network/rest/server.js
```diff
@@ -0,0 +1,332 @@
...
+ * You have to call setup() to initialize the storage backend, then
+ * start() to start listening to the configured port.
+ */
+class RESTServer extends httpServer {
+
+    /**
+     * @constructor
+     * @param {Object} params - constructor params
+     * @param {Number} params.port - TCP port where the server listens to
+     * @param {Object} [params.log] - logger configuration
```
ttvi: No `params.dataStore`?

lib/storage/data/file/store.js
jonathan-gramain: I don't think it will be necessary for the need.

If number of inodes is a problem, a deeper and better directory hashing can be a solution (not planned for now).

package.json
jonathan-gramain: oops, good catch

lib/network/rest/server.js
jonathan-gramain: fixed

lib/network/rest/server.js
jonathan-gramain: Yes, a `switch` is a good alternative.

The `bind` is necessary in the current implementation to get the proper `this` (don't ask me why...), but with a `switch` we can get rid of it.

jonathan-gramain: As an alternative, `reqMethods` may be initialized in the constructor to avoid creating it each time (after, this or a `switch..case` is a matter of taste, I tend to like this function mapping style slightly better though).

lib/network/rest/server.js
```diff
@@ -0,0 +1,332 @@
...
+     * @return {undefined}
+     */
+    _onPut(req, res, log) {
+        const urlObj = url.parse(req.url);
+        const pathInfo = utils.explodePath(urlObj.path);
+        if (!pathInfo) {
+            return sendError(res, 403, 'invalid URL');
+        }
+        if (pathInfo.service !== constants.dataFileURL) {
+            return sendError(res, 403,
```
rahulreddy: It's ideal if you return Arsenal errors instead of custom errors. This ensures consistency amongst the projects.

lib/network/rest/server.js
```diff
@@ -0,0 +1,332 @@
...
+        }
+        this.dataStore.put(req, size, log, (err, key) => {
+            if (err) {
+                return sendError(res, err.code, err.message);
+            }
+            log.debug('sending back 201 response');
+            res.setHeader('Location', `${constants.dataFileURL}/${key}`);
+            setContentLength(res, 0);
+            res.writeHead(201);
+            return res.end();
```
rahulreddy: You can use end logger in res.end()'s callback to log end of request.

lib/network/rest/server.js
```diff
@@ -0,0 +1,332 @@
...
+     *
+     * @param {http.IncomingMessage} req - HTTP request object
+     * @param {http.ServerResponse} res - HTTP response object
+     * @param {Logger} log - logger object
+     * @return {undefined}
+     */
+    _onGet(req, res, log) {
+        const urlObj = url.parse(req.url);
+        const pathInfo = utils.explodePath(urlObj.path);
+        if (!pathInfo) {
```
rahulreddy: There is quite a bit of code which is being duplicated for request parsing. Can you abstract this into a method instead?

lib/network/rest/server.js
```diff
@@ -0,0 +1,332 @@
...
+                             `unsupported service '${pathInfo.service}'`);
+        }
+        if (pathInfo.key === undefined) {
+            return sendError(res, 403, 'URL is missing key');
+        }
+        const rangeHeader = req.headers.range;
+        let rangeSpec = undefined;
+        if (rangeHeader !== undefined) {
+            try {
+                rangeSpec = parseRangeSpec(rangeHeader);
```
rahulreddy: From the looks of it `parseRangeSpec` does not throw? If that's true, why use a try...catch here?

lib/network/rest/utils.js
```diff
@@ -0,0 +1,13 @@
...
+module.exports.explodePath = function explodePath(path) {
+    const pathMatch = /^(\/[a-zA-Z0-9]+)(\/([0-9a-f]*))?$/.exec(path);
+    if (pathMatch) {
+        return {
+            service: pathMatch[1],
+            key: (pathMatch[3] !== undefined && pathMatch[3].length > 0 ?
+                  pathMatch[3] : undefined),
+        };
+    }
+    return undefined;
```
rahulreddy: If you return an error here as `new Error('invalid path')` the logic where you use this method will be more readable.

lib/storage/data/file/store.js
```diff
@@ -0,0 +1,274 @@
...
+const async = require('async');
+const assert = require('assert');
+
+const Logger = require('werelogs').Logger;
+const errors = require('../../../errors');
+const stringHash = require('../../../stringHash');
+const jsutil = require('../../../jsutil');
+const storageUtils = require('../../utils');
+
+const FOLDER_HASH = 3511;
```
rahulreddy: Can you add a comment here on what this constant does?

lib/storage/data/file/store.js
```diff
@@ -0,0 +1,274 @@
...
+     * files in it.
+     *
+     * The function ensures that dataPath is accessible and
+     * pre-creates the directory hashes under dataPath.
+     *
+     * @param {function} callback - called when done with no argument
+     * @return {undefined}
+     */
+    setup(callback) {
+        fs.accessSync(this.dataPath, fs.F_OK | fs.R_OK | fs.W_OK);
```
rahulreddy: Why not use the async version to catch the error if it throws (since `setup` is async anyways)

lib/storage/data/file/store.js
```diff
@@ -0,0 +1,274 @@
...
+            fs.mkdir(`${this.dataPath}/${subDirName}`, err => {
+                // If already exists, move on
+                if (err && err.code !== 'EEXIST') {
+                    return next(err);
+                }
+                return next();
+            });
+        },
+        err => {
+            assert.strictEqual(err, null, `Error creating data files ${err}`);
```
rahulreddy: It's not ideal to throw in async functions. I recommend sending the error to the caller and taking care of it.

lib/storage/data/file/store.js
```diff
@@ -0,0 +1,274 @@
...
+        const key = crypto.pseudoRandomBytes(20).toString('hex');
+        const filePath = this.getFilePath(key);
+        log.debug('starting to write data', { key, filePath });
+        dataStream.pause();
+        fs.open(filePath, 'wx', (err, fd) => {
+            if (err) {
+                log.error('error opening filePath', { error: err });
+                return callback(err);
+            }
+            const cbOnce = jsutil.once(callback);
```
rahulreddy: This may mask an actual situation where bad code is calling callback twice. I recommend fixing the callback usage instead.

lib/storage/data/file/store.js
```diff
@@ -0,0 +1,274 @@
...
+                    return cbOnce(null, key);
+                }
+                if (this.noSync) {
+                    fs.close(fd);
+                    return ok();
+                }
+                fs.fsync(fd, err => {
+                    fs.close(fd);
+                    if (err) {
+                        log.error('fsync error', { error: err });
```
rahulreddy: Please annotate the logs with `method` param so that it's easier when debugging.

lib/storage/data/file/store.js
```diff
@@ -0,0 +1,274 @@
...
+                return undefined;
+            }).on('error', err => {
+                log.error('error streaming data on write',
+                          { error: err });
+                // destroying the write stream forces a close(fd)
+                fileStream.destroy();
+                return cbOnce(err);
+            });
+            dataStream.resume();
+            dataStream.pipe(fileStream);
```
rahulreddy: Instead of making `pipe` end `fileStream` automatically, why not call `fd.close()` explicitly after `pipe`. This also ensures that the callback is not called twice.

lib/storage/data/file/store.js
jonathan-gramain: (of course if the number of inodes on the whole file system is the issue, directory hashing does not help)

lib/network/rest/client.js
```diff
@@ -0,0 +1,267 @@
...
+        this.logging = new Logger('RESTClient', options);
+    }
+
+    createLogger(reqUids) {
+        return reqUids ?
+            this.logging.newRequestLoggerFromSerializedUids(reqUids) :
+            this.logging.newRequestLogger();
+    }
+
+    createRequest(method, headers, key = '') {
```
jonathan-gramain: Well, not exactly, actually it creates the fields necessary to issue the request head. So maybe createRequestHeadFields can be a better name.

jonathan-gramain: Note that it also returns the agent used to manage connections.

jonathan-gramain: The result contains a `headers` field which correspond to the headers to send.

lib/network/rest/client.js
```diff
@@ -0,0 +1,267 @@
...
+
+        const reqHeader = this.createRequest('DELETE', headers, key);
+        log.debug('about to send DELETE request', {
+            hostname: reqHeader.hostname,
+            port: reqHeader.port,
+            path: reqHeader.path,
+            headers: reqHeader.headers });
+        const request = http.request(reqHeader, response => {
+            response.once('readable', () => {
+                if (response.statusCode !== 200) {
```
jonathan-gramain: AFAIK, 204 has a specific meaning where the client should not update its view of a requested document contents. I think it should only be used in this particular case. So in doubt, I use 200 instead. And it's not excluded that we may add some meta-information in the request body later.

lib/network/rest/server.js
```diff
@@ -0,0 +1,332 @@
...
+/**
+ * Parse the Range header into a two-integers array with possibly
+ * undefined values
+ *
+ * @param {String} rangeHeader - The 'Range' header value
+ * @return {(Number|undefined) []} An array of [left, right] integer
+ *   values provided in the range specification if present, or
+ *   undefined otherwise.
+ */
+function parseRangeSpec(rangeHeader) {
```
jonathan-gramain: I prefer having separate functions for separate concerns of 1. parsing the range spec and 2. getting the absolute byte range to read from the object.

Also `parseRange` looks a bit too complex and error-prone for what it does, and it's in S3 code (which would mean moving it to arsenal).

Though I agree it's not ideal to have duplication. Ideally (to me!) we would have both functions in Arsenal, used by both the data server and S3 server. Do you think it's worth doing this change?

Overall, those two functions are small and (hopefully) well tested so I'm not too concerned about keeping as-is.

lib/network/rest/server.js
```diff
@@ -0,0 +1,332 @@
...
+    const rangeMatch = /^bytes=([0-9]+)?-([0-9]+)?$/.exec(rangeHeader);
+    // ignore header if syntax is invalid
+    if (rangeMatch) {
+        return rangeMatch.slice(1, 3).map(
+            str => (str === undefined ?
+                    undefined : Number.parseInt(str, 10)));
+    }
+    return undefined;
```
jonathan-gramain: It would allow to log something when catching it, indeed. But in order to be HTTP compliant (we not necessarily have to for a local daemon) we have to catch the exception and return a 200 OK anyway.

jonathan-gramain: I can do this change for the log.

jonathan-gramain: by return, do you mean to throw? Otherwise we need to check the type on return, not sure it helps.

jonathan-gramain: The way it is may throw in `Number.parseInt` if range boundaries are not correct numbers. So i think it's consistent to throw also if the regexp is not happy.

lib/network/rest/server.js
```diff
@@ -0,0 +1,332 @@
...
+    return byteRange;
+}
+
+function setContentLength(response, contentLength) {
+    response.setHeader('Content-Length', contentLength.toString());
+}
+
+function setContentRange(response, byteRange, objectSize) {
+    const [start, end] = byteRange;
+    assert(start !== undefined && end !== undefined);
```
jonathan-gramain: I always found asserts a good tool to catch programming errors early. At least in C they are very much used in the ring code. Here, the previous `getByteRange` function ought to have returned a complete byte range, since that is its job, so we're checking this here.

Though if the practice is different in javascript for various reasons, I don't mind turning them to errors (here also, returning something means checking the return value, which may also add unnecessary complexity, if we can throw I find it better).

Please come back if you want it to be errors instead (though I will probably not be convinced still :))

lib/network/rest/server.js
```diff
@@ -0,0 +1,332 @@
...
+     */
+    constructor(params) {
+        assert(params.port);
+
+        const logging = createLoggingObject(params.log);
+        super(params.port, logging);
+        this.logging = logging;
+        this.dataStore = params.dataStore;
+
+        this.onRequest(this._onRequest);
```
jonathan-gramain: done

lib/network/rest/server.js
jonathan-gramain: `sendError` is actually forging the HTTP response, nothing more, it does not actually create an Error object.

I see a value of using arsenal errors when they match a specific error type. In this case, as an example, no arsenal error tells that the user has entered an invalid URI, so it turns that either I have to create a new arsenal error containing this message and a 403 error code (there are already a dozen, most S3-oriented), or I have to return an "AccessDenied" message of some sort coming from an existing error, that does not contain the reason, which can make it harder to debug invalid use. Though maybe it can be used to reconstruct the matching arsenal error on client side, but only if we parse the description, not sure it's practical to do.

lib/network/rest/server.js
jonathan-gramain: This is a nice idea

lib/network/rest/server.js
jonathan-gramain: Actually, there is an "invalid uri", my bad, it's a 400 though.

I will browse the existing errors sent and see if they can match an existing message. But I doubt we can reconstruct an actual arsenal error on the other side.

jonathan-gramain: Or what I could do is instead of returning a bare message, return a serialized form of the arsenal error which contains the string code ("InvalidURI")

jonathan-gramain: done :)

lib/network/rest/server.js
jonathan-gramain: sure

lib/network/rest/server.js
jonathan-gramain: it does (`Number.parseInt`), see above discussion

lib/network/rest/utils.js
jonathan-gramain: Shouldn't I use `throw` instead? I agree though that giving back an error is better.

lib/storage/data/file/store.js
jonathan-gramain: We have to guard against calling multiple times here, because we're not controlling how errors events are generated, both from `dataStream` and from `fileStream`. Both may actually generate errors until they are destroyed, and multiple errors are also possible AFAIK.

lib/storage/data/file/store.js
jonathan-gramain: This is the case: `fs.close(fd)` is called when done. We set `autoClose: false` option so that the write stream is not closed automatically and we can control the time we do the close.

The common case where it's done is in the callback of `fs.fsync()`, other cases where we close fd exist in case of errors. It has to be done only when all users have finished using the fd, so it cannot be done just after `pipe`.

lib/storage/data/file/store.js
jonathan-gramain: done

lib/storage/data/file/store.js
jonathan-gramain: Now it's more modular it may indeed be a better idea. I just kept how it was done before, since if we can't initialize the backend we can't do much anyway it's not a big deal to assert, but returning a proper async error would be better I agree.

lib/storage/data/file/store.js
jonathan-gramain: same comment than above

lib/storage/data/file/store.js
jonathan-gramain: done

lib/network/rest/server.js
jonathan-gramain: Edit: I'm wrong, `Number.parseInt` does not throw but returns `NaN`, so your point is valid.

jonathan-gramain: Just came back to me after more reading: the numbers parsed are necessarily valid positive numbers because already checked by the regexp. (We don't check that they don't overflow though).

lib/network/rest/client.js
jonathan-gramain: Apparently either a 200 or 204 is ok for a DELETE status code. So in our case it can indeed make more sense to return 204 since we don't return any body.

I like this chart ^^:
https://i.stack.imgur.com/whhD1.png

lib/network/rest/client.js
```diff
@@ -0,0 +1,267 @@
...
+        if (range) {
+            setRange(headers, range);
+        }
+        const reqHeader = this.createRequest('GET', headers, key);
+        log.debug('about to send GET request', {
+            hostname: reqHeader.hostname,
+            port: reqHeader.port,
+            path: reqHeader.path,
+            headers: reqHeader.headers });
+        const request = http.request(reqHeader, response => {
```
jonathan-gramain: done

lib/network/rest/client.js
```diff
@@ -0,0 +1,267 @@
...
+
+    setupLogging(config) {
+        let options = undefined;
+        if (config !== undefined) {
+            options = {
+                level: config.logLevel,
+                dump: config.dumpLevel,
+            };
+        }
+        this.logging = new Logger('RESTClient', options);
```
jonathan-gramain: renamed to 'DataFileRESTClient', same for server

lib/storage/data/file/store.js
jonathan-gramain: done

lib/network/rest/server.js
```diff
@@ -0,0 +1,332 @@
+'use strict'; // eslint-disable-line
+
+const httpServer = require('../http/server');
+const Logger = require('werelogs').Logger;
+const url = require('url');
+const assert = require('assert');
```
jonathan-gramain: done

jonathan-gramain:
I should be done with the requested changes, please have a look @rahulreddy when you have time . Thanks for your useful comments!

lib/network/rest/server.js
rahulreddy: I think it's ideal if we move this piece into Arsenal and make S3 use the same one. We will have one fully tested piece. I have seen quite a few instances in the past where the code is duplicated for simple utilities where there is a bug in one and not the other. One becomes a fully tested piece and the other one not. Also that's the whole purpose of Arsenal. It's meant to have this collection of utilities (big/small) which can be shared across.

lib/network/rest/server.js
rahulreddy: Instead of throwing and wrapping a try..catch around it, I find the pattern of returning an Error instance cleaner and clearer. 
```js
return new Error('invalid range header');
```
This would not throw.
You can add stack, message properties if you need them. Also, returning an error here would mean, you have to define the error only once and return it. While the other way would mean you are defining the error case multiple times.

lib/network/rest/server.js
rahulreddy: throwing runtime would mean crashing the process if the exception is not caught. It means penalizing other requests being processed by that process when only one request was actually incorrect. We hunted down a ton of run-time throws last time where they were crashing the process at a customer and it was ideal. We can use Errors instead and check for error in the caller.

lib/network/rest/server.js
rahulreddy: Yes, serialized Arsenal errors are the best course. You can sum any custom error that you may not find in the error list and we can come up with an agreement whether to add them or find some other compromise.

lib/network/rest/utils.js
rahulreddy: Ultimately, throw would end being try..catch 'd. Just returning the error directly when we know it's an error would make sense.

lib/storage/data/file/store.js
rahulreddy: Thanks.

lib/storage/data/file/store.js
rahulreddy: Ah I see. Is it possible to log when a double callback occurs. It would help us understanding it's behavior better.

lib/storage/data/file/store.js
rahulreddy: Thanks.

lib/storage/data/file/store.js
rahulreddy: > The common case where it's done is in the callback of fs.fsync(), other cases where we close fd exist in case of errors. It has to be done only when all users have finished using the fd, so it cannot be done just after pipe.

That's true 😄 . I missed that point.

jonathan-gramain:
S3 side depends on this PR to be compatible with this one (because of metadata file name changes): https://github.com/scality/S3/pull/686

lib/network/rest/client.js
jonathan-gramain: As far as I can remember, sproxyd returns 200 + a small message body. But whichever works anyway.

lib/network/rest/utils.js
jonathan-gramain: So in order to know if an error occurs, do we have to check the type of what's returned? (`if (typeof ret === 'Error') {...}` for example) I personally find the `try..catch` a better structure (it is actually made for that and can possibly be placed further up in the stack if required). But if you have a better way to handle returned errors, I'd be glad to know! (real question, no sarcasm)

jonathan-gramain: Of course the other side of it is that you *have* to catch at some point to not crash your process, but when using `return` if you're not checking the returned error, you don't control what will be going on later (maybe stuck a client or even take the error contents for payload and return some garbage...)

lib/network/rest/server.js
jonathan-gramain: To be honest I'm not very satisfied with the current implementation of the S3 function (but this is probably my picky nature :p) hence my reluctance to use it as-is. I find it a bit too complex for what it is for, as it does not decouple the parsing concern from the processing, and the parsing could be both more robust and simpler if it was using a regexp (unless there's a performance concern about regexp, though when compiled they should be rather fast).

See e.g.:
```
    const rangePortion = rangeHeader.replace('bytes=', '').split('-');
    if (rangePortion.length > 2) {
        return { range };
    }
```

This would not prevent e.g. `bytes=bytes=5-10bytes=` to be processed as a valid range. Of course it's not likely to happen unless someone forges a request like that, so it's nothing critical, just noticed it on the way.

```
    if (start >= totalLength && end >= totalLength) {
        return { range, error: errors.InvalidRange };
    }
```
This may be the answer to my unrelated question about how we handle errors that are returned and not thrown.

lib/jsutil.js
```diff
@@ -0,0 +1,32 @@
+'use strict'; // eslint-disable-line
+
+const debug = require('util').debuglog('jsutil');
+
+// JavaScript utility functions
+
+/**
+ * force @a func to be called only once, even if actually called
```
LaurenSpiegel: @a?

lib/network/http/utils.js
```diff
@@ -0,0 +1,109 @@
...
+ * Parse the Range header into an object
+ *
+ * @param {String} rangeHeader - The 'Range' header value
+
+ * @return {Object} object containing a range specification, with
+ * either of:
+ * - start and end attributes: a fully specified range request
+ * - a single start attribute: no end is specified in the range request
+ * - a suffix attribute: suffix range request
+ * - an error attribute of type errors.InvalidArgument if the range
```
LaurenSpiegel: we should run the ceph tests against s3 server with this branch before merging this. I believe AWS has specific rules on when it ignores an invalid range (just sends the full object) and when it returns an error.

lib/network/rest/RESTClient.js
```diff
@@ -0,0 +1,255 @@
...
+        this.logging = new Logger('DataFileRESTClient', options);
+    }
+
+    createLogger(reqUids) {
+        return reqUids ?
+            this.logging.newRequestLoggerFromSerializedUids(reqUids) :
+            this.logging.newRequestLogger();
+    }
+
+    doRequest(method, headers, key, log, responseCb) {
```
LaurenSpiegel: wouldn't configureRequest or createRequest be more applicable?

lib/jsutil.js
jonathan-gramain: I'm not sure if this is valid JSDoc syntax. In doxygen it means a link to some other object in the doxygen, so that func refers to the actual parameter. No problem to remove the `@a` if it does not make sense here.

lib/network/http/utils.js
jonathan-gramain: Any test you know of to validate this is good.

I believe AWS rules are those of HTTP RFC regarding range support. But those are sometimes not completely unambiguous when it comes to uncommon cases such as empty object support and suffix range.

lib/network/rest/RESTClient.js
jonathan-gramain: This function actually sends the request (at line 119)

jonathan-gramain: But I'm lacking a better name so a better suggestion is still welcome.

lib/network/http/utils.js
LaurenSpiegel: these are the tests I am referring to: https://github.com/ceph/s3-tests
See integration for how to run them: https://github.com/scality/Integration/blob/master/genericStaas/env/run_cephtests.bash

lib/network/rest/RESTClient.js
LaurenSpiegel: I guess doRequest is fine.  makeRequest also works

lib/network/http/utils.js
jonathan-gramain: Tests are happy now (except the unrelated zero maxKeys).

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine retry

PR #236, opened 2017-03-24
fix: werelogs as a dependency

ttvi:


LaurenSpiegel:
@ironman-machine go

LaurenSpiegel:
end to end passed here: http://ci.ironmann.io/gh/scality/Integration/11068

PR #238, opened 2017-03-28
S3C-35 change metadata timeout from 5 to 30s

jonathan-gramain:
The rationale is that tests have failed with timeout errors, it seems 5 seconds is too short for our test VMS. 

Hopefully 30s is long enough that these errors will not occur again.


rahulreddy:
@ironman-machine try

PR #240, opened 2017-04-03
fix: wrong check of delimiter index

ttvi:
This fixes a false assumption in #228 that the index of the last character of a string equaled to the length of the string https://github.com/scality/Arsenal/pull/228#discussion_r109256720.
Thanks @jonathan-gramain for pointing out this bug.

### DIFF
--- a/lib/algos/list/delimiterMaster.js
+++ b/lib/algos/list/delimiterMaster.js
@@ -82,8 +82,10 @@ class DelimiterMaster extends Delimiter {
             // - foo/ : skipping foo/
             // - foo  : skipping foo.
             const index = this.NextMarker.lastIndexOf(this.delimiter);
-            return this.NextMarker +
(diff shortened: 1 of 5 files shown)


rahulreddy:
@ironman-machine try

jonathan-gramain:
BTW, any test exists to catch this case?

rahulreddy:
Is `skipping` method used anywhere? When is search for skipping in S3/MetaData/Arsenal I don't see any references.

ttvi:
Expectation: test should fail and fix should not.

ttvi:
Mr. Linter now also doing the job of checking double quote in JSON stringified objects. Good for you, it's good.

ttvi:
Fixes #241.

rahulreddy:
@ironman-machine try

PR #242, opened 2017-04-05
rf: change encoding of version ids [S3C-184]

electrachong:
Removing encryption and switching to hex for URL-safetyness.

### DIFF
--- a/lib/versioning/VersionID.js
+++ b/lib/versioning/VersionID.js
@@ -1,5 +1,3 @@
-const crypto = require('crypto');
-
 // VersionID format:
 //         timestamp  sequential_position  site_id  other_information
 // where:
@@ -45,16 +43,6 @@ const MAX_SQ = Math.pow(10, LENGTH_SQ) - 1; // good for 1 billion ops
 
 // the earliest versionId, used for versions before versioning
 const VID_INF = pad(MAX_TS, TEMPLATE_TS) + pad(MAX_SQ, TEMPLATE_SQ) + SITE_ID;
-
-// constants for versionId encryption; these values (including the encryption
-// password) should never be reconfigured otherwise existing versionIds will
-// not work; think about using Federation's keyfile as the encryption password
-const VID_CRYPTO_ALGORITHM = process.env.VID_CRYPTO_ALGORITHM || 'aes256';
-const VID_CRYPTO_ENCODING = process.env.VID_CRYPTO_ENCODING || 'base64';
-const VID_CRYPTO_ENCODING_TXT = process.env.VID_CRYPTO_ENCODING_TXT || 'utf8';
-// no need to be secured because users cannot actively create a specific version
-// but better off specifying it if there is a need for additional security layer
-const VID_CRYPTO_PASSWORD = process.env.VID_CRYPTO_PASSWORD || 'hangngon';
 
 // internal state of the module
 let prvts = 0; // epoch of the last versionId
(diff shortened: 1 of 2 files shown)


ttvi:
I'm OK with the change.
Here is just FYI:

1. The versionIds may look a lot the same, look at the sample below where each line is a tuple of `raw_vid` `hex_vid`:
```
98508619053651999999PARIS 39383530383631393035333635313939393939395041524953
98508619053546999999PARIS 39383530383631393035333534363939393939395041524953
98508619053446999999PARIS 39383530383631393035333434363939393939395041524953
98508619053346999999PARIS 39383530383631393035333334363939393939395041524953
98508619053245999999PARIS 39383530383631393035333234353939393939395041524953
98508619053145999999PARIS 39383530383631393035333134353939393939395041524953
98508619053044999999PARIS 39383530383631393035333034343939393939395041524953
98508619052944999999PARIS 39383530383631393035323934343939393939395041524953
98508619052844999999PARIS 39383530383631393035323834343939393939395041524953
98508619052743999999PARIS 39383530383631393035323734333939393939395041524953
98508619052643999999PARIS 39383530383631393035323634333939393939395041524953
```
This may make users confused.

2. Existing encryption can also generate hex output with the strings being totally random; encryption also provides a protection layer, the same to what AWS is doing.
Here is the sample of encryption with hex output:
```
98508618840180999999PARIS 45c3d6931d7343f7b8c6aca9361229f4e3dfec0d4778e3ee2e40ad0e399b3197
98508618840072999999PARIS 7596ceb91794e5f6bd4c8ecb90705598923a33b80de6374576d81b84cc826f17
98508618839970999999PARIS 77e32da1d56c2026b8c9d37845c0bc9764d25f8df24a857ff8cfb4838ca67a55
98508618839870999999PARIS 19e25787b4d7077c6f77b231fbee5ed5977337e5edccdd340074402c6e540d05
98508618839769999999PARIS abce55d3a27ca688bab71c77eb686239842c7c21451bcd0ec68711fbe0fb5c5f
98508618839669999999PARIS afedd7d0c398cf04100575a03a15b9dd0cd3ffd2499009dc8d8825902817883e
98508618839568999999PARIS 8cbecf8dbe196c4c5099ac33e3e6eb64e03c9757084adb7d687c9eff01526cad
98508618839467999999PARIS 7d2bdee3ef3630c70e5d8e3450528a9d4d086e99fb691b46e5d3fdc52808453e
98508618839367999999PARIS 21e9cb16622e28fbc60de889594ce7482951bc4162003b2929eb7c45bc53dda5
98508618839267999999PARIS c44c6f96f0b2bc01d38920221274aaa35283b57c8835d310bae074c99ec33972
98508618839166999999PARIS fcff62de47fa6aa3b65bb14f86b2a10d9c559476e11e4e862d624a44d464d3e9
98508618839066999999PARIS 99f62254c95282cf567cb67b88a1e4b231e5052af39e0903b6172cf40604297d
98508618838966999999PARIS ae7c55face6543ffad934e0c0d17bce54ce867c17ebe929a637b8a95a408d354
```

3. Performance-wise, it more than enough for each S3 server to calculate several thousands of encryption activities per second (the throughput of the whole system at the moment); it's 100K ops on my laptop.

4. And finally, if the purpose was to remove the '+' sign from the versionIds then trading all the advantages for that purpose, instead of dealing with the '+' sign, may not be the best approach.

package.json
```diff
@@ -3,7 +3,7 @@
   "engines": {
     "node": "6.9.5"
   },
-  "version": "7.0.0",
+  "version": "7.0.1",
```
LaurenSpiegel: This should be dropped before merging.

lib/versioning/VersionID.js
```diff
@@ -117,35 +105,29 @@ function generateVersionId(info) {
...
  */
-function decrypt(str) {
-    const decipher = crypto.createDecipher(
-            VID_CRYPTO_ALGORITHM, VID_CRYPTO_PASSWORD);
-    let decrypted = decipher.update(str,
-            VID_CRYPTO_ENCODING, VID_CRYPTO_ENCODING_TXT);
-    decrypted += decipher.final(VID_CRYPTO_ENCODING_TXT);
-    return decrypted;
+function decode(str) {
+    const result = Buffer.from(str, 'hex').toString('utf8');
```
LaurenSpiegel: note that this will throw on its own if you give it a number which is ok since we know this function can throw.

lib/versioning/VersionID.js
```diff
@@ -117,35 +105,29 @@ function generateVersionId(info) {
...
+    if (result === '') {
+        throw new TypeError('invalid hex string');
```
rahulreddy: Why do you want to throw in this function. Why not return an error so that it can be checked from the caller function.
```js
return new Error('invalid hex string');
```
```js
if (decode(str) instanceof Error) {
   // do something
}
```

lib/versioning/VersionID.js
rahulreddy: Since we know that this will throw `TypeError` if str is not a string, let's add a check which returns an error
```js
function decode(str) {
    if (typeof str !== 'string') {
        return new Error('invalid input, str is not a string');
    }
    const result = Buffer.from(str, 'hex').toString('utf8');
     ...
}
```

electrachong: This throws a `TypeError` not only if not a string, but also if the string does not meet some conditions as a hex string (might be the number of characters). 
Currently we do a try-except in all the cases we currently call this function; I agree it makes more sense to do a try-except here and return an error in case of exception.

rahulreddy: Cool, yeah wrapping it up try...catch is better.

electrachong: updated

electrachong:
end to end running here:
http://ci.ironmann.io/gh/scality/Integration/11474

electrachong:
@ttvi I like the randomized output in your example and the additional layer of security. But the consensus has been the information stored in the version ids does not justify adding the extra calculations involved in encryption, and the non-randomized hex ids are acceptable.

electrachong:
end-to-end passed; squashed commits and dropped arsenal version bump

tests/unit/versioning/VersionID.js
```diff
@@ -9,7 +9,7 @@ function randkey(length) {
     return key;
 }
 
-describe('test generating versionIds', () => {
+describe.only('test generating versionIds', () => {
```
rahulreddy: Can you remove .only

electrachong: thanks, I removed it

PR #244, opened 2017-04-11
S3C-35 bucketfile versioning support in Arsenal

jonathan-gramain:
What it does: provides different layers of processing requests
- VersioningRequestProcessor: to process versioning information
- WriteCache: to ensure the atomicity and isolation of requests
- WriteGatheringManager: bucketfile's operation batching layer

This versioning support is meant to be used eventually for MetaData as
well, it is a slightly modified port of the original versioning code
in MetaData.

Author: Vinh Tao <vinh.tao@scality.com>

Changes made by: Jonathan Gramain <jonathan.gramain@scality.com>

### DIFF
--- a/lib/versioning/VersionID.js
+++ b/lib/versioning/VersionID.js
@@ -8,45 +8,52 @@
 
 // the lengths of the components in bytes
 const LENGTH_TS = 14; // timestamp: epoch in ms
-const LENGTH_SQ = 6; // position in ms slot
-const LENGTH_ST = 5; // site identifier
+const LENGTH_SEQ = 6; // position in ms slot
+const LENGTH_SITE = 5; // site identifier
 
 // empty string template for the variables in a versionId
 const TEMPLATE_TS = new Array(LENGTH_TS + 1).join('0');
-const TEMPLATE_SQ = new Array(LENGTH_SQ + 1).join('0');
-const TEMPLATE_ST = new Array(LENGTH_ST + 1).join(' ');
+const TEMPLATE_SEQ = new Array(LENGTH_SEQ + 1).join('0');
+const TEMPLATE_SITE = new Array(LENGTH_SITE + 1).join(' ');
 
 /**
- * Padding a string representation of a value with a given template.
- * For example: pad('foo', '00000') equals '00foo' for left padding
- * and 'foo00' for right padding.
+ * Left-pad a string representation of a value with a given template.
+ * For example: pad('foo', '00000') gives '00foo'.
  *
  * @param {any} value - value to pad
  * @param {string} template - padding template
- * @param {boolean} [right] - padding right (default is padding left)
  * @return {string} - padded string
  */
-function pad(value, template, right = false) {
-    if (right) {
-        return `${value}${template}`.slice(0, template.length);
-    }
+function padLeft(value, template) {
     return `${template}${value}`.slice(-template.length);
 }
 
+/**
+ * Right-pad a string representation of a value with a given template.
+ * For example: pad('foo', '00000') gives 'foo00'.
+ *
+ * @param {any} value - value to pad
+ * @param {string} template - padding template
+ * @return {string} - padded string
+ */
+function padRight(value, template) {
+    return `${value}${template}`.slice(0, template.length);
+}
+
 // site identifier, like PARIS, TOKYO; will be trimmed if exceeding max length
-const SITE_ID = pad(process.env.SITE_ID, TEMPLATE_ST, true);
-// const SITE_ID = `${process.env.SITE_ID}${TEMPLATE_ST}`.slice(0, LENGTH_ST);
+const SITE_ID = padRight(process.env.SITE_ID, TEMPLATE_SITE);
 
 // constants for max epoch and max sequential number in the same epoch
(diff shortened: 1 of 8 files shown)


tests/unit/versioning/VersioningRequestProcessor.js
```diff
@@ -0,0 +1,153 @@
...
+            db: 'foo',
+            key: `bar${i}`,
+            value: '{"qux":"quz"}',
+            options: { versioning: true },
+        };
+        setTimeout(() => vsp.put(request, logger, next), i);
+    }, callback);
+}
+
+describe.only('test VSP', () => {
```
LaurenSpiegel: please remove the only

lib/versioning/VersionID.js
```diff
@@ -8,45 +8,52 @@
...
+ * @return {string} - padded string
+ */
+function padRight(value, template) {
+    return `${value}${template}`.slice(0, template.length);
+}
+
 // site identifier, like PARIS, TOKYO; will be trimmed if exceeding max length
-const SITE_ID = pad(process.env.SITE_ID, TEMPLATE_ST, true);
-// const SITE_ID = `${process.env.SITE_ID}${TEMPLATE_ST}`.slice(0, LENGTH_ST);
+const SITE_ID = padRight(process.env.SITE_ID, TEMPLATE_SITE);
```
LaurenSpiegel: this should be 'replication_group_token'. It would be good to have some default beyond undefined and some explanation in the s3 readme for open source users.  Instead of an env variable, could we use the s3 config and pass an argument when calling the generateVersionId function?

tests/unit/versioning/VersioningRequestProcessor.js
jonathan-gramain: Funny, I removed it in the first place, then added it back, my bad :(

lib/versioning/VersioningRequestProcessor.js
```diff
@@ -0,0 +1,404 @@
...
+        this.writeCache.get({ db, key }, logger, (err, data) => {
+            if (err && !err.ObjNotFound) {
+                return callback(err);
+            }
+            // delete the specific version
+            const versionId = options.versionId;
+            const versionKey = formatVersionKey(key, versionId);
+            const ops = [{ key: versionKey, type: 'del' }];
+            // update the master version as PHD if it is the deleting version
+            if (Version.isPHD(data) ||
```
jonathan-gramain: I get the following trace at this point, when doing a batch delete of versioned objects:

```
/git/ironman/S3/node_modules/arsenal/lib/versioning/Version.js:46
        if (value.indexOf('isPHD') < 0) {
                 ^

TypeError: Cannot read property 'indexOf' of undefined
    at Function.isPHD (/git/ironman/S3/node_modules/arsenal/lib/versioning/Version.js:46:18)
    at writeCache.get (/git/ironman/S3/node_modules/arsenal/lib/versioning/VersioningRequestProcessor.js:390:25)
    at queue.(anonymous function).queue.forEach.callback (/git/ironman/S3/node_modules/arsenal/lib/versioning/WriteCache.js:117:28)
    at Array.forEach (native)
    at WriteCache._dequeue (/git/ironman/S3/node_modules/arsenal/lib/versioning/WriteCache.js:112:40)
    at wgm.get (/git/ironman/S3/node_modules/arsenal/lib/versioning/WriteCache.js:65:18)
    at subDb.get (/git/ironman/S3/node_modules/arsenal/lib/network/level-net/index.js:48:28)
    at /git/ironman/S3/node_modules/level-sublevel/shell.js:101:15
    at /git/ironman/S3/node_modules/level-sublevel/nut.js:121:19
```

The direct cause is that the WriteCache returns `ObjNotFound` and a `null` data (as expected with `ObjNotFound`) but we're skipping this error explicitly from error checking and we're using the `data` object afterwards.

I don't know what would be the correct fix here, @ttvi any clue?

jonathan-gramain: `processVersionSpecificPut` seems to have the same issue.

jonathan-gramain: Note that I previously used a non-versioning-aware branch when doing a batch of deletes, so it may have caused inconsistency in my local store. The ObjNotFound error should be handled gracefully though.

ttvi: It's correct not to process `'ObjNotFound'` here because metadata backends don't return any error when deleting a non-existing object. It's just that `Version.isPHD` was not rigid enough; seems like we've never tested the invalid inputs for the function before, I have included that in the new commit.

jonathan-gramain:
**Attention to reviewers**

I'm finishing a refactoring on the level-net side related to the changes done in this PR - the code which handles the packing/unpacking of arguments, creating server and client stubs for remote communication etc. - so don't focus on this side of things in this PR for now.

The goal is two fold: reduce the code mix between RPC logic, levelDB logic and API-specific logic, some of which are currently all in the same place/file leading to code hard to understand (and this is mostly due to the current level-net design, which was a hard fit for the new logic needed here which doesn't help). And it's also to prepare the ground for the future additions that will be needed for backbeat and replication log service.

The versioning specific code is not touched in this refactoring.

jonathan-gramain:
Replaced by: https://github.com/scality/Arsenal/pull/245

PR #245, opened 2017-04-20
S3C-193 bucketfile versioning support in Arsenal

jonathan-gramain:
This replaces PR https://github.com/scality/Arsenal/pull/244 by adding a prior refactoring of level-net, then versioning changes on top of it. The ticket number has been fixed too.

Reviews should be easier reviewing each commit independently rather than as a whole:
 - first commit just reorganizes code around
 - second commit for the level-net refactoring (It might be easier to review the original changed files rather than the diff for this commit)
 - third commit ports the versioning code to bucketfile


lib/network/rpc/level-net.js
```diff
@@ -1,343 +1,109 @@
...
-                        // that async DB sublevel calls expect as
-                        // their last argument.
-                        decodedArgs.rpcArgs.push((err, data) => {
-                            cb(flattenError(err),
-                               mdStreams.encodeStreams(data));
-                        });
+    openSub(subName) {
+        const subDbClient = new LevelDbClient({ url: this.url,
+                                                logger: this.logger });
+        Object.assign(subDbClient, this);
```
electrachong: now that we create the subDbClient as a new instance of LevelDbClient, do we still need to assign properties of `this` to subDbClient?

lib/network/rpc/rpc.js
```diff
@@ -0,0 +1,475 @@
...
+     */
+    connect(cb) {
+        this.socket = ioClient(this.url);
+        this.socketStreams = sioStream.createSocket(
+            this.socket,
+            this.logger,
+            this.streamMaxPendingAck,
+            this.streamAckTimeoutMs);
+        this.socket.on('error', err => {
+            this.logger.warn('connectivity error to storage daemon',
```
electrachong: do you still want to refer to the rpc server as storage daemon?

lib/network/rpc/level-net.js
jonathan-gramain: Yes, we need this because we have to provide the original exposed RPC calls to the new object.

jonathan-gramain: Maybe I should comment this to make it clear.

lib/network/rpc/rpc.js
jonathan-gramain: I makes sense for me, do you think about data-metadata daemon as an alternate name?

electrachong: I was thinking the applications of the rpc server might be more general than accessing the storage (is backbeat queue the same application?). If the only use case is connecting to data storage and it makes sense to you, then feel free to stick with it.

jonathan-gramain: Now I think I get what you mean, because it's in the common RPC layer, it it not necessarily talking to a storage daemon. That is right, even though for now this is what we use it for.

I could change the message to "connectivity error to the RPC server", adding the URL to the error params so that people know where it goes to.

lib/network/rpc/level-net.js
```diff
@@ -0,0 +1,110 @@
...
+     * calls server.registerServices() automatically
+     */
+    constructor(params) {
+        assert(params.rootDb);
+        super(params);
+        this.rootDb = params.rootDb;
+
+        this.addRequestInfoConsumer((dbService, params) => {
+            const env = {};
+            env.subLevel = params.subLevel;
```
electrachong: did you include `params.subLevel` in the jsDoc?

lib/versioning/Version.js
```diff
@@ -48,6 +59,39 @@ class Version {
...
+    /**
+     * Put versionId into an object in the (cheap) way of string manipulation,
+     * instead of the more expensive alternative parsing and stringification.
+     *
+     * @param {string} value - stringification of the object to append versionId
+     * @param {string} versionId - the versionId to append
+     * @return {string} - the object with versionId appended
+     */
+    static appendVersionId(value, versionId) {
+        // assuming value has the format of '{...}'
```
jonathan-gramain: I'm not sure on this one -> @ttvi?

I think this code is also in the MetaData code as-is. I think we can keep it for post-merge, if there's a sanity check to add.

jonathan-gramain: A good timing for the change could be when making MetaData use this common code too (probably in the coming month or two).

lib/network/rpc/level-net.js
jonathan-gramain: Actually this is not the same `params` than in the englobing function (naming confusion could be solved by using different names then).

jonathan-gramain: I realize that it could be useful to have the additional request params passed in the RPC requests documented somewhere in the service definition where it applies.

electrachong: I see. maybe we can call the `params` here `callParams` or `rpcCallParams`

LaurenSpiegel:
end to end passed: http://ci.ironmann.io/gh/scality/Integration/12191

PR #247, opened 2017-04-26
Ft/replication group token

electrachong:
Dependency of https://github.com/scality/S3/pull/708 and https://github.com/scality/MetaData/pull/1075

Provide a way to specify the replication group token (formerly 'site id') for versioning and replication without depending on environmental variables.

The replication group token identifies a group or unit of deployment sites relevant for versioning and replication; for example, it is used to generate version IDs that can be sorted by the identified groups after being replicated in another deployment. We use the term 'replication group' to avoid confusion with regions and location constraints.

The value should be set in the configuration of S3 if using in-memory or bucketfile backends, or the configuration of Metadata if using real metadata backend, so the appropriate backend can pass the token as an argument when generating version IDs.

### DIFF
--- a/lib/versioning/Version.js
+++ b/lib/versioning/Version.js
@@ -1,7 +1,5 @@
 'use strict'; // eslint-disable-line strict
 
-const VID = require('./VersionID');
-
 /**
  * Class for manipulating an object version.
  * The format of a version: { isNull, isDeleteMarker, versionId, otherInfo }
@@ -64,14 +62,11 @@ class Version {
      * to indicate that the master version of an object has been deleted and it
      * needs to be updated by the latest version if any.
      *
-     * @param {string} versionId - the versionId of the version, this expects
-     *                             a fixed length versionId (without info part)
-     *                             to facilitate easy checking if a version is
-     *                             a PHD version; see PHD_VER_LENGTH above
+     * @param {string} versionId - versionId of the PHD version
      * @return {string} - the serialized version
      */
-    static generatePHDVersion() {
-        return `{ "isPHD": true, "versionId": "${VID.generateVersionId('')}" }`;
+    static generatePHDVersion(versionId) {
+        return `{ "isPHD": true, "versionId": "${versionId}" }`;
     }
 
     /**
--- a/lib/versioning/VersionID.js
+++ b/lib/versioning/VersionID.js
@@ -1,20 +1,20 @@
 // VersionID format:
-//         timestamp  sequential_position  site_id  other_information
+//         timestamp  sequential_position  rep_group_id  other_information
 // where:
 // - timestamp              14 bytes        epoch in ms (good untill 5138)
 // - sequential_position    06 bytes        position in the ms slot (1B ops)
-// - site_id                05 bytes        site identifier (like PARIS)
+// - rep_group_id           07 bytes        replication group identifier
 // - other_information      arbitrary       user input, such as a unique string
 
 // the lengths of the components in bytes
 const LENGTH_TS = 14; // timestamp: epoch in ms
 const LENGTH_SEQ = 6; // position in ms slot
-const LENGTH_SITE = 5; // site identifier
+const LENGTH_RG = 7; // replication group id
 
 // empty string template for the variables in a versionId
 const TEMPLATE_TS = new Array(LENGTH_TS + 1).join('0');
 const TEMPLATE_SEQ = new Array(LENGTH_SEQ + 1).join('0');
-const TEMPLATE_SITE = new Array(LENGTH_SITE + 1).join(' ');
+const TEMPLATE_RG = new Array(LENGTH_RG + 1).join(' ');
 
 /**
  * Left-pad a string representation of a value with a given template.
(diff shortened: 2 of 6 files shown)


lib/versioning/Version.js
```diff
@@ -64,14 +64,12 @@ class Version {
      * to indicate that the master version of an object has been deleted and it
      * needs to be updated by the latest version if any.
      *
-     * @param {string} versionId - the versionId of the version, this expects
```
electrachong: Note: This might have been a leftover after refactoring - I did not see this param used anywhere in the function, so I removed it from the jsdoc.

lib/versioning/VersionID.js
```diff
@@ -75,15 +80,19 @@ function wait(span) {
...
  * function is stateful which means it keeps some values in the memory and the
  * next call depends on the previous call.
  *
  * @param {string} info - the additional info to ensure uniqueness if desired
+ * @param {string} replicationGroupToken - replication unit identifier
  * @return {string} - the formated versionId string
  */
-function generateVersionId(info) {
+function generateVersionId(info, replicationGroupToken) {
+    // deployment identifier, like PARIS; will be trimmed if exceed max length
```
LaurenSpiegel: if exceed LENGTH_TOKEN

lib/versioning/VersionID.js
```diff
@@ -9,12 +9,12 @@
 // the lengths of the components in bytes
 const LENGTH_TS = 14; // timestamp: epoch in ms
 const LENGTH_SEQ = 6; // position in ms slot
-const LENGTH_SITE = 5; // site identifier
+const LENGTH_TOKEN = 5; // replication group identifier
```
LaurenSpiegel: @ttvi, can we make this longer?  Why 5?

lib/versioning/VersionID.js
```diff
@@ -40,16 +40,21 @@ function padRight(value, template) {
...
-// the earliest versionId, used for versions before versioning
-const VID_INF = (padLeft(MAX_TS, TEMPLATE_TS) +
-                 padLeft(MAX_SEQ, TEMPLATE_SEQ) + SITE_ID);
+/**
+ * Generates the earliest versionId, used for versions before versioning
+ *
+ * @param {string} replicationGroupToken - replication unit identifier
+ * @return {string} version ID for versions before versionig
+ */
+function generateDummyVersionId(replicationGroupToken) {
```
jonathan-gramain: This function should rather be called `generateEarliestVersionId`, dummy sounds to me like it's only for testing (but I'm not a native English speaker!)

lib/versioning/VersionID.js
ttvi: Yes. It's just an arbitrary, but reasonable, value I chose. It should have a fixed length so that versionIds will have the same length, while not making a versionId too long.

lib/versioning/VersionID.js
```diff
@@ -9,12 +9,12 @@
...
 const LENGTH_TS = 14; // timestamp: epoch in ms
 const LENGTH_SEQ = 6; // position in ms slot
-const LENGTH_SITE = 5; // site identifier
+const LENGTH_TOKEN = 5; // replication group identifier
 
 // empty string template for the variables in a versionId
 const TEMPLATE_TS = new Array(LENGTH_TS + 1).join('0');
 const TEMPLATE_SEQ = new Array(LENGTH_SEQ + 1).join('0');
-const TEMPLATE_SITE = new Array(LENGTH_SITE + 1).join(' ');
+const TEMPLATE_TOKEN = new Array(LENGTH_TOKEN + 1).join(' ');
```
ttvi: I don't see why you need to change the name of the internal variable; using the name you prefer in Federation is enough.

lib/versioning/VersionID.js
ttvi: Please don't change the name I made to something meaningless. Is `getInfVid` ok?

lib/versioning/VersionID.js
```diff
@@ -108,7 +117,7 @@ function generateVersionId(info) {
     // reversed chronological order---newest versions first. This is because of
     // the limitation of leveldb for listing keys in the reverse order.
     return padLeft(MAX_TS - lastTimestamp, TEMPLATE_TS) +
-           padLeft(MAX_SEQ - lastSeq, TEMPLATE_SEQ) + SITE_ID + info;
+           padLeft(MAX_SEQ - lastSeq, TEMPLATE_SEQ) + rgToken + info;
```
ttvi: No need to change the name here.

lib/versioning/VersioningRequestProcessor.js
```diff
@@ -50,8 +50,8 @@ class VersioningRequestProcessor {
         this.repairing = {};
     }
 
-    generateVersionId() {
-        return genVID(this.uidCounter++);
+    generateVersionId(replicationGroupToken) {
+        return genVID(this.uidCounter++, replicationGroupToken);
```
ttvi: You can get the thing you want from config and use it here without the need to get it from input. Then you can keep the rest unchanged. For PHD versions, you can generate the versionId here then pass it to the function `generatePHDVersion()`, making Version.js independent from VersionID.js.

ttvi:
My thrashing comments on style does not mean this is not a very simple but effective PR. Much better than what I intended to do.

lib/versioning/VersioningRequestProcessor.js
electrachong: This is Arsenal, how to you intend to get the value from S3 or Metadata config without passing it along through the API? Do you mean there is an Arsenal-specific config we can use so we can import the value directly?

lib/versioning/VersionID.js
electrachong: I increased to 7 bytes for now, to give just a little more room for naming the replication group.

lib/versioning/VersionID.js
electrachong: It's not something I feel strongly about. I will use `getInfVid`.

lib/versioning/VersionID.js
electrachong: NOTE: `TOKEN` here is a bit vague so I changed it to `LENGTH_RG`.

---
It is to be consistent in our use of semantics. 

Although you intended 'SITE' to be a general notion of geography (not necessarily specific to the datacenter), we opted for replication group over site ID to limit confusion because there is a scenario where there may be two nearby deployment sites considered one unit in terms of replication. Is that correct?

If we name the value `replication_group_token` everywhere else and use the same value to build the version ID, it's strange that we call it `SITE_ID` here.

That said, as you're the author of the original code, I'll respect your wishes if you want to retain `SITE_ID` instead.

electrachong:
end-to-end passed: http://ci.ironmann.io/gh/scality/Integration/12421

lib/versioning/VersioningRequestProcessor.js
ttvi: You have to pass the config value from S3 and metadata, take it as the input when instantiating this class, then set it as the global value to use thereafter.
This may be used by both metadata and bucketfile so no changes to the APIs down here is better.

lib/versioning/VersionID.js
```diff
@@ -75,15 +80,19 @@ function wait(span) {
 /**
  * This function returns a "versionId" string indicating the current time as a
  * combination of the current time in millisecond, the position of the request
- * in that millisecond, and the identifier of the local site (which could be
+ * in that millisecond, and a replication unit identifier (which could be a
```
ttvi: If you don't like the term 'site' you can use 'replica' instead. It's standard.

electrachong:
end-to-end passed here: http://ci.ironmann.io/gh/scality/Integration/12549

PR #252, opened 2017-05-03
Ft/rpc service rest api

jonathan-gramain:
Add a REST server for metadata RPC calls
    
This is meant to be an ease for scripting/debugging, using traditional HTTP tools (curl etc.) rather than a socket.io client. The actual client implementations should still use the socket.io client.

PR #253, opened 2017-05-03
Ft/s3 c 301 record log api

jonathan-gramain:
This is basically implementing the record log functionality in the dmd necessary to enable async replication with backbeat.

A compatible API is being implemented also for MetaData, fetching the updates from the existing transaction log.

The current PR is based on top of https://github.com/scality/Arsenal/pull/252, for conflict management.

### DIFF
--- a/lib/storage/metadata/file/MetadataFileServer.js
+++ b/lib/storage/metadata/file/MetadataFileServer.js
@@ -5,6 +5,7 @@ const assert = require('assert');
 const uuid = require('uuid');
 const level = require('level');
 const sublevel = require('level-sublevel');
+const debug = require('debug')('MetadataFileServer');
 
 const Logger = require('werelogs').Logger;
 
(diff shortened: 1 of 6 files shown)


jonathan-gramain:
This PR is review only for now until we have a clear view of how the log format should look like, to match MetaData log format.

jonathan-gramain:
@ironman-machine try

rahulreddy:
@ironman-machine try

lib/storage/metadata/file/MetadataFileServer.js
```diff
@@ -187,9 +198,24 @@ class MetadataFileServer {
...
-                            err => callback(err));
+                const ops = request.array.map(
+                    op => Object.assign({}, op, { prefix: dbPath }));
+                if (this.recordLog.enabled) {
+                    this.recordLogService
+                        .withRequestParams(
+                            { logName: this.recordLog.recordLogName })
+                        .createLogRecordOps(
+                            ops, (err, logEntries) => {
+                                debug('ops to batch:', ops);
```
jonathan-gramain: The `env` param is special, it is present only on the server-side backend call to pass some environment attributes. It gets passed as the first argument of every backend API function. The client does not see this parameter and does not pass it in the call.

jonathan-gramain: To complete this answer, the `openLog` object from the env is opened on the server from the `logName` attribute sent by the client through `withRequestParams()`.

jonathan-gramain: The same kind of things is done to pass logging attributes like the request ID to the server, so that they don't have to be passed as an extra argument in all the calls.

rahulreddy:
@ironman-machine

PR #256, opened 2017-05-12
bf: use arsenal errors in data store API

jonathan-gramain:
While adding consistency on error management, the goal is also to be able to transmit
these errors properly to the upper layers (above wrapper.js) which expect proper
arsenal errors, so that specific error types can be checked (e.g. ObjNotFound).

+ unrelated: minor fix in trace message in auth module (not worth
doing a separate PR for this one)

Linked to backbeat routes handling on S3 connector, so that data fetching routes can handle proper "not found" error types.

sproxydclient does seem to already provide Arsenal errors, so there should be nothing to do on it.

### DIFF
--- a/lib/storage/data/file/DataFileStore.js
+++ b/lib/storage/data/file/DataFileStore.js
@@ -130,7 +130,8 @@ class DataFileStore {
             if (err) {
                 log.error('error opening filePath',
                           { method: 'put', key, filePath, error: err });
-                return callback(err);
+                return callback(errors.InternalError.customizeDescription(
+                    `filesystem error: open() returned ${err.code}`));
             }
             const cbOnce = jsutil.once(callback);
             // disable autoClose so that we can close(fd) only after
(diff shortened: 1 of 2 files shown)


lib/storage/data/file/DataFileStore.js
```diff
@@ -165,7 +169,8 @@ class DataFileStore {
                           { method: 'put', key, filePath, error: err });
                 // destroying the write stream forces a close(fd)
                 fileStream.destroy();
-                return cbOnce(err);
+                return cbOnce(errors.InternalError.customizeDescription(
+                    `write stream error: ${err}`));
```
rahulreddy: Is this an object of Error instance? If yes, wouldn't this be an `{}` if you stringify it?

rahulreddy:
@ironman-machine

rahulreddy:
End to end passed here http://ci.ironmann.io/gh/scality/Integration/13406

PR #257, opened 2017-05-15
S3C-380: Get security token

alexandre-merle:
This PR introduce the security token needed for
authenticate request with temporary credentials.

See https://scality.atlassian.net/browse/S3C-380

### DIFF
--- a/lib/auth/v4/queryAuthCheck.js
+++ b/lib/auth/v4/queryAuthCheck.js
@@ -1,5 +1,6 @@
 'use strict'; // eslint-disable-line strict
 
+const constants = require('../../constants');
 const errors = require('../../errors');
 
 const constructStringToSign = require('./constructStringToSign');
@@ -23,6 +24,15 @@ function check(request, log, data) {
     if (Object.keys(authParams).length !== 5) {
         return { err: errors.InvalidArgument };
     }
+
+    // Query params are not specified in AWS documentation as case-insensitive,
+    // so we use case-sensitive
+    const token = data['X-Amz-Security-Token'];
+    if (token && !constants.iamSecurityToken.pattern.test(token)) {
+        log.debug('invalid security token', { token });
+        return { err: errors.InvalidToken };
+    }
+
     const signedHeaders = authParams.signedHeaders;
     const signatureFromRequest = authParams.signatureFromRequest;
     const timestamp = authParams.timestamp;
@@ -95,6 +105,7 @@ function check(request, log, data) {
                 authType: 'REST-QUERY-STRING',
                 signatureVersion: 'AWS4-HMAC-SHA256',
                 signatureAge: Date.now() - convertAmzTimeToMs(timestamp),
+                securityToken: token,
             },
         },
     };
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -1,6 +1,21 @@
 'use strict'; // eslint-disable-line strict
 
+// The min value here is to manage further backward compat if we
+// need it
+const iamSecurityTokenSizeMin = 128;
+const iamSecurityTokenSizeMax = 128;
+// Security token is an hex string (no real format from amazon)
+const iamSecurityTokenPattern =
+    new RegExp(`^[a-f0-9]{${iamSecurityTokenSizeMin},` +
+        `${iamSecurityTokenSizeMax}}$`);
+
 module.exports = {
+    // info about the iam security token
+    iamSecurityToken: {
+        min: iamSecurityTokenSizeMin,
+        max: iamSecurityTokenSizeMax,
+        pattern: iamSecurityTokenPattern,
+    },
     // PublicId is used as the canonicalID for a request that contains
     // no authentication information.  Requestor can access
     // only public resources
--- a/lib/policyEvaluator/RequestContext.js
+++ b/lib/policyEvaluator/RequestContext.js
@@ -160,6 +160,7 @@ function _buildArn(service, generalResource, specificResource, requesterInfo) {
  * @param {string} signatureVersion - auth signature type used
  * @param {string} authType - type of authentication used
  * @param {number} signatureAge - age of signature in milliseconds
+ * @param {string} securityToken - auth security token (temporary credentials)
  * @return {RequestContext} a RequestContext instance
  */
 
@@ -167,7 +168,7 @@ class RequestContext {
     constructor(headers, query, generalResource, specificResource,
         requesterIp, sslEnabled, apiMethod,
         awsService, locationConstraint, requesterInfo,
-        signatureVersion, authType, signatureAge) {
+        signatureVersion, authType, signatureAge, securityToken) {
         this._headers = headers;
         this._query = query;
         this._requesterIp = requesterIp;
@@ -192,6 +193,7 @@ class RequestContext {
         this._signatureVersion = signatureVersion;
         this._authType = authType;
         this._signatureAge = signatureAge;
+        this._securityToken = securityToken;
 
         return this;
     }
@@ -218,6 +220,7 @@ class RequestContext {
             signatureAge: this._signatureAge,
             locationConstraint: this._locationConstraint,
             tokenIssueTime: this._tokenIssueTime,
+            securityToken: this._securityToken,
         };
         return JSON.stringify(requestInfo);
     }
@@ -238,7 +241,7 @@ class RequestContext {
             obj.specificResource, obj.requesterIp, obj.sslEnabled,
             obj.apiMethod, obj.awsService, obj.locationConstraint,
             obj.requesterInfo, obj.signatureVersion,
-            obj.authType, obj.signatureAge);
+            obj.authType, obj.signatureAge, obj.securityToken);
     }
 
     /**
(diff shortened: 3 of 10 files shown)


lib/auth/v4/queryAuthCheck.js
```diff
@@ -23,6 +24,13 @@ function check(request, log, data) {
     if (Object.keys(authParams).length !== 5) {
         return { err: errors.InvalidArgument };
     }
+
+    const token = data['X-Amz-Security-Token'];
```
rahulreddy: `X-Amz-Security-Token` << Why is this not lowercased?

lib/constants.js
```diff
@@ -1,6 +1,21 @@
 'use strict'; // eslint-disable-line strict
 
+// The min value here is to manage further backward compat if we
+// need it
+const iamSecurityTokenSizeMin = 128;
```
rahulreddy: sha256 hex length is 64, shouldn't that be sufficient?

lib/auth/v4/queryAuthCheck.js
alexandre-merle: because the name is not lowercased

lib/constants.js
alexandre-merle: the security token is not a hash, i put 128 for now, if we want to encode some informations inside it, we can

lib/constants.js
```diff
@@ -1,6 +1,21 @@
...
+const iamSecurityTokenSizeMax = 128;
+// Security token is an hex string (no real format from amazon)
+const iamSecurityTokenSizePattern =
```
jonathan-gramain: Don't you mean `iamSecurityTokenPattern` for this variable?

tests/unit/auth/v2/headerAuthCheck.js
```diff
@@ -0,0 +1,40 @@
...
+            headers: {
+                'x-amz-security-token': 'invalid-token',
+            },
+        };
+        const res = headerAuthCheck(request, log, {});
+        assert.notStrictEqual(res.err, undefined);
+        assert.strictEqual(res.err.InvalidToken, true);
+    });
+
+    it('should succeed if the token is valid', () => {
```
jonathan-gramain: minor: maybe precise that the token is only *syntactically* valid, not an actual valid token.

tests/unit/auth/v4/headerAuthCheck.js
```diff
@@ -36,6 +36,35 @@ const request = {
...
+    });
+
+    it('shoud succeed on valid token', () => {
+        const clock = lolex.install(1454962445000);
+        const alteredRequest = createAlteredRequest(
+            { 'x-amz-security-token': 'a'.repeat(128) },
+            'headers', request, headers);
+        const res = headerAuthCheck(alteredRequest, log);
+        clock.uninstall();
+        if (res.err) {
```
jonathan-gramain: Why not testing the expected value of `res.err` here

lib/auth/v4/queryAuthCheck.js
alexandre-merle: it is a query parameter :) i don't think this deserve a comment ^^

lib/constants.js
```diff
@@ -1,6 +1,21 @@
...
+    new RegExp(`^[a-f0-9]{${iamSecurityTokenSizeMin},` +
```
alexandre-merle: not sure if its necessary to allow uppercases letters, since its only internal, we can generate what we want, but why not

tests/unit/auth/v4/headerAuthCheck.js
alexandre-merle: same ^^ if i test it here i have to throw in the next line, instead of the current assert

tests/unit/auth/v2/headerAuthCheck.js
alexandre-merle: the notion of valid / invalid token will not be manage by arsenal, from here a token is valid when the syntaxe is correct :)

lib/constants.js
alexandre-merle: right :) will change it

lib/policyEvaluator/RequestContext.js
```diff
@@ -509,6 +512,26 @@ class RequestContext {
...
+        return this._securityToken;
+    }
+
+    /**
+     * Set the authentication security token
+     *
+     * @param {string} token - Security token
+     * @return {RequestContext} itself
+     */
+    setSecurityToken(token) {
```
AntoninCoulibaly: it could'nt be a good idea here to test the pattern?

tests/unit/auth/v2/headerAuthCheck.js
```diff
@@ -0,0 +1,40 @@
+const assert = require('assert');
+
+const headerAuthCheck =
+    require('../../../../lib/auth/v2/headerAuthCheck').check;
+const DummyRequestLogger = require('../../helpers').DummyRequestLogger;
+
+const log = new DummyRequestLogger();
+
+describe('v2: headerAuthCheck', () => {
+    it('should not fail if the token is missing', () => {
```
AntoninCoulibaly: can you factorize here?

```js
// 0 -> assert.notStrictEqual
// 1 -> assert.StrictEqual
const tests = [
    ['should do X', {}, 'strictEqual'],
    ['should not fail if the token is missing', {}, 'notStrictEqual'],
]

tests.forEach(test => {
    // your test
    assert[test[2]](x, true);
});
```
You can replace array by object for making it more readable

tests/unit/auth/v2/queryAuthCheck.js
```diff
@@ -0,0 +1,33 @@
+const assert = require('assert');
+
+const queryAuthCheck =
+    require('../../../../lib/auth/v2/queryAuthCheck').check;
+const DummyRequestLogger = require('../../helpers').DummyRequestLogger;
+
+const log = new DummyRequestLogger();
+
+describe('v2: queryAuthCheck', () => {
+    it('should not fail if the token is missing', () => {
```
AntoninCoulibaly: same as before 
```js
// 0 -> assert.notStrictEqual
// 1 -> assert.StrictEqual
const tests = [
    ['should do X', {}, 'strictEqual'],
    ['should not fail if the token is missing', {}, 'notStrictEqual'],
]

tests.forEach(test => {
    // your test
    assert[test[2]](x, true);
});
```

tests/unit/auth/v4/headerAuthCheck.js
```diff
@@ -36,6 +36,35 @@ const request = {
 };
 
 describe('v4 headerAuthCheck', () => {
+    it('should succeed if missing token', () => {
```
AntoninCoulibaly: same 
```js
// 0 -> assert.notStrictEqual
// 1 -> assert.StrictEqual
const tests = [
    ['should do X', {}, 'strictEqual'],
    ['should not fail if the token is missing', {}, 'notStrictEqual'],
]

tests.forEach(test => {
    // your test
    assert[test[2]](x, true);
});
```

lib/policyEvaluator/RequestContext.js
alexandre-merle: no need to :) it will be already check by vault :)

lib/auth/v4/queryAuthCheck.js
jonathan-gramain: It's not obvious to me that AWS (hence our implementation as well) expects query params to be case sensitive, while headers are not wrt HTTP spec, so I don't think a comment would hurt (not necessarily on this particular header though).

This being said, it's fine for me if you don't feel the need for a comment.

tests/unit/auth/v2/queryAuthCheck.js
```diff
@@ -0,0 +1,28 @@
...
+        const data = {
+            SecurityToken: test.token,
+        };
+        const res = queryAuthCheck(request, log, data);
+        if (test.error) {
+            assert.notStrictEqual(res.err, undefined);
+            assert.strictEqual(res.err.InvalidToken, true);
+        } else {
+            assert.notStrictEqual(res.err, undefined);
+            assert.notStrictEqual(res.err.InvalidToken, true);
```
jonathan-gramain: Is it possible to check that the error is `AccessDenied` here like done for v4?

lib/auth/v4/queryAuthCheck.js
alexandre-merle: all the actual params are case-sensitive, nothing in AWS says its case-insensitive, from that, since we do not have any issue identified for that, i assume all the tools follow this rule :)

tests/unit/auth/v2/queryAuthCheck.js
alexandre-merle: sure :)

lib/auth/v4/queryAuthCheck.js
alexandre-merle: comment added :)

alexandre-merle:
@jonathan-gramain updated

alexandre-merle:
@rahulreddy can you approve if you are okay ? :)

alexandre-merle:
@ironman-machine clean try

PR #262, opened 2017-06-06
ft: add a new mode in versioning put request for backbeat replication

jonathan-gramain:
When both 'versioning' and 'versionId' options are provided, write a
new version with the specified versionId, and also create or update
the master version like done for new versioned puts.

This is for backbeat to write new specific versions to the destination, while updating master as well

@ttvi Is it a reasonable implementation for you, or do you think it should be made another way?

### DIFF
--- a/lib/versioning/VersioningRequestProcessor.js
+++ b/lib/versioning/VersioningRequestProcessor.js
@@ -253,21 +253,14 @@ class VersioningRequestProcessor {
         // valid combinations of versioning options:
         // - !versioning && !versionId: normal non-versioning put
         // - versioning && !versionId: create a new version
-        // - !versioning && versionId: update (PUT/DELETE) an existing version
-        //                             if versionId === '' update master version
-        // other cases are invalid and metadata does nothing
+        // - versionId: update (PUT/DELETE) an existing version, and
+        //              also update master version in case the put
+        //              version is newer or same version than master.
+        //              if versionId === '' update master version
         const versioning = options &&
             (options.versioning || options.versioning === '');
         const versionId = options &&
             (options.versionId || options.versionId === '');
-        if (versioning && versionId) {
-            return callback(errors.BadRequest.message);
-        }
-        // no versioning or versioning configuration off
-        if (!versioning && !versionId) {
-            return this.writeCache.batch(
-                    { db, array: [{ key, value }] }, logger, callback);
-        }
 
         const versioningCb = (err, array, vid) => {
             if (err) {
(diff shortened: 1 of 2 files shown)


ttvi:
Before going into the details of the code, could it be clear that:
1. Applying remote updates should be throttled so that not affecting local operations; this implies there should be a different route somewhere for remote updates so that we can control them separately (at the replication processing layer for example).
2. Versioning processing layers are to be used commonly for both metadata and file backends. So it's good if you think this structure is general enough and can provide the previously mentioned route for both backends.

ttvi:
For the code itself, I think remote updates should be classified as version specific puts. You need to take care of the order between local and remote updates as well.

What you need to do is changing the check [`Version.from(data).getVersionId() === versionId`](https://github.com/scality/Arsenal/pull/262/files#diff-8a95352c1a2cc2406f6dc4cb77a4f23bR349) into `Version.from(data).getVersionId() <= versionId`.

jonathan-gramain:
Thanks for your feedback @ttvi. I modified the code to reflect this.

For the throttling, I agree, though I think we should implement this as an additional option/hint, not based on what versioning parameters are passed to the request. S3 uses a separate set of routes for backbeat puts, so those could add an additional option to the versioning calls to deprioritize the requests over local ones.

rahulreddy:
@jonathan-gramain @ttvi 
I agree that we need to do some throttling down the line. It will be easier if it is implemented on a higher level like Backbeat where we can leverage operational metrics to make decisions on the throttling numbers.
@ttvi Thanks for the input on the implementation.

rahulreddy:
@jonathan-gramain Can you add a unit test asserting the behavior?

jonathan-gramain:
@rahulreddy Thanks for suggesting a test, I realized that my version comparison test was inverted (greater versions are actually older). We would probably have caught this during the next round of end-to-end tests but still good to catch it earlier.

jonathan-gramain:
@ironman-machine try

lib/versioning/VersioningRequestProcessor.js
```diff
@@ -255,19 +255,13 @@ class VersioningRequestProcessor {
         // - versioning && !versionId: create a new version
         // - !versioning && versionId: update (PUT/DELETE) an existing version
         //                             if versionId === '' update master version
-        // other cases are invalid and metadata does nothing
+        // - versioning && versionId: create a new version which has
+        //   the specified versionId (e.g. for async replication)
+
         const versioning = options &&
             (options.versioning || options.versioning === '');
```
electrachong: I know this is from before but if we expect versioning to be true/false, why do we interpret `versioning: ''` as `versioning: true`?

lib/versioning/VersioningRequestProcessor.js
```diff
@@ -255,19 +255,13 @@ class VersioningRequestProcessor {
         // - versioning && !versionId: create a new version
         // - !versioning && versionId: update (PUT/DELETE) an existing version
         //                             if versionId === '' update master version
-        // other cases are invalid and metadata does nothing
+        // - versioning && versionId: create a new version which has
+        //   the specified versionId (e.g. for async replication)
```
electrachong: I'm not sure I agree with this description, from what I understand the `versioning` option is not checked to create a new version that also updates the master. Isn't it more like the following?
```
         // - !versioning && versionId: update or create (PUT/DELETE) a version
         //                             if versionId === '' update master version
         //                             if versionId => master version ID, update master version
         // - versioning && versionId: same as !versioning && versionId
```
?

electrachong: But as we were discussing, I think the extension of the behavior is OK and doesn't impact existing operations in S3.

jonathan-gramain: comment updated

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
Integration tests OK

PR #263, opened 2017-06-06
ft: add new action type 'objectReplicate'

jonathan-gramain:
This action maps to a standard policy action 's3:ReplicateObject', that will be
granted to backbeat on the destination.


rahulreddy:
@ironman-machine 
DEFAULT_BRANCH=master
SCALITY_S3_BRANCH=ft/S3-REP.1

PR #267, opened 2017-06-11
fix(net): http server remove listeners before to rebind events

ThibaultRiviere:
In case of error the old listeners was not removed before to restart the
server leading to some leaks and useless calls

### DIFF
--- a/lib/network/http/server.js
+++ b/lib/network/http/server.js
@@ -325,15 +325,17 @@ class Server {
                 this._server = http.createServer(
                     (req, res) => this._onRequest(req, res));
             }
+
+            this._server.on('error', err => this._onError(err));
+            this._server.on('secureConnection',
+                sock => this._onSecureConnection(sock));
+            this._server.on('tlsClientError', (err, sock) =>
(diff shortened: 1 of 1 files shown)


ThibaultRiviere:
```js
const arsenal = require('arsenal');
var s = new arsenal.network.http.server(4242, {
    info: () => {},
    error: () => {},
}).onError(() => {
    console.log('Err');
    return true;
}).start();

```

With binding the port 4242 before to run this script we can see that the log `err` will be called more times at each errors

ThibaultRiviere:
The listen part have been move since the function was added to the listeners leading to that 
```
(node:12999) Warning: Possible EventEmitter memory leak detected. 11 listening listeners added. Use emitter.setMaxListeners() to increase limit
```

ThibaultRiviere:
@rahulreddy if you can merge this one

rahulreddy:
@ThibaultRiviere Thanks. I will review this.

rahulreddy:
@ironman-machine

PR #268, opened 2017-06-12
Rf/extract s3routes [S3C-511]

electrachong:
Dependency of https://github.com/scality/S3/pull/768

To better reference changes from S3 code, view https://github.com/scality/Arsenal/commit/1ea0e96273183965458ffe5e67a4590ad1ab7a56 .

### DIFF
--- a/lib/s3routes/routes.js
+++ b/lib/s3routes/routes.js
@@ -1,0 +1,210 @@
+const assert = require('assert');
+
+const errors = require('../errors');
+const routeGET = require('./routes/routeGET');
+const routePUT = require('./routes/routePUT');
+const routeDELETE = require('./routes/routeDELETE');
+const routeHEAD = require('./routes/routeHEAD');
+const routePOST = require('./routes/routePOST');
+const routeOPTIONS = require('./routes/routeOPTIONS');
+const routesUtils = require('./routesUtils');
+const routeWebsite = require('./routes/routeWebsite');
+
+const routeMap = {
+    GET: routeGET,
+    PUT: routePUT,
+    POST: routePOST,
+    DELETE: routeDELETE,
+    HEAD: routeHEAD,
+    OPTIONS: routeOPTIONS,
+};
+
+function checkUnsupportedRoutes(reqMethod, reqQuery, unsupportedQueries, log) {
+    const method = routeMap[reqMethod];
+    if (!method) {
+        return { error: errors.MethodNotAllowed };
+    }
+
+    if (routesUtils.isUnsupportedQuery(reqQuery, unsupportedQueries)) {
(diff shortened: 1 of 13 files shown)


DavidPineauScality:
Naive question as to the context of this code extraction: What is this code going to be used for, aside S3 itself ? (Side thought is: Arsenal is becoming fatter and fatter, and we're going to pull so much code with it that I'm not sure it will still make sense to share it between all projects)

electrachong:
@DavidPineauScality We're working on an S3-to-Azure translator. It's the only project that will utilize the routes other than S3 at least for now. Arsenal was a convenient place to stick the shared code, but if you have a better idea, :)

DavidPineauScality:
Oh, so this means that the S3-to-Azure  translator will not be part of S3 but another component/service then ? I thought it'd be a backend to S3, not a full fledged implementation :)

Rather than an idea, it's just an observation for now. The ideal practice would be to split Arsenal into various projects, according to the subject/aim of each bit (much like the usual nodeJS habits); but that would mean multiplying libs and repositories, which may not be ideal at the moment; so I don't think it's really a thing; I was mostly raising that Arsenal has left the domain of the "utility library" long ago, and it's becoming closer and closer to a "Core library" for a lot of our components (meaning that we might pull a lot of code for only one part of it).

My question mostly aimed at understanding what's happening around S3/Arsenal currently, so my remark was nothing that should prevent any kind of ongoing work; simple curiosity. Thanks for the insight :)

lib/s3routes/routes.js
```diff
@@ -0,0 +1,136 @@
...
+    if ((reqQuery.partNumber || reqQuery.uploadId)
+        && objectKey === undefined) {
+        return errors.InvalidRequest
+            .customizeDescription('A key must be specified');
+    }
+    return undefined;
+}
+
+function routes(req, res, params, logger) {
+    const {
```
LaurenSpiegel: do you think we should do an assertion that each of these are here so the first request fails if anything is missing rather than waiting for some individual request later to fail?  also, since this is a major exported API, we should document the signature really well and assertions would help do that in addition to the jsdoc.

electrachong: Yeah, I think that's not a bad idea.

electrachong:
end-to-end passed: http://ci.ironmann.io/gh/scality/Integration/13542

rahulreddy:
@DavidPineauScality We are temporarily parking routes, bucketfile backend in Arsenal. Ideally, once we have time we would want to split them into their own modules. It's not ideal right now as the setup would mean orchestrating a lot of changes across the projects. Arsenal should be back to becoming a utility library after that.

PR #271, opened 2017-06-20
ft: extract s3auth utils

electrachong:
Dependency of https://github.com/scality/S3/pull/779

Similar to the recent extracting routes pull request: we are extracting some of the authentication functions from S3 for use in common projects.


tests/unit/auth/s3/indexer.js
```diff
@@ -0,0 +1,78 @@
+const assert = require('assert');
+
+const Indexer = require('../../../../lib/auth/auth').s3.inMemory.indexer;
+const ref = require('./sample_authdata.json');
+const { should } = require('./checker');
+
+describe('lib/auth/in_memory/index.js', () => {
```
electrachong: need to update this

electrachong:
@ironman-machine try

lib/auth/in_memory/indexer.js
```diff
@@ -0,0 +1,165 @@
...
+            accessKey: {},
+            email: {},
+        };
+        this.usersBy = {
+            accessKey: {},
+            email: {},
+        };
+
+        /*
+         * This may happen if the file backend is not configured for S3.
```
LaurenSpiegel: should update this

electrachong: You mean to delete it since we check the authdata before we create the indexer, right?

lib/auth/in_memory/backend.js
```diff
@@ -3,16 +3,37 @@
...
-const accountsKeyedbyCanID =
-    require('./vault.json').accountsKeyedbyCanID;
-const accountsKeyedbyEmail =
-    require('./vault.json').accountsKeyedbyEmail;
 const calculateSigningKey = require('./vaultUtilities').calculateSigningKey;
 const hashSignature = require('./vaultUtilities').hashSignature;
+const Indexer = require('./indexer');
+
+function _buildArn(generalResource, specificResource) {
+    return `arn:aws:s3:::${generalResource}/${specificResource}`;
```
rahulreddy: If you make service, here s3 in the arn configurable, this can be used by some other lightweight S3 we may build (e.g. Backbeat)

lib/auth/in_memory/backend.js
```diff
@@ -3,16 +3,37 @@
...
+ *
+ * @class Backend
+ */
+const backend = class Backend {
+    /**
+     * @constructor
+     * @param {object} authdata - the authentication config file's data
+     */
+    constructor(authdata) {
+        this.indexer = new Indexer(authdata);
```
rahulreddy: The constructor could take service as an input and use it for building ARN.

lib/auth/in_memory/backend.js
```diff
@@ -3,16 +3,37 @@
...
+    };
+}
+
+/**
+ * Class that provides a memory backend for verifying signatures and getting
+ * emails and canonical ids associated with an account.
+ *
+ * @class Backend
+ */
+const backend = class Backend {
```
rahulreddy: It's unnecessary to create a const as well as define a named class. I recommend 
```js
class Backend {

}
module.exports = Backend;
```
Btw, make sure you name the file Backend.js, as by convention,  the capitalization of filenames  signifies classes.

lib/auth/in_memory/backend.js
```diff
@@ -3,16 +3,37 @@
...
+    /**
+     * @constructor
+     * @param {object} authdata - the authentication config file's data
```
rahulreddy: What are the properties of this object?

lib/auth/in_memory/checker.js
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    authdata.accounts.forEach(account => {
+        checkAccount(checkData, account, checkSas);
+    });
+
+    return dumpErrors(checkData, log);
+}
+
+module.exports = check;
```
rahulreddy: I think `check` is too generic, may be call this validateAuthConfig or something like that. Make sure your filename reflects the same name as the method you are exporting.

lib/auth/in_memory/indexer.js
```diff
@@ -0,0 +1,166 @@
+/**
+ * Class that provides an internal indexing over the simple data provided by
+ * the authentication configuration file for the memory backend. This allows
+ * accessing the different authentication entities through various types of
+ * keys.
+ *
+ * @class Indexer
+ */
+module.exports = class Indexer {
```
rahulreddy: Same here.
```js
class Indexer {

}
module.exports = Indexer;
```

lib/auth/in_memory/checker.js
```diff
@@ -0,0 +1,195 @@
...
+                value: key,
+                count: obj[key],
+            });
+            ++count;
+        }
+    });
+    return count;
+}
+
+function dumpErrors(checkData, log) {
```
rahulreddy: Please use `_` to indicate internal function names.

lib/auth/in_memory/indexer.js
```diff
@@ -0,0 +1,166 @@
...
+            accessKey: {},
+            email: {},
+        };
+
+        /*
+         * This may happen if the application is configured to use another
+         * authentication backend than in-memory.
+         * As such, we're managing the error here to avoid screwing up there.
+         */
+        if (!authdata) {
```
rahulreddy: Since this happens at startup time, doesn't having an assertion solve the problem?

lib/auth/vault.js
```diff
@@ -0,0 +1,265 @@
...
+                code: 200,
+                message: 'OK',
+                body: obj,
+            };
+            return callback(null, respBody);
+        });
+    }
+}
+
+module.exports = Vault;
```
rahulreddy: Please use the same letter case ~~vault~~ Vault for the filename.

tests/unit/auth/in_memory/checker.js
```diff
@@ -0,0 +1,161 @@
...
+    done();
+}
+
+function shouldSuccess(obj, checkSas, done) {
+    const res = checker(obj, loggerConfig, checkSas);
+    assert.strictEqual(res, false);
+    done();
+}
+
+const should = {
```
rahulreddy: This is beautiful 💯

lib/auth/in_memory/backend.js
electrachong: I made an underlying class that also takes an indexer / formatter as constructor parameters since it's possible we'll want to extend the backend / indexer for Mystique. Not sure, just in case.

lib/auth/in_memory/backend.js
electrachong: updated

lib/auth/in_memory/indexer.js
electrachong: I think we always start up the in-memory backend, and initialize the indexer, even if the service is configured to use actual Vault and there is no `config.authData`. So we don't want to crash the service in that case.

PR #274, opened 2017-06-22
cleanup: generalize internal routes handling

jonathan-gramain:
Every url starting with '/_/' is now routed through a registered internal service handler in the routing code, instead of being specifically checked for a particular service. This will be useful to integrate backbeat routes properly in the next step.

### DIFF
--- a/lib/s3routes/routes/routePUT.js
+++ b/lib/s3routes/routes/routePUT.js
@@ -62,7 +62,7 @@ function routePUT(request, response, api, log, statsClient) {
                     routesUtils.responseNoBody(err, corsHeaders, response, 200,
                         log);
                 });
-        } else if (request.query.acl === undefined) {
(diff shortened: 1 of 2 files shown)


lib/s3routes/routes/routePUT.js
```diff
@@ -62,7 +62,7 @@ function routePUT(request, response, api, log, statsClient) {
                     routesUtils.responseNoBody(err, corsHeaders, response, 200,
                         log);
                 });
-        } else if (request.query.acl === undefined) {
```
rahulreddy: Is this change accidental?

electrachong:
then can you change [this line in Mystique](https://github.com/scality/Mystique/blob/master/lib/server.js#L36) to
`internalHandlers: {},`
? since it imports routes as well :)

PR #276, opened 2017-06-23
fix: missing parameter in buildArn call

electrachong:
Dependency of https://github.com/scality/S3/pull/779

Changes in s3 to use this module have not been merged yet since this fix was needed before regular ci tests (in-memory auth backend) would pass.


electrachong:
@rahulreddy updated to add a test

PR #278, opened 2017-06-23
Rf/unsupported check

electrachong:
Dependency of https://github.com/scality/S3/pull/788
Since S3-related projects may diverge on support and rely on different parameters, move check for unsupported queries and headers to api.


electrachong:
@ironman-machine

electrachong:
@ironman-machine try

electrachong:
End to end with s3 branch here: http://ci.ironmann.io/gh/scality/Integration/13690

PR #280, opened 2017-06-26
add raft client on top of bucketclient API

jonathan-gramain:
Use this client to fetch logs from bucketd in a way consistent between MetaData and bucketfile.

### DIFF
--- a/index.js
+++ b/index.js
@@ -64,6 +64,8 @@ module.exports = {
             require('./lib/storage/metadata/file/MetadataFileServer'),
             MetadataFileClient:
             require('./lib/storage/metadata/file/MetadataFileClient'),
+            LogConsumer:
+            require('./lib/storage/metadata/bucketclient/LogConsumer'),
         },
         data: {
             file: {
--- a/lib/storage/metadata/bucketclient/LogConsumer.js
+++ b/lib/storage/metadata/bucketclient/LogConsumer.js
@@ -1,0 +1,125 @@
+'use strict'; // eslint-disable-line
+
+const stream = require('stream');
+
+const Logger = require('werelogs').Logger;
+
+const errors = require('../../../errors');
+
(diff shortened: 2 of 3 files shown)


DavidPineauScality:
I was wary of the name `RaftClient`, but it seems it's only a way to keep the state of the RaftSessionID along with a bucketclient instance, right ? Then it's okay (though I'd have prefered another name such as `RaftLogConsumer` or something showing a difference between 'RAFT' and the usage of 'Raft logs'

lib/storage/metadata/bucketclient/RecordLogProxy.js
```diff
@@ -0,0 +1,108 @@
...
+                    if (err.code === 404) {
+                        // no such raft session, log and ignore
+                        this.logger.warn(
+                            'This raft session does not exist (yet)',
+                            { raftId: this.raftSession });
+                        return cbOnce(null, { info: { start: null,
+                                                      end: null } });
+                    }
+                    if (err.code === 416) {
+                        // requested range not satisfiable
```
rahulreddy: Let's have a trace log here.

lib/storage/metadata/bucketclient/RecordLogProxy.js
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    _transform(itemObj, encoding, callback) {
+        itemObj.entries.forEach(entry => {
+            // eslint-disable-next-line no-param-reassign
+            entry.type = entry.type || 'put';
+        });
+        this.logger.debug('processing log item', { itemObj });
+        this.push(itemObj);
+        callback();
```
jonathan-gramain: This should be ok, it's called by the node engine internally and the doc examples just call the callback directly as a way to notify that we processed the items.

jonathan-gramain:
@ironman-machine try

PR #290, opened 2017-06-30
fix: continue handling req after writeContinue

electrachong:
After we send the write continue response to the client, we should handle the request normally.


electrachong:
@ironman-machine

electrachong:
@rahulreddy end-to-end passed

PR #291, opened 2017-07-03
ft: add JSON error response support

jonathan-gramain:
In addition to XML error response, JSON response will be used by
backbeat routes, because their success responses is in JSON
format. Having JSON as success and XML as error format confuses the
AWS client going to be used as the client for backbeat routes, as it
expects and can be configured for one or the other, not both.

### DIFF
--- a/lib/s3routes/routesUtils.js
+++ b/lib/s3routes/routesUtils.js
@@ -49,71 +49,145 @@ function okHeaderResponse(headers, response, httpCode, log) {
     });
 }
 
-/**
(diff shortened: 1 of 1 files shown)


jonathan-gramain:
@ironman-machine try

jonathan-gramain:
Integration passed in http://ci.ironmann.io/gh/scality/Integration/13856#build-parameters

jonathan-gramain:
replaced by https://github.com/scality/Arsenal/pull/292

PR #292, opened 2017-07-05
ft: add JSON error response support

jonathan-gramain:
In addition to XML error response, JSON response will be used by
backbeat routes, because their success responses is in JSON
format. Having JSON as success and XML as error format confuses the
AWS client going to be used as the client for backbeat routes, as it
expects and can be configured for one or the other, not both.

### DIFF
--- a/lib/s3routes/routesUtils.js
+++ b/lib/s3routes/routesUtils.js
@@ -49,71 +49,135 @@ function okHeaderResponse(headers, response, httpCode, log) {
     });
 }
 
-/**
- * okXMLResponse - Response with XML body
- * @param {string} xml - XML body as string
- * @param {object} response - http response object
- * @param {object} log - Werelogs logger
- * @param {object} additionalHeaders -- additional headers to add to response
- * @return {object} response - response object with additional headers
- */
-function okXMLResponse(xml, response, log, additionalHeaders) {
-    const bytesSent = Buffer.byteLength(xml);
-    log.trace('sending success xml response');
-    log.addDefaultFields({
-        bytesSent,
-    });
-    setCommonResponseHeaders(additionalHeaders, response, log);
-    response.writeHead(200, { 'Content-type': 'application/xml' });
-    log.debug('response http code', { httpCode: 200 });
-    log.trace('xml response', { xml });
-    return response.end(xml, 'utf8', () => {
-        log.end().info('responded with XML', {
-            httpCode: response.statusCode,
-        });
-    });
-}
+const XMLResponseBackend = {
 
-function errorXMLResponse(errCode, response, log, corsHeaders) {
-    log.trace('sending error xml response', { errCode });
-    /*
-    <?xml version="1.0" encoding="UTF-8"?>
-     <Error>
-     <Code>NoSuchKey</Code>
-     <Message>The resource you requested does not exist</Message>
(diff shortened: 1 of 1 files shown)


jonathan-gramain:
Integration passed in http://ci.ironmann.io/gh/scality/Integration/13856#build-parameters (on same branch with previous name ft/jsonErrorResponse)

lib/s3routes/routesUtils.js
```diff
@@ -49,71 +49,145 @@ function okHeaderResponse(headers, response, httpCode, log) {
...
+        log.trace('sending error json response', { errCode });
+        /*
+         {
+             "code": "NoSuchKey",
+             "message": "The resource you requested does not exist",
+             "resource": "/mybucket/myfoto.jpg",
+             "requestId": "4442587FB7D0A2F9"
+         }
+         */
+        const jsonStr =
```
rahulreddy: Why can't we just stringify the Arsenal error? `requestId` should be present in the response headers, I don't see the benefit of duplicating it in the body.

lib/s3routes/routesUtils.js
```diff
@@ -49,71 +49,145 @@ function okHeaderResponse(headers, response, httpCode, log) {
...
+                  `"message":"${errCode.description}",` +
+                  '"resource":null,' +
+                  `"requestId":"${log.getSerializedUids()}"}`;
+        const bytesSent = Buffer.byteLength(jsonStr);
+        log.addDefaultFields({
+            bytesSent,
+        });
+        if (corsHeaders) {
+            // eslint-disable-next-line no-param-reassign
+            corsHeaders['Content-Type'] = 'application/json';
```
rahulreddy: Is this needed since we are going to set it anyways?

lib/s3routes/routesUtils.js
```diff
@@ -209,6 +283,19 @@ function retrieveData(locations, dataRetrievalFn,
         });
 }
 
+function responseBody(responseBackend, errCode, payload, response, log,
```
rahulreddy: Since this an internal call you can namespace it with `_` for clarity.

lib/s3routes/routesUtils.js
jonathan-gramain: We need to stringify because AWS client expects the `code` and `message` fields, they are not a direct mapping of arsenal error properties.

I agree for the request ID, I basically copied what is output in the XML version, but if you think it's not needed (which I think too) I can remove. I was thinking it was in the XML because some clients expect the request id to be there. (for JSON since it's for internal needs it does not apply).

lib/s3routes/routesUtils.js
jonathan-gramain: I just mapped the XML version. But it may be that the XML generating code is also doing it twice, will remove and see if tests are still happy.

jonathan-gramain:
@ironman-machine try

jonathan-gramain:
E2E passed in http://ci.ironmann.io/gh/scality/Integration/13876.

lib/s3routes/routesUtils.js
rahulreddy: Ah I see, makes sense.
For request id, I forgot that it shows up in the request body, you can leave it as is.

PR #296, opened 2017-07-07
fix: send accurate content-length for error xml

electrachong:
Before we were sending the length of the xml string before it was converted to utf-8 encoding; we should send the accurate byte length of the string when it has been utf-8 encoded in case a character conversion results in different lengths.

This issue is causing an existing functional test in Mystique to fail: we were sending a weird character by accident in an error response, and the inaccurate content-length post-encoding caused the xml to be trimmed, resulting in an XmlParserError from the sdk cilent.

If tests are desired, I might be able to add one to the s3 repo where I can make a functional test or use the node-mocks-http dependency for a unit test, or follow other suggestions.


electrachong:
@ironman-machine

electrachong:
end-to-end passed

PR #298, opened 2017-07-12
Dev/delimiter s3sofs

jeremyds:


lib/algos/list/delimiter.js
```diff
@@ -56,6 +62,8 @@ class Delimiter extends Extension {
         this.prefix = parameters.prefix;
         this.marker = parameters.marker;
         this.maxKeys = parameters.maxKeys || 1000;
+        this.alphabeticalOrder = parameters.alphabeticalOrder || true;
```
LaurenSpiegel: won't this always be true since if parameters.alphaticalOrder is false or undefined this will return true?

LaurenSpiegel: adding a unit test would be good

jeremyds: test in progress to check the constructor and the filter method behavior.

jeremyds: done

tests/unit/algos/list/delimiter.js
```diff
@@ -274,6 +306,18 @@ describe('Delimiter listing algorithm', () => {
         done();
     });
 
+    it('Should set Delimiter alphabeticalOrder field to the good value',
```
LaurenSpiegel: 'Should set Delimiter alphabeticalOrder field to the expected value'

tests/unit/algos/list/delimiter.js
```diff
@@ -274,6 +306,18 @@ describe('Delimiter listing algorithm', () => {
...
+       done =>
```
LaurenSpiegel: this is not async so don't need done

tests/unit/algos/list/delimiter.js
jeremyds: done

tests/unit/algos/list/delimiter.js
jeremyds: done

LaurenSpiegel: ironic 😄

jeremyds:
@ironman-machine go

PR #299, opened 2017-07-12
Add Bucket and ObjecTMD models in Arsenal

jeremyds:
This two models are needed for the S3SOFS feature into the cdmiclient
repository. The S3 one does not export modules and shared ones are in
Arsenal so move these two model modules to it.

### DIFF
--- a/lib/models/BucketInfo.js
+++ b/lib/models/BucketInfo.js
@@ -1,0 +1,502 @@
+const assert = require('assert');
+const { WebsiteConfiguration } = require('./WebsiteConfiguration');
+const ReplicationConfiguration = require('./ReplicationConfiguration');
+
+// WHEN UPDATING THIS NUMBER, UPDATE MODELVERSION.MD CHANGELOG
+const modelVersion = 5;
+
+class BucketInfo {
+    /**
+    * Represents all bucket information.
+    * @constructor
+    * @param {string} name - bucket name
+    * @param {string} owner - bucket owner's name
+    * @param {string} ownerDisplayName - owner's display name
+    * @param {object} creationDate - creation date of bucket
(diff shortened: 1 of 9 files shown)


lib/models/BucketInfo.js
```diff
@@ -0,0 +1,503 @@
+const assert = require('assert');
+const { WebsiteConfiguration } = require('./WebsiteConfiguration');
```
jeremyds: should be moved too :(

lib/models/BucketInfo.js
```diff
@@ -0,0 +1,503 @@
...
+const ReplicationConfiguration =
```
jeremyds: same here

lib/models/BucketInfo.js
jeremyds: done

lib/models/BucketInfo.js
jeremyds: done

rahulreddy:
We need some units tests to go along with this and a PR in S3 removing the models and using them from Arsenal.

jeremyds:
@rahulreddy The related S3 PR is here: https://github.com/scality/S3/pull/820
What kind of tests do you think about to test a code move?
I guessed these source files are used and tested in the S3 repository but if you think more tests may help to check the code move do not hesitate ;)

rahulreddy:
No, I mean generally when we move these files from S3 we move the accompanying tests as well. If there were none, we add some. Otherwise it would not make sense to have code in the project that is not tested.

rahulreddy:
@jeremyds This PR is missing some property changes that were made for backbeat replication. I was going to make the changes manually. Please reach out to me next time before merging manually. We usually need two approvals and run an end to end before merging.

PR #300, opened 2017-07-12
AUTH: Handle expect header stripping

LaurenSpiegel:
If load balancer strips off expect header but
expect header was included in signed headers for v4 auth,
authentication will fail. We add back the header value here since
there is only 1 specified expect header value.


LaurenSpiegel:
@ironman-machine go

LaurenSpiegel:
end to end passed

PR #302, opened 2017-07-14
Ft/disk usage

RachedBenMustapha:
Adds `getDiskUsage` APIs to the data and metadata servers.

### DIFF
--- a/lib/network/rest/RESTClient.js
+++ b/lib/network/rest/RESTClient.js
@@ -210,6 +210,47 @@ class RESTClient {
     }
 
     /**
+     * Send a GET request to the REST server, for a specific action rather
+     * than an object. Response will be truncated at the high watermark for
+     * the internal buffer of the stream, which is 16KB.
+     *
+     * @param {String} action - The action to query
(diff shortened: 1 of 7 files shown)


RachedBenMustapha:
Added tests for the data service, seems md doesn't have a suite.

lib/network/rest/RESTClient.js
```diff
@@ -210,6 +210,44 @@ class RESTClient {
...
+            path: reqParams.path,
+            headers: reqParams.headers });
+
+        const request = http.request(reqParams, response => {
+            response.once('readable', () => {
+                if (response.statusCode !== 200 &&
+                    response.statusCode !== 206) {
+                    return callback(makeErrorFromHTTPResponse(response));
+                }
+                return callback(null, response.read().toString());
```
LaurenSpiegel: Note that this should only work for responses 16kb or less (the highwatermark for the internal buffer of the stream). You might want to document that limit for the function?

rahulreddy:
@ironman-machine

PR #303, opened 2017-07-17
bf: pull response content headers from query

electrachong:
Dependency of https://github.com/scality/Mystique/pull/18

We used to pull them from the headers object, but they are supposed to be specified as query parameters.

It worked because of an accidental side effect with V2 auth where we assigned query parameters to request.headers. But this meant that getting the response content headers was failing with v4 auth.

### DIFF
--- a/lib/s3routes/routes/routeGET.js
+++ b/lib/s3routes/routes/routeGET.js
@@ -108,7 +108,7 @@ function routerGET(request, response, api, log, statsClient, dataRetrievalFn) {
                     }
                     log.end().addDefaultFields({ contentLength });
                     routesUtils.statsReport500(err, statsClient);
-                    return routesUtils.responseStreamData(err, request.headers,
+                    return routesUtils.responseStreamData(err, request.query,
                         resMetaHeaders, dataGetInfo, dataRetrievalFn, response,
                         range, log);
                 });
--- a/lib/s3routes/routesUtils.js
+++ b/lib/s3routes/routesUtils.js
@@ -181,8 +181,8 @@ const JSONResponseBackend = {
 
 /**
  * Modify response headers for an objectGet or objectHead request
- * @param {object} overrideHeaders - headers in this object override common
- * headers. These are extracted from the request object
+ * @param {object} overrideParams - parameters in this object override common
+ * headers. These are extracted from the request's query object
  * @param {object} resHeaders - object with common response headers
  * @param {object} response - router's response object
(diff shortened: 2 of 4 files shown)


electrachong:
@ironman-machine

lib/s3routes/routes/routeGET.js
```diff
@@ -108,7 +108,7 @@ function routerGET(request, response, api, log, statsClient, dataRetrievalFn) {
                     }
                     log.end().addDefaultFields({ contentLength });
                     routesUtils.statsReport500(err, statsClient);
-                    return routesUtils.responseStreamData(err, request.headers,
+                    return routesUtils.responseStreamData(err, request.query,
                         resMetaHeaders, dataGetInfo, dataRetrievalFn, response,
```
LaurenSpiegel: what about where responseContentHeaders is used?  in routeHead and routeWebsite

lib/s3routes/routesUtils.js
```diff
@@ -181,8 +181,8 @@ const JSONResponseBackend = {
 
 /**
  * Modify response headers for an objectGet or objectHead request
- * @param {object} overrideHeaders - headers in this object override common
- * headers. These are extracted from the request object
+ * @param {object} overrideHeaders - parameters in this object override common
```
LaurenSpiegel: change name to overrideParams?

lib/s3routes/routesUtils.js
```diff
@@ -337,8 +337,8 @@ const routesUtils = {
 
     /**
      * @param {string} errCode - S3 error Code
-     * @param {object} overrideHeaders - headers in this object override common
-     * headers. These are extracted from the request object
+     * @param {object} overrideHeaders - parameters in this object override
```
LaurenSpiegel: same

lib/s3routes/routesUtils.js
```diff
@@ -365,8 +365,8 @@ const routesUtils = {
...
```
LaurenSpiegel: same

lib/s3routes/routes/routeGET.js
electrachong: for both of those calls, just a empty object `{}` is being passed -- makes sense since I don't think you can set the response content headers in those api's.

lib/s3routes/routesUtils.js
electrachong: sure, I updated it

rahulreddy:
@ironman-machine

PR #306, opened 2017-07-20
Rf/parameterize retrieve data

electrachong:
Dependency of https://github.com/scality/Mystique/pull/18

For getting data from an Azure backend, it's easiest if we can pass the response object directly to the Azure getBlobToStream function to write to.


electrachong:
@ironman-machine

electrachong:
end-to-end retried after rebasing from master: http://ci.ironmann.io/gh/scality/Integration/14082

PR #308, opened 2017-07-24
DelimiterTools export path

jeremyds:
Export DelimiterTools outside the "algorithms.list" one which contains
only delimiter classes.

Furthermore, a S3 unit test expects all the class exported in this path
to be delimiter classes and iterate on them. Instead of adding
conditional in test, give to this tool class a specific export path.


LaurenSpiegel:
@ironman-machine go

PR #313, opened 2017-08-03
ft: extract MD5Sum util

electrachong:
Dependency of https://github.com/scality/Mystique/pull/33

No modifications from source code in s3.


electrachong:
@ironman-machine try

electrachong:
end-to-end retry after dropping accidental .only: http://ci.ironmann.io/gh/scality/Integration/14212

PR #317, opened 2017-08-07
ft: round-robin helper class in network/utils/RoundRobin

jonathan-gramain:
This is meant to be a generic round-robin manager (aka. bootstrap
list) to contact hosts in turn.

Blacklisting should be implemented in a next iteration.

### DIFF
--- a/index.js
+++ b/index.js
@@ -51,6 +51,7 @@ module.exports = {
             RESTServer: require('./lib/network/rest/RESTServer'),
             RESTClient: require('./lib/network/rest/RESTClient'),
         },
+        RoundRobin: require('./lib/network/RoundRobin'),
     },
     s3routes: {
         routes: require('./lib/s3routes/routes'),
--- a/lib/network/RoundRobin.js
+++ b/lib/network/RoundRobin.js
@@ -1,0 +1,163 @@
+const DEFAULT_STICKY_COUNT = 100;
+
+/**
+ * Shuffle an array in-place
+ *
+ * @param {Array} array - The array to shuffle
+ * @return {undefined}
+ */
+function shuffle(array) {
+    for (let i = array.length - 1; i > 0; i--) {
+        const randIndex = Math.floor(Math.random() * (i + 1));
+        /* eslint-disable no-param-reassign */
+        const randIndexVal = array[randIndex];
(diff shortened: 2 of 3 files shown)


lib/network/utils/RoundRobin.js
```diff
@@ -0,0 +1,128 @@
...
+    }
+}
+
+class RoundRobin {
+    /**
+     * @constructor
+     * @param {object[]} hostsList - list of hosts to query in
+     *   round-robin fashion
+     * @param {string} hostsList[].host - host name or IP address
+     * @param {number} hostsList[].port - port number to contact
```
rahulreddy: What do you think of making `port` optional or using `host:port` strings as an array? This way we can support both configurations where port is not used.

index.js
```diff
@@ -51,6 +51,9 @@ module.exports = {
             RESTServer: require('./lib/network/rest/RESTServer'),
             RESTClient: require('./lib/network/rest/RESTClient'),
         },
+        utils: {
```
jonathan-gramain: I think `utils != utils` in this case :)

Here I put it in utils, because it does not do any network but is rather a helper to support network-related operations. Now if you think it's better to put it under the network/ directory anyway, I can change it, let me know.

lib/network/utils/RoundRobin.js
jonathan-gramain: Why not, technically the round-robin logic does not need items to have any specific fields (items like `{ banana: 42 }` would work as well), the sanity checks done are also part of the helper logic to make sure the user provides valid input, so we can be quite flexible on this.

One way of dealing with missing ports could be to provide a default port value at construction, which would get appended in case the item has no port specified. Or we can let the application deal with missing port if it does not require it.

jonathan-gramain: We can also be more flexible on the input format, and convert `host:port` strings as a common `{host, port}` JS object to the user (or the other way, but I tend to prefer the object form since it's more easily extensible).

jonathan-gramain:
@ironman-machine try

rahulreddy:
@ironman-machine r+

rahulreddy:
Merging manually.

PR #320, opened 2017-08-09
bf: force pickNextHost() to pick all hosts in turn

jonathan-gramain:
Previously the shuffling may have returned the same host or at least
not go through all hosts once in turn. Fix this by not shuffling the
hosts array and by doing round-robin prior to returning the next host.


jonathan-gramain:
@ironman-machine try

PR #324, opened 2017-08-09
fix: update replication configuration format

rahulreddy:
This commit updates the format of the expected replication config
from S3 from name/endpoint to site/servers to support the new
bootstrap list changes.


rahulreddy:
@ironman-machine try

PR #327, opened 2017-08-21
ft: log uid from sent X-Scal-Request-Uids header

jonathan-gramain:
backbeat will send this header to provide the request uid set while
processing an entry to S3 routes.


jonathan-gramain:
@ironman-machine try

rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine r+

PR #328, opened 2017-08-23
Add MetadataFileServer.rawListKeys()

RachedBenMustapha:
This new DB method lists the keys for the root DB, which makes inspecting all databases at once more convenient.


rahulreddy:
@ironman-machine r+

PR #329, opened 2017-08-25
ft: blacklist object prefixes

electrachong:
Dependency of https://github.com/scality/Mystique/pull/40

Add blacklisting for object prefixes.

### DIFF
--- a/lib/s3routes/routesUtils.js
+++ b/lib/s3routes/routesUtils.js
@@ -782,6 +782,22 @@ const routesUtils = {
     },
 
     /**
+     * Validate object key per naming rules and restrictions
+     * @param {string} objectKey - object key
+     * @param {string[]} prefixBlacklist - prefixes reserved for internal use
+     * @return {object} - object containing true/false result and invalidPrefix
+     *  if false
+     */
+    isValidObjectKey(objectKey, prefixBlacklist) {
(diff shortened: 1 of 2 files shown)


electrachong:
@ironman-machine try

lib/s3routes/routesUtils.js
```diff
@@ -782,6 +782,28 @@ const routesUtils = {
...
     /**
+     * Validate object key per naming rules and restrictions
+     * @param {string} objectKey - object key
+     * @param {string[]} prefixBlacklist - prefixes reserved for internal use
+     * @return {object} - object containing true/false result and invalidPrefix
+     *  if false
+     */
+    isValidObjectKey(objectKey, prefixBlacklist) {
+        let invalidPrefix;
+        prefixBlacklist.some(prefix => {
```
jonathan-gramain: If you don't use the return value of `some` you don't need to return any value from the test function, and then there's no benefit of using `some`. A more concise way can be to use the `Array.find` function like `invalidPrefix = prefixBlacklist.find(prefix => objectKey.startsWith(prefix))`

electrachong: you're right, thanks for the tip!

LaurenSpiegel:
@ironman-machine r+

PR #330, opened 2017-08-29
Add azure putpart utils

dora-korpar:
Pulls utils needed by Zenko for Azure putPart API from Mystique and makes them accessible in s3middleware.

### DIFF
--- a/index.js
+++ b/index.js
@@ -65,6 +65,15 @@ module.exports = {
             require('./lib/s3middleware/validateConditionalHeaders')
             .validateConditionalHeaders,
         MD5Sum: require('./lib/s3middleware/MD5Sum'),
+        objectUtils: require('./lib/s3middleware/objectUtils'),
+        azureHelper: {
+            mpuUtils:
+                require('./lib/s3middleware/azureHelpers/mpuUtils'),
+            ResultsCollector:
+                require('./lib/s3middleware/azureHelpers/ResultsCollector'),
+            SubStreamInterface:
+                require('./lib/s3middleware/azureHelpers/SubStreamInterface'),
+        },
     },
     storage: {
         metadata: {
--- a/lib/s3middleware/objectUtils.js
+++ b/lib/s3middleware/objectUtils.js
@@ -1,0 +1,9 @@
+const objectUtils = {};
+
+objectUtils.getHexMD5 = base64MD5 =>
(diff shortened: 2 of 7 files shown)


dora-korpar:
@ironman-machine try

index.js
```diff
@@ -65,6 +65,10 @@ module.exports = {
             require('./lib/s3middleware/validateConditionalHeaders')
             .validateConditionalHeaders,
         MD5Sum: require('./lib/s3middleware/MD5Sum'),
+        azureMpuUtils: require('./lib/s3middleware/azureMpuUtils'),
```
LaurenSpiegel: for the azureMpuUtils, ResultsCollector and SubStreamInterface (and maybe objectUtils but they seem pretty generic) I would nest them under an azure subheading so they are separated since these only apply to putting to azure.

lib/s3middleware/SubStreamInterface.js
```diff
@@ -0,0 +1,103 @@
...
+            subPartIndex: this._subPartIndex,
+        };
+    }
+
+    /**
+     * SubStreamInterface.write - write to the current stream
+     * @param {Buffer} chunk - a chunk of data
+     * @return {undefined}
+     */
+    write(chunk) {
```
LaurenSpiegel: I don't think this file is the most up to date version.

lib/s3middleware/objectUtils.js
```diff
@@ -0,0 +1,9 @@
+const objectUtils = {};
+
+objectUtils.getHexMD5 = base64MD5 =>
+    Buffer.from(base64MD5, 'base64').toString('hex');
+
+objectUtils.getBase64MD5 = hexMD5 =>
+    Buffer.from(hexMD5, 'hex').toString('base64');
+
+module.exports = objectUtils;
```
LaurenSpiegel: unit tests should come along for the ride.

index.js
dora-korpar: I ended up not needing the ResultsCollector and SubStreamInterface in S3, so have removed them. Is it worth putting just the azureMpuUtils in it's own directory?

dora-korpar:
@ironman-machine try

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r- r+

PR #333, opened 2017-09-07
ARN object model class

jonathan-gramain:
Will be used by backbeat for roles, and potentially can be used for other components that manipulate ARNs.

### DIFF
--- a/lib/models/ARN.js
+++ b/lib/models/ARN.js
@@ -1,0 +1,106 @@
+const errors = require('../errors');
+
+const validServices = {
+    aws: ['s3', 'iam', 'sts', 'ring'],
+    scality: ['utapi', 'sso'],
+};
+
+class ARN {
+    /**
+     *
+     * Create an ARN object from its individual components
+     *
+     * @constructor
+     * @param {string} partition - ARN partition (e.g. 'aws')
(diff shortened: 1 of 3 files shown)


jonathan-gramain:
@ironman-machine try

lib/models/ARN.js
```diff
@@ -0,0 +1,94 @@
+const errors = require('../errors');
+
+const validServices = {
+    aws: ['s3', 'iam'],
```
rahulreddy: We now have `sso`, `ring`, `utapi`, `sts` as valid services for ARN.
https://github.com/scality/Arsenal/blob/master/lib/policyEvaluator/RequestContext.js#L133
May be we should include them as well.

lib/models/ARN.js
```diff
@@ -0,0 +1,94 @@
...
+};
+
+class ARN {
+    constructor(partition, service, region, accountId, resource) {
```
rahulreddy: Please document the constructor params.

lib/models/ARN.js
jonathan-gramain: updated

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r- r+

PR #337, opened 2017-09-11
Chore: upgrade eslint

philipyoo:
I tried to separate my commits by each type of eslint change, so it might be easiest to view by each commit.

Let me know what you think, if we should add or remove certain rules. Current rules can be viewed:
https://github.com/scality/Guidelines/blob/chore/upgradeEslint/index.js

List of Eslint rules: https://eslint.org/docs/rules/

Note: Indent fixes were mainly done by the linter `--fix` command

Note: Will squash commits after making any changes based on peoples' feedback


lib/s3routes/routes/routeGET.js
```diff
@@ -74,6 +74,7 @@ function routerGET(request, response, api, log, statsClient, dataRetrievalFn) {
                 });
         }
     } else {
+        /*eslint-disable no-lonely-if*/
```
bennettbuchanan: Why disable it here?

lib/storage/utils.js
```diff
@@ -6,6 +6,7 @@ const assert = require('assert');
 
 function trySetDirSyncFlag(path) {
     // may throw if ioctl is not available
+    // eslint-disable-next-line
```
bennettbuchanan: Can drop this rule.

lib/s3routes/routes/routeGET.js
philipyoo: I wasn't sure how to refactor this part because each nested if seemed like it served a purpose :/

philipyoo:
@ironman-machine try plz

lib/auth/in_memory/Indexer.js
```diff
@@ -126,7 +126,8 @@ class Indexer {
      * @return {Object} entity.email - The entity's lowercased email
      */
     getEntityByKey(key) {
-        if (this.accountsBy.accessKey.hasOwnProperty(key)) {
+        if (Object.prototype.hasOwnProperty.call(
```
rahulreddy: I think at this point, it's just easy to call 
```js
if (this.accountsBy.accessKey[key] !== undefined) {
```

philipyoo:
@ironman-machine try

rahulreddy:
@ironman-machine r+

PR #340, opened 2017-09-14
Allow CORS for Get Service

vrancurel:
- allow OPTIONS for get service
- return corsHeaders for get service

### DIFF
--- a/lib/s3routes/routes.js
+++ b/lib/s3routes/routes.js
@@ -35,12 +35,12 @@ function checkUnsupportedRoutes(reqMethod) {
 
 function checkBucketAndKey(bucketName, objectKey, method, reqQuery,
     blacklistedPrefixes, log) {
-    // if empty name and request not a list Buckets
-    if (!bucketName && !(method === 'GET' && !objectKey)) {
+    // bucketName should also be undefined, but is checked below anyway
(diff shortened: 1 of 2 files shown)


LaurenSpiegel:
@electrachong to finish

lib/s3routes/routes.js
```diff
@@ -35,12 +35,11 @@ function checkUnsupportedRoutes(reqMethod) {
 
 function checkBucketAndKey(bucketName, objectKey, method, reqQuery,
     blacklistedPrefixes, log) {
-    // if empty name and request not a list Buckets
-    if (!bucketName && !(method === 'GET' && !objectKey)) {
+    const getServiceCall = (method === 'GET' && !bucketName && !objectKey);
```
jonathan-gramain: you may remove the superfluous `&& !bucketName` test (as far as I understand the naming would be still ok since a GET with a bucket name is also a service call).

electrachong: `getService` is the name AWS uses for the list buckets call, when you make a GET request without defining a bucketname or object key resource in the request http://docs.aws.amazon.com/AmazonS3/latest/API/RESTServiceGET.html
but I know I can remove the test since the check below also checks for bucketName, I was just upset about naming the variable `getServiceCall` without including the bucketName check in there. :(

PR #341, opened 2017-09-15
ft: update mpu utils with optimizations

electrachong:


electrachong:
@ironman-machine try

lib/s3middleware/azureHelpers/mpuUtils.js
```diff
@@ -188,12 +188,22 @@ log, cb) => {
     hashedStream.on('data', data => {
         const currentLength = streamInterface.getLengthCounter();
         if (currentLength + data.length > azureMpuUtils.maxSubPartSize) {
+            const bytesToMaxSize = azureMpuUtils.maxSubPartSize - currentLength;
+            const firstChunk = bytesToMaxSize === 0 ? data :
+                data.slice(bytesToMaxSize);
```
dora-korpar: Should this `data.slice` be from 0 - bytesToMaxSize?

electrachong: We write 0 - bytesToMaxSize in the current stream on https://github.com/scality/Arsenal/pull/341/files/53c88cf4c415d0284b490d9ded207aa5f8d0b245#diff-95424c3fe467f0aebd5fb790122f83c1R197
and then write the remainder (from bytesToMaxSize to the end) as the first chunk of the next stream

lib/s3middleware/azureHelpers/mpuUtils.js
```diff
@@ -111,12 +111,15 @@ log, cb) => {
         if (err) {
             log.error('Error from Azure data backend uploadPart',
                 { error: err.message, dataStoreName });
+            if (err.code === 'InvalidMd5') {
```
electrachong: @LaurenSpiegel should we consider porting the translate errors to Arsenal?
If someone deletes the container/bucket (although I think we should return `MpuInProgress` if someone tried that from ScalityConnect) before we get to this call, we would respond InternalError instead of NoSuchBucket for this call.

LaurenSpiegel: Let's discuss tomorrow. I don't fully understand what you mean.

electrachong:
@ironman-machine try

electrachong:
after making change to expect base64 md5s for putting a single part
@ironman-machine try

electrachong:
after rebasing...
@ironman-machine try

electrachong:
test for safety's sake
@ironman-machine try

electrachong:
end-to-end auto-retried here: http://ci.ironmann.io/gh/scality/Integration/15007

PR #342, opened 2017-09-22
Fix: S3C-905 Quiet health check request logs

LaurenSpiegel:


LaurenSpiegel:
@ironman-machine try

lib/s3routes/routes.js
```diff
@@ -163,7 +163,10 @@ function routes(req, res, params, logger) {
     const log = (reqUids !== undefined ?
                  logger.newRequestLoggerFromSerializedUids(reqUids) :
                  logger.newRequestLogger());
-    log.info('received request', clientInfo);
+
+    if (!req.url.startsWith('/_/healthcheck')) {
```
DavidPineauScality: shouldn't we at least log those as debug, just in case ? (error logging would dump it, and it would be seen when bumping the log level for more informations)

LaurenSpiegel: since the clientInfo is added as a default field do we really gain much?

DavidPineauScality: I wasn't even thinking about the client informations; mostly about logging the entry point of the request. But I guess this specific request would rarely if ever dump logs, and the real log that is filling the logs is the last log (`end()` call), right ?

LaurenSpiegel: This is the only log line that appears on a successful health check. I don't think we want to see it even in debug mode. If someone has to go into the system to debug, the first thing they will often do is just comment this out because it is hard to zoom in on the problem with the healthcheck logs.

DavidPineauScality: oh? Ok, maybe that changed from previous versions then., I see. thanks for taking the time to answer my questions :)

LaurenSpiegel:
@ironman-machine r+

PR #344, opened 2017-09-29
MD-191: Support slash for access keys

alexandre-merle:
Add support for slashs in access keys, since the compatibility for rs2
allow / inside an access key

Fixes https://scality.atlassian.net/browse/MD-191

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -7,6 +7,8 @@ general:
 machine:
   node:
     version: 4.5.0
+  environment:
+    CXX: g++-4.9
 
 dependencies:
   pre:
--- a/lib/auth/v4/validateInputs.js
+++ b/lib/auth/v4/validateInputs.js
@@ -109,6 +109,13 @@ function extractQueryParams(queryObj, log) {
     const credential = queryObj['X-Amz-Credential'];
     if (credential && credential.length > 28 && credential.indexOf('/') > -1) {
         authParams.credential = credential.split('/');
+        // in case access keys contains a '/', reduce the credentials array to
+        // the desire size by recomputing the access key. This step is based on
+        // fact than no other array component will have a '/' character.
+        while (authParams.credential.length > 5) {
+            authParams.credential[0] += `/${authParams.credential[1]}`;
+            authParams.credential.splice(1, 1);
+        }
     } else {
         log.warn('invalid credential param', { credential });
         return authParams;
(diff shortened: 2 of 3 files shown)


alexandre-merle:
@ironman-machine try

lib/auth/v4/validateInputs.js
```diff
@@ -109,6 +109,12 @@ function extractQueryParams(queryObj, log) {
     const credential = queryObj['X-Amz-Credential'];
     if (credential && credential.length > 28 && credential.indexOf('/') > -1) {
         authParams.credential = credential.split('/');
+        // in case access keys contains a '/', reduce the credentials array to
+        // the desire size by recomputing the access key
+        while (authParams.credential.length > 5) {
```
DavidPineauScality: So we're assuming that in this "credential", where AccessKey is supposed to be first; we concatenate all the early items in the array, until we only have 5 left ?

ie, assuming an array length of X after splitting (python example for ease of explanation):
```py
'/'.join(creds[0:X-4])
```

Does this mean that no other component can have a '/' within ? If so, it might deserve an additional comment to explain the assumptions made in this code, and validate its correctness.

lib/auth/v4/validateInputs.js
```diff
@@ -140,6 +146,13 @@ function extractAuthItems(authHeader, log) {
         && credentialStr.indexOf('/') > -1) {
         authItems.credentialsArr = credentialStr
             .trim().replace('Credential=', '').split('/');
+        // in case access keys contains a '/', reduce the credentials array to
+        // the desire size by recomputing the access key
+        while (authItems.credentialsArr.length > 5) {
```
DavidPineauScality: Same comment here :)

circle.yml
```diff
@@ -7,6 +7,8 @@ general:
 machine:
   node:
     version: 4.5.0
+  environment:
+    CXX: g++-4.9
```
DavidPineauScality: Oh. this is a new dep for the ajv update ? or unrelated ?

ThibaultRiviere:
@alexandre-merle a unit test as well can be a good idea

lib/auth/v4/validateInputs.js
alexandre-merle: yes, not other component will have a '/' :)

circle.yml
alexandre-merle: nope, this one is for compiling with -std=c++11 :)

DavidPineauScality: yes but I meant: what is requiring that in Arsenal 6.4 ?

lib/auth/v4/validateInputs.js
DavidPineauScality: okay, then a comment could be nice. I'm also guessing there's no specification of the AccessKeyLength, that could help double-checking that we're not messing with the creds ?

circle.yml
alexandre-merle: a compilation issue with 'diskusage' :)

lib/auth/v4/validateInputs.js
alexandre-merle: the length can change, but the signature computation is secure enought :)

circle.yml
DavidPineauScality: ok, I guess the version was bumped and we're not depending on a fixed one, leading to this being un-caught till now

alexandre-merle: absolutly no idea, just fixing ^^

alexandre-merle:
i think i dream about that '/', can't find it anymore, closing this one ^^'

PR #345, opened 2017-10-04
service accounts

jonathan-gramain:
ft: service account identification support

Add support in AuthInfo for service accounts, designated by a
canonical ID starting with the service accounts namespace URL (newly
defined). We can ask if the authenticated account is a service account
or a particular one by its name.

### DIFF
--- a/lib/auth/AuthInfo.js
+++ b/lib/auth/AuthInfo.js
@@ -49,6 +49,14 @@ class AuthInfo {
     isRequesterPublicUser() {
         return this.canonicalID === constants.publicId;
     }
+    isRequesterAServiceAccount() {
+        return this.canonicalID.startsWith(
+            `${constants.zenkoServiceAccount}/`);
+    }
+    isRequesterThisServiceAccount(serviceName) {
+        return this.canonicalID ===
+            `${constants.zenkoServiceAccount}/${serviceName}`;
+    }
 }
 
 module.exports = AuthInfo;
--- a/lib/auth/in_memory/Backend.js
+++ b/lib/auth/in_memory/Backend.js
@@ -171,11 +171,11 @@ class Backend {
     * @param {object} requestContextParams.constantParams -
     * params that have the
     * same value for each requestContext to be constructed in Vault
-    * @param {object} requestContextParams.paramaterize - params that have
+    * @param {object} requestContextParams.parameterize - params that have
     * arrays as values since a requestContext needs to be constructed with
     * each option in Vault
-    * @param {object[]} requestContextParams.paramaterize.specificResource -
-    * specific resources paramaterized as an array of objects containing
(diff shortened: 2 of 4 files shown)


lib/auth/AuthInfo.js
```diff
@@ -49,6 +49,14 @@ class AuthInfo {
     isRequesterPublicUser() {
         return this.canonicalID === constants.publicId;
     }
+    isRequesterAServiceAccount() {
+        return this.canonicalID.startsWith(
+            `${constants.zenkoServiceAccountsNs}/`);
```
LaurenSpiegel: Ns for namespace? How about zenkoServiceAccountId since we use publicId?

tests/unit/auth/AuthInfo.js
```diff
@@ -72,4 +72,20 @@ describe('AuthInfo class constructor', () => {
...
+    it('should have a working isRequesterAServiceAccount() method', () => {
+        assert.strictEqual(authInfo.isRequesterAServiceAccount(), false);
+        const serviceAccount = new AuthInfo({
+            canonicalID: `${constants.zenkoServiceAccountsNs}/Clueso` });
+        assert.strictEqual(serviceAccount.isRequesterAServiceAccount(), true);
+    });
+
+    it('should have a working isRequesterThisServiceAccount() method', () => {
+        const serviceAccount = new AuthInfo({
+            canonicalID: `${constants.zenkoServiceAccountsNs}/Clueso` });
```
LaurenSpiegel: I think we should do lowercase since we're doing a sort of url

lib/auth/AuthInfo.js
jonathan-gramain: I prefer to name it a namespace (or prefix), because it's not an account id but the prefix at which all service accounts will go, the full account id must have the service name appended to the prefix. As for publicId it's a complete account ID.

tests/unit/auth/AuthInfo.js
jonathan-gramain: I agree, will change the test.

lib/auth/in_memory/Backend.js
```diff
@@ -171,11 +171,11 @@ class Backend {
...
     * params that have the
     * same value for each requestContext to be constructed in Vault
-    * @param {object} requestContextParams.paramaterize - params that have
+    * @param {object} requestContextParams.parameterize - params that have
     * arrays as values since a requestContext needs to be constructed with
     * each option in Vault
-    * @param {object[]} requestContextParams.paramaterize.specificResource -
-    * specific resources paramaterized as an array of objects containing
+    * @param {object[]} requestContextParams.parameterize.specificResource -
+    * specific resources parameterized as an array of objects containing
```
electrachong: 😄

rahulreddy:
@ironman-machine r+

PR #347, opened 2017-10-05
FT: Add storageType to replicationInfo

bennettbuchanan:
Depends on https://github.com/scality/S3/pull/922.

* Add the `storageType` field to the `replicationInfo` field of an object's metadata.
* The definition of a valid `storageClass` is expanded to include values that exist in the locationConfig.json of CloudServer.

### DIFF
--- a/lib/models/ReplicationConfiguration.js
+++ b/lib/models/ReplicationConfiguration.js
@@ -58,6 +58,8 @@ class ReplicationConfiguration {
         this._role = null;
         this._destination = null;
         this._rules = null;
+        this._prevStorageClass = null;
+        this._isExternalLocation = null;
     }
 
     /**
@@ -148,10 +150,15 @@ class ReplicationConfiguration {
         }
         const role = parsedRole[0];
         const rolesArr = role.split(',');
-        if (rolesArr.length !== 2) {
+        if (!this._isExternalLocation && rolesArr.length !== 2) {
             return errors.InvalidArgument.customizeDescription(
                 'Invalid Role specified in replication configuration: ' +
                 'Role must be a comma-separated list of two IAM roles');
+        }
+        if (this._isExternalLocation && rolesArr.length > 1) {
+            return errors.InvalidArgument.customizeDescription(
+                'Invalid Role specified in replication configuration: ' +
+                'Role may not contain a comma separator');
         }
         const invalidRole = rolesArr.find(r => !this._isValidRoleARN(r));
         if (invalidRole !== undefined) {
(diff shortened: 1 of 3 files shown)


lib/models/ReplicationConfiguration.js
```diff
@@ -148,10 +148,10 @@ class ReplicationConfiguration {
         }
         const role = parsedRole[0];
         const rolesArr = role.split(',');
-        if (rolesArr.length !== 2) {
+        if (rolesArr.length > 2) {
```
jonathan-gramain: While you're at this, I'm thinking we should allow no role at all for zenko, since it will use a special account to do the operations, it does not make sense to set a dummy role for it. It may break AWS strict specs, and we may want to allow only if we're not using vault. @rahulreddy any concern?

bennettbuchanan: I agree that is preferable to setting a dummy role. However I think the scope of changes for replicating to AWS is already wide. Maybe we should add a TODO for modifying this along with changes related to zenko configuration?

lib/models/ReplicationConfiguration.js
```diff
@@ -357,7 +367,11 @@ class ReplicationConfiguration {
      * @return {undefined}
      */
     parseConfiguration() {
-        const err = this._parseRole() || this._parseRules();
+        let err = this._parseRules();
+        if (err) {
+            return err;
+        }
+        err = this._parseRole();
```
rahulreddy: You can `return this._parseRole();` for brevity and remove the if clause below it.

bennettbuchanan: True! Updated.

lib/models/ReplicationConfiguration.js
jonathan-gramain: Yes you can add a TODO if you wish it can help tracking.

bennettbuchanan:
@ironman-machine 
SCALITY_S3_BRANCH=ft/S3C-972/add-storageType-to-replicationInfo
try

rahulreddy:
@ironman-machine clean r+

lib/models/ReplicationConfiguration.js
```diff
@@ -267,32 +274,35 @@ class ReplicationConfiguration {
...
+        if (locationConstraints.includes(storageClass)) {
+            if (this._prevStorageClass !== null &&
+                this._prevStorageClass !== storageClass) {
+                return errors.InvalidRequest.customizeDescription(
+                    'The storage class must be same for all rules when ' +
+                    'replicating objects to an external location');
+            }
+            this._isExternalLocation = storageClass;
+        }
+        if (!replicationEndpoints.includes(storageClass) &&
```
jonathan-gramain: minor: you can save a test by using `else`

lib/models/ReplicationConfiguration.js
```diff
@@ -267,32 +274,35 @@ class ReplicationConfiguration {
...
+        const locationConstraints =
+            Object.keys(this._config.locationConstraints);
+        if (locationConstraints.includes(storageClass)) {
+            if (this._prevStorageClass !== null &&
+                this._prevStorageClass !== storageClass) {
+                return errors.InvalidRequest.customizeDescription(
+                    'The storage class must be same for all rules when ' +
+                    'replicating objects to an external location');
+            }
+            this._isExternalLocation = storageClass;
```
jonathan-gramain: I guess you want to assign `true` here instead of `storageClass`

bennettbuchanan: Thanks! Updated.

PR #351, opened 2017-10-12
ft: policy arn condition

mvaude:
MD-286
There was a bug in `checkArnMatch`: it was making a ternary that is not necessary (`toLowerCase()` is done when the elements are joined).
Furthermore, it was calling this last method on an array.

### DIFF
--- a/lib/policyEvaluator/RequestContext.js
+++ b/lib/policyEvaluator/RequestContext.js
@@ -188,6 +188,7 @@ function _buildArn(service, generalResource, specificResource, requesterInfo) {
  * @param {string} authType - type of authentication used
  * @param {number} signatureAge - age of signature in milliseconds
  * @param {string} securityToken - auth security token (temporary credentials)
+ * @param {string} policyArn - policy arn
  * @return {RequestContext} a RequestContext instance
  */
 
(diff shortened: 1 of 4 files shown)


lib/policyEvaluator/RequestContext.js
```diff
@@ -579,6 +582,27 @@ class RequestContext {
         this._securityToken = token;
         return this;
     }
+
+    /**
+     * Get the policy arn
+     *
+     * @param {string} policyArn - Policy arn
```
alexandre-merle: to be removed :)

lib/policyEvaluator/RequestContext.js
```diff
@@ -579,6 +582,27 @@ class RequestContext {
...
+     * @return {RequestContext} itself
```
alexandre-merle: return the arn :)

lib/policyEvaluator/RequestContext.js
mvaude: [x] remove doc for wrong parameter

lib/policyEvaluator/RequestContext.js
mvaude: - [x] update doc with good return type and name

rahulreddy:
@ironman-machine r+

PR #354, opened 2017-10-17
chore: move stats and redisclient from s3

philipyoo:
Moving StatsClient and RedisClient from S3 repo to reuse classes


rahulreddy:
@ironman-machine r+

PR #355, opened 2017-10-17
ft: add getters to storageClass/storageType

jonathan-gramain:
These getters were originally in backbeat QueueEntry class, move them
to the ObjectMD model.


jonathan-gramain:
@ironman-machine try

PR #358, opened 2017-10-18
ft: add data store version id to objectMD

electrachong:
Dependency of https://github.com/scality/S3/pull/938


electrachong:
@ironman-machine try

PR #359, opened 2017-10-18
ft: add getter for dataStoreVersionId

electrachong:


electrachong:
@ironman-machine try

lib/models/ObjectMD.js
```diff
@@ -753,6 +753,19 @@ class ObjectMD {
     }
 
     /**
+     * Get dataStoreVersionId
+     *
+     * @return {string} external backend version id for data
+     */
+    getDataStoreVersionId() {
+        if (!this._data.location || !Array.isArray(this._data.location)
```
jonathan-gramain: You can call `getLocation()` which always returns an array (possibly empty), and get rid of the first two checks in this test.

jonathan-gramain: When metadata is loaded from the blob, `_convertToLatestModel()` ensures that `location` is or becomes an array or `null`.

electrachong: updated -- you're right, but don't I still one of the two checks in case it is null? Otherwise we would get a `Cannot read property '0' of null` error

electrachong: nvm I see what you are saying and will update again

rahulreddy:
@ironman-machine try

electrachong:
@ironman-machine try

electrachong:
replication tests timed out; auto-retried here: http://ci.ironmann.io/gh/scality/Integration/15954

PR #361, opened 2017-10-21
FWD: rel/7.1 to master

bennettbuchanan:


rahulreddy:
@ironman-machine try

bennettbuchanan:
Retried here and tests passed: http://ci.ironmann.io/gh/scality/Integration/15889.

PR #363, opened 2017-10-25
FT: No need action mapping

alexandre-merle:
We currently use an action mapping with api methods, allowing the option
to send directly the action to not have to modify arsenal each new
action managed.

Need for https://scality.atlassian.net/browse/MD-292


rahulreddy:
@ironman-machine try

PR #365, opened 2017-10-31
bf: sanity check on returned content-length

jonathan-gramain:
In responseStreamData() helper, ensure the content-length is correct
with respect to data locations aggregated size. Log an error and
return internal error to the client if not.

This should catch off-by-one errors when computing ranges of data
locations to fetch and return.

### DIFF
--- a/lib/s3routes/routesUtils.js
+++ b/lib/s3routes/routesUtils.js
@@ -310,7 +310,8 @@ function _contentLengthMatchesLocations(contentLength, dataLocations) {
         (sum, location) => (sum !== undefined && location.size ?
                             sum + Number.parseInt(location.size, 10) :
                             undefined), 0);
-    return sumSizes === undefined || sumSizes === contentLength;
+    return sumSizes === undefined ||
+        sumSizes === Number.parseInt(contentLength, 10);
 }
 
 const routesUtils = {


jonathan-gramain:
E2E passed in http://ci.ironmann.io/gh/scality/Integration/16632

lib/s3routes/routesUtils.js
```diff
@@ -402,6 +411,21 @@ const routesUtils = {
...
                                                     resHeaders);
         }
+        if (dataLocations !== null && !response.headersSent) {
+            // sanity check of content length against individual data
+            // locations to fetch
+            const contentLength = resHeaders && resHeaders['Content-Length'];
+            if (contentLength !== undefined &&
+                !_contentLengthMatchesLocations(contentLength,
+                                                dataLocations)) {
+                log.error('logic error: total length of fetched data ' +
```
rahulreddy: May be remove `logic error:` and keep the rest of the message?

jonathan-gramain: ok

jonathan-gramain: Well, maybe not going to change it since master branch is already merged.

rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine r+

PR #369, opened 2017-11-06
Ft: extend StatsClient to increase by given amount

philipyoo:
Allow StatsClient to increment a key by a given amount, needed for https://github.com/scality/backbeat/pull/126 (to increment by byte size)

Support random order optional args case with `callback` and `amount` arg. 
I am not checking to see if the optional args passed are valid types. Instead, building on top of what exists, if a callback and/or integer are passed, they are assigned to appropriate variables for use

### DIFF
--- a/lib/metrics/StatsClient.js
+++ b/lib/metrics/StatsClient.js
@@ -65,16 +65,29 @@ class StatsClient {
     /**
     * report/record a new request received on the server
     * @param {string} id - service identifier
-    * @param {callback} cb - callback
+    * @param {number} incr - optional param increment
+    * @param {function} cb - callback
     * @return {undefined}
     */
-    reportNewRequest(id, cb) {
+    reportNewRequest(id, incr, cb) {
         if (!this._redis) {
             return undefined;
         }
-        const callback = cb || this._noop;
+
+        let callback;
+        let amount;
+        if (typeof incr === 'function') {
+            // In case where optional `incr` is not passed, but `cb` is passed
+            callback = incr;
+            amount = 1;
+        } else {
+            callback = (cb && typeof cb === 'function') ? cb : this._noop;
+            amount = (incr && typeof incr === 'number') ? incr : 1;
+        }
+
         const key = this._buildKey(`${id}:requests`, new Date());
-        return this._redis.incrEx(key, this._expiry, callback);
+
+        return this._redis.incrbyEx(key, amount, this._expiry, callback);
     }
 
     /**
--- a/tests/unit/metrics/StatsClient.js
+++ b/tests/unit/metrics/StatsClient.js
@@ -1,6 +1,7 @@
 'use strict'; // eslint-disable-line strict
 
 const assert = require('assert');
+const async = require('async');
 
 const RedisClient = require('../../../lib/metrics/RedisClient');
 const StatsClient = require('../../../lib/metrics/StatsClient');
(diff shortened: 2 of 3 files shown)


lib/metrics/StatsClient.js
```diff
@@ -65,15 +65,20 @@ class StatsClient {
     /**
     * report/record a new request received on the server
     * @param {string} id - service identifier
+    * @param {number} amount - amount to increment by (default = 1)
```
jonathan-gramain: use optional syntax: `[amount=1]`

lib/metrics/StatsClient.js
```diff
@@ -65,15 +65,20 @@ class StatsClient {
...
     * @param {callback} cb - callback
     * @return {undefined}
     */
-    reportNewRequest(id, cb) {
+    reportNewRequest(id, amount = 1, cb) {
```
jonathan-gramain: You should make sure the existing API is not broken with the changes, here the existing callers that may provide `cb` still expect their callback to be called.

I saw this pattern already in some standard node code, where arguments with default values are not the trailing arguments, I think it's fine but requires proper handling in the function. I would think a proper way to deal with this in a clean way would be to declare the arguments with spread syntax `...args`, pop the last argument as the callback function if it's of type `function`, otherwise skip it (consider callback is not provided), and keep going with the preceding args to fit the exposed API. Then any existing caller of `reportNewRequest(id, cb)` would not be broken.

Another more straightforward option is to create a new API function with an additional mandatory `amount` argument.

tests/unit/metrics/StatsClient.js
```diff
@@ -27,7 +27,7 @@ describe('StatsClient class', () => {
 
     afterEach(() => redisClient.clear(() => {}));
 
-    it('should correctly record a new request', () => {
+    it('should correctly record a new request by default one increment', () => {
```
jonathan-gramain: Here, I suggest you add an actual callback argument and call it at the end of `reportNewRequest`'s callback, to make sure it gets called or tests will timeout.

tests/unit/metrics/StatsClient.js
```diff
@@ -47,6 +47,22 @@ describe('StatsClient class', () => {
         });
     });
 
+    it('should record new requests by defined amount increments', () => {
```
jonathan-gramain: same here

lib/metrics/StatsClient.js
jonathan-gramain: Sorry I missed your comment in the PR about this :)

You can still implement the above or leave as is, up to you, but if you don't you should fix the test(s) that use the old API with a callback, see my comment below :)

rahulreddy:
We will merge this after 7.2 is built to avoid potential breakage of compatibility.

lib/metrics/StatsClient.js
```diff
@@ -65,15 +65,30 @@ class StatsClient {
...
     * @return {undefined}
     */
-    reportNewRequest(id, cb) {
+    reportNewRequest(id, incr, cb) {
         if (!this._redis) {
             return undefined;
         }
-        const callback = cb || this._noop;
+
+        let callback = this._noop;
```
jonathan-gramain: may just declare `let callback` here and at line 83 do `callback = ... : this._noop;`

Same thing for `amount` if you add `amount = 1;` in the `typeof incr === 'function'` block (I think it's clearer not to reassign to the same variable previously assigned so one does not have to backtrack to the original value).

lib/metrics/StatsClient.js
```diff
@@ -65,15 +65,30 @@ class StatsClient {
...
+            // In case where optional `incr` is not passed, but `cb` is passed
+            callback = incr;
+        } else {
+            callback = (cb && typeof cb === 'function') ? cb : callback;
+            amount = (incr && typeof incr === 'number') ? incr : amount;
+        }
+
         const key = this._buildKey(`${id}:requests`, new Date());
+
+        if (amount !== 1) {
```
jonathan-gramain: maybe you can use `incrbyEx` in all cases, I'm sure if there's any optimization for the incr by one case the redis API will have implemented it also in `incbyEx`.

tests/unit/metrics/StatsClient.js
```diff
@@ -27,52 +28,110 @@ describe('StatsClient class', () => {
...
+            next => {
+                statsClient.report500(id, err => {
+                    assert.ifError(err);
+                    next();
+                });
+            },
+            next => {
+                statsClient.getStats(fakeLogger, id, (err, res) => {
+                    assert.ifError(err);
+                    assert.equal(typeof res, 'object');
```
jonathan-gramain: You could get rid of the following lines since you're checking the whole object just next. (for debugging failed tests, no worry since mocha will dump the obtained and expected objects when it fails, so we know exactly what has been returned in the callback).

I also saw a couple other instances above where using `deepEqual` makes the other tests redundant.

tests/unit/metrics/StatsClient.js
```diff
@@ -27,52 +28,110 @@ describe('StatsClient class', () => {
...
 
-            const expected = { 'requests': 1, '500s': 1, 'sampleDuration': 30 };
-            assert.deepEqual(res, expected);
-        });
+                    const expected = {
+                        'requests': 1,
+                        '500s': 1,
+                        'sampleDuration': 30,
+                    };
+                    assert.deepEqual(res, expected);
```
jonathan-gramain: I'm not sure of the exact difference but we're mostly using `deepStrictEqual` for comparing objects deeply, maybe stick to this function to avoid surprises (it also applies to a few cases above).

tests/unit/metrics/StatsClient.js
```diff
@@ -27,52 +28,110 @@ describe('StatsClient class', () => {
...
+                    assert.equal(Object.keys(res).length, 3);
+                    assert.equal(res.sampleDuration, STATS_EXPIRY);
 
-            const expected = { 'requests': 1, '500s': 1, 'sampleDuration': 30 };
-            assert.deepEqual(res, expected);
-        });
+                    const expected = {
+                        'requests': 1,
+                        '500s': 1,
+                        'sampleDuration': 30,
```
jonathan-gramain: you may replace 30 by `STATS_EXPIRY` here

rahulreddy:
@ironman-machine r+

PR #374, opened 2017-11-21
FT: Export constructStringToSignV2 module

alexanderchan-scality:
Exports v2 version of constructStringtoSign for use with the coming GCP
backend.

### DIFF
--- a/lib/auth/auth.js
+++ b/lib/auth/auth.js
@@ -7,6 +7,7 @@ const AuthInfo = require('./AuthInfo');
 const v2 = require('./v2/authV2');
 const v4 = require('./v4/authV4');
 const constants = require('../constants');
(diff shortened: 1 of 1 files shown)


lib/auth/auth.js
```diff
@@ -212,6 +212,7 @@ module.exports = {
     },
     client: {
         generateV4Headers,
+        constructStringToSignV2,
```
nicolas2bert: Don't you also need `constructStringToSignV4`?

alexanderchan-scality: Don't think so. The Google Storage backend only supports v2 signing at the moment.

rahulreddy: @alexanderchan-scality I though GCP also supported V4 but we don't need to patch the signing module?

PR #377, opened 2017-11-22
ft: S3C-1065: Null part generator

ploki:
This provides a new class, NullStream, that will be used to read null parts.
(null parts are parts with null keys which are generated by the NFS server when
growing a file with truncate)

### DIFF
--- a/lib/s3middleware/nullStream.js
+++ b/lib/s3middleware/nullStream.js
@@ -1,0 +1,42 @@
+const Readable = require('stream').Readable;
+
+
+/**
+ * This class is used to produce zeros filled buffers for a reader consumption
+ */
+class NullStream extends Readable {
+
+    /**
+     * Construct a new zeros filled buffers producer that will
+     * produce as much bytes as specified by the range parameter, or the size
+     * parameter if range is null or not constituted of 2 elements
+     * @constructor
+     * @param {integer} size - the number of null bytes to produce
+     * @param {array} range - a range specification to override to size
+     */
+    constructor(size, range) {
+        super({});
+        if (Array.isArray(range) && range.length === 2) {
+            this.bytesToRead = range[1] - range[0] + 1;
+        } else {
+            this.bytesToRead = size;
+        }
+    }
+
+    /**
+     * This function generates the stream of null bytes
+     *
+     * @param {integer} size - advisory amount of data to produce
+     * @returns {undefined}
+     */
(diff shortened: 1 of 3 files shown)


lib/s3middleware/nullStream.js
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Construct a new zeros filled buffers producer that will
+     * produce as much bytes as specified by the range parameter, or the size
+     * parameter if range is null or not constituted of 2 elements
+     * @constructor
+     * @param {integer} size - the number of null bytes to produce
+     * @param {array} range - a range specification to override to size
+     */
+    constructor(size, range) {
+        super({});
```
philipyoo: Just curious, why pass empty object here?

ploki: To be consistent with other stream specializations across all the projects

lib/s3middleware/nullStream.js
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Construct a new zeros filled buffers producer that will
+     * produce as much bytes as specified by the range parameter, or the size
+     * parameter if range is null or not constituted of 2 elements
+     * @constructor
+     * @param {integer} size - the number of null bytes to produce
+     * @param {array} range - a range specification to override to size
+     */
+    constructor(size, range) {
```
jonathan-gramain: IMO, this class should not have to deal with ranges, with the goal of making this API as simple as it could/should be for a generic implementation I would recommend to only support the `size` parameter and let the user do his range to size conversion on his own if he has to.

lib/s3middleware/nullStream.js
```diff
@@ -0,0 +1,42 @@
...
+        super({});
+        if (Array.isArray(range) && range.length === 2) {
+            this.bytesToRead = range[1] - range[0] + 1;
+        } else {
+            this.bytesToRead = size;
+        }
+    }
+
+    /**
+     * This function generate the stream of null bytes
```
jonathan-gramain: generate*s*

lib/s3middleware/nullStream.js
ploki: Makes sense. But in this particular case, it mimics the API of the data wrapper client ```get``` method.

rahulreddy:
@ironman-machine r+

PR #379, opened 2017-11-29
bf: stream response from getRaftLog()

jonathan-gramain:
Adapt LogConsumer.readRecords() to use the stream returned by the modified
BucketClient.getRaftLog() function. That allows end-to-end streaming, hence
supporting arbitrary-sized responses, which should avoid toString() exceptions
or consume excessive amounts of memory.

Goes with https://github.com/scality/bucketclient/pull/126


rahulreddy:
@ironman-machine 
SCALITY_BUCKETCLIENT_BRANCH=bf/S3C-1134-streamGetRaftLog
r+

PR #385, opened 2017-12-07
rf: avoid hacky trick to store number subparts

electrachong:
Dependency of https://github.com/scality/Mystique/pull/112 and https://github.com/scality/S3/pull/1049

### DIFF
--- a/lib/s3middleware/azureHelpers/mpuUtils.js
+++ b/lib/s3middleware/azureHelpers/mpuUtils.js
@@ -1,3 +1,4 @@
+const assert = require('assert');
 const crypto = require('crypto');
 const stream = require('stream');
 
@@ -76,6 +77,7 @@ azureMpuUtils.getSubPartInfo = dataContentLength => {
         numberFullSubParts + 1 : numberFullSubParts;
     const lastPartSize = remainder || azureMpuUtils.maxSubPartSize;
     return {
+        expectedNumberSubParts: numberSubParts,
         lastPartIndex: numberSubParts - 1,
         lastPartSize,
     };
@@ -95,7 +97,6 @@ azureMpuUtils.putSinglePart = (errorWrapperFn, request, params, dataStoreName,
 log, cb) => {
     const { bucketName, partNumber, size, objectKey, contentMD5, uploadId }
         = params;
-    const totalSubParts = 1;
     const blockId = azureMpuUtils.getBlockId(uploadId, partNumber, 0);
     const passThrough = new stream.PassThrough();
     const options = {};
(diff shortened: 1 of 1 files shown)


lib/s3middleware/azureHelpers/mpuUtils.js
```diff
@@ -172,18 +172,21 @@ dataStoreName, log, cb) => {
                 `Error returned from Azure: ${err}`));
         }
         const numberSubParts = results.length;
+        if (numberSubParts !== subPartInfo.expectedNumberSubParts) {
+            return cb(errors.InternalError.customizeDescription('Unexpected ' +
+            'result streaming part to Azure. Please retry.'));
+        }
```
LaurenSpiegel: what is left in azure at this point? will a new put overwrite or are we going to have a size issue?

electrachong: A new put will overwrite; but if somehow we had streamed one more subpart than expected, that block will remain and take up one block of the total block limit. It shouldn't happen anyway, but it's a safety check. Even if it does happen, we would still be able to retrieve only the correct data.

electrachong: @LaurenSpiegel does returning an InternalError make sense here or still return ServiceUnavailable?

LaurenSpiegel: should be an assertion since it is a code problem.

electrachong: updated

lib/s3middleware/azureHelpers/mpuUtils.js
```diff
@@ -172,18 +173,27 @@ dataStoreName, log, cb) => {
                 `Error returned from Azure: ${err}`));
         }
         const numberSubParts = results.length;
+        // check if we have streamed more parts than calculated; should not
+        // occur, but do a sanity assertion to detect any coding logic error
+        if (numberSubParts !== subPartInfo.expectedNumberSubParts) {
```
LaurenSpiegel: if we're going to do the assertion, we should just do it. No need for the conditional check and the log line. If we crash it will be in the logs.

electrachong: updated so it's just the assertion now

rahulreddy:
@ironman-machine 
r=4c282c4

rahulreddy:
@ironman-machine
r=4c282c4

rahulreddy:
@ironman-machine r-

rahulreddy:
@ironman-machine
SCALITY_S3_BRANCH=rf/hacky-subpart-trick
r=4c282c4

PR #389, opened 2017-12-27
Ft/s3 c 1155 put bucket lifecycle

dora-korpar:
Creates LifecycleConfiguration class for S3 put bucket lifecycle API

### DIFF
--- a/lib/models/LifecycleConfiguration.js
+++ b/lib/models/LifecycleConfiguration.js
@@ -1,0 +1,653 @@
+const assert = require('assert');
+const UUID = require('uuid');
+
+const errors = require('../errors');
+
+/**
+ * Format of xml request:
+
+ <LifecycleConfiguration>
+    <Rule>
+        <ID>id1</ID>
+        <Filter>
+            <Prefix>logs/</Prefix>
+        </Filter>
+        <Status>Enabled</Status>
+        <Expiration>
+            <Days>365</Days>
+        </Expiration>
+    </Rule>
+    <Rule>
+        <ID>DeleteAfterBecomingNonCurrent</ID>
+        <Filter>
+            <And>
+                <Prefix>logs/</Prefix>
+                <Tag>
+                    <Key>key1</Key>
+                    <Value>value1</Value>
+                </Tag>
+            </And>
+        </Filter>
+        <Status>Enabled</Status>
+        <NoncurrentVersionExpiration>
+            <NoncurrentDays>1</NoncurrentDays>
+        </NoncurrentVersionExpiration>
+        <AbortIncompleteMultipartUploads>
+            <DaysAfterInitiation>1</DaysAfterInitiation>
+        </AbortIncompleteMultipartUploads>
+    </Rule>
+</LifecycleConfiguration>
+ */
+
+/**
+ * Format of config:
+
+  config = {
+    rules = [
+        {
+            ruleID: <value>,
+            ruleStatus: <value>,
+            filter: {
+                rulePrefix: <value>,
+                tags: [
+                    {
+                        key: <value>,
+                        val: <value>
+                    },
+                    {
+                        key: <value>,
+                        val: <value>
+                    }
+                ]
+            },
+            actions: [
+                {
+                    actionName: <value>,
+                    days: <value>,
+                    date: <value>,
+                    deleteMarker: <value>
+                },
+                {
+                    actionName: <value>,
+                    days: <value>,
+                    date: <value>,
+                    deleteMarker: <value>,
+                },
+            ]
+        }
+    ]
+  };
+ */
+
+class LifecycleConfiguration {
+    /**
+     * Create a Lifecycle Configuration instance
+     * @param {string} xml - the parsed xml
+     * @return {object} - LifecycleConfiguration instance
+     */
+    constructor(xml) {
+        this._parsedXML = xml;
+        this._ruleIDs = [];
+        this._tagKeys = [];
+        this._config = {};
+    }
+
+    /**
+     * Get the lifecycle configuration
+     * @return {object} - the lifecycle configuration
+     */
+    getLifecycleConfiguration() {
+        const rules = this._buildRulesArray();
+        if (rules.error) {
+            this._config.error = rules.error;
+        }
+        return this._config;
+    }
+
+    /**
+     * Build the this._config.rules array
+     * @return {object} - contains error if any rule returned an error
+     * or parsing failed
+     */
+    _buildRulesArray() {
+        const rules = {};
+        this._config.rules = [];
+        if (!this._parsedXML || this._parsedXML === '') {
+            rules.error = errors.MalformedXML.customizeDescription(
+                'request xml is undefined or empty');
+            return rules;
+        }
+        if (!this._parsedXML.LifecycleConfiguration &&
+            this._parsedXML.LifecycleConfiguration !== '') {
+            rules.error = errors.MalformedXML.customizeDescription(
+                'request xml does not include LifecycleConfiguration');
+            return rules;
+        }
+        const lifecycleConf = this._parsedXML.LifecycleConfiguration;
+        const rulesArray = lifecycleConf.Rule;
+        if (!rulesArray || !Array.isArray(rulesArray)
+        || rulesArray.length === 0) {
+            rules.error = errors.MalformedXML.customizeDescription(
+                'request xml does not include at least one rule');
+            return rules;
+        }
+        if (rulesArray.length > 1000) {
+            rules.error = errors.MalformedXML.customizeDescription(
+                'request xml includes over max limit of 1000 rules');
+            return rules;
+        }
+        for (let i = 0; i < rulesArray.length; i++) {
+            const rule = this._parseRule(rulesArray[i]);
+            if (rule.error) {
+                rules.error = rule.error;
+                break;
+            } else {
+                this._config.rules.push(rule);
+            }
+        }
+        return rules;
+    }
+
+    /**
+     * Check that each xml rule is valid
+     * @param {object} rule - a rule object from Rule array from this._parsedXml
+     * @return {object} - contains error if any component returned an error
+     * or parsing failed, else contains parsed rule object
+     *
+     * Format of ruleObj:
+     * ruleObj = {
+     *      ruleID: <value>,
+     *      ruleStatus: <value>,
+     *      filter: {
+     *          rulePrefix: <value>,
+     *          tags: [
+     *              {
+     *                  key: <value>,
+     *                  val: <value>,
+     *              }
+     *          ]
+     *      }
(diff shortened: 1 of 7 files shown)


lib/models/LifecycleConfiguration.js
```diff
@@ -0,0 +1,623 @@
...
+     */
+    _parseRule(rule) {
+        const ruleObj = {};
+        if (rule.Transition || rule.NoncurrentVersionTransition) {
+            ruleObj.error = errors.NotImplemented.customizeDescription(
+                'Transition lifecycle action not yet implemented');
+            return ruleObj;
+        }
+        if ((!rule.Filter && rule.Filter !== '') &&
+        (!rule.Prefix && rule.Prefix !== '')) {
```
philipyoo: maybe `(!rule.Filter || rule.Filter === '')` and also for `Prefix`. If `rule.Filter` and `rule.Prefix` are both empty strings, this will pass

lib/models/LifecycleConfiguration.js
```diff
@@ -0,0 +1,623 @@
...
+            return ruleObj;
+        }
+        if (!rule.Status) {
+            ruleObj.error = errors.MalformedXML.customizeDescription(
+                'Rule xml does not include Status');
+            return ruleObj;
+        }
+        const subFilter = rule.Filter ? rule.Filter[0] : rule.Prefix;
+
+        const id = this._parseID(rule.ID);
```
philipyoo: just checking here because I'm not exactly sure, is `rule.ID` not an array like `Filter` or `Status`?

lib/models/LifecycleConfiguration.js
```diff
@@ -0,0 +1,623 @@
...
+            if (filter.length > 1) {
+                filterObj.error = errors.MalformedXML.customizeDescription(
+                    'Rule includes more than one Prefix');
+                return filterObj;
+            }
+            filterObj.rulePrefix = filter[0];
+            return filterObj;
+        }
+        if (filter.And && (filter.Prefix || filter.Tag) ||
+        (filter.Prefix && filter.Tag)) {
```
philipyoo: I'm guessing checking from left to right, but just to be safe, maybe add an outer parens around the first `&&` check

lib/models/LifecycleConfiguration.js
```diff
@@ -0,0 +1,623 @@
...
+    /**
+     * Check that each tag object is valid
+     * @param {object} tag - a tag object from a filter object
+     * @return {boolean} - indicates whether tag object is valid
+     */
+    _checkTagError(tag) {
+        if (!tag.Key || !tag.Value) {
+            return true;
+        }
+        if (tag.Key.length !== tag.Value.length) {
```
philipyoo: also checking here, why must the tag key and value be the same length?

lib/models/LifecycleConfiguration.js
```diff
@@ -0,0 +1,623 @@
...
+        }
+        if (!this._parsedXML.LifecycleConfiguration &&
+            this._parsedXML.LifecycleConfiguration !== '') {
+            rules.error = errors.MalformedXML.customizeDescription(
+                'request xml does not include LifecycleConfiguration');
+            return rules;
+        }
+        const lifecycleConf = this._parsedXML.LifecycleConfiguration;
+        const rulesArray = lifecycleConf.Rule;
+        if (!rulesArray || rulesArray.length === 0) {
```
bennettbuchanan: May be worthwhile to check `rulesArray` is an array before comparing its length.

lib/models/LifecycleConfiguration.js
dora-korpar: I know this logic is confusing >.< but it's meant to be that way because the conditional is supposed to pass if either Filter or Prefix is an empty string. I'll add a comment to hopefully clarify a bit

lib/models/LifecycleConfiguration.js
dora-korpar: It is, but it can also be totally omitted from the passed config, so I check for an array in the `_parseID` function :)

lib/models/LifecycleConfiguration.js
```diff
@@ -0,0 +1,623 @@
...
+        }
+        if (rulesArray.length > 1000) {
+            rules.error = errors.MalformedXML.customizeDescription(
+                'request xml includes over max limit of 1000 rules');
+            return rules;
+        }
+        for (let i = 0; i < rulesArray.length; i++) {
+            const rule = this._parseRule(rulesArray[i]);
+            if (rule.error) {
+                rules.error = rule.error;
```
bennettbuchanan: If I'm not mistaken, we could return immediately here because we will send an error response and don't care about the remaining rules.

lib/models/LifecycleConfiguration.js
```diff
@@ -0,0 +1,623 @@
...
+
+        const id = this._parseID(rule.ID);
+        const status = this._parseStatus(rule.Status[0]);
+        const filter = this._parseFilter(subFilter);
+        const actions = this._parseAction(rule);
+
+        const rulePropArray = [id, status, filter, actions];
+        rulePropArray.forEach(prop => {
+            if (prop.error) {
+                ruleObj.error = prop.error;
```
bennettbuchanan: Same comment about returning here immediately here.

lib/models/LifecycleConfiguration.js
dora-korpar: the Key and Value array lengths should be the same because there has to be one value for every key. wouldn't do to have 5 keys and only 4 values, or the other way around :)

dora-korpar:
@ironman-machine try

lib/models/LifecycleConfiguration.js
```diff
@@ -0,0 +1,623 @@
...
+        return filterObj;
+    }
+
+    /**
+     * Check that each tag object is valid
+     * @param {object} tag - a tag object from a filter object
+     * @return {boolean} - indicates whether tag object is valid
+     */
+    _checkTagError(tag) {
+        if (!tag.Key || !tag.Value) {
```
bennettbuchanan: Object tagging can have a `''` value. If true also for Lifecycle, we could add a check for this.

dora-korpar: Even if the key has a `''` value, since `Value` is an array, it would still return true. But there do need to be a few more checks here

bennettbuchanan: Oops, yes I misread that.

dora-korpar:
@ironman-machine try

lib/models/LifecycleConfiguration.js
```diff
@@ -0,0 +1,653 @@
...
+
+    /**
+     * Build the this._config.rules array
+     * @return {object} - contains error if any rule returned an error
+     * or parsing failed
+     */
+    _buildRulesArray() {
+        const rules = {};
+        this._config.rules = [];
+        if (!this._parsedXML || this._parsedXML === '') {
```
jonathan-gramain: you can go with the first test only, as an empty string is true-ish (`!''` is true), though considering that you access attributes of `_parsedXML` next, maybe a better validation can be to check that it is a JS object (`typeof this._parsedXML === 'object'`) otherwise it would raise an exception when getting the `LifecycleConfiguration` attribute if the user passes a non-empty string.

As a general comment for this whole function, using the `joi` module to do the first pass of validation with declarative syntax, then doing the few domain-specific checks not supported by joi would likely lead to much more compact and maintainable code (if you decide to go this way, have a look at how it's used in backbeat + reference at https://www.npmjs.com/package/joi, otherwise let me know I will happily review the remaining parsing code, I'll skip it for now as it's a significant chunk to review that will be wasted time if you choose to rewrite with `joi` :wink: )

lib/models/LifecycleConfiguration.js
```diff
@@ -0,0 +1,653 @@
...
+    _parseExpiration(rule) {
+        const expObj = {};
+        const subExp = rule.Expiration[0];
+        if (!subExp.Date && !subExp.Days && !subExp.ExpiredObjDeleteMarker) {
+            expObj.error = errors.MalformedXML.customizeDescription(
+                'Expiration action does not include an action time');
+            return expObj;
+        }
+        if (subExp.Date) {
+            const isoRegex = new RegExp('^(-?(?:[1-9][0-9]*)?[0-9]{4})-' +
```
jonathan-gramain: This regexp can deserve to be put in `constants.js`.

tests/unit/models/LifecycleConfiguration.js
```diff
@@ -0,0 +1,271 @@
...
+        <NoncurrentVersionExpiration>
+            <NoncurrentDays>1</NoncurrentDays>
+        </NoncurrentVersionExpiration>
+    </Rule>
+</LifecycleConfiguration>
+ */
+
+const requiredTags = [
+    { tag: 'LifecycleConfiguration',
+        errMessage: 'request xml is undefined or empty' },
```
jonathan-gramain: Checking error messages sounds a little bit overkill, compared to the additional cost to maintain the error messages both in the feature code and the tests (to fix typos/reformulate). I would rather check only the error codes (names) which are less likely to change than the full message.

lib/models/LifecycleConfiguration.js
dora-korpar: Yeah, I saw that module in backbeat, but I thought it would be better not to use it since it's not used anywhere else in the Arsenal repo. I can rewrite it if you think it's worth it though :)

tests/unit/models/LifecycleConfiguration.js
dora-korpar: I agree, but since so many of the errors returned are `MalformedXML`, I wanted to be sure that the error was being thrown in the correct place.
That was especially true when I was ironing out the logic for the parsing, but if you still think it should be changed, let me know :)

lib/models/LifecycleConfiguration.js
jonathan-gramain: I personally think it's worth it, because it can both simplify the code a lot and get rid of potential issues in the current code (because of its length and relative complexity) and I think it was on Rahul's mind concerning the refactoring projects (using joi where configuration parsing occurs).

Now, I find it rather simple and pleasant to use joi module but there's some small effort anyway to get into it, and it's always a bit annoying to rewrite a piece of code we just wrote, so feel free to do the change if you have some time for it but I'm not pushing either, as long as we have working code :)

tests/unit/models/LifecycleConfiguration.js
jonathan-gramain: No you can leave as is no worry :)

dora-korpar:
@ironman-machine try

bennettbuchanan:
@ironman-machine r+

bennettbuchanan:
@ironman-machine r+

PR #392, opened 2018-01-02
bf: close/end readable/response streams on errors

rahulreddy:
This fixes the leakage of sockets in CLOSE_WAIT state by closing the streams
and destroying the sockets when the client has abruptly closed the connection.

Upstream requests to Azure/AWS need to be aborted in
AzureClient/AWSClient implementation. Currently azure-storage module doesn't
have a clear way of aborting a request.

### DIFF
--- a/lib/s3routes/routesUtils.js
+++ b/lib/s3routes/routesUtils.js
@@ -1,6 +1,11 @@
 const url = require('url');
+const { eachSeries } = require('async');
 const ipCheck = require('../ipCheck');
 const errors = require('../errors');
+
+const responseErr = new Error();
+responseErr.code = 'ResponseError';
+responseErr.message = 'response closed by client request before all data sent';
 
 /**
  * setCommonResponseHeaders - Set HTTP response headers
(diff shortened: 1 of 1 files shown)


LaurenSpiegel:
Is there a change to be made to the azure client code or is nothing exposed in their api?

rahulreddy:
@LaurenSpiegel There is a workaround filter which looked way to complicated to implement for abort. I think they are in the process of implemented a simple abort method. For now, I think I will dig through the object to find the socket and destroy it. I will make a PR in S3 .

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r- r+

alexandre-merle:
This should have not been merged, its an ultron bug, leading to broken rel/7.2, can someone revert this ?

DavidPineauScality:
My bad, it looks like my playing with Ultron triggered this bug... :/

alexandre-merle:
@DavidPineauScality already fixed :p

DavidPineauScality:
Yup I know, just wanted to apologize for the inconveniences :(

PR #393, opened 2018-01-08
Rf/livy

JianqinWang:
Still a work in progress (need to add more tests)
Combine functions for sessions and batches if similar operation
Additional functionality to fully support the REST API
Re-factored existing tests so they can be executed consistently on different machines

### DIFF
--- a/lib/livyClient/lib/client.js
+++ b/lib/livyClient/lib/client.js
@@ -6,50 +6,72 @@ const https = require('https');
 const querystring = require('querystring');
 const Backoff = require('backo');
 // min and max in ms
-const GET_BACKOFF_PARAMS = { min: 100, max: 200, maxAttempts: 300,
+const defaultGetBackoffParams = { min: 100, max: 200, maxAttempts: 300,
     factor: 1.01 };
-const POST_BACKOFF_PARAMS = { min: 1000, max: 3000, maxAttempts: 20,
+const defaultPostBackoffParams = { min: 1000, max: 3000, maxAttempts: 20,
     factor: 1.1 };
-
 
 class LivyClient {
     /**
      * Constructor for REST client to apache livy
      *
-     * @param {string} host - hostname or IP of the livy server
-     * @param {number} [port=8998] - port of the livy server
-     * @param {object} [logger=console] - logger object
-     * @param {boolean} [useHttps] - whether to use https or not
-     * @param {string} [key] - https private key content
-     * @param {string} [cert] - https public certificate content
-     * @param {string} [ca] - https authority certificate content
+     * @param {object} params - object for all parameters to initialize Livy
      * @return {undefined}
      */
-    constructor(host, port = 8998, logger = console, useHttps, key, cert, ca) {
-        assert(typeof host === 'string' && host !== '', 'host is required');
-        assert(Number.isInteger(port), 'port must be an integer');
-        assert(typeof logger === 'object', 'logger must be an object');
-        assert(typeof logger.error === 'function', 'logger must have' +
-        'error method');
-        assert(typeof logger.info === 'function', 'logger must have' +
-        'info method');
-        assert(key === undefined || typeof key === 'string',
-                'key must be a string');
-        assert(cert === undefined || typeof cert === 'string',
-                'cert must be a string');
-        assert(ca === undefined || typeof ca === 'string',
-                'ca must be a string');
-        this.serverHost = host;
-        this.serverPort = port;
-        this.logger = logger;
-        this._key = key;
-        this._cert = cert;
-        this._ca = ca;
-        this.useHttps = (useHttps === true);
+    constructor(params) {
+        assert(typeof params.host === 'string' && params.host !== '',
+            'host is required');
+        assert(params.port === undefined || Number.isInteger(params.port),
+            'port must be an integer');
+        assert(params.logger === undefined || typeof params.logger === 'object',
+            'logger must be an object');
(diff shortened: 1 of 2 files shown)


lib/livyClient/lib/client.js
```diff
@@ -11,45 +11,82 @@ const GET_BACKOFF_PARAMS = { min: 100, max: 200, maxAttempts: 300,
...
+            'logger must have info method');
+        assert(params.key === undefined || typeof params.key === 'string',
+            'key must be a string');
+        assert(params.cert === undefined || typeof params.cert === 'string',
+            'cert must be a string');
+        assert(params.ca === undefined || typeof params.ca === 'string',
+            'ca must be a string');
+        assert(params.getBackoffParams === undefined ||
+            typeof params.getBackoffParams === 'object',
+            'get_backoff_params must be an object');
```
LaurenSpiegel: let's be consistent and use camelcase throughout.

lib/livyClient/lib/client.js
```diff
@@ -11,45 +11,82 @@ const GET_BACKOFF_PARAMS = { min: 100, max: 200, maxAttempts: 300,
...
+        assert(params.postBackoffParams === undefined ||
+            typeof params.postBackoffParams === 'object',
+            'put_backoff_params must be an object');
```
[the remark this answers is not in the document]
LaurenSpiegel: postBackoffParams

lib/livyClient/lib/client.js
```diff
@@ -11,45 +11,82 @@ const GET_BACKOFF_PARAMS = { min: 100, max: 200, maxAttempts: 300,
...
+            'postBackoffParams should have valid numerical values for ' +
+            'the min, max, maxAttempts, and factor attributes');
+        this.serverHost = params.host;
+        this.serverPort = params.port !== undefined ? params.port : 8998;
+        this.logger = params.logger !== undefined ? params.logger : console;
+        this._key = params.key;
+        this._cert = params.cert;
+        this._ca = params.ca;
+        this._getBackoffParams = params.getBackoffParams !== undefined ?
+            params.getBackoffParams : GET_BACKOFF_PARAMS;
```
LaurenSpiegel: let's call this defaultGetBackoffParams now

lib/livyClient/lib/client.js
```diff
@@ -11,45 +11,82 @@ const GET_BACKOFF_PARAMS = { min: 100, max: 200, maxAttempts: 300,
...
+        this._postBackoffParams = params.postBackoffParams !== undefined ?
+            params.postBackoffParams : POST_BACKOFF_PARAMS;
```
LaurenSpiegel: same idea here

lib/livyClient/lib/client.js
```diff
@@ -11,45 +11,82 @@ const GET_BACKOFF_PARAMS = { min: 100, max: 200, maxAttempts: 300,
 const POST_BACKOFF_PARAMS = { min: 1000, max: 3000, maxAttempts: 20,
     factor: 1.1 };
 
+function inputCheck(batchOrSession, id, callback) {
```
LaurenSpiegel: move this into the class as a method and call _inputCheck to make clearer that internal function

lib/livyClient/lib/client.js
```diff
@@ -62,27 +99,43 @@ class LivyClient {
...
+     * @param {number} [numOfSessionOrBatch] - number of sessions to
      * return
      * @param {function} callback - callback
      * @return {undefined}
      */
-    getSessions(startIndex, numOfSessions, callback) {
-        assert(typeof callback === 'function', 'callback must be a function');
+    getSessionsOrBatches(batchOrSession, startIndex, numOfSessionOrBatch,
+    callback) {
+        inputCheck(batchOrSession, null, callback);
```
LaurenSpiegel: doesn't inputCheck return whether batch or session so don't have to check again below?

lib/livyClient/lib/client.js
```diff
@@ -235,31 +216,84 @@ class LivyClient {
...
     }
 
+    /** Run a statement in a session
+     * @param {number} sessionId - options for session
+     * @param {string} codeToExecute - actual code to be executed by spark
+     * @param {function} callback - callback
+    * @param {object} [backoff] - backoff instance
+     * @return {undefined}
+     */
+    postStatement(sessionId, codeToExecute, callback, backoff) {
```
LaurenSpiegel: was this just moved?

JianqinWang: Yes

lib/livyClient/lib/client.js
JianqinWang: Right now I'm returning the strings 'batch' or 'session', to use to check which request prefix I will be using ('batch' -> '/batches' vs 'session' -> '/sessions'). It makes more sense though to simply return the correct prefix to avoid the second check, thanks for pointing it out!

lib/livyClient/lib/client.js
```diff
@@ -279,19 +300,50 @@ class LivyClient {
...
+     */
+    deleteSessionOrBatch(batchOrSession, id, callback) {
+        const pathPrefix = this._inputCheck(batchOrSession, id, callback);
+        this._request('DELETE', `${pathPrefix}/${id}`,
             null, null, callback);
         return undefined;
     }
 
+    _inputCheck(batchOrSession, id, callback) {
+        if (batchOrSession) {
```
LaurenSpiegel: remove if

PR #395, opened 2018-01-12
FIX: v4 signing tool

LaurenSpiegel:


LaurenSpiegel:
@ironman-machine try

LaurenSpiegel:
@ironman-machine try

rahulreddy:
I believe we encountered this with Utapi as well but I thought we had a fix for it. We may have fixed it in Utapi itself.

rahulreddy:
@ironman-machine r+

PR #397, opened 2018-01-12
Ft/s3 c 1156 get bucket lifecycle

dora-korpar:
Requires https://github.com/scality/Arsenal/pull/389 to be merged first

PR #398, opened 2018-01-12
Ft/s3 c 1157 delete bucket lifecycle

dora-korpar:
Requires https://github.com/scality/Arsenal/pull/397 to be merged first


bennettbuchanan:
@ironman-machine r+

PR #401, opened 2018-01-15
ft(health): add Too many requests error

ThibaultRiviere:


package.json
```diff
@@ -24,7 +24,7 @@
     "ioredis": "2.4.0",
     "ipaddr.js": "1.2.0",
     "joi": "^10.6",
-    "level": "~1.6.0",
+    "level": "scality/level#proto/custom-env",
```
DavidPineauScality: dropme ?

ThibaultRiviere: Yes it's because of the conflict between the deps of arsenal and MD . Will remove that when everything is finished !

DavidPineauScality: Ok, but could we not already merge this PR ? It does not break anything, and at least we'll be one PR down :)

ThibaultRiviere: Ok :)

ThibaultRiviere:
@DavidPineauScality can be merge

DavidPineauScality:
Merging, as this is non-colliding with any other changes in the project; and allow related PRs to make progress.
@ironman-machine r+

PR #409, opened 2018-02-07
EVE-817 - Setting up EVE pipeline

tcarmet:


eve/main.yml
```diff
@@ -0,0 +1,48 @@
+---
+version: 0.2
+
+branches:
+  default:
+    stage: pre-merge
+
+models:
```
sylvain-killian: you don't need models since it is used only once, it would be clearer with a flat yaml

eve/main.yml
```diff
@@ -0,0 +1,48 @@
...
+    worker: &master-worker
+      type: docker
+      path: eve/workers/master
+      volumes:
+        - '/home/eve/workspace'
+    steps:
+      - Git: *clone
+      - ShellCommand:
+          name: install
+          command: npm install
```
sylvain-killian: external deps must be included in the docker image to reduce depedency to outside world during the build

eve/main.yml
```diff
@@ -0,0 +1,48 @@
...
+      - ShellCommand:
+          name: run test
+          command: npm run --silent test
+      - ShellCommand:
+          name: run ft_test
+          command: npm run ft_test
+      - ShellCommand:
+          name: run executables tests
+          command: |
+            cd lib/executables/pensieveCreds &&
```
sylvain-killian: you could use workdir ShellCommand option

eve/main.yml
```diff
@@ -0,0 +1,48 @@
...
+            ln -s ../../../node_modules . &&
+            npm install && npm test
```
sylvain-killian: putting the package file in the worker dir, then link it at root, may give better code?

eve/workers/master/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+# Misc
+#
+
+# RUN echo '127.0.0.1    bucketwebsitetester.s3-website-us-east-1.amazonaws.com' >> /etc/hosts
+
+#
+# Run buildbot-worker on startup
+#
+
+ARG BUILDBOT_VERSION=0.9.7
```
sylvain-killian: 0.9.12 currently

tcarmet: Done

eve/main.yml
tcarmet: Now using workdir

eve/main.yml
tcarmet: I've removed the model

eve/main.yml
tcarmet: In this case, it's another `package.json` file that is called. I've linked the main one which is located on the root directory to the worker. Not sure if will do the same with this one.

eve/main.yml
tcarmet: There's no more npm install, all set with the docker image

eve/main.yml
tcarmet: and btw, this step takes exactly 4 seconds.

PR #413, opened 2018-02-08
EVE-817 - Setting up EVE pipeline

tcarmet:


ThibaultRiviere:
I think you can close the others PR, once merged, you will have to forward this one until master

tcarmet:
@ThibaultRiviere Ok! will do!

tcarmet:
@ThibaultRiviere all the other PR are now closed. We can merge this PR if everything sounds good to you and your team.

eve/main.yml
```diff
@@ -0,0 +1,43 @@
...
+          command: npm run --silent lint_md
+      - ShellCommand:
+          name: run test
+          command: npm run --silent test
+      - ShellCommand:
+          name: run ft_test
+          command: npm run ft_test
+      - ShellCommand:
+          name: run executables tests
+          command: npm install && npm test
```
ThibaultRiviere: Why an other npm install ?
Should we change the order ? like first `npm run --silent test` and then `npm run ft_test`

tcarmet: Actually the `npm install && npm test`  was defined in the `master` version of the `circle.yml`

Indeed it's not available in the 6.4 version so I don't have any good explanation. But it works so I've decided to leave it. (we can remove it in this version if you think it doesn't make sense here).

If you take a look at the master version, there's the following line:
```shell
cd lib/executables/pensieveCreds && npm install && npm test
```
Which correspond to this step.

DavidPineauScality:
@ironman-machine r+

eve/main.yml
DavidPineauScality: MHh this actually appeared much later than 6.4, and should be added afterwards on later versions/branches.

DavidPineauScality:
@ironman-machine r-

eve/main.yml
ThibaultRiviere: As discuss offline. Let's remove it here, forward this change and create and other PR targeting rel/7.2 to add it where it make sens . 
Thanks :)

tcarmet:
@ThibaultRiviere @DavidPineauScality I've removed the requested part.

DavidPineauScality:
@ironman-machine r+

tcarmet:
need your approval @DavidPineauScality

DavidPineauScality:
Merging manually to avoid a full re-cycle; my bad ~!

tcarmet:
Thank you :heart:

PR #422, opened 2018-02-13
ft(test): EVE-817 add pensieveCreds tests

ThibaultRiviere:
Work from tcarmet, just adding it


DavidPineauScality:
Merging manually without E2E as this only impact EVE builds for Arsenal.

PR #423, opened 2018-02-13
bf: fix get lifecycle xml

dora-korpar:
There was an issue using AWS cli's in getting a bucket's lifecycle due to xml formatting.

### DIFF
--- a/lib/models/LifecycleConfiguration.js
+++ b/lib/models/LifecycleConfiguration.js
@@ -672,9 +672,15 @@ class LifecycleConfiguration {
                     return Tag;
                 }).join('');
             }
-            const Filter = ((rulePrefix && tags) || (tags && tags.length > 1)) ?
-                `<Filter><And>${Prefix}${tagXML}</And></Filter>` :
-                `<Filter>${Prefix}${tagXML}</Filter>`;
+            let Filter;
+            if (rulePrefix && !tags) {
+                Filter = Prefix;
+            } else if (tags && (rulePrefix || tags.length > 1)) {
+                Filter = `<Filter><And>${Prefix}${tagXML}</And></Filter>`;
+            } else {
+                // remaining condition is if only one or no tag
+                Filter = `<Filter>${tagXML}</Filter>`;
+            }
 
             const Actions = actions.map(action => {
                 const { actionName, days, date, deleteMarker } = action;


dora-korpar:
@ironman-machine try

lib/models/LifecycleConfiguration.js
```diff
@@ -672,9 +672,15 @@ class LifecycleConfiguration {
...
-            const Filter = ((rulePrefix && tags) || (tags && tags.length > 1)) ?
-                `<Filter><And>${Prefix}${tagXML}</And></Filter>` :
-                `<Filter>${Prefix}${tagXML}</Filter>`;
+            let Filter;
+            if (rulePrefix && !tags) {
+                Filter = Prefix;
+            } else if ((rulePrefix && tags) || (tags && tags.length > 1)) {
+                Filter = `<Filter><And>${Prefix}${tagXML}</And></Filter>`;
+            } else {
+                // remaining condition is if only one tag
```
jonathan-gramain: I guess it could also be no tag (but should work anyway)

lib/models/LifecycleConfiguration.js
```diff
@@ -672,9 +672,15 @@ class LifecycleConfiguration {
                     return Tag;
                 }).join('');
             }
-            const Filter = ((rulePrefix && tags) || (tags && tags.length > 1)) ?
-                `<Filter><And>${Prefix}${tagXML}</And></Filter>` :
-                `<Filter>${Prefix}${tagXML}</Filter>`;
+            let Filter;
+            if (rulePrefix && !tags) {
+                Filter = Prefix;
```
jonathan-gramain: Does this fix because the prefix was not included in the previous case? Asking because before I see `${Prefix}` always part of the `Filter` output in the previous code.

lib/models/LifecycleConfiguration.js
```diff
@@ -672,9 +672,15 @@ class LifecycleConfiguration {
...
+            } else if ((rulePrefix && tags) || (tags && tags.length > 1)) {
```
jonathan-gramain: You may replace with the slightly simpler form `if (tags && (rulePrefix || tags.length > 1))`

jonathan-gramain:
The fix should target `rel/7.4` instead of `master` right?

lib/models/LifecycleConfiguration.js
dora-korpar: It fixes it because before `${Prefix}` was always within the `<Filter>` xml tag, but the cli doesn't support Filter, so it wasn't parsing the Prefix from it

dora-korpar:
Rebased to target `rel/7.4` branch

dora-korpar:
@ironman-machine try

lib/models/LifecycleConfiguration.js
jonathan-gramain: :+1:

bennettbuchanan:
@ironman-machine r+

bennettbuchanan:
@ironman-machine r-

bennettbuchanan:
@ironman-machine
SCALITY_DEFAULT_BRANCH=rel/7.4
r+

bennettbuchanan:
@ironman-machine r-

bennettbuchanan:
@ironman-machine
SCALITY_DEFAULT_BRANCH=
DEFAULT_BRANCH=rel/7.4
r+

bennettbuchanan:
@ironman-machine r-

bennettbuchanan:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ultron/rel/7.4
DEFAULT_BRANCH=rel/7.4
r+

bennettbuchanan:
@ironman-machine r+

PR #424, opened 2018-02-15
S3C-1026: S3 consumes too much tcp memory

flavien-scality:
retrieveData rework :
Use the proven 6.4 code with an eachSeries, replacing the recursion.
Isolate Azure in a separate function.

### DIFF
--- a/lib/s3routes/routesUtils.js
+++ b/lib/s3routes/routesUtils.js
@@ -1,6 +1,11 @@
 const url = require('url');
 const ipCheck = require('../ipCheck');
 const errors = require('../errors');
+const { eachSeries } = require('async');
+
+const responseErr = new Error();
+responseErr.code = 'ResponseError';
+responseErr.message = 'response closed by client request before all data sent';
 
 /**
  * setCommonResponseHeaders - Set HTTP response headers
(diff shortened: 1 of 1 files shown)


lib/s3routes/routesUtils.js
```diff
@@ -1,6 +1,11 @@
 const url = require('url');
 const ipCheck = require('../ipCheck');
 const errors = require('../errors');
+const { eachSeries } = require('async');
+
+const responseErr = new Error();
```
jonathan-gramain: I would suggest using one of the ArsenalError duplicated with `.customizeDescription('...')`

lib/s3routes/routesUtils.js
```diff
@@ -242,54 +247,97 @@ function okContentHeadersResponse(overrideParams, resHeaders,
     return response;
 }
 
-function retrieveData(locations, dataRetrievalFn,
-    response, logger, errorHandlerFn) {
+function retrieveDataAzure(locations, dataRetrievalFn, response, logger) {
```
jonathan-gramain: Rename `dataRetrievalFn` to `retrieveDataFn` to match the retrieveData function?

lib/s3routes/routesUtils.js
```diff
@@ -242,54 +247,97 @@ function okContentHeadersResponse(overrideParams, resHeaders,
...
+                    _destroyResponse();
+                    return next(err);
                 });
-                return errorHandlerFn(err);
+                currentStream = readable;
+                return readable.pipe(response, { end: false });
+            }), err => {
+                currentStream = null;
+                if (err) {
+                    log.debug('abort response due to client error', {
```
jonathan-gramain: minor, especially since it's a debug log, but you can also have an error due to the readable object here.

lib/s3routes/routesUtils.js
rahulreddy: This should be ok since the error is not sent to the client but we use it internally in the waterfall.

rahulreddy:
@ironman-machine r+

DavidPineauScality:
kafka build flaky
@ironman-machine r- r+

PR #428, opened 2018-02-20
Fwd/6.4 7.2

alexandre-merle:
no diff expect


DavidPineauScality:
0-diff, no need for testing

PR #425, opened 2018-02-17
Ft/move md backends from S3

JianqinWang:
Extract metadata backends from S3 to here so backbeat can use these features.
Merge this PR before https://github.com/scality/S3/pull/1129 and https://github.com/scality/backbeat/pull/187
JIRA ticket: https://scality.atlassian.net/browse/ZENKO-140

PR #433, opened 2018-02-28
Ft/mongo oplog

JianqinWang:
Merge after https://github.com/scality/Arsenal/pull/425

Based off of https://github.com/scality/Arsenal/blob/master/lib/storage/metadata/bucketclient/LogConsumer.js

### DIFF
--- a/lib/storage/metadata/mongoclient/LogConsumer.js
+++ b/lib/storage/metadata/mongoclient/LogConsumer.js
@@ -1,0 +1,190 @@
+'use strict'; // eslint-disable-line
+
+const stream = require('stream');
+const MongoClient = require('mongodb').MongoClient;
+const format = require('util').format;
+const { Timestamp } = require('bson');
+
+let lastEndID = undefined;
+
+const ops = {
+    i: 'put',
+    u: 'put',
+    d: 'delete',
+};
+
+class ListRecordStream extends stream.Transform {
+    constructor(logger) {
+        super({ objectMode: true });
+        this.logger = logger;
+        this.hasStarted = false;
+        this.start = undefined;
+        this.end = undefined;
+        this.lastUniqID = undefined;
+        // this.unpublishedListing is true once we pass the oplog that has the
+        // start seq timestamp and uniqID 'h'
+        this.unpublishedListing = undefined;
+    }
+
+    _transform(itemObj, encoding, callback) {
+        if (!itemObj) {
+            this.push(null);
+            this.emit('info', {
+                start: this.start,
+                end: this.end,
+                uniqID: this.lastUniqID,
+            });
+            return callback();
+        }
+
+        // always update to most recent uniqID
+        this.lastUniqID = itemObj.h.toString();
+
+        if (this.end === undefined || itemObj.ts.toNumber() > this.end) {
+            this.end = itemObj.ts.toNumber();
+        }
+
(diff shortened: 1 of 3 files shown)


JianqinWang:
@ironman-machine try

JianqinWang:
@ironman-machine try

vrancurel:
@ironman-machine r+

vrancurel:
@ironman-machine r-

vrancurel:
@ironman-machine try

vrancurel:
@ironman-machine r+

lib/storage/metadata/mongoclient/LogConsumer.js
```diff
@@ -0,0 +1,187 @@
...
+    readRecords(params, cb) {
+        const recordStream = new ListRecordStream(this.logger);
+        const limit = params.limit || 10000;
+        const startIDandSeq = params.startSeq.toString().split('|');
+        const startSeq = parseInt(startIDandSeq[0], 10) || 0;
+        lastEndID = startIDandSeq[1];
+
+        this.coll = this.db.collection('oplog.rs');
+        return this.coll.find({
+            ns: /^(?!.*metadata.*(?:__)).*metadata\.\w+.*/,
```
rahulreddy: What is this regular expression? Adding a comment of what the expression is trying to filter out would be helpful. Cleaner way would be writing it as a constant on the top of the file
```js
/*
** This is a comment to explain this filter
*/
const METADATA_FILTER  = /^(?!.*metadata.*(?:__)).*metadata\.\w+.*/
```
and then use it in the code as

```js
 return this.coll.find({
            ns: METADATA_FILTER,
...
```

lib/storage/metadata/mongoclient/LogConsumer.js
```diff
@@ -0,0 +1,187 @@
...
+            ts: { $gte: Timestamp.fromNumber(startSeq) },
+        }, {
+            limit,
+            tailable: false,
+            awaitData: false,
+            noCursorTimeout: true,
+            OplogReplay: true,
+            numberOfRetries: Number.MAX_VALUE,
```
rahulreddy: Why is retry configured to such a high number?

lib/storage/metadata/mongoclient/LogConsumer.js
```diff
@@ -0,0 +1,187 @@
...
+        }, (err, res) => {
+            const stream = res.stream();
+            stream.on('data', data => {
+                recordStream.write(data);
+            });
+            stream.on('end', () => {
+                recordStream.write(undefined);
```
rahulreddy: Shouldn't this be `null` to indicate the end of stream?

lib/storage/metadata/mongoclient/LogConsumer.js
```diff
@@ -0,0 +1,189 @@
...
+            });
+            recordStream.once('info', info => {
+                recordStream.removeAllListeners('error');
```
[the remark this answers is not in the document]
rahulreddy: It's important to understand why that was put in LogConsumer in the first place. I don't like the idea of removing all error listeners. I believe the reason it was used in RecordStream was because Metadata was returning InternalError when we request the next sequence id but there are no logs available. @jonathan-gramain was the author of that class, can you check with him?

PR #434, opened 2018-03-05
ft: extend StatsClient to query by list of ids

philipyoo:
Extend support for querying a list of ids and
returning a total sum of results for that list.

Also add wrapper for redis method `keys`


lib/metrics/RedisClient.js
```diff
@@ -22,6 +22,16 @@ class RedisClient {
...
     }
 
+    /**
+    * get keys for given pattern
+    * @param {string} pattern - string pattern to match with all existing keys
+    * @param {callback} cb - callback (error, result)
+    * @return {undefined}
+    */
+    keys(pattern, cb) {
+        return this._client.keys(pattern, cb);
```
rahulreddy: Instead of `keys`, I would recommend using `scan` https://redis.io/commands/scan since `keys` can accidentally cause a performance hit on the system.
From the docs

> Warning: consider KEYS as a command that should only be used in production environments with extreme care. It may ruin performance when it is executed against large databases. This command is intended for debugging and special operations, such as changing your keyspace layout. Don't use KEYS in your regular application code. If you're looking for a way to find keys in a subset of your keyspace, consider using SCAN or sets.

lib/metrics/RedisClient.js
```diff
@@ -22,6 +22,36 @@ class RedisClient {
...
+    * @return {undefined}
+    */
+    scan(pattern, count, cb) {
+        let callback;
+        // set default count to 10
+        const params = { match: pattern, count: 10 };
+        const keys = [];
+
+        if (!cb) {
+            callback = count;
```
bennettbuchanan: I guess since the `count` param is optional, this could actually end up being the callback function? Wondering if there's a more clear way of handling this case rather than a reassignment.

lib/metrics/StatsClient.js
```diff
@@ -105,6 +105,49 @@ class StatsClient {
...
+
+        const statsRes = {
+            'requests': 0,
+            '500s': 0,
+            'sampleDuration': this._expiry,
+        };
+        let requests = 0;
+        let errors = 0;
+
+        return async.each(ids, (id, done) => {
```
bennettbuchanan: Wondering if there should be a limit imposed on the number of connections to the redis DB. Perhaps using `eachLimit` with a concurrency of something like 10 would be safer.

philipyoo: good call, will push up this change soon

lib/metrics/RedisClient.js
philipyoo: For this, I ran into a similar issue in the past, actually it was for:
https://github.com/scality/Arsenal/pull/434/files#diff-9c8ee1136532e66fc124018674bec5e0R79
I think this might've been the simplest solution without breaking anything that relied on it.

That said, this use here is slightly different, but still has a `cb` as a mandatory argument. I could set a default value on count, and call this function passing a null argument maybe?

PR #435, opened 2018-03-06
Ft/objdel add modified header check

dora-korpar:
Extracts check of date modified headers to separate function for use in S3 objectDelete API

### DIFF
--- a/lib/s3middleware/validateConditionalHeaders.js
+++ b/lib/s3middleware/validateConditionalHeaders.js
@@ -69,6 +69,31 @@ function _checkUnmodifiedSince(ifUnmodifiedSinceTime, lastModified) {
 }
 
 /**
+ * checks 'if-modified-since' and 'if-unmodified-since' headers if included in
+ * request against last-modified date of object
+ * @param {object} headers - headers from request object
+ * @param {string} lastModified - last modified date of object
+ * @return {object} contains modifiedSince and unmodifiedSince res objects
+ */
+function checkDateModifiedHeaders(headers, lastModified) {
+    let lastModifiedDate = new Date(lastModified);
+    lastModifiedDate.setMilliseconds(0);
+    lastModifiedDate = lastModifiedDate.getTime();
+
+    const ifModifiedSinceHeader = headers['if-modified-since'] ||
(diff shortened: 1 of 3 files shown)


lib/s3middleware/validateConditionalHeaders.js
```diff
@@ -68,6 +68,31 @@ function _checkUnmodifiedSince(ifUnmodifiedSinceTime, lastModified) {
...
+ * checks 'if-modified-since' and 'if-unmodified-since' headers if included in
+ * request against last-modified date of object
+ * @param {object} headers - headers from request object
+ * @param {string} lastModified - last modified date of object
+ * @return {object} contains modifiedSince and unmodifiedSince res objects
+ */
+function checkDateModifiedHeaders(headers, lastModified) {
+    let lastModifiedDate = new Date(lastModified);
+    lastModifiedDate.setMilliseconds(0);
+    lastModifiedDate = lastModifiedDate.getTime();
```
rahulreddy: You should probably send `lastModifiedDate` as a parameter so that this method can be used as a true utility.

dora-korpar: I'm not sure what you mean by this... I put the `lastModifiedDate` manipulation here so that it wouldn't have to be repeated in the `validateConditionalHeaders` function and the `objectDelete` api

rahulreddy: So instead of checking it against current date it would make it easy to check it against any date. Writing different cases for unit tests mimicking old dates would be easier. The intention of checking utilities in Arsenal is that it may be used by many projects (S3, Backbeat, Connect etc.) and since it's hard to anticipate every need making a utility as flexible as possible would help.

dora-korpar: Discusses, change unneeded

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r- r+

rahulreddy:
@ironman-machine r- r+

bennettbuchanan:
@ironman-machine r-

PR #437, opened 2018-03-07
Ft/S3C-1148 statsclient multiple ids

philipyoo:
Extend support for querying a list of ids and
returning a total sum of results for that list.
(Plan is to query all keys using wildcard, and pass
the list of keys to `getAllStats`)

Also add wrapper for redis method `keys`

### DIFF
--- a/lib/metrics/RedisClient.js
+++ b/lib/metrics/RedisClient.js
@@ -23,6 +23,28 @@ class RedisClient {
     }
 
     /**
+    * scan a pattern and return matching keys
+    * @param {string} pattern - string pattern to match with all existing keys
+    * @param {number} [count=10] - scan count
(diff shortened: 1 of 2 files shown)


philipyoo:
@ironman-machine try

philipyoo:
@ironman-machine try

lib/metrics/RedisClient.js
```diff
@@ -22,6 +22,28 @@ class RedisClient {
         return this;
     }
 
+    /**
+    * scan a pattern and return matching keys
+    * @param {string} pattern - string pattern to match with all existing keys
+    * @param {number} count - scan count, defaults to 10
```
jonathan-gramain: You can write the shorter form `[count=10] - scan count`

PR #439, opened 2018-03-09
Ft/zenko 19

anurag4DSB:
Wrapper for Prometheus Client - https://github.com/siimon/prom-client


lib/monitoring/monitoringClientInterface.js
```diff
@@ -0,0 +1,128 @@
...
+
+const assert = require('assert')
+const client = require('prom-client');
+
+/**
+ * Wrapper for Prometheus client - https://github.com/siimon/prom-client
+ */
+
+var references = {};
+var namespace = '';
```
NicolasT: I don't see why you'd ever want/need to 'namespace' metrics. All filtering can/should be done based on Prometheus' dynamic labels (added by the scraper).

Even to the contrary: you *don't* want to 'namespace' metrics because then you're certainly not following the de-facto 'standard' metric names used by various services and exporter.

lib/monitoring/monitoringClientInterface.js
```diff
@@ -0,0 +1,128 @@
...
+   * @param {string} metricsHelp - description of metrics
+   * @param {string} metricsLabels - Static labels applied to every metric.
+   * Optional Parameter - metricsLabels 
+   */
+  createCounter: function (metricsName, metricsHelp, metricsLabels) {
+    if (metricsLabels === undefined) {
+        metricsLabels = [];
+    }
+    references[metricsName] = new client.Counter({
+        name: metricsName,
```
NicolasT: Even though I reject the thought of using 'namespaces' (see comment above), I guess this lacks the corresponding code to modify the given `metricsName`, as done for other metric types.

lib/monitoring/monitoringClientInterface.js
```diff
@@ -0,0 +1,128 @@
+'use strict';
+
+const assert = require('assert')
+const client = require('prom-client');
```
NicolasT: Not declared as a dependency.

lib/monitoring/monitoringClientInterface.js
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Wrapper for Prometheus client - https://github.com/siimon/prom-client
+ */
+
+var references = {};
```
NicolasT: No need to keep another layer of references, `prom-client` has `getSingleMetric` on a `Registry`.

lib/monitoring/monitoringClientInterface.js
```diff
@@ -0,0 +1,128 @@
...
+   * this can however be overriden.
+   * @param {string} metricsName - name of the metrics
+   * @param {string} metricsHelp - description of metrics
+   * @param {string} metricsBuckets - size of buckets for cumulative counters
+   * @param {string} metricsLabels - Static labels applied to every metric
+   * Optional Parameter - metricsBuckets and metricsLabels
+   */
+  createHistogram: function (metricsName, metricsHelp, metricsBuckets, metricsLabels) {
+    if (metricsBuckets === undefined) {
+        metricsBuckets = [0.1, 5, 15, 50, 100, 500]
```
NicolasT: These don't match the defaults used by `prom-client`. Why these values instead?

NicolasT:
I'm not re-posting previous comments which are now no longer visible in the diff view because of whitespace updates. They still apply.

lib/monitoring/monitoringClientInterface.js
anurag4DSB: Changed to the default values: [0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10]

lib/monitoring/monitoringClientInterface.js
anurag4DSB: Removed namespace

lib/monitoring/monitoringClientInterface.js
anurag4DSB: removed namespace.

lib/monitoring/monitoringClientInterface.js
anurag4DSB: adding it

NicolasT:
There's still https://github.com/scality/Arsenal/pull/439#discussion_r173340571

Also, TBH, I really don't see what this brings but being a slightly different API for a subset of `prom-client`'s API, but I guess that's for @vrancurel to decide.

lib/monitoring/monitoringClientInterface.js
anurag4DSB: Makes sense. I have removed references.

lib/monitoring/monitoringClientInterface.js
```diff
@@ -0,0 +1,137 @@
...
+     * @param {string} metricsLabels - Static labels applied to every metric
+     * @return {Object}
+     * Optional Parameter - metricsBuckets and metricsLabels
+     */
+    createHistogram(metricsName, metricsHelp, metricsBuckets, metricsLabels) {
+        let label;
+        let bucket;
+        if (metricsBuckets === undefined) {
+            bucket =
+            [0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10];
```
NicolasT: Why do defaults from `prom-client` need to be duplicated here?

anurag4DSB: Because we are supplying a parameter, it cannot be empty.

anurag4DSB: If the user does not supply the buckets parameter, then an empty variable will be sent.

rahulreddy:
One upside I see to this is, if we are expecting this to be used in other components as well, Vault/Metadata for example, they can share this class instead of re-implementing from scratch. All the common logic can be encapsulated here.

vrancurel:
The point of @NicolasT is that to do a proper wrapper we would need to also return a wrapped object (and not a prometheus object) and things can be complex very easily.

PR #440, opened 2018-03-10
bf: minor lifecycle fixes

dora-korpar:
A few things had to be changed in the lifecycle model in order to be totally AWS compliant:
- added MissingRequiredParameter error
- don't throw error if multiple Prefixes are included, just use the last one
- fixed returned error in various situations
- fix expiration date regex
- add check for rule including both Tags and ExpiredObjectDeleteMarker (not allowed)
- tests galore!

### DIFF
--- a/lib/models/LifecycleConfiguration.js
+++ b/lib/models/LifecycleConfiguration.js
@@ -127,8 +127,8 @@ class LifecycleConfiguration {
         const rulesArray = lifecycleConf.Rule;
         if (!rulesArray || !Array.isArray(rulesArray)
         || rulesArray.length === 0) {
-            rules.error = errors.MalformedXML.customizeDescription(
-                'request xml does not include at least one rule');
+            rules.error = errors.MissingRequiredParameter.customizeDescription(
+                'missing required key \'Rules\' in LifecycleConfiguration');
             return rules;
         }
         if (rulesArray.length > 1000) {
@@ -197,8 +197,8 @@ class LifecycleConfiguration {
             return ruleObj;
         }
         if (!rule.Status) {
-            ruleObj.error = errors.MalformedXML.customizeDescription(
-                'Rule xml does not include Status');
+            ruleObj.error = errors.MissingRequiredParameter.
+                customizeDescription('Rule xml does not include Status');
             return ruleObj;
         }
         const subFilter = rule.Filter ? rule.Filter[0] : rule.Prefix;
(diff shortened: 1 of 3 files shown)


tests/unit/models/LifecycleConfiguration.js
```diff
@@ -173,99 +252,111 @@ function generateXml(errorTag, tagValue) {
...
-describe('LifecycleConfiguration class getLifecycleConfiguration', () => {
-    const tagValue = {};
-    requiredTags.forEach(tag => {
-        it(`should return MalformedXML error if ${tag.tag} tag is missing`,
-        done => {
-            generateParsedXml(tag.tag, tagValue, parsedXml => {
-                checkError(parsedXml, 'MalformedXML', tag.errMessage, done);
-            });
-        });
+describe.only('LifecycleConfiguration class getLifecycleConfiguration', () => {
```
JianqinWang: :thinking: sneaky

dora-korpar: omg i always do that

JianqinWang: same, hopefully the linter gets added soon xD

lib/models/LifecycleConfiguration.js
```diff
@@ -539,39 +530,59 @@ class LifecycleConfiguration {
...
-                '(1[0-2]|0[1-9])-(3[0-1]|0[1-9]|[1-2][0-9])T(2[0-3]|[0-1]' +
-                '[0-9]):([0-5][0-9]):([0-5][0-9])(\.[0-9]+)?(Z|[+-](?:2[0-3]' +
-                '|[0-1][0-9]):[0-5][0-9])?$');
+                '(1[0-2]|0[1-9])-(3[0-1]|0[1-9]|[1-2][0-9])T00:00:00' +
+                '(.[0-9]+)?(Z)?$');
             if (!isoRegex.test(subExp.Date[0])) {
-                expObj.error = errors.MalformedXML.customizeDescription(
-                    'Date is not in proper ISO 8601 format');
+                expObj.error = errors.InvalidArgument.customizeDescription(
+                    'Date must be at midnight GMT in ISO 8601 format');
```
jonathan-gramain: I think we don't have to comply exactly with AWS spec on this one, because we can provide more flexibility, and this sounds more like an AWS technical constraint than a real spec (and our system can handle it without more work), so it should be fine to allow any valid date string (can be constrained to only accept UTC / ISO 8601 though)

dora-korpar: sounds good 😸

bennettbuchanan:
@ironman-machine r+

bennettbuchanan:
@ironman-machine r-

bennettbuchanan:
@ironman-machine try

dora-korpar:
@ironman-machine try

dora-korpar:
@ironman-machine try

bennettbuchanan:
@ironman-machine r+

PR #444, opened 2018-03-15
fix: StatsClient zero byte increment

philipyoo:
For metrics, there might be a chance byte size might be 0


philipyoo:
@ironman-machine try

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r-

PR #446, opened 2018-03-19
ft: Add uid property to BucketInfo

philipyoo:
Currently only setting `uid` when using `putBucketLifecycle`

### DIFF
--- a/lib/models/BucketInfo.js
+++ b/lib/models/BucketInfo.js
@@ -1,10 +1,13 @@
 const assert = require('assert');
+const uuid = require('uuid/v4');
+
 const { WebsiteConfiguration } = require('./WebsiteConfiguration');
 const ReplicationConfiguration = require('./ReplicationConfiguration');
 const LifecycleConfiguration = require('./LifecycleConfiguration');
 
 // WHEN UPDATING THIS NUMBER, UPDATE MODELVERSION.MD CHANGELOG
-const modelVersion = 6;
+// MODELVERSION.MD can be found in S3 repo: lib/metadata/ModelVersion.md
+const modelVersion = 7;
 
 class BucketInfo {
     /**
@@ -47,12 +50,14 @@ class BucketInfo {
     * @param {string[]} [cors[].exposeHeaders] - headers expose to applications
     * @param {object} [replicationConfiguration] - replication configuration
     * @param {object} [lifecycleConfiguration] - lifecycle configuration
+    * @param {string} [uid] - unique identifier for the bucket, necessary
+    * addition for use with lifecycle operations
     */
     constructor(name, owner, ownerDisplayName, creationDate,
                 mdBucketModelVersion, acl, transient, deleted,
                 serverSideEncryption, versioningConfiguration,
                 locationConstraint, websiteConfiguration, cors,
-                replicationConfiguration, lifecycleConfiguration) {
+                replicationConfiguration, lifecycleConfiguration, uid) {
         assert.strictEqual(typeof name, 'string');
         assert.strictEqual(typeof owner, 'string');
         assert.strictEqual(typeof ownerDisplayName, 'string');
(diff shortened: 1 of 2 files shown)


lib/models/BucketInfo.js
```diff
@@ -521,6 +533,13 @@ class BucketInfo {
         return this._versioningConfiguration &&
             this._versioningConfiguration.Status === 'Enabled';
     }
+    /**
+     * Get unique id of bucket.
+     * @return {string} - unique id
+     */
+    getUid() {
```
bennettbuchanan: Seems like the pattern is to also ship a setter (i.e., `setUid`) for each bucketInfo property.

rahulreddy: I think this is one setter where we will carefully implement the setter when we need use it throughout the project.

lib/models/BucketInfo.js
```diff
@@ -112,6 +117,10 @@ class BucketInfo {
         if (lifecycleConfiguration) {
             LifecycleConfiguration.validateConfig(lifecycleConfiguration);
         }
+        if (uid) {
+            assert.strictEqual(typeof uid, 'string');
+            assert.strictEqual(uid.length, 36);
```
philipyoo: not sure if we should validate length of uid, but I was thinking it should be something directly from the `uuid` module

tests/unit/models/BucketInfo.js
```diff
@@ -381,3 +389,25 @@ Object.keys(acl).forEach(
         });
     })
 );
+
+describe('uid default', () => {
+    it('should not set uid if none is specified by constructor params', () => {
```
jonathan-gramain: :+1:

lib/models/BucketInfo.js
jonathan-gramain: Could be, I think it's fine to validate the length as it should be 36 (uuid length) in all cases.

lib/models/BucketInfo.js
```diff
@@ -521,6 +535,20 @@ class BucketInfo {
...
+     * @return {string} - unique id
+     */
+    getUid() {
+        return this._uid;
+    }
+    /**
+     * Set unique id for bucket.
+     * @return {BucketInfo} - bucket info instance
+     */
+    setUid() {
```
jonathan-gramain: This is not technically a setter, so to avoid confusion I would call it e.g. `genUid()`.

jonathan-gramain: or `generateUid` which is more self-explanatory

jonathan-gramain: or `createUid()` which also carries the notion of setting it

lib/models/BucketInfo.js
```diff
@@ -137,6 +146,8 @@ class BucketInfo {
         this._replicationConfiguration = replicationConfiguration || null;
         this._cors = cors || null;
         this._lifecycleConfiguration = lifecycleConfiguration || null;
+        // Using undefined because limited to certain features at the moment
+        this._uid = uid || undefined;
```
rahulreddy: You can generate uid if it doesn't exist. This makes sure that the buckets always get uid old and new.

jonathan-gramain: Beware, you then need to make sure all callers of BucketInfo constructor pass the existing uid, otherwise it will generate a new one each time. That was my concern and I though it was safer to just set it to undefined here and let callers set it on a case-by-case basis. Now if we want to make this change everywhere for good reason, why not.

jonathan-gramain: For the sake of generating a build today, I think it's best to leave it as is for now (tested and works) and do this change for end of March (if needed).

philipyoo: For now, sticking with current implementation as ideas for adding uuid on all buckets should not affect this PR's changes. A follow-up PR will be made soon for testing and adding to a future build

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r-

PR #447, opened 2018-03-20
FIX: Mongo Client - countItems

nicolas2bert:


bennettbuchanan:
@ironman-machine r+

bennettbuchanan:
@ironman-machine r- r+

PR #452, opened 2018-03-28
add bson to dependencies

jeremyds:


rahulreddy:
@jeremyds Why do we need this?

jeremyds:
@rahulreddy it is used in mongo client.

PR #453, opened 2018-03-28
Forward/orbit

rahulreddy:


lib/storage/metadata/mongoclient/backend.js
```diff
@@ -0,0 +1,835 @@
...
+ * writes. They may fail in between. To some extend those situations
+ * may generate orphans but not alter the proper conduct of operations
+ * (what he user wants and what we acknowledge to the user).
+ *
+ * Orphan situations may be recovered by the Lifecycle.
+ *
+ * We use proper atomic operations when needed.
+ */
+const async = require('async');
+const arsenal = require('arsenal');
```
JianqinWang: I'm confused, can repos have themselves as dependencies?

rahulreddy: This was a bad port. I brought an old file somehow.

PR #454, opened 2018-03-29
transient source - nonTransientLocation attribute in ObjectMD

jonathan-gramain:
Will be needed for transient source POC implementation.


jonathan-gramain:
POC only

PR #457, opened 2018-03-30
Ft/initial instance

RachedBenMustapha:
Allows pre-provisioning an instance id like with dmd


RachedBenMustapha:
@ironman-machine try

PR #459, opened 2018-04-02
Fix/stuck replication

RachedBenMustapha:
Fixes mongdb oplog consumer:

* Un-hardcode mongodb database name
* The mongodb transform stream would never actually emit any objects to the extensions
* Remove use of global variable

### DIFF
--- a/lib/storage/metadata/mongoclient/LogConsumer.js
+++ b/lib/storage/metadata/mongoclient/LogConsumer.js
@@ -48,7 +48,6 @@ class ListRecordStream extends stream.Transform {
             if (lastEndID === itemObj.h.toString()) {
                 this.unpublishedListing = true;
             }
-            return callback();
         }
 
         if (!this.hasStarted) {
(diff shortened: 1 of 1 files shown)


lib/storage/metadata/mongoclient/LogConsumer.js
```diff
@@ -150,15 +149,18 @@ class LogConsumer {
      * @return {undefined}
      */
     readRecords(params, cb) {
-        const recordStream = new ListRecordStream(this.logger);
         const limit = params.limit || 10000;
         const startIDandSeq = params.startSeq.toString().split('|');
         const startSeq = parseInt(startIDandSeq[0], 10) || 0;
-        lastEndID = startIDandSeq[1];
+        const lastEndID = startIDandSeq[1];
```
alexanderchan-scality: should this change the `this.lastEndID`? just wondering, in the the old code, the global lastEndID is reassigned.

RachedBenMustapha: They are in 2 separate classes, and the global was removed.

PR #460, opened 2018-04-03
ft; add data managed metrics

alexanderchan-scality:
Adds total data managed information to the countItems method.

structure of data managed object:
```
{
    objects: 10,
    versions: 100,
    buckets: 1,
    bucketList: [...]
    dataManaged: {
        total: { curr: 1024, prev: 128 },
        backend-site-1: { curr: 512, prev: 64 }
        (...)
    }
}
```

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -862,7 +862,29 @@ class MongoClientInterface {
             versions: 0,
             buckets: 0,
             bucketList: [],
+            dataManaged: {
+                total: { curr: 0, prev: 0 },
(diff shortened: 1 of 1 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -862,7 +862,31 @@ class MongoClientInterface {
             versions: 0,
             buckets: 0,
             bucketList: [],
+            /* eslint-disable quote-props*/
+            dataManaged: {
+                total: { curr: 0, prev: 0 },
+                'by-location': {},
```
RachedBenMustapha: If this upsets the linter, let's just call it `byLocation`

PR #463, opened 2018-04-05
Fix/mongo tests

ssalaues:
The mongo backend was found to be failing S3 functional in https://github.com/scality/S3/pull/1188

This fixes the failing versioning tests.

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -374,46 +374,72 @@ class MongoClientInterface {
     }
 
     /*
-     * In this case the aller provides a versionId. This function will
-     * sequentially update the object with given versionId THEN the
-     * master iff the provided versionId matches the one of the master
+     * In this case the caller provides a versionId. This function will
+     * check if the object exists then sequentially update the object
+     * (or create if doesn't exist) with given versionId THEN the master
+     * if the provided versionId matches the one of the master. There
+     * is a potential race condition where if two putObjectVerCase3 are
+     * occurring at the same time and the master version doesn't exist,
+     * then one will upsert and update the master and one will fail with
+     * the KeyAlreadyExists error.
      */
     putObjectVerCase3(c, bucketName, objName, objVal, params, log, cb) {
         // eslint-disable-next-line
         objVal.versionId = params.versionId;
         const vObjName = formatVersionKey(objName, params.versionId);
-        c.bulkWrite([{
-            updateOne: {
-                filter: {
-                    _id: vObjName,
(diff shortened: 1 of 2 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -374,46 +374,65 @@ class MongoClientInterface {
     }
 
     /*
-     * In this case the aller provides a versionId. This function will
-     * sequentially update the object with given versionId THEN the
-     * master iff the provided versionId matches the one of the master
+     * In this case the caller provides a versionId. This function will
+     * check if the object exists then sequentially update the object
+     * (or create if doesn't exist) with given versionId THEN the master
+     * if the provided versionId matches the one of the master
```
vrancurel: Please mention the race condition case

ssalaues:
Add info on potential race condition

ssalaues:
Changed the cb return for 11000 error to an internal error so that in the case of the race condition, retries can happen automatically and therefore prevent the race condition from impacting versionid put objects.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -374,46 +374,72 @@ class MongoClientInterface {
...
+                        // case of the race condition mentioned above, the
+                        // InternalError will allow for automatic retries
+                        log.error('putObjectVerCase3:', errors.KeyAlreadyExist);
+                        return cb(errors.InternalError);
+                    }
+                    return cb(errors.NoSuchVersion);
+                }
+                return cb(null, `{"versionId": "${objVal.versionId}"}`);
+            });
+            return null;
```
dora-korpar: Would be best practice to `return c.bulkWrite` instead of a null return

ssalaues:
Changed the return to c.bulkWrite instead of the return null as recommended by @dora-korpar

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -374,46 +374,71 @@ class MongoClientInterface {
...
+            }, err => {
+                if (err) {
+                    log.error(
+                        'putObjectVerCase3: error putting object version',
+                    { error: err.message });
+                    if (err.code === 11000) {
+                        // We want duplicate key error logged however in
+                        // case of the race condition mentioned above, the
+                        // InternalError will allow for automatic retries
+                        log.error('putObjectVerCase3:', errors.KeyAlreadyExist);
```
rahulreddy: There is a typo `KeyAlreadyExist` instead of KeyAlreadyExist**s**

PR #464, opened 2018-04-07
Fixes flakiness in S3 functional tests with mongo backend

ssalaues:
Commit b3103e1 that fixes versioning introduced some flakiness into the S3 functional tests with mongo backend. This reduces the flakiness on the tests.

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -393,7 +393,7 @@ class MongoClientInterface {
                 return cb(errors.InternalError);
             }
             const objUpsert = !checkObj;
-            return c.bulkWrite([{
+            c.bulkWrite([{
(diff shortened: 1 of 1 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -440,6 +440,7 @@ class MongoClientInterface {
                 }
                 return cb(null, `{"versionId": "${objVal.versionId}"}`);
             });
+            return null;
```
philipyoo: Is the caller not expecting the `cb` to be invoked?

rahulreddy: Yeah I don't understand how this solves the issue.

rahulreddy:
Merging this but let's discuss what the issue is and how this solves it.

PR #465, opened 2018-04-09
fx: correct data managed count

alexanderchan-scality:
the original code had locations only count replicated data.
this fix adds count for data stored directly on the external backends

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -895,20 +895,22 @@ class MongoClientInterface {
         };
 
         const consolidateData = dataManaged => {
-            const locations = dataManaged.locations;
-            res.dataManaged.total.curr += dataManaged.total.curr;
-            res.dataManaged.total.prev += dataManaged.total.prev;
-            Object.keys(locations).forEach(site => {
-                if (!res.dataManaged.byLocation[site]) {
-                    res.dataManaged.byLocation[site] =
-                        Object.assign({}, locations[site]);
-                } else {
-                    res.dataManaged.byLocation[site].curr +=
-                        locations[site].curr;
-                    res.dataManaged.byLocation[site].prev +=
-                        locations[site].prev;
-                }
-            });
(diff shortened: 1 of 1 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1069,28 +1118,55 @@ class MongoClientInterface {
...
-                (key, index, self) => self.indexOf(key) === index);
-            // for each uniqKey get master version id
-            return async.map(uniqKeys, (key, done) => {
-                this.getLatestVersion(c, key, log, (err, mst) => {
+        const retBucketInfo = {
+            name: bucketName,
+            location: bucketInfo.getLocationConstraint(),
+        };
+        if (bucketInfo._versioningConfiguration &&
+            bucketInfo._versioningConfiguration.Status) {
```
RachedBenMustapha: Can `Status` be truthy and not `Enabled`?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -929,27 +931,28 @@ class MongoClientInterface {
...
                 // in $regex
 
-                return async.parallel({
-                    bucketList: done => this.getBucketAttributes(
+                return async.waterfall([
+                    next => this.getBucketAttributes(
                     bucketName, log, (err, bucketInfo) => {
                         if (err) {
-                            return done(errors.InternalError);
+                            log.debug('error occured in countItems');
```
RachedBenMustapha: Should log the error here, and use level error instead of debug, since this is going to be a real error that needs troubleshooting (goes for all returns that swallow errors)

alexanderchan-scality: will change the log levels

lib/storage/metadata/mongoclient/MongoClientInterface.js
alexanderchan-scality: `Status` can either be values `Suspended` or `Enabled`.

PR #473, opened 2018-04-23
bf: do not include current UploadIdMarker in list

philipyoo:
for in_memory only
listMultipartUpload should not list current marker in the
listing. Previously, it would include the marker as the
first item in the list

### DIFF
--- a/lib/storage/metadata/in_memory/bucket_utilities.js
+++ b/lib/storage/metadata/in_memory/bucket_utilities.js
@@ -1,33 +1,16 @@
 function markerFilterMPU(allMarkers, array) {
     const { keyMarker, uploadIdMarker } = allMarkers;
-    for (let i = 0; i < array.length; i++) {
-        // If the keyMarker is the same as the key,
-        // check the uploadIdMarker.  If uploadIdMarker is the same
-        // as or alphabetically after the uploadId of the item,
-        // eliminate the item.
-        if (uploadIdMarker && keyMarker === array[i].key) {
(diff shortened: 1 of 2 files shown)


lib/storage/metadata/in_memory/bucket_utilities.js
```diff
@@ -6,9 +6,9 @@ function markerFilterMPU(allMarkers, array) {
         // as or alphabetically after the uploadId of the item,
         // eliminate the item.
         if (uploadIdMarker && keyMarker === array[i].key) {
```
jonathan-gramain: The logic of the whole function seems quite convoluted for what it does, what about going to a simpler:

```javascript
const firstUnfilteredIndex = array.findIndex(
    item => (uploadIdMarker && item.key === keyMarker ?
             item.uploadId > uploadIdMarker :
             item.key > keyMarker));
    return firstUnfilteredIndex !== -1 ? array.slice(firstUnfilteredIndex) : array;
```

philipyoo: nice!

philipyoo:
@ironman-machine
RUN_BACKBEAT_CRR_TESTS=true
try

philipyoo:
@ironman-machine
RUN_BACKBEAT_CRR_TESTS=true
try

philipyoo:
@ironman-machine
RUN_BACKBEAT_CRR_TESTS=true
try

PR #476, opened 2018-04-30
Bf/mongo list object limit

alexanderchan-scality:
https://github.com/scality/S3/issues/1235

### DIFF
--- a/lib/algos/list/delimiter.js
+++ b/lib/algos/list/delimiter.js
@@ -1,7 +1,9 @@
 'use strict'; // eslint-disable-line strict
 
 const Extension = require('./Extension').default;
-const { inc, FILTER_END, FILTER_ACCEPT, FILTER_SKIP } = require('./tools');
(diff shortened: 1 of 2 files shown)


lib/algos/list/delimiter.js
```diff
@@ -1,7 +1,9 @@
 'use strict'; // eslint-disable-line strict
 
 const Extension = require('./Extension').default;
-const { inc, FILTER_END, FILTER_ACCEPT, FILTER_SKIP } = require('./tools');
+const { checkLimit, inc, FILTER_END, FILTER_ACCEPT, FILTER_SKIP } =
+    require('./tools');
+const DEFAULT_MAX_KEYS = 10000;
```
rahulreddy: `DEFAULT_MAX_KEYS` could be confused with `  this.maxKeys = parameters.maxKeys || 1000;`
Something like `MAX_KEYS_HARDLIMIT` is more appropriate.

PR #477, opened 2018-04-30
ft: interpret healthcheck status from bucketd

rahulreddy:
As discussed in S3C-1412, it is necessary for S3 to interpret Bucketd health
status in order to provide more flexibility (relatively to the health check
mechanism) than failing due to a light partial unavailability of the platform.

This is done in the bucketclient backend's healthcheck method, in order to
comply with all the other backends.

Fixes S3C-1412

Signed-off-by: David Pineau <david.pineau@scality.com>

### DIFF
--- a/lib/storage/metadata/bucketclient/BucketClientInterface.js
+++ b/lib/storage/metadata/bucketclient/BucketClientInterface.js
@@ -111,17 +111,81 @@ class BucketClientInterface {
         return null;
     }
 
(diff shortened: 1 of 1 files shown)


rahulreddy:
@DavidPineauScality  I agree. It would be inline with S3's usage of Metadata as a black box without subscribing to its internals.

rahulreddy:
@ironman-machine r+

PR #479, opened 2018-05-02
FX: constructing v4 query auth signature with proxyPath

nicolas2bert:
If `proxyPath` header is specified, we use it to construct our v4 query auth signature.

### DIFF
--- a/lib/auth/v4/headerAuthCheck.js
+++ b/lib/auth/v4/headerAuthCheck.js
@@ -126,8 +126,17 @@ function check(request, log, data, awsService) {
     if (isTimeSkewed) {
         return { err: errors.RequestTimeTooSkewed };
     }
-    const proxyPath = request.headers.proxy_path ?
-    decodeURIComponent(request.headers.proxy_path) : null;
+
+    let proxyPath = null;
+    if (request.headers.proxy_path) {
+        try {
+            proxyPath = decodeURIComponent(request.headers.proxy_path);
+        } catch (err) {
+            log.debug('invalid proxy_path header', { proxyPath, err });
+            return { err: errors.InvalidArgument.customizeDescription(
+              'invalid proxy_path header') };
+        }
+    }
 
     const stringToSign = constructStringToSign({
         log,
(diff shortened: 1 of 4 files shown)


tests/unit/auth/v4/queryAuthCheck.js
```diff
@@ -225,4 +225,18 @@ describe('v4 queryAuthCheck', () => {
         assert.strictEqual(res.params.version, 4);
         done();
     });
+
+    it('should successfully return no error if proxy_path header is added',
```
dora-korpar: Will there ever be a case where proxy_path header is added but invalid?

rahulreddy:
What is `proxyPath` header? Where is this coming from?

nicolas2bert:
It comes from https://github.com/scality/pensieve-front/pull/145/files#diff-fa6b1f2ed66280c30886fd06d3faa03aR233. We added a proxy, and now when we make a request to S3, it fails. Indeed, S3 tries to construct v4 query auth signature using `request.path` but since our added proxy has modified the path, signature generated by the client (that used the original path) does not match the signature generated by S3 (that used the modified path). So we pass the original path using `proxyPath` header.

lib/auth/v4/headerAuthCheck.js
```diff
@@ -126,8 +126,15 @@ function check(request, log, data, awsService) {
...
     }
-    const proxyPath = request.headers.proxy_path ?
-    decodeURIComponent(request.headers.proxy_path) : null;
+
+    let proxyPath = null;
+    if (request.headers.proxy_path) {
+        try {
+            proxyPath = decodeURIComponent(request.headers.proxy_path);
+        } catch (err) {
+            log.debug('invalid proxy_path header', { proxyPath });
```
nicolas2bert: @rahulreddy I was not sure if I should return an error here or not to keep consistency.

nicolas2bert: Also, if we want to return an error, I was not sure on which one should we return. :)

tests/unit/auth/v4/queryAuthCheck.js
nicolas2bert: Good catch!! Thanks, I addressed your comment.

lib/auth/v4/headerAuthCheck.js
rahulreddy: I recommend logging the `err` as a debug log and returning InvalidArgument as error.

nicolas2bert: Thanks, I addressed your comment :)

PR #483, opened 2018-05-04
bf: remove mongo cursor limit

alexanderchan-scality:
removed the mongo cursor hard coded limit as that introduced undesired
behavior with small max-keys


alexanderchan-scality:
renaming branch with JIRA ticket

PR #488, opened 2018-05-10
bf: ZENKO-336 zero fill interval stats if none present

philipyoo:
Fixes:
- check if `_zip` receives an empty array
- check if there are data intervals is less than total number of intervals. If so, zero fill empty intervals

### DIFF
--- a/lib/metrics/StatsModel.js
+++ b/lib/metrics/StatsModel.js
@@ -17,7 +17,10 @@ class StatsModel extends StatsClient {
     * @return {array} converted array
     */
     _zip(arrays) {
-        return arrays[0].map((_, i) => arrays.map(a => a[i]));
+        if (arrays.length > 0 && arrays.every(a => Array.isArray(a))) {
+            return arrays[0].map((_, i) => arrays.map(a => a[i]));
+        }
+        return [];
     }
 
     /**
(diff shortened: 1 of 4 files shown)


bennettbuchanan (dismissed):

lib/metrics/StatsModel.js
```diff
@@ -88,10 +96,22 @@ class StatsModel extends StatsClient {
...
+            errors = this._zip(errors).map(arr =>
                 arr.reduce((acc, i) => acc + i));
+
+            // replace any existing values in arrays "requests" and "errors"
+            // in statsRes.requests and statsRes['500s']
+            // i.e. statsRes.requests = [0,0,0,0]
+            //      requests = [1, 2]
+            //    The splice should change:
+            //      statsRes.requests = [1,2,0,0]
+            Array.prototype.splice.apply(statsRes.requests,
```
jonathan-gramain: For clarity I think you could call the function normally and pass the remaining arguments as `...requests`.

lib/metrics/StatsModel.js
```diff
@@ -17,7 +17,10 @@ class StatsModel extends StatsClient {
     * @return {array} converted array
     */
     _zip(arrays) {
-        return arrays[0].map((_, i) => arrays.map(a => a[i]));
+        if (arrays.length > 0 && arrays.every(a => Array.isArray(a))) {
+            return arrays[0].map((_, i) => arrays.map(a => a[i]));
+        }
+        return arrays;
```
jonathan-gramain: Can you comment why you return `arrays` here if items are not arrays? I think this is to support the zero-fill case though not crystal clear for me.

philipyoo: I actually think I should return an empty array (or validate after `_zip` returns). The `StatsClient.getStats` on error will return an integer. The original comment says that Redis for stats isn't a critical component, so the default response is a 0.
Or if no redis keys, `_getCount` will return an array of 0's.

philipyoo (commented):
Since the override of `_getCount` is used by `StatsClient.getStats`,  instead of making validations within `StatsModel.getAllStats`, it would be best to validate proper length and data from the overridden `_getCount`

I'll be pushing a commit to reflect this

philipyoo:
re-opened for bert-e and retargeting 8.0

PR #496, opened 2018-05-31
MD-654: fix algo delimiter

ghost:
This PR is a "Hardening" PR related to MD-654.

As I'm not very knowledgeable about S3 cloud-server, I'm counting on reviews from CloudServer-knowledgeable people to tell me whether this fix is at risk to create issues there or not.

Please mind the fact that this may slightly change the behavior of the listing extensions, in a way that does not affect MetaData, but may affect other usages.

Note also that I dind't check other extensions yet for similar behavior, but those may require a similar fix.
Note also that this assumes that we're using backends that have an alphabetical order (and is not checking that flag), as I'm not aware of us using a backend which does non-ordered key storage for our listing. Please comment if it's wrong.

### DIFF
--- a/lib/algos/list/delimiter.js
+++ b/lib/algos/list/delimiter.js
@@ -131,7 +131,6 @@ class Delimiter extends Extension {
         this.Contents.push({ key, value });
         this.NextMarker = key;
         ++this.keys;
-        console.log('Adding entry..');
         return FILTER_ACCEPT;
     }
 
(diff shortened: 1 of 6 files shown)


ghost:
Side note: Testing these changes applied onto z/1.0 with S3 z/1.0 basis: http://ci.ironmann.io/gh/scality/S3/12133

lib/algos/list/delimiter.js
```diff
@@ -148,10 +148,18 @@ class Delimiter extends Extension {
     filter(obj) {
         const key = obj.key;
         const value = obj.value;
-        if ((this.prefix && !key.startsWith(this.prefix))
-            || (this.alphabeticalOrder
-                && typeof this.NextMarker === 'string'
-                && key <= this.NextMarker)) {
+        if (this.prefix && !key.startsWith(this.prefix)) {
```
philipyoo, line 151: you mention we assume we work with backends that are alphabetically ordered. If possible and if it makes sense here, can we add  `this.aphabeticalOrder`. Would be nice here to make it more apparent

philipyoo (commented):
just commenting as I'm not fully sure, but from what I understand of listing in S3, lgtm

lib/algos/list/delimiter.js
ghost: Sure, the point was that I have no idea if this alphabeticalOrder flag is actually necessary or not, given that I have no idea why we'd not be using it.
I'll add the check.

ghost:
Things have evolved too much in there to keep this.

PR #497, opened 2018-05-31
ft: ZENKO-433 add item count support incremental update and refresh

alexanderchan-scality:


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -480,18 +490,35 @@ class MongoClientInterface {
...
+        const mongoCall = {
+            args: [c, bucketName, objName, objVal, params, log],
+            cb,
+        };
+        let isUpdate = false;
         if (params && params.versioning && !params.versionId) {
-            return this.putObjectVerCase1(c, bucketName, objName, objVal,
-                                          params, log, cb);
+            countParams.isVer = true;
+            mongoCall.op = this.putObjectVerCase1.bind(this);
```
rachedbenmustapha: We should avoid binding functions in the per-request code path, as it puts extra pressure on the GC that can be detrimental under high load.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -480,18 +490,35 @@ class MongoClientInterface {
...
+            countParams.isVer = false;
+            mongoCall.op = this.putObjectNoVer.bind(this);
+        }
+        if (isUpdate) {
+            return dataCounterHelper.updateObjWrapper(
+            this.dataCount, c, countParams, log, mongoCall);
         }
-        return this.putObjectNoVer(c, bucketName, objName, objVal,
-                                   params, log, cb);
+        return dataCounterHelper.addObjWrapper(
```
rachedbenmustapha: This inverts control, even if it's a bit more repetitive the following pattern:

- do operation
- if no error, then account for it

would be more readable and would avoid conditioning the mongo call to the wrapper not failing. Also it incurs a lot of arguments serialization that can be a performance penalty.

rachedbenmustapha (changes requested):

alexanderchan-scality:
closing in favor of #499

PR #499, opened 2018-06-07
ft: ZENKO-433 add item count support incremental update and refresh

alexanderchan-scality:
Old version: #497

### DIFF
--- a/lib/storage/metadata/mongoclient/DataCounter.js
+++ b/lib/storage/metadata/mongoclient/DataCounter.js
@@ -1,0 +1,254 @@
+const NEW_OBJ = 0;
+const NEW_VER = 1;
+const UPDATE_VER = 2;
+const UPDATE_MST = 3;
+const RESTORE = 4;
+
+const DEL_VER = 0;
+const DEL_MST = 1;
+
+const CURR = 'curr';
+const PREV = 'prev';
+
+function deepCopyObject(obj) {
+    return JSON.parse(JSON.stringify(obj));
+}
+
+class DataCounter {
+    /**
+     * DataCounter - class for keeping track of the ItemCount metrics
+     * @return {DataCounter} DataCounter object
+     */
+    constructor() {
+        this.objects = 0;
+        this.versions = 0;
+        this.buckets = 0;
+        this.bucketList = [];
+        this.dataManaged = {
+            total: { curr: 0, prev: 0 },
+            byLocation: {},
+        };
+        this.populated = false;
+    }
+
+    /**
+     * set - set DataCounter values
+     * @param {Object} setVal - object containing values to be used for setting
+     * DataCounter
+     * @param {number} setVal.objects - number of master objects
+     * @param {number} setVal.versions - number of versioned objects
+     * @param {number} setVal.buckets - number of buckets
+     * @param {string[]} setVal.bucketList - list of bucket names
+     * @param {Object} setVal.dataManaged - object containing information about
+     * all the data managed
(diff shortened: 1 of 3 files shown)


philipyoo (dismissed):

alexanderchan-scality:
@bert-e  help

alexanderchan-scality:
@bert-e wait

rahulreddy:
@bert-e wait

dora-korpar (dismissed):

lib/storage/metadata/mongoclient/DataCounter.js
```diff
@@ -0,0 +1,254 @@
...
+const RESTORE = 4;
+
+const DEL_VER = 0;
+const DEL_MST = 1;
+
+const CURR = 'curr';
+const PREV = 'prev';
+
+function deepCopyObject(obj) {
+    return JSON.parse(JSON.stringify(obj));
```
ploki, line 14: that looks costly, is it the best practice when it comes to deep copy of objects attributes without all the methods?

lib/storage/metadata/mongoclient/DataCounter.js
```diff
@@ -0,0 +1,254 @@
...
+     * @param {Object} setVal.byLocaton - object containing the information
+     * about data managed on each location
+     * @return {undefined}
+     */
+    set(setVal) {
+        if (setVal) {
+            this.objects = setVal.objects;
+            this.versions = setVal.versions;
+            this.buckets = setVal.buckets;
+            this.bucketList = [...setVal.bucketList];
```
ploki, line 59: what is the benefit of the spread syntax instead of a direct affectation of this.bucketList? deep copy?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -477,21 +494,110 @@ class MongoClientInterface {
         });
     }
 
+    _getPrevMD(c, params, isUpdate, log, cb) {
+        const collectionName = c.s.name;
+        const { objName, versionId } = params;
+        if (isUserBucket(collectionName)) {
+            if (isUpdate) {
+                return c.find({
```
ploki, line 502: What is the complexity of this find operation? it looks like it perform a database request.

lib/storage/metadata/mongoclient/DataCounter.js
rachedbenmustapha: Suprising as it is, the native JSON parser is pretty efficient, more efficient than recursively iterating over object properties and array elements (which creates temp JS-side objects in the process).

lib/storage/metadata/mongoclient/DataCounter.js
rachedbenmustapha: It's a shallow copy and is equivalent to loop-pushing into an empty new array. Since this is setting a new counter snapshot it allows being "consistent" and not have a dangling reference to an array that can still be pushed into by who knows what callback is in the queue.

lib/storage/metadata/mongoclient/MongoClientInterface.js
rachedbenmustapha: It helps maintain counters accurate (because the update can have a different size than the version it's overwriting) and does a get on a version id. It's either this with full-scan every hour or full-scan way more often. So PUT/DELETE operations on versioned buckets go from 1 read + 1 write to 2 read + 1 write.

ploki (approved):

philipyoo (approved):

PR #503, opened 2018-06-11
bf: ZENKO-336 zero fill interval stats if none present

philipyoo:
Re-opened for bert-e. Original PR https://github.com/scality/Arsenal/pull/488

Fixes:
- check if `_zip` receives an empty array
- check if there are data intervals is less than total number of intervals. If so, zero fill empty intervals

### DIFF
--- a/lib/metrics/StatsModel.js
+++ b/lib/metrics/StatsModel.js
@@ -17,7 +17,10 @@ class StatsModel extends StatsClient {
     * @return {array} converted array
     */
     _zip(arrays) {
-        return arrays[0].map((_, i) => arrays.map(a => a[i]));
+        if (arrays.length > 0 && arrays.every(a => Array.isArray(a))) {
+            return arrays[0].map((_, i) => arrays.map(a => a[i]));
+        }
+        return [];
     }
 
     /**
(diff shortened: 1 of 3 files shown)


lib/metrics/StatsModel.js
```diff
@@ -88,14 +102,11 @@ class StatsModel extends StatsClient {
...
-            if (requests.length) {
-                statsRes.requests = this._zip(requests).map(arr =>
-                    arr.reduce((acc, i) => acc + i, 0));
-            }
-            if (errors.length) {
-                statsRes['500s'] = this._zip(errors).map(arr =>
-                    arr.reduce((acc, i) => acc + i, 0));
-            }
+            statsRes.requests = this._zip(requests).map(arr =>
+                arr.reduce((acc, i) => acc + i));
```
jonathan-gramain: It does not hurt to set the initial value explicitly, you can keep the last `reduce()` argument.

rahulreddy:
@bert-e wait

tmacro (dismissed):

jonathan-gramain (dismissed):

rahulreddy:
@philipyoo Careful with drop commits. BertE will merge automatically if it has two approvals.

jonathan-gramain:
I know, I approved because you asked him to wait ;) but yes we should take care of this.

What about a rule to prevent bert-e to merge based on regexp to detect drop/squash commits?

jonathan-gramain:
Sorry I realized your comment was for Phil, but I felt involved as well :)

rahulreddy:
It all depends on what @tcarmet can support.

tmacro (approved):

rahulreddy (approved):

rahulreddy:
@bert-e help

rahulreddy:
@bert-e status

rahulreddy:
@bert-e force_reset

tcarmet:
Hi @rahulreddy I've created a ticket related to your issue, it's going to be adressed on our next sprint (normally). here's the link https://scality.atlassian.net/browse/BERTE-450

I think I described the need and situation pretty well, but feel free to add a comment on something you would like to have!

lib/metrics/StatsModel.js
rachedbenmustapha: thanks @jonathan-gramain 
@philipyoo not only does it not hurt, but it was required to fix ZENKO-346 (cloudserver crashing on empty stats, on recent node versions iirc) - so you would have effectively reverted the fix for the crash. It's important to check (with git blame for example) why things are done in a certain way before changing.

PR #508, opened 2018-06-19
ft: ZENKO-561: bucket attributes handling fixed

ploki:
on putBucketAttributes and getBucketAttributes

### DIFF
--- a/lib/storage/metadata/proxy/BucketdRoutes.js
+++ b/lib/storage/metadata/proxy/BucketdRoutes.js
@@ -32,15 +32,24 @@ class BucketdRoutes {
 
     _getBucketAttributes(req, res, bucketName, logger) {
         return this._metadataWrapper.client.getBucketAttributes(
-            bucketName, logger, (err, data) =>
-                                sendResponse(req, res, logger, err,
-                                    BucketInfo.fromObj(data).serialize())
-                            );
+            bucketName, logger, (err, data) => {
+                if (err) {
+                    logger.error('Failed to get bucket attributes',
+                                 { bucket: bucketName, error: err });
+                    return sendResponse(req, res, logger, err);
+                }
+                if (data === undefined) {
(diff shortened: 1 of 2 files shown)


JianqinWang (dismissed):

ploki:
@bert-e wait

alexanderchan-scality (dismissed):

nicolas2bert (dismissed):

lib/storage/metadata/proxy/BucketdRoutes.js
```diff
@@ -32,15 +32,19 @@ class BucketdRoutes {
...
-            bucketName, logger, (err, data) =>
-                                sendResponse(req, res, logger, err,
-                                    BucketInfo.fromObj(data).serialize())
-                            );
+            bucketName, logger, (err, data) => {
+                if (data === undefined) {
+                    return sendResponse(req, res, logger,
+                                        errors.NoSuchBucket);
+                }
+                return sendResponse(req, res, logger, err,
```
rahulreddy: If there was a TCP error making a request, would data be `undefined` or `null` and `err` be set?

rahulreddy (changes requested):

lib/storage/metadata/proxy/BucketdRoutes.js
ploki: err and data come from the MetadataWrapper and are not supposed to carry information about the underlying TCP connection, if any.
But I might want to check for an error before checking the payload validity

rahulreddy (dismissed):

lib/storage/metadata/proxy/BucketdRoutes.js
rahulreddy: Since we follow error first callbacks style, it's inline.

bennettbuchanan (approved):

rahulreddy (approved):

PR #511, opened 2018-06-22
bugfix: list version is incorrect because sometimes

vrancurel:
we replace the master with an incorrect last version
because inserts are sometimes swapped. Add a check
to be sure we always replace the master with a
smaller (thus more recent) version.

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -335,7 +335,11 @@ class MongoClientInterface {
 
     /*
      * In this case we generate a versionId and
-     * sequentially create the object THEN update the master
+     * sequentially create the object THEN update the master.
+     *
+     * It is possible that 2 version creations are inverted
+     * in flight so we also check that we update a master only
+     * if the version in place is greater that the one we set.
      */
     putObjectVerCase1(c, bucketName, objName, objVal, params, log, cb) {
         const versionId = generateVersionId(this.replicationGroupId);
@@ -354,8 +358,12 @@ class MongoClientInterface {
             },
         }, {
             updateOne: {
+                // eslint-disable-next-line
                 filter: {
                     _id: objName,
+                    'value.versionId': {
+                        $gt: objVal.versionId,
+                    },
                 },
                 update: {
                     _id: objName, value: objVal,


jeremyds (dismissed):
tested on my scality.cloud RC-zenko-1.0 env

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -335,7 +335,11 @@ class MongoClientInterface {
 
     /*
      * In this case we generate a versionId and
-     * sequentially create the object THEN update the master
+     * sequentially create the object THEN update the master.
+     *
+     * It is possible that 2 version creations are inverted
+     * in flight so we also check that we never update a
+     * master if the version is greater that the one we replace.
```
jeremyds: nice comment

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -357,8 +361,12 @@ class MongoClientInterface {
                 filter: {
                     _id: objName,
                 },
+                // eslint-disable-next-line
                 update: {
                     _id: objName, value: objVal,
+                    'value.versionId': {
```
ploki: this modifies the object pointed by the objVal reference. is it okay to do so? (seems also to be the reason for the eslint-disable comment

ploki: might have unexpected effect on line 355 too

vrancurel: The comment is just there because the linter does not like the quotes surrounding the dotted expression. Also if I don't quote the linter says 'unexpected token .'

ghost:
Is this change confirmed not to break existing MD tests ?

vrancurel:
Even though this fix solves 90% of issues, we need to solve ZENKO-578 which causes the other 10% before unskipping tests/functional/aws-node-sdk/test/versioning/versioningGeneral1.js

giacope (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -354,8 +358,12 @@ class MongoClientInterface {
             },
         }, {
             updateOne: {
+                // eslint-disable-next-line
```
rahulreddy, line 361: This can be removed by using the following code instead
```js
                    '_id': objName,
                    'value.versionId': {
                        $gt: versionId,
                    },
```

rahulreddy (commented):
A few recommendations
- Commit message: Please keep it short < 50 charts and add the rest to the commit body

ploki (approved):

PR #513, opened 2018-06-26
bf: ZENKO-578 mongoDB error replies

jeremyds:
Do not raise an internalError to upper layer when mongoDB fails to
update the master version with a specific error code. This fix is
related to the mongoDB issue: SERVER-19600.

This commit fixes too the message field name of the mongoDB error, it is
'errmsg' and not 'message'.


jeremyds:
bad branch naming

PR #519, opened 2018-06-28
Bugfix/zenko 603/mongo item count

alexanderchan-scality:


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1139,168 +1134,152 @@ class MongoClientInterface {
...
+        const total = { curr: 0, prev: 0 };
+        const locations = {};
 
-        const addLocation = (site, size) => {
-            dataManaged.total += size;
-            if (!dataManaged.locations[site]) {
-                dataManaged.locations[site] = 0;
+        Object.keys(res.nullData).forEach(loc => {
+            const bytes = res.nullData[loc];
+            const locName = loc === 'mem' || loc === 'file' ?
```
nicolas2bert: **IDEA**: you could have make a function that returns `locName` and use it every time we need `locName` so if one day we decide to change our logic, it would have to be done in one sigle place.

nicolas2bert (dismissed):

nicolas2bert (dismissed):

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -850,6 +850,10 @@ class MongoClientInterface {
                     { error: err.message });
                 return cb(errors.InternalError);
             }
+            if (isUserBucket(c.s.name) && result.value) {
```
rahulreddy: Not sure I get much context around this but before you access a property's property, please make sure the parent one is defined. Otherwise it will throw an exception.

rahulreddy (changes requested):

philipyoo (approved):

jonathan-gramain (approved):

PR #520, opened 2018-06-28
bugfix: fix versioningGeneral2 test failing with Mongo

vrancurel:
When an object has been created without versioning and
the versioning has been enabled, when creating a version
we must consider the case that the object doesn't have
the versionId property.


jeremyds (approved):

philipyoo (approved):

tcarmet:
due to work getting all branches synchronized on S3/zenko projects, manual push/merge were made, I'll restart bert-e to get everything in order.

tcarmet:
should be good now

PR #521, opened 2018-06-28
bf: ZENKO-612 use correct MD value for counter

alexanderchan-scality:
context: 
when versioned object is deleted, the master MD is for the counter calculation instead of the versioned MD.


alexanderchan-scality:
merging with #519

PR #525, opened 2018-06-29
Feature/zenko 597/transient mongo count

alexanderchan-scality:


lib/storage/metadata/mongoclient/DataCounter.js
```diff
@@ -29,6 +29,22 @@ class DataCounter {
...
+        this.transientList = {};
+    }
+
+    /**
+     * updateTransientList - update data counter list of transient locations
+     * @param  {Object} newLocations - list of locations constraint details
+     * @return {undefined}
+     */
+    updateTransientList(newLocations) {
+        if (newLocations) {
```
philipyoo: can `newLocations` be an empty object?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1304,7 +1341,9 @@ class MongoClientInterface {
             isVersioned: !!bucketInfo.getVersioningConfiguration(),
             ownerCanonicalId: bucketInfo.getOwner(),
         };
-
+        const isTransient =
+            this.config.getLocationConstraint(retBucketInfo.location)
```
philipyoo: just to be safe, `this.config` can be undefined (given how it is set in constructor), so I feel we should check and make sure it is a valid object we have access to here

philipyoo (dismissed):

philipyoo (approved):

rahulreddy (approved):

PR #528, opened 2018-07-02
feature: ZENKO-483 Monitor CRR upload

bennettbuchanan:
See feature docs here: https://github.com/scality/backbeat/pull/356

### DIFF
--- a/lib/backbeat/Metrics.js
+++ b/lib/backbeat/Metrics.js
@@ -5,6 +5,7 @@ const RedisClient = require('../../lib/metrics/RedisClient');
 const StatsModel = require('../../lib/metrics/StatsModel');
 const INTERVAL = 300; // 5 minutes
 const EXPIRY = 900; // 15 minutes
+const OBJECT_MONITORING_EXPIRY = 86400; // 24 hours.
 
 class Metrics {
     constructor(config, logger) {
@@ -23,10 +24,13 @@ class Metrics {
      * Query StatsClient for all ops given
      * @param {array} ops - array of redis key names to query
      * @param {string} site - site name or '*' wildcard
+     * @param {string} bucketName - the name of the bucket
+     * @param {string} objectKey - the object key name
+     * @param {string} versionId - the object version ID
      * @param {function} cb - callback(err, res)
      * @return {undefined}
      */
-    _queryStats(ops, site, cb) {
+    _queryStats(ops, site, bucketName, objectKey, versionId, cb) {
         return async.map(ops, (op, done) => {
             if (site === 'all') {
                 const queryString = `*:${op}:*`;
@@ -61,6 +65,12 @@ class Metrics {
                     method: 'Metrics._queryStats',
                 });
             }
+            if (bucketName && objectKey && versionId) {
+                const queryString =
+                    `${site}:${bucketName}:${objectKey}:${versionId}:${op}`;
+                return this._objectStatsClient.getStats(this._logger,
+                    queryString, done);
+            }
             const queryString = `${site}:${op}`;
             return this._statsClient.getStats(this._logger, queryString, done);
         }, cb);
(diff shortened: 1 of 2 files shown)


bennettbuchanan:
@bert-e wait

lib/backbeat/Metrics.js
```diff
@@ -301,6 +312,107 @@ class Metrics {
...
+    /**
+     * Get current throughput for an object in bytes/sec. Throughput is the
+     * number of bytes transferred in a given time.
+     * @param {object} details - route details from lib/api/routes.js
+     * @param {function} cb - callback(error, data)
+     * @return {undefined}
+     */
+    getObjectThroughput(details, cb) {
+        this._getData(details, undefined, (err, res) => {
+            if (err && err.type) {
```
jonathan-gramain, line 324: I'm not convinced this code is really useful (I think it's there to create a new instance of an arsenal error returned by `_getData`) and if so, probably not the right place or the right way to do this duplication. In doubt we can keep this, but we should consider reworking error handling in the module later on.

lib/backbeat/Metrics.js
```diff
@@ -301,6 +312,107 @@ class Metrics {
...
+                this._logger.error('error getting metric: object throughput', {
+                    origin: err.method,
+                    method: 'Metrics.getObjectThroughput',
+                });
+                return cb(err.type.customizeDescription(err.message));
+            }
+            if (err) {
+                this._logger.error('error getting metrics: object throughput', {
+                    method: 'Metrics.getObjectThroughput',
```
jonathan-gramain, line 333: Let's log the error message with `error: err.message`

lib/backbeat/Metrics.js
```diff
@@ -301,6 +312,107 @@ class Metrics {
...
+                });
+                return cb(errors.InternalError);
+            }
+            const now = new Date();
+            const timeSinceStart = (now - this._internalStart) / 1000;
```
jonathan-gramain, line 339: I have seen this touchy computation code copied in multiple functions, as I commented in a previous PR it should be feasible to put it into a helper function or in some other file as a library function, so it can be tested independently and be confident it works ok across the various stats gathering functions. Ok to skip for now and keep for later.

lib/backbeat/Metrics.js
```diff
@@ -301,6 +312,107 @@ class Metrics {
...
+            // time
+            const timeSinceStart = (Date.now() - this._internalStart) / 1000;
+            const timeDisplay = timeSinceStart < OBJECT_MONITORING_EXPIRY ?
+                timeSinceStart : OBJECT_MONITORING_EXPIRY;
+            const numOfIntervals = Math.ceil(timeDisplay / INTERVAL);
+            const [totalBytesToComplete, bytesComplete] = res.map(r => (
+                r.requests.slice(0, numOfIntervals).reduce((acc, i) =>
+                    acc + i, 0)
+            ));
+            let percentage = bytesComplete / totalBytesToComplete;
```
jonathan-gramain: May be better to check if `totalBytesToComplete === 0` than relying on `isNaN`, though not sure this case can actually happen there.

Also could be clearer with a separate var `const ratio = ...`, then `const percentage = ...` below.

jonathan-gramain (changes requested):
Commented to add the error message in the log, the rest for me can wait a future refactor.

lib/backbeat/Metrics.js
bennettbuchanan: Okay, I see your concern though since this is the common error handling pattern for all the methods Metrics.js, I'll leave as-is for now.

jonathan-gramain (dismissed):

lib/backbeat/Metrics.js
```diff
@@ -23,10 +24,13 @@ class Metrics {
      * Query StatsClient for all ops given
      * @param {array} ops - array of redis key names to query
      * @param {string} site - site name or '*' wildcard
+     * @param {string} bucketName - the name of the bucket
+     * @param {string} objectKey - the object key
```
nicolas2bert: `the object key name`?

tmacro (dismissed):

nicolas2bert (dismissed):

lib/backbeat/Metrics.js
bennettbuchanan: Updated! 😸

nicolas2bert (approved):

jonathan-gramain (approved):

PR #531, opened 2018-07-10
bugfix: ZENKO-621 Make _buildKey public method

bennettbuchanan:
Timestamp key calculation is needed for https://github.com/scality/backbeat/pull/368 so make `_buildKey` a public method.

### DIFF
--- a/lib/metrics/StatsClient.js
+++ b/lib/metrics/StatsClient.js
@@ -41,11 +41,11 @@ class StatsClient {
     /**
     * build redis key to get total number of occurrences on the server
     * @param {string} name - key name identifier
-    * @param {object} d - Date instance
+    * @param {Date} date - Date instance
     * @return {string} key - key for redis
     */
-    _buildKey(name, d) {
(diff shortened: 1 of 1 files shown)


bennettbuchanan:
@bert-e wait

lib/metrics/StatsClient.js
```diff
@@ -44,7 +44,7 @@ class StatsClient {
     * @param {object} d - Date instance
```
jonathan-gramain: While you're at it, maybe you can rename `d` to `date`? And maybe specify `{Date}` instead as the type in the JSDoc.

rahulreddy (dismissed):

lib/metrics/StatsClient.js
bennettbuchanan: Updated.

rahulreddy (approved):

alexanderchan-scality (approved):

tmacro (approved):

PR #536, opened 2018-07-20
improvement: ZENKO-760 add connection timeout for mongoclient

rahulreddy:
Mongoclient checks each node in the replica set to see which one's the primary,
this check has a default timeout of 30s which delays the startup of Cloudserver
when one of the nodes is unavailable. Cutting down the timeout makes it go through the
list of nodes in the replica set quicker to find the primary. MONGO_CONNECT_TIMEOUT_MS env
var is introduced to adjust the timeout in deployments.


vrancurel (approved):

tmacro (approved):

bennettbuchanan:
Just wanted to confirm the default timeout as 5s. In the PR comment it sounds like the intention was 30s but in the code it is 5s (if I'm understanding that correctly as what is meant by default). Maybe one of those values was meant to be updated?

rahulreddy:
The default set by mongodb npm module is 30s. So we introduced our custom timeout which is 5s.

PR #541, opened 2018-07-25
bf: ZENKO-763 fix object tag replication

jonathan-gramain:
Add extra logic to consume mongo oplog correctly (commit message contains more details)

### DIFF
--- a/lib/storage/metadata/mongoclient/LogConsumer.js
+++ b/lib/storage/metadata/mongoclient/LogConsumer.js
@@ -1,94 +1,8 @@
 'use strict'; // eslint-disable-line
 
-const stream = require('stream');
 const MongoClient = require('mongodb').MongoClient;
+const ListRecordStream = require('./ListRecordStream');
 const { Timestamp } = require('bson');
-
-const ops = {
-    i: 'put',
-    u: 'put',
-    d: 'delete',
-};
-
-class ListRecordStream extends stream.Transform {
-    constructor(logger, lastEndID) {
-        super({ objectMode: true });
-        this.logger = logger;
-        this.lastEndID = lastEndID;
-        this.hasStarted = false;
-        this.start = null;
-        this.end = null;
-        this.lastUniqID = null;
-        // this.unpublishedListing is true once we pass the oplog that has the
-        // start seq timestamp and uniqID 'h'
-        this.unpublishedListing = false;
-    }
-
-    _transform(itemObj, encoding, callback) {
-        if (!itemObj) {
-            this.push(null);
-            this.emit('info', {
-                start: this.start,
-                end: this.end,
-                uniqID: this.lastUniqID,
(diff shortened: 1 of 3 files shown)


jonathan-gramain:
I will rework a bit as some other log entry cases are unclear, and add some unit tests.

jonathan-gramain:
@bert-e wait

bennettbuchanan (approved):
Thanks for this @jonathan-gramain! :shipit:

lib/storage/metadata/mongoclient/LogConsumer.js
```diff
@@ -63,16 +57,39 @@ class ListRecordStream extends stream.Transform {
         }
 
         const dbName = itemObj.ns.split('.');
+        let entry;
+        if (itemObj.op === 'i') {
+            entry = {
+                type: 'put',
+                key: itemObj.o._id,
```
rahulreddy: We should check for the existence of `o` and `o2` properties before accessing their children. I know that the format is probably guaranteed by Mongodb but it's best if we check for the properties cleanly.

JianqinWang (approved):
thank you  @jonathan-gramain !

rahulreddy (changes requested):

lib/storage/metadata/mongoclient/LogConsumer.js
jonathan-gramain: done

lib/storage/metadata/mongoclient/LogConsumer.js
```diff
@@ -151,15 +66,28 @@ class LogConsumer {
...
+                // previous 'info' event
+                startSeq = JSON.parse(params.startSeq);
+            } catch (err) {
+                this.logger.error('malformed startSeq', {
+                    startSeq: params.startSeq,
+                });
+                // start over
+                startSeq = { ts: 0 };
+            }
+        } else {
```
bennettbuchanan: Minor: we can avoid the else block here if we initially set `startSeq` to `{ ts: 0 }`.

bennettbuchanan (approved):

lib/storage/metadata/mongoclient/LogConsumer.js
jonathan-gramain: done

JianqinWang (approved):

jonathan-gramain:
E2E tests have some unrelated flakiness issues during install phase, but all replication tests pass (including the re-enabled CRR tagging tests, see https://github.com/scality/Zenko/pull/227). In this respect it should be good to merge, and updated E2E tests will run again in zenko with updated images.

One build where CRR tests pass: https://eve.devsca.com/github/scality/zenko/#/builders/10/builds/3532

rahulreddy (approved):

PR #543, opened 2018-07-29
feat: add checkHealth to mongodb interface

giacope:
This adds a simple health check function to the `MongoClientInterface`.
The main purpose for this is https://github.com/scality/S3/pull/1442 (reviews required), which checks if mongodb is connected to cloudserver when the route `/_/healthcheck` is requested.

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -1046,6 +1046,14 @@ class MongoClientInterface {
                                        log, cb);
     }
 
+    checkHealth(implName, log, cb) {
+        if (this.client.isConnected()) {
+            return cb();
+        }
+        log.error('disconnected from mongodb');
+        return cb(errors.InternalError, []);
+    }
+
     readUUID(log, cb) {
         const i = this.getCollection(INFOSTORE);
         i.findOne({


giacope:
@bert-e help

giacope:
@bert-e help

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1046,6 +1046,20 @@ class MongoClientInterface {
...
     }
 
+    checkHealth(implName, log, cb) {
+        return this.client.connect(this.mongoUrl, err => {
+            const respBody = {
+                code: 200,
+                error: null,
+            };
+            if (err) {
+                respBody[implName].code = err.code || 400;
```
rachedbenmustapha: 400 means that the error is the client's fault. 503 would be appropriate here

rachedbenmustapha (changes requested):

rachedbenmustapha (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1046,6 +1046,23 @@ class MongoClientInterface {
                                        log, cb);
     }
 
+    checkHealth(implName, log, cb) {
+        return MongoClient.connect(this.mongoUrl, err => {
```
rahulreddy: Since we would already have made a `connect` attempt using `setup` method before we call `checkHealth`, I think calling `MongoClient.isConnected()` would be sufficient. Otherwise, the way I see it is that it's establishing a new connection every time this method is called.

rahulreddy (changes requested):
Thanks for the effort, this was tech debt that needed addressing at some point in time.

giacope:
@rahulreddy Updated PR.

JianqinWang (approved):

PR #548, opened 2018-08-06
bf: ZENKO-922 add redis disconnect method

alexanderchan-scality:
Adds disconnect method to allow closing of the backbeat metrics redis
client

### DIFF
--- a/lib/backbeat/Metrics.js
+++ b/lib/backbeat/Metrics.js
@@ -461,6 +461,24 @@ class Metrics {
             });
         });
     }
+
+    /**
+     * Close redis client
+     * @param {function} cb - callback(error, data)
+     * @return {undefined}
+     */
+    disconnect(cb) {
+        return this._redisClient.disconnect(cb);
+    }
+
+    /**
+     * Retrieve the list of redis client connectiosn
+     * @param {function} cb - callback(error, data)
+     * @return {undefined}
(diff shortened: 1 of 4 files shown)


lib/backbeat/Metrics.js
```diff
@@ -461,6 +461,24 @@ class Metrics {
...
         });
     }
+
+    /**
+     * Close redis client
+     * @param {function} cb - callback(error, data)
+     * @return {undefined}
+     */
+    disconnect(cb) {
+        return this._redisClient.disconnect(cb);
```
philipyoo: I'm thinking we could just have the wrapper functions on RedisClient and remove these methods on Metrics class. Any place we use an instance of this Metrics class, we will also have a RedisClient instance available. So I'm thinking just calling directly to the RedisClient would make more sense

philipyoo (approved):

JianqinWang (approved):

tmacro (approved):

jonathan-gramain (commented):
changes look good to me, maybe you should either squash the version bump commit or tag it with the jira number?

rahulreddy (approved):

rahulreddy:
I missed @jonathan-gramain 's comment. We should squash the version bump or tag it with Jira next time.

PR #549, opened 2018-08-07
ft: increase crr metrics expiry window to 24hrs

philipyoo:
Changes in this PR:
- Increase metrics expiry, but keep throughput to 15 minute
  averages.
- Add helper method `_getMaxUptime` to find # of intervals
- Update tests to reflect the extra intervals fetched from
  Redis/StatsModel
- Remove `OBJECT_MONITORING_EXPIRY` and use `EXPIRY` instead
  as values are now same
- Use single instance of `StatsModel`
- Remove extra interval in `StatsModel` expiry field. Not
  needed anymore as throughput uses a 15 minute window and
  the extra interval for it will be available by default
- Use explicit variable names when data is fetched from
  `StatsClient`

### DIFF
--- a/lib/backbeat/Metrics.js
+++ b/lib/backbeat/Metrics.js
@@ -5,7 +5,6 @@ const RedisClient = require('../../lib/metrics/RedisClient');
 const StatsModel = require('../../lib/metrics/StatsModel');
 const INTERVAL = 300; // 5 minutes
 const EXPIRY = 86400; // 24 hours
-const OBJECT_MONITORING_EXPIRY = 86400; // 24 hours.
 const THROUGHPUT_EXPIRY = 900; // 15 minutes
 
 class Metrics {
@@ -15,8 +14,7 @@ class Metrics {
         this._redisClient = new RedisClient(redisConfig, this._logger);
         // Redis expiry increased by an additional interval so we can reference
         // the immediate older data for average throughput calculation
-        this._statsClient = new StatsModel(this._redisClient, INTERVAL,
-            (EXPIRY + INTERVAL));
+        this._statsClient = new StatsModel(this._redisClient, INTERVAL, EXPIRY);
         this._validSites = validSites;
         this._internalStart = internalStart;
     }
@@ -66,13 +64,13 @@ class Metrics {
                     method: 'Metrics._queryStats',
                 });
             }
+            let queryString;
             if (bucketName && objectKey && versionId) {
-                const queryString =
+                queryString =
                     `${site}:${bucketName}:${objectKey}:${versionId}:${op}`;
-                return this._objectStatsClient.getStats(this._logger,
-                    queryString, done);
+            } else {
+                queryString = `${site}:${op}`;
             }
-            const queryString = `${site}:${op}`;
             return this._statsClient.getStats(this._logger, queryString, done);
         }, cb);
     }
@@ -136,14 +134,14 @@ class Metrics {
             }
             const uptime = this._getMaxUptime(EXPIRY);
             const numOfIntervals = Math.ceil(uptime / INTERVAL);
-            const d = res.map(r => (
+            const [ops, opsDone, bytes, bytesDone] = res.map(r => (
                 r.requests.slice(0, numOfIntervals).reduce((acc, i) =>
                     acc + i, 0)
             ));
 
-            let opsBacklog = d[0] - d[1];
+            let opsBacklog = ops - opsDone;
             if (opsBacklog < 0) opsBacklog = 0;
-            let bytesBacklog = d[2] - d[3];
+            let bytesBacklog = bytes - bytesDone;
             if (bytesBacklog < 0) bytesBacklog = 0;
             const response = {
                 backlog: {
@@ -185,7 +183,7 @@ class Metrics {
             }
             const uptime = this._getMaxUptime(EXPIRY);
             const numOfIntervals = Math.ceil(uptime / INTERVAL);
-            const d = res.map(r => (
+            const [opsDone, bytesDone] = res.map(r => (
                 r.requests.slice(0, numOfIntervals).reduce((acc, i) =>
                     acc + i, 0)
             ));
@@ -196,8 +194,8 @@ class Metrics {
                         '(count) and number of bytes transferred (size) in ' +
                         `the last ${Math.floor(uptime)} seconds`,
                     results: {
-                        count: d[0],
-                        size: d[1],
+                        count: opsDone,
+                        size: bytesDone,
                     },
                 },
             };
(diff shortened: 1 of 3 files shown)


philipyoo:
@bert-e wait

lib/backbeat/Metrics.js
```diff
@@ -4,7 +4,8 @@ const errors = require('../../lib/errors');
 const RedisClient = require('../../lib/metrics/RedisClient');
 const StatsModel = require('../../lib/metrics/StatsModel');
 const INTERVAL = 300; // 5 minutes
-const EXPIRY = 900; // 15 minutes
+const EXPIRY = 86400; // 24 hours
+const THROUGHPUT_EXPIRY = 900; // 15 minutes
 const OBJECT_MONITORING_EXPIRY = 86400; // 24 hours.
```
bennettbuchanan: I think we can go ahead and get rid of `OBJECT_MONITORING_EXPIRY` since it was added only because it was different from the original expiry time.

lib/backbeat/Metrics.js
```diff
@@ -14,8 +15,7 @@ class Metrics {
         this._redisClient = new RedisClient(redisConfig, this._logger);
         // Redis expiry increased by an additional interval so we can reference
         // the immediate older data for average throughput calculation
-        this._statsClient = new StatsModel(this._redisClient, INTERVAL,
-            (EXPIRY + INTERVAL));
+        this._statsClient = new StatsModel(this._redisClient, INTERVAL, EXPIRY);
```
bennettbuchanan, line 17: How come we don't want to pass in `EXPIRY + INTERVAL` as the third argument to the `StatsModel` constructor?

lib/backbeat/Metrics.js
```diff
@@ -96,6 +96,16 @@ class Metrics {
         return cb(null, data);
     }
 
+    /**
+     * Uptime of server based on this._internalStart up to max of expiry
+     * @param {number} expiry - max expiry
+     * @return {number} uptime of server up to expiry time
+     */
+    _getMaxUptime(expiry) {
+        const timeSinceStart = (Date.now() - this._internalStart) / 1000;
```
bennettbuchanan: Minor: Renaming `timeSinceStart` to `secondsSinceStart` might help with readability here.

lib/backbeat/Metrics.js
```diff
@@ -336,20 +336,18 @@ class Metrics {
...
                 .reduce((acc, i) => acc + i, 0);
-            // if timeDisplay !== OBJECT_MONITORING_EXPIRY, use internal timer
+            // if timeDisplay !== THROUGHPUT_EXPIRY, use internal timer
             // and do not include the extra 4th interval
-            if (timeDisplay === EXPIRY) {
+            if (uptime === THROUGHPUT_EXPIRY) {
                 // all intervals apply, including 4th interval
                 const lastInterval =
-                    this._statsClient._normalizeTimestamp(new Date(now));
+                    this._statsClient._normalizeTimestamp(now);
```
bennettbuchanan, line 349: 👍 Nice improvement along the way!

tests/functional/backbeat/Metrics.js
```diff
@@ -57,15 +61,16 @@ describe('Metrics class', () => {
                 completions: {
                     description: 'Number of completed replication operations' +
                         ' (count) and number of bytes transferred (size) in ' +
-                        'the last 900 seconds',
```
bennettbuchanan, line 60: Let's be sure to update the docs to reflect this change. For example, [here](https://github.com/scality/backbeat/blob/development/8.0/docs/metrics.md#_backbeatapimetricscrrlocation-namecompletions) and other references to the default window of 15 minutes. That should help prevent a perception that the routes aren't working as intended.

lib/backbeat/Metrics.js
```diff
@@ -181,10 +190,10 @@ class Metrics {
                 completions: {
                     description: 'Number of completed replication operations ' +
                         '(count) and number of bytes transferred (size) in ' +
-                        `the last ${Math.floor(timeDisplay)} seconds`,
+                        `the last ${Math.floor(uptime)} seconds`,
```
bennettbuchanan, line 195: If `uptime` is `1` here, then `Math.floor` will make the value `0` which seems inaccurate.

bennettbuchanan (approved):
Awesome speedy turnaround on this change! Overall, LGTM.

lib/backbeat/Metrics.js
philipyoo: the extra interval was used for throughput to get a more accurate average. Keeping throughput to still use a 15 minute range, so we no longer need this extra interval 👍

lib/backbeat/Metrics.js
philipyoo: I think `Math.floor` here will still leave it at 1, `uptime` will be guaranteed at least 1

rahulreddy (changes requested):
@philipyoo This is just to make our maintenance easier. Can you separate the commits, one for changing the expiry and two for the refactor you are doing here and add context in the commit body.

lib/backbeat/Metrics.js
bennettbuchanan: Wow, apparently I don't know how `Math.floor` works! It does, however, look like `uptime` could be a decimal < 1. It's definitely a minor case though, since probably no requests will come in before one second has elapsed and it only affects the description 😸.

JianqinWang (approved):

lib/backbeat/Metrics.js
bennettbuchanan: Oh, I see, since it will be included anyway in the new expiry time frame. 👍

lib/backbeat/Metrics.js
philipyoo: Ah you're right! I pushed a change to guarantee minimum value of 1 in the helper method `_getMaxUptime`

lib/backbeat/Metrics.js
```diff
@@ -96,6 +97,18 @@ class Metrics {
...
 
+    /**
+     * Uptime of server based on this._internalStart up to max of expiry
+     * @param {number} expiry - max expiry
+     * @return {number} uptime of server up to expiry time
+     */
+    _getMaxUptime(expiry) {
+        let secondsSinceStart = (Date.now() - this._internalStart) / 1000;
+        // allow only a minimum value of 1 for uptime
+        if (secondsSinceStart < 1) secondsSinceStart = 1;
```
jonathan-gramain: I think the style guidelines advocate to always use curly braces and put the conditional statement(s) in this block on separate lines.

jonathan-gramain (approved):

jonathan-gramain:
BTW, is this change relevant for S3 connector? Not sure but since the plan is to have retry feature available in S3 connector it could be worth targeting development/7.4.

philipyoo:
@jonathan-gramain Ya, I think this should also apply for s3 connector, but there is quite a big difference how metrics looks in 7.4. For 7.4, these functions only exists on backbeat, so I'll need to create separate PRs for these changes on backbeat. Will create a jira ticket for tracking this. Also, I have a few other things I need to port for metrics

rahulreddy (approved):

PR #550, opened 2018-08-13
MD-661, ZENKO-945: fix delimiterMaster::filter

jeremyds:
The return values of the delimiterMaster::filter function are used by
its client, metadata back-end like the mongoClient or Metadata, to
implement a skipScan mechanism: after a number of consecutive SKIP
return values, the clients changes their dataset to the next key range.

This algo, by allowing to skip values unwanted in the results
efficiently gives good performance to master version listing.

The previous algo was broken, preventing it client to perform the
skipScan: it returns ACCEPT for values like PHD, delete marker, and
versions, reseting the SKIP counter of clients.

This commit changes the return values of this function to allow
delimiter clients to use the skipScan mechanism.

### DIFF
--- a/lib/algos/list/delimiterMaster.js
+++ b/lib/algos/list/delimiterMaster.js
@@ -22,6 +22,8 @@ class DelimiterMaster extends Delimiter {
      */
     constructor(parameters) {
         super(parameters);
+        // non-PHD master version or a version whose master is a PHD version
+        this.prvKey = undefined;
         this.prvPHDKey = undefined;
     }
 
@@ -39,31 +41,68 @@ class DelimiterMaster extends Delimiter {
     filter(obj) {
         let key = obj.key;
         const value = obj.value;
+
+        /* Skip keys not starting with the prefix or not alphabetically
+         * ordered. */
         if ((this.prefix && !key.startsWith(this.prefix))
                 || (typeof this.NextMarker === 'string' &&
                     key <= this.NextMarker)) {
             return FILTER_SKIP;
         }
+
+        /* Skip version keys (<key><versionIdSeparator><version>) if we already
+         * have a master version. */
         const versionIdIndex = key.indexOf(VID_SEP);
         if (versionIdIndex >= 0) {
-            // generally we do not accept a specific version,
-            // we only do when the master version is a PHD version
             key = key.slice(0, versionIdIndex);
-            if (key !== this.prvPHDKey) {
-                return FILTER_ACCEPT; // trick repd to not increase its streak
+            /* - key === this.prvKey is triggered when a master version has
+             *   been accepted for this key,
+             * - key === this.NextMarker is triggered when a listing page ends
+             *   on a accepted obj and the next page starts with a version of
+             *   this object. In that case prvKey is default set to undefined
+             *   in the constructor) and comparing to NextMarker is the only
+             *   way to know  we should not accept this version. This test is
+             *   not redundant with the one at the beginning of this function,
+             *   we are comparing here the key without the version suffix,
+             * - key startsWith the previous NextMarker happens because we set
+             *   NextMarker to the common prefix instead of the whole key
+             *   value. (TODO: remove this test once ZENKO-1048 is fixed. ).
+             *   */
+            if (key === this.prvKey || key === this.NextMarker ||
+                (this.delimiter && key.startsWith(this.NextMarker))) {
+                /* master version already filtered */
+                return FILTER_SKIP;
             }
         }
         if (Version.isPHD(value)) {
-            // master version is a PHD version: wait for the next version
+            /* master version is a PHD version, we want to wait for the next
+             * one:
+             * - Set the prvKey to undefined to not skip the next version,
+             * - return accept to avoid users to skip the next values in range
+             *   (skip scan mechanism in metadata backend like Metadata or
+             *   MongoClient). */
+            this.prvKey = undefined;
             this.prvPHDKey = key;
-            return FILTER_ACCEPT; // trick repd to not increase its streak
+            return FILTER_ACCEPT;
         }
         if (Version.isDeleteMarker(value)) {
-            // version is a delete marker, ignore
-            return FILTER_ACCEPT; // trick repd to not increase its streak
+            /* This entry is a deleteMarker which has not been filtered by the
+             * version test. Either :
+             * - it is a deleteMarker on the master version, we want to SKIP
+             *   all the following entries with this key (no master version),
+             * - or a deleteMarker following a PHD (setting prvKey to undefined
+             *   when an entry is a PHD avoids the skip on version for the
+             *   next entry). In that case we expect the master version to
+             *   follow. */
+            if (key === this.prvPHDKey) {
+                this.prvKey = undefined;
+                return FILTER_ACCEPT;
+            }
+            this.prvKey = key;
+            return FILTER_SKIP;
         }
-        // non-PHD master version or a version whose master is a PHD version
-        this.prvPHDKey = undefined;
+
+        this.prvKey = key;
         if (this.delimiter) {
             // check if the key has the delimiter
             const baseIndex = this.prefix ? this.prefix.length : 0;
--- a/tests/unit/algos/list/delimiterMaster.js
+++ b/tests/unit/algos/list/delimiterMaster.js
@@ -1,0 +1,400 @@
+'use strict'; // eslint-disable-line strict
+
+const assert = require('assert');
+
+const DelimiterMaster =
+    require('../../../../lib/algos/list/delimiterMaster').DelimiterMaster;
+const {
+    FILTER_ACCEPT,
+    FILTER_SKIP,
+    SKIP_NONE,
+} = require('../../../../lib/algos/list/tools');
+const VSConst =
+    require('../../../../lib/versioning/constants').VersioningConstants;
+const Version = require('../../../../lib/versioning/Version').Version;
+const { generateVersionId } = require('../../../../lib/versioning/VersionID');
+
+
+const VID_SEP = VSConst.VersionId.Separator;
+const EmptyResult = {
+    CommonPrefixes: [],
+    Contents: [],
+    IsTruncated: false,
+    NextMarker: undefined,
+    Delimiter: undefined,
+};
+
+
+describe('Delimiter All masters listing algorithm', () => {
+    it('should return SKIP_NONE for DelimiterMaster when NextMarker is ' +
+       'undefined', () => {
+        const delimiter = new DelimiterMaster({ delimiter: '/' });
+
+        assert.strictEqual(delimiter.NextMarker, undefined);
+
+        /* When there is no NextMarker, it should return SKIP_NONE. */
+        assert.strictEqual(delimiter.skipping(), SKIP_NONE);
+    });
+
+    it('should return <key><VersionIdSeparator> for DelimiterMaster when ' +
+       'NextMarker is set and there is a delimiter', () => {
+        const key = 'key';
+        const delimiter = new DelimiterMaster({ delimiter: '/', marker: key });
+
+        /* Filter a master version to set NextMarker. */
+        // TODO: useless once S3C-1628 is fixed.
+        delimiter.filter({ key, value: '' });
+        assert.strictEqual(delimiter.NextMarker, key);
+
+        /* With a delimiter skipping should return previous key + VID_SEP
+         * (except when a delimiter is set and the NextMarker ends with the
+         * delimiter) . */
+        assert.strictEqual(delimiter.skipping(), key + VID_SEP);
+    });
+
+    it('should return NextMarker for DelimiterMaster when NextMarker is set' +
+       ', there is a delimiter and the key ends with the delimiter', () => {
+        const delimiterChar = '/';
+        const keyWithEndingDelimiter = `key${delimiterChar}`;
+        const delimiter = new DelimiterMaster({
+            delimiter: delimiterChar,
+            marker: keyWithEndingDelimiter,
+        });
+
+        /* When a delimiter is set and the NextMarker ends with the
+         * delimiter it should return the next marker value. */
+        assert.strictEqual(delimiter.NextMarker, keyWithEndingDelimiter);
+        assert.strictEqual(delimiter.skipping(), keyWithEndingDelimiter);
+    });
+
+    it('should skip entries not starting with prefix', () => {
+        const delimiter = new DelimiterMaster({ prefix: 'prefix' });
+
+        assert.strictEqual(delimiter.filter({ key: 'wrong' }), FILTER_SKIP);
+        assert.strictEqual(delimiter.NextMarker, undefined);
+        assert.strictEqual(delimiter.prvKey, undefined);
+        assert.deepStrictEqual(delimiter.result(), EmptyResult);
+    });
+
+    it('should skip entries superior to next marker', () => {
+        const delimiter = new DelimiterMaster({ marker: 'b' });
+
+        assert.strictEqual(delimiter.filter({ key: 'a' }), FILTER_SKIP);
+        assert.strictEqual(delimiter.NextMarker, 'b');
+        assert.strictEqual(delimiter.prvKey, undefined);
+        assert.deepStrictEqual(delimiter.result(), EmptyResult);
+    });
+
+    it('should accept a master version', () => {
+        const delimiter = new DelimiterMaster({});
+        const key = 'key';
+        const value = '';
+
+        assert.strictEqual(delimiter.filter({ key, value }), FILTER_ACCEPT);
+        assert.strictEqual(delimiter.prvKey, key);
+        assert.strictEqual(delimiter.NextMarker, key);
+        assert.deepStrictEqual(delimiter.result(), {
+            CommonPrefixes: [],
+            Contents: [{ key, value }],
+            IsTruncated: false,
+            NextMarker: undefined,
+            Delimiter: undefined,
+        });
+    });
+
+    it('should accept a PHD version as first input', () => {
+        const delimiter = new DelimiterMaster({});
+        const keyPHD = 'keyPHD';
+        const objPHD = {
+            key: keyPHD,
+            value: Version.generatePHDVersion(generateVersionId('', '')),
+        };
+
+        /* When filtered, it should return FILTER_ACCEPT and set the prvKey.
+         * to undefined. It should not be added to result the content or common
+         * prefixes. */
+        assert.strictEqual(delimiter.filter(objPHD), FILTER_ACCEPT);
+        assert.strictEqual(delimiter.prvKey, undefined);
+        assert.strictEqual(delimiter.NextMarker, undefined);
+        assert.deepStrictEqual(delimiter.result(), EmptyResult);
+    });
+
+    it('should accept a PHD version', () => {
+        const delimiter = new DelimiterMaster({});
+        const key = 'keyA';
+        const value = '';
+        const keyPHD = 'keyBPHD';
+        const objPHD = {
+            key: keyPHD,
+            value: Version.generatePHDVersion(generateVersionId('', '')),
+        };
+
+        /* Filter a master version to set the NextMarker, the prvKey and add
+         * and element in result content. */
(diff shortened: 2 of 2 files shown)


jeremyds:
@bert-e wait

lib/algos/list/delimiterMaster.js
```diff
@@ -46,24 +47,22 @@ class DelimiterMaster extends Delimiter {
         }
         const versionIdIndex = key.indexOf(VID_SEP);
         if (versionIdIndex >= 0) {
-            // generally we do not accept a specific version,
-            // we only do when the master version is a PHD version
             key = key.slice(0, versionIdIndex);
-            if (key !== this.prvPHDKey) {
-                return FILTER_ACCEPT; // trick repd to not increase its streak
+            if (key === this.prvPHDKey || key === this.NextMarker) {
```
rahulreddy: Isn't this conditional `key === this.NextMarker` already taken care of on line 43?

lib/algos/list/delimiterMaster.js
```diff
@@ -46,24 +47,22 @@ class DelimiterMaster extends Delimiter {
...
+                /* master version already filtered */
+                return FILTER_SKIP;
             }
         }
         if (Version.isPHD(value)) {
             // master version is a PHD version: wait for the next version
-            this.prvPHDKey = key;
-            return FILTER_ACCEPT; // trick repd to not increase its streak
+            this.prvPHDKey = undefined;
```
rahulreddy: ~~Doesn't this mean PHD along with master gets added to the result set?~~

tests/unit/algos/list/delimiterMaster.js
```diff
@@ -0,0 +1,224 @@
...
+    require('../../../../lib/versioning/constants').VersioningConstants;
+const Version = require('../../../../lib/versioning/Version').Version;
+const { generateVersionId } = require('../../../../lib/versioning/VersionID');
+
+
+const VID_SEP = VSConst.VersionId.Separator;
+
+
+describe('Delimiter All masters listing algorithm', () => {
+    it('Should return good skipping value for DelimiterMaster', done => {
```
rahulreddy: This can be split into two different tests one testing with NextMarker and the other one without it.

tests/unit/algos/list/delimiterMaster.js
```diff
@@ -0,0 +1,224 @@
...
+    it('Should return good filtering value for PHD version', done => {
+        const delimiter = new DelimiterMaster({});
+        const key = 'key';
+        const keyPHD = 'keyPHD';
+        const objPHD = {
+            key: keyPHD,
+            value: Version.generatePHDVersion(generateVersionId('', '')),
+        };
+
+        assert.strictEqual(delimiter.filter({ key, value: '' }), FILTER_ACCEPT);
```
rahulreddy: This is the normal case. It can be written as a test itself. This test should be testing if it's skipping PHD.

tests/unit/algos/list/delimiterMaster.js
```diff
@@ -0,0 +1,224 @@
...
+        /* When filtered, it should return FILTER_ACCEPT and set the prvPHDKey.
+         * to undefined. */
+        assert.strictEqual(delimiter.filter(objPHD), FILTER_SKIP);
+        assert.strictEqual(delimiter.prvPHDKey, undefined);
+        assert.strictEqual(delimiter.NextMarker, key);
+
+        done();
+    });
+
+    it('Should return good filtering value for delete marker', done => {
```
rahulreddy: This can be `should skip a delete marker`

tests/unit/algos/list/delimiterMaster.js
```diff
@@ -0,0 +1,224 @@
...
+        };
+
+        assert.strictEqual(delimiter.filter(obj), FILTER_SKIP);
+        assert.strictEqual(delimiter.NextMarker, undefined);
+        assert.strictEqual(delimiter.prvPHDKey, undefined);
+
+        done();
+    });
+
+    it('Should return good filtering value for version', done => {
```
rahulreddy: `should accept a version`

tests/unit/algos/list/delimiterMaster.js
```diff
@@ -0,0 +1,224 @@
...
+
+        const result = JSON.stringify(delimiter.result());
+        const expected = JSON.stringify({
+            CommonPrefixes: [],
+            Contents: [{ key: 'key', value: '' }],
+            IsTruncated: false,
+            NextMarker: undefined,
+            Delimiter: undefined,
+        });
+        assert.equal(result, expected);
```
rahulreddy: You can assert.deepStrictEqual instead of converting them into strings.

jeremyds:
this Arsenal branch tested in : 
- S3: [here](https://github.com/scality/cloudserver/commits/wip/ZENKO-945-delimitermaster-filter)
- MD:[here]( https://github.com/scality/MetaData/tree/wip/ZENKO-945-delimitermaster-filter)

tests/unit/algos/list/delimiterMaster.js
```diff
@@ -0,0 +1,224 @@
...
+       done => {
+           const delimiter = new DelimiterMaster({ delimiter: '/' });
+           const keyWithEndingDelimiter = 'key/';
+
+           /* With a delimiter set and a NextMarker value ending with this
+            * delimiter, it should return nextMarker. */
+           delimiter.filter({ key: keyWithEndingDelimiter, value: '' });
+           assert.strictEqual(delimiter.NextMarker, keyWithEndingDelimiter);
+
+           assert.strictEqual(delimiter.skipping(), keyWithEndingDelimiter);
```
rahulreddy: Can you add an assertion showing that it has been added to `CommonPrefixes`?

tests/unit/algos/list/delimiterMaster.js
```diff
@@ -0,0 +1,224 @@
...
+
+        const result = JSON.stringify(delimiter.result());
+        const expected = JSON.stringify({
+            CommonPrefixes: [],
+            Contents: [{ key: 'key', value: '' }],
+            IsTruncated: false,
+            NextMarker: undefined,
+            Delimiter: undefined,
+        });
+        assert.equal(result, expected);
```
rahulreddy: This can be assert.deepStrictEqual instead of stringifying the results.

rahulreddy (changes requested):
In the tests whenever `.filter` is called, it changes the state of the delimiter instance. I recommend confining the tests to be specific, i.e. a unit test should test only one thing.

Can you add a test case where the entries may come out of order, for example
```
delimiter.filter(versionedKey);
delimiter.filer(deletemarker);
delimiter.filter(masterKey);
delimiter.filter(PHDKey);
```

lib/algos/list/delimiterMaster.js
jeremyds: not exactly, here key is the object name without the version suffix. It handles a specific case and ~~may~~ deserve a comment: if the previous listing page ends with the major version, the next page will have the nextMarker set to the master key. In that case, prvPHDKey is not set (undefined default value) and comparing the nextMarker is the only way to identify this case.

tests/unit/algos/list/delimiterMaster.js
jeremyds: you re right, it will ease the read

tests/unit/algos/list/delimiterMaster.js
jeremyds: done

lib/algos/list/delimiterMaster.js
jeremyds: comment added

MathieuCassagne (approved):

lib/algos/list/delimiterMaster.js
```diff
@@ -39,31 +40,52 @@ class DelimiterMaster extends Delimiter {
...
-            // version is a delete marker, ignore
-            return FILTER_ACCEPT; // trick repd to not increase its streak
+            /* version is a delete marker, ignore and return accept for the
+             * same reason than for PHD version.  */
+            return FILTER_ACCEPT;
         }
-        // non-PHD master version or a version whose master is a PHD version
-        this.prvPHDKey = undefined;
+
+        this.prvPHDKey = key;
```
lamphamsy: I think it's better to rename `prvPHDKey` according to the change :)

lamphamsy (approved):
LGTM

lib/algos/list/delimiterMaster.js
```diff
@@ -39,31 +40,52 @@ class DelimiterMaster extends Delimiter {
...
+             *   MongoClient). */
+            this.prvPHDKey = undefined;
+            return FILTER_ACCEPT;
         }
         if (Version.isDeleteMarker(value)) {
-            // version is a delete marker, ignore
-            return FILTER_ACCEPT; // trick repd to not increase its streak
+            /* version is a delete marker, ignore and return accept for the
+             * same reason than for PHD version.  */
+            return FILTER_ACCEPT;
```
lamphamsy: I found sth after my approval.
This point and https://github.com/scality/Arsenal/pull/550/files#diff-d72bb66c5574d68c654584b499fe9b9aR80 return `FILTER_ACCEPT`, then will reset the skip-scan mechanism. Hence, as the skipping of the same common prefix is performed later: https://github.com/scality/Arsenal/pull/550/files#diff-d72bb66c5574d68c654584b499fe9b9aR95, we will list (all) objects that are either `isPHD` (Master) or `isDeleteMarker` (Version) for a comment prefix.

lamphamsy: To avoid this issue, could we move the block that check `commonPrefixes` (https://github.com/scality/Arsenal/pull/550/files#diff-d72bb66c5574d68c654584b499fe9b9aR89) to be performed before checking `isPHD` (https://github.com/scality/Arsenal/pull/550/files#diff-d72bb66c5574d68c654584b499fe9b9aR72)?

jeremyds: Returning FILTER_ACCEPT to reset the skip mechanism here is intentional, we want to avoid the client to skip the next entries. 
We reach these two tests when the first one (line 55 & 67) fails to skip the entry, meaning either the entry key does not have a version part or the key has a version and the key part is different from the previous master accepted.
- If the key does have a version part, it is the master version and it may be a PHD. When it is the case we don't want to call addContents or addCOmmonPrefix with this value and, to be able to find the latest version of this key, we want to avoid the client skipping the next entries in its current key range.
- If the entry key has a version part but the entry has not been skipped with the previous test, it is the same principle, we don't want to add it to result and we want to be sure to have a find the latest version.

This kind of input leads to a buggy result by returning FILTER_SKIP in these cases:

> given the SKIP number to activate the skip scan mechanism is SKIP_NUMBER
> a listing page ending with SKIP_NUMBER -1 FILTER_SKIP results
> the next listing page starting with an entry either a PHD or a delete marker

Here, if we return SKIP for the first entry of the second listing, it will trigger the skip scan mechanism and we will miss the version for this key.

Or I miss something :(
At least I should comment more the code, some corner cases are still non obvious.

lib/algos/list/delimiterMaster.js
jeremyds: right!

lib/algos/list/delimiterMaster.js
```diff
@@ -39,31 +40,51 @@ class DelimiterMaster extends Delimiter {
...
         if (Version.isPHD(value)) {
-            // master version is a PHD version: wait for the next version
-            this.prvPHDKey = key;
-            return FILTER_ACCEPT; // trick repd to not increase its streak
+            /* master version is a PHD version, we want to wait for the next
+             * one:
+             * - Set the prvKey to undefined to not skip the next version,
+             * - return accept to avoid users to skip the next values in range
+             *   (skip scan mechanism in metata backend like Metadata or
+             *   MongoClient). */
```
rahulreddy: ~~metata~~ metadata

jeremyds: oops

jeremyds: done

tests/unit/algos/list/delimiterMaster.js
```diff
@@ -0,0 +1,346 @@
...
+        assert.strictEqual(delimiter.NextMarker, key);
+
+        /* With a delimiter skipping should return previous key + VID_SEP
+         * (except when a delimiter is set and the NextMarker ends with the
+         * delimiter) . */
+        assert.strictEqual(delimiter.skipping(), key + VID_SEP);
+    });
+
+    it('should return NextMarker for DelimiterMaster when NextMarker is set' +
+       ', there is a delimiter and the key ends with thedelimiter', () => {
```
rahulreddy: ~~thedelimiter~~ the delimiter

jeremyds: done

vrancurel (approved):

rahulreddy (approved):

lib/algos/list/delimiterMaster.js
MathieuCassagne: We investigated with @lamphamsy, and we found out that the extension returns the expected result, but resetting the skip scan when a PHD/isDeleteMarker version is encountered can make it not working at all, depending on the data inside the bucket.

For example you have 10000 master key in a bucket, but with every 90 master key is a PHD one.
You want to do a listBucket operation, the extension will return FILTER_ACCEPT for each PHD version it encounters, which reset the skip scan. So every 90 master keys the skip scan get resetted, and we end up listing the 10000keys instead of ~100.

I think a good solution would be to have a FILTER_CONTINUE flag, that will go to the next entry without resetting the streakLength counter, and use it when we encounter a PHD/isDeleteMarker version.

lamphamsy: I add sth to @MathieuCassagne 's example: All the keys have a same prefix, e.g. `prefix1/`. The expected result of a list operation should contain an array of common prefixes that is exactly [`prefix1`]. Thanks to the skip-scan mechanism, we should obtain the expected result after scanning several hundreds (200 in Metadata) of keys.
In the case @MathieuCassagne mentioned above, we have to scan all keys as skip-scan's counter is reset for every a PHD Master key.

lamphamsy:
There is one more case we should handle:

Given an object having several versions, e.g. `v1`, `v2`, we delete it without specifying versionId => we delete its Master key (if I understand correctly) => 
- Set Master key` isDeleteMarker = true` (but not `isPHD`),  and
- All its version keys `v1`, `v2` are not changed.

When we list the bucket, keys of the object are received:
- Master key `isDeleteMarker = true` and `isPHD = false` => return `FILTER_ACCEPT`. And `prvKey` is still `underfined` or equal to Master key of the previously listed object: https://github.com/scality/Arsenal/blob/464ab87903bf17962774ef523b1ea374ea43ad48/lib/algos/list/delimiterMaster.js#L84
- Version key `v1` => pass all checks, go to https://github.com/scality/Arsenal/blob/464ab87903bf17962774ef523b1ea374ea43ad48/lib/algos/list/delimiterMaster.js#L87 => add it to the result => It should not be there.

lib/algos/list/delimiterMaster.js
jeremyds: You re absolutely right, we missed this use case. It should be fixed.

jeremyds: my previous comment was addressed to your other comment. about the incorrect listing result.

rahulreddy (commented):
Seems to cover the issues raised. @MathieuCassagne and @lamphamsy . Can you review and approve if it works. You can remove berte wait comment to merge.

lib/algos/list/delimiterMaster.js
```diff
@@ -39,31 +41,63 @@ class DelimiterMaster extends Delimiter {
...
+            /* - key === this.prvKey is triggered when a master version has
+             *   been accepted for this key,
+             * - key === this.NextMarker is triggered when a listing page ends
+             *   on a accepted obj and the next page starts with a version of
+             *   this object. In that case prvKey is default set to undefined
+             *   in the constructor) and comparing to NextMarker is the only
+             *   way to know  we should not accept this version. This test is
+             *   not redundant with the one at the beginning of this function,
+             *   we are comparing here the key without the version suffix. */
+            if (key === this.prvKey || key === this.NextMarker) {
```
lamphamsy: As `this.NextMarker` can be either
- case1: its Master key if that key is added by `addContents()`: https://github.com/scality/Arsenal/blob/2db904952bddcc9a40f786444db44514ae29908a/lib/algos/list/delimiter.js#L132 
- case2: only a common prefix extracted from a Master key added by addCommonPrefix(): https://github.com/scality/Arsenal/blob/2db904952bddcc9a40f786444db44514ae29908a/lib/algos/list/delimiter.js#L184

Case1 is checked correctly here but Case2 might be not. Should we check this case?

lib/algos/list/delimiterMaster.js
```diff
@@ -39,31 +41,63 @@ class DelimiterMaster extends Delimiter {
...
             this.prvPHDKey = key;
-            return FILTER_ACCEPT; // trick repd to not increase its streak
+            return FILTER_ACCEPT;
         }
         if (Version.isDeleteMarker(value)) {
-            // version is a delete marker, ignore
-            return FILTER_ACCEPT; // trick repd to not increase its streak
+            /* This entry is a deleteMarker which has not been filtered by the
+             * version test. Either :
+             * - it is a deleteMarker on the master version, we want to SKIP
```
lamphamsy, line 91: 👍

lamphamsy (commented):
Do we skip the case mentioned in https://github.com/scality/Arsenal/pull/550#discussion_r210588132?
Moreover, there is a small case that might be verified.

lib/algos/list/delimiterMaster.js
MathieuCassagne: For the case 2, if a master key has been added by addCommonPrefix(), this.prvKey is set to the master key, and so the check at line 67 is true, and we return FILTER_SKIP.
I think we're good ?

lamphamsy: yes, it's true but only if these keys are on the same page.
We need `this.NextMarker` only for the next page starts: https://github.com/scality/Arsenal/blob/2db904952bddcc9a40f786444db44514ae29908a/lib/algos/list/delimiterMaster.js#L60
At that time, `this.prvKey` is reset to be `undefined` in the constructor.

lamphamsy: I think we could add an extra condition that checks `NextMarker` is a common prefix and `key` starting with `NextMarker`, e.g.

```javascript
if (key === this.prvKey || key === this.NextMarker ||
    (this.delimiter && this.NextMarker && 
     this.NextMarker.endsWith(this.delimiter) && key.startsWith(this.NextMarker)) {
```

MathieuCassagne: After more investigation it appear that our NextMarker might not be set to the good value when it is a CommonPrefixes, we're setting it to the common prefix we found, but we should set it to the key which we extract the common prefix.

The AWS documentation about NextMarker:
"
When the response is truncated (that is, the IsTruncated element value in the response is true), you can use the key name in this field as a marker in the subsequent request to get next set of objects. Amazon S3 lists objects in UTF-8 character encoding in lexicographical order. 
"
It is said that NextMarker, if exists, should be a key, but we set it to a CommonPrefix, which is not a key.
If we can change this, the check line 67, is good as it is for both case 1 and 2.

Also there is a transformation done on NextMarker in the constructor of delimiter.js, we might want to use the marker(which is NextMarker we sent back without the modification) parameter instead.

jeremyds: wow, amazing review again. I don't want to handle the change of NextMarker value for common prefix in delimiter.js in this PR. I am going to open an issue, adds the test mentioned by @lamphamsy here https://github.com/scality/Arsenal/pull/550#discussion_r211982038 and tracked this issue by a comment in the code.

rahulreddy (approved):
I cannot review with the same depth as @MathieuCassagne and @MathieuCassagne (thanks guys for the reviews and raising the issues). Going by their reviews, doing my own review and seeing that @jeremyds addressed the corner case raised, I am merging this PR. 
@jeremyds will follow up with the common prefix fix for delimiter.js in another PR.

PR #555, opened 2018-08-23
bugfix: ZENKO-1024 Add pending counters

bennettbuchanan:
Opening this PR for high-level reviews. Extensive testing for this new metric is still needed.

See associated backbeat PR https://github.com/scality/backbeat/pull/412.

### DIFF
--- a/lib/backbeat/Metrics.js
+++ b/lib/backbeat/Metrics.js
@@ -6,6 +6,7 @@ const StatsModel = require('../../lib/metrics/StatsModel');
 const INTERVAL = 300; // 5 minutes
 const EXPIRY = 86400; // 24 hours
 const THROUGHPUT_EXPIRY = 900; // 15 minutes
+const isTest = process.env.CI === 'true';
 
 class Metrics {
     constructor(config, logger) {
@@ -31,8 +32,9 @@ class Metrics {
      */
     _queryStats(ops, site, bucketName, objectKey, versionId, cb) {
         return async.map(ops, (op, done) => {
+            const hasGlobalKey = this._hasGlobalKey(op);
             if (site === 'all') {
-                const queryString = `*:${op}:*`;
+                const queryString = hasGlobalKey ? `*:${op}` : `*:${op}:*`;
                 return this._redisClient.scan(queryString, undefined,
                 (err, res) => {
                     if (err) {
@@ -42,6 +44,10 @@ class Metrics {
                             type: errors.InternalError,
                             method: 'Metrics._queryStats',
                         });
+                    }
+                    if (hasGlobalKey) {
+                        return this._statsClient.getAllGlobalStats(res,
+                            this._logger, done);
                     }
                     const allKeys = res.map(key => {
                         const arr = key.split(':');
@@ -71,8 +77,34 @@ class Metrics {
             } else {
                 queryString = `${site}:${op}`;
             }
+            if (hasGlobalKey) {
+                return this._redisClient.get(queryString, (err, res) => {
+                    if (err) {
+                        return done({
+                            message: `Redis error: ${err.message}`,
+                            type: errors.InternalError,
+                            method: 'Metrics._queryStats',
+                        });
+                    }
+                    return done(null, { requests: [res || 0] });
+                });
+            }
             return this._statsClient.getStats(this._logger, queryString, done);
         }, cb);
+    }
+
+    /**
+     * Determines whether the Redis op uses a global counter or interval key.
+     * @param {String} op - The Redis operation
+     * @return {Boolean} true if a global counter, false otherwise
+     */
+    _hasGlobalKey(op) {
+        if (isTest) {
+            return op.includes('test:bb:bytespending') ||
+                op.includes('test:bb:opspending');
+        }
+        return op.includes('bb:crr:bytespending') ||
+            op.includes('bb:crr:opspending');
     }
 
     /**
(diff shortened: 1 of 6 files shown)


bennettbuchanan:
@bert-e wait

lib/backbeat/Metrics.js
```diff
@@ -410,6 +410,50 @@ class Metrics {
...
 
+    /**
+     * Get pending replication stats by ops count and size in bytes
+     * @param {object} details - route details from lib/backbeat/routes.js
+     * @param {function} cb - callback(error, data)
+     * @return {undefined}
+     */
+    getPending(details, cb) {
+        const { site, dataPoints } = details;
+        // Query only a single given site.
```
dora-korpar: Are there plans to do a 'getPending' for all sites?

bennettbuchanan: Yes, I'm working on including it in the response of https://github.com/scality/backbeat/blob/development/8.0/docs/metrics.md#_backbeatapimetricscrrlocation-name

philipyoo (approved):

lib/backbeat/Metrics.js
```diff
@@ -71,10 +77,36 @@ class Metrics {
             } else {
                 queryString = `${site}:${op}`;
             }
+            if (hasGlobalKey) {
+                return this._redisClient.get(queryString, (err, res) => {
+                    if (err) {
+                        return done({
+                            message: `Redis error: ${err.message}`,
```
rahulreddy, line 84: This is an anti-pattern, why is this isn't an instance of `Error`?

lib/backbeat/Metrics.js
```diff
@@ -410,6 +442,43 @@ class Metrics {
...
+            if (err && err.type) {
+                this._logger.error('error getting metric: pending', {
+                    origin: err.method,
+                    method: 'Metrics.getPending',
+                });
+                return cb(err.type.customizeDescription(err.message));
+            }
+            if (err || res.length !== details.dataPoints.length) {
+                this._logger.error('error getting metrics: pending', {
+                    method: 'Metrics.getPending',
```
rahulreddy: Can you log `err`, `res` and `details.dataPoints` here?

lib/backbeat/Metrics.js
```diff
@@ -410,6 +442,43 @@ class Metrics {
...
+                });
+                return cb(errors.InternalError);
```
rahulreddy: Add a custom description here.

rahulreddy (changes requested):
The stats calculation LGTM

lib/backbeat/Metrics.js
philipyoo: This is debt that will need to be refactored eventually. The error log if we just return the error returned by redis will not properly log from the BackbeatServer (also I guess logged in BackbeatAPI as well)

lib/metrics/StatsClient.js
```diff
@@ -105,6 +129,37 @@ class StatsClient {
...
     }
 
+    /**
+     * Handles getting a list of global keys.
+     * @param {array} ids - Service identifiers
+     * @param {object} log - Werelogs request logger
+     * @param {function} cb - Callback
+     * @return {undefined}
+     */
+    getAllGlobalStats(ids, log, cb) {
```
philipyoo: Just a suggestion, but maybe move this to StatsModel to separate out methods that will be used by utapi and methods that will be used by backbeat

lib/metrics/StatsClient.js
```diff
@@ -105,6 +129,37 @@ class StatsClient {
...
+            if (err) {
+                log.error('error getting metrics', {
+                    error: err,
+                    method: 'StatsClient.getAllGlobalStats',
+                });
+                return cb(null, statsRes);
+            }
+            statsRes.requests = res.reduce((sum, curr) => {
+                if (curr[0]) {
+                    log.error('error getting metrics', {
```
philipyoo: i'm wondering if this could potentially be more confusing to a user. In a global metrics case, the only time this would error would be if the pending metrics for a given site does not exist yet. So either a brand new location, or maybe somehow the metrics key was deleted. Or the new location was setup for replication and never was replicated to yet. These errors may show up frequently as cloudserver will query metrics often.

Maybe best to ignore these logs

lib/metrics/StatsClient.js
bennettbuchanan: Okay, I've moved this. I didn't realize there were plans to have another client for Utapi.

lib/backbeat/Metrics.js
bennettbuchanan: Indeed, I've followed the pattern of logging that is used in this file. We could eventually refactor those as well.

rahulreddy (approved):

lib/metrics/StatsClient.js
bennettbuchanan: If the global key has not been created, the response will be `[ null, null ]` for each response of the batched request. I was thinking it would be helpful to log any error that could occur in an individual request from the batch. 

I've updated the variable names and added a comment to make this more explicit. Does this work for you, or do you think it's still best to omit these error logs?

lib/metrics/StatsClient.js
philipyoo: sorry bad wording on my part. I meant StatsClient is used by utapi, and StatsModel was adjusted to fit for backbeat metrics. Since this method will be only used for backbeat, I thought it'd be nice to keep it in StatsModel

philipyoo (approved):

PR #556, opened 2018-09-01
bf: ZENKO-1024 use pending metrics for backlog

philipyoo:
Pending metrics don't expire which was a cause for problems
with current backlog. This quick fix is to use pending
metrics in place of backlog but keeping the same names
and routes in place to avoid regression.

To be merged with https://github.com/scality/backbeat/pull/415

### DIFF
--- a/lib/backbeat/Metrics.js
+++ b/lib/backbeat/Metrics.js
@@ -164,25 +164,16 @@ class Metrics {
                 });
                 return cb(errors.InternalError);
             }
-            const uptime = this._getMaxUptime(EXPIRY);
-            const numOfIntervals = Math.ceil(uptime / INTERVAL);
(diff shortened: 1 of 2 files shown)


anurag4DSB (approved):

alexanderchan-scality (approved):

philipyoo:
Sorry, I'd like to make a small change to this PR. Will be removing "backlog" from the all metrics route and only displaying pending as both are now the same.

philipyoo:
@bert-e wait

philipyoo:
Hmm, actually this change will require changes in cloudserver and orbit :/
For now, keeping as is and will prioritize testing

bennettbuchanan (approved):

PR #558, opened 2018-09-05
fx: ZENKO-1112 Incomplete management client error logging

nicolas2bert:
`errorUtils` is imported in:
- https://github.com/scality/cloudserver/pull/1528/files
- https://github.com/scality/backbeat/pull/420/files

### DIFF
--- a/lib/errorUtils.js
+++ b/lib/errorUtils.js
@@ -1,0 +1,13 @@
+function reshapeExceptionError(error) {
+    const { message, code, stack, name } = error;
+    return {
+        message,
+        code,
+        stack,
+        name,
+    };
+}
+
+module.exports = {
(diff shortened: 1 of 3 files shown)


lib/errorUtils.js
```diff
@@ -0,0 +1,10 @@
+function reshapeExceptionError(error) {
+    return {
+        message: error.message,
+        code: error.code,
```
rachedbenmustapha: From the docs, some attributes that can be useful to capture are also `stack` and `name`.

rachedbenmustapha (changes requested):

tmacro (approved):

rachedbenmustapha (approved):

PR #559, opened 2018-09-07
Bugfix/zenko 1124 fix delete marker handling

jeremyds:


jeremyds:
@bert-e reset

ghost (approved):
As discussed offline, this fix looks okay, and I didn't identify any risk related to this change for MetaData.

That being said, I really think we should add more regression testing whatever the layer, for this; rather than only one case in an unit test.

bennettbuchanan (approved):

bennettbuchanan:
@jeremyds It looks like change is affecting other delimiter tests.

jeremyds:
@bert-e reset

jeremyds:
@bert-e wait

lib/algos/list/delimiterMaster.js
```diff
@@ -116,7 +117,7 @@ class DelimiterMaster extends Delimiter {
     }
 
     skipping() {
-        if (this.NextMarker) {
+        if (this.NextMarker && this.NextMarker !== this.markerParams) {
```
ghost, line 120: Does this mean that if we haven't updated the NextMarker, `skipping()` will return `SKIP_NONE` ?
I don't see how that improves the situation. Was there a new corner case identified since the last version of the fix ?

ghost (commented):
questions about new fix code

vrancurel:
wontfix

PR #563, opened 2018-09-10
ZENKO-1124: mongo listing, avoid to loop

jeremyds:


ghost (commented):
I have a hard time understanding the issue here.
My guess is that it's related to the changes in https://github.com/scality/Arsenal/pull/559
That being said, with what I read on the other PR, I fear that we are created new corner cases where NextMarker isn't properly updated, and we end up with a `skipping() == SKIP_NONE`; and I'm also unsure about the effect of that on this code.
Need an offline conversation to clarify this.

rahulreddy:
@DavidPineauScality This is related to https://scality.atlassian.net/browse/ZENKO-1124 , observed at a customer.

ghost:
I understand that, but my issue is that I don't understand how the new/alternative fix in  #559 works, and I'm worried whether the fix here should also land in MetaData.
So I'm a little lost.

rahulreddy:
Oh I was only commenting on what this issue is related to. As for the changes, I don't understand them either, it's better to schedule a call with @jeremyds.

vrancurel:
In theory @DavidPineauScality you indeed have to change Metadata code.

We may factorize this function in Arsenal though since it is pure common code.

rahulreddy (commented):
I will be discussing the solution offline with @jeremyds . I built a custom image and verified that the fix works.

ghost:
Ok, thanks for the heads up.
I'll leave that up to you guys, but if my feedback/understanding/review is still necessary, I'm available for a sync-up

jeremyds:
Sorry guys, I did not communicate well on the 2 PR. The other one, the fix in delimiterMaster, does not work. I tried to find a solution in the delimiterMaster code but each solution I found introduced (sometime subtle) regressions or behavior changes I was not confident with.
I finally fix the infinite loop on the same range bug in metadata in this PR. We don't use the skipping range returned by the skipping function when it is the same that the actual one, avoiding to loop on it.

Vianney closed the other PR, I should have done it yesterday.

vrancurel (approved):

anurag4DSB (approved):

vrancurel (approved):

lib/algos/list/skip.js
```diff
@@ -0,0 +1,88 @@
...
+
+const MAX_STREAK_LENGTH = 100;
+
+/**
+ * Handle the filtering and the skip mechanism of a listing result.
+ */
+class Skip {
+    /**
+     * @param {Object} params    - skip parameters
+     * @param {Object} extension - delimiter extension used (required)
```
bennettbuchanan: Minor: This looks like additional arguments to the class. Can you update to make it apparent they are properties of `params`?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -962,47 +962,34 @@ class MongoClientInterface {
...
+        skip.setSkipRangeCb(range => {
+            // stop listing this key range
+            stream.destroy();
+            // update the new listing parameters here
+            // eslint-disable-next-line no-param-reassign
+            params.start = undefined; // 'start' is deprecated
+            // eslint-disable-next-line no-param-reassign
+            params.gt = undefined;
+            // eslint-disable-next-line no-param-reassign
+            params.gte = range;
```
bennettbuchanan: Minor: Could be a bit cleaner to create a new object and reassign these properties rather than disabling the linter.

jeremyds: Done. I saw it so much in the code than I thought it is a good practice :S

lib/algos/list/skip.js
jeremyds: done

bennettbuchanan (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -962,47 +962,34 @@ class MongoClientInterface {
...
+        const skip = new Skip({
+            extension,
+            gte: params.gte,
+        });
         let cbDone = false;
-        let streakLength = 0;
-        const stream = new MongoReadStream(c, params,
-                                           params.mongifiedSearch);
+
+        skip.setListingEndCb(() => {
```
jonathan-gramain, line 972: minor: if those are mandatory callbacks, maybe they should be set directly in the constructor params instead of setters?

lib/algos/list/skip.js
```diff
@@ -0,0 +1,88 @@
...
+
+        /* Used to count consecutive FILTER_SKIP returned by the extension
+         * filter method. Once this counter reaches MAX_STREAK_LENGTH, the
+         * filter function tries to skip unwanted values by defining a new
+         * range. */
+        this.streakLength = 0;
+    }
+
+    setListingEndCb(cb) {
+        this.listingEndCb = cb;
```
jonathan-gramain, line 35: `listingEndCb` and `skipRangeCb` contracts should be documented explicitly, since they are essential to the behavior of the listing extension.

jonathan-gramain (approved):
I'm good with the logic, I have minor comments on the style though, feel free to ignore if time is short.

ghost:
Note that I can still identify one corner-case (I don't see a solution ATM, tbh):
Lots of deleted-master-versions means that we'll continue listing. If the number is high enough, we might be able to hit a listing timeout, Thus answering with a 500 instead of answering with a truncated listing.

PR #564, opened 2018-09-11
S3C-1665: delimiter*. return FILTER_END ASAP

jeremyds:
For all the delimiter extension, once we reach the maxKey parameter
entries added to the result, we should return FILTER_END on the next
entry.

Currently, we wait to find the next entry to add to the result to
return FILTER_END (and we don't add this entry to the result).

This commit moves the check to test is maxKeys is reached from
addContent/addCommon methods to the begining of the filter method.

### DIFF
--- a/lib/algos/list/delimiter.js
+++ b/lib/algos/list/delimiter.js
@@ -1,7 +1,13 @@
 'use strict'; // eslint-disable-line strict
 
 const Extension = require('./Extension').default;
-const { inc, FILTER_END, FILTER_ACCEPT, FILTER_SKIP } = require('./tools');
+const {
+    inc,
+    FILTER_END,
+    FILTER_ACCEPT,
+    FILTER_SKIP,
+    AWS_LISTING_DEFAULT_MAX_KEYS,
+} = require('./tools');
 
 /**
  * Find the next delimiter in the path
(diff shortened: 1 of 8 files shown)


jeremyds:
@bert-e wait

ghost (approved):
I don't see any problem with this fix.
That being said, this is a minor fix, right ? It's not fixing anything major ?

MathieuCassagne (approved):

lib/algos/list/delimiter.js
```diff
@@ -146,6 +143,10 @@ class Delimiter extends Extension {
      *  @return {number}          - indicates if iteration should continue
      */
     filter(obj) {
+        if (this._reachedMaxKeys()) {
+            return FILTER_END;
```
lamphamsy, line 139: When maxKeys is reached, the filtering `obj` would not be listed in the result. So we could miss the `obj`.

jeremyds: It was already the case before. It was not added in the result in addContents (this.Contents.push({ key, value}) not called if _reachedMaxKeys true). On that aspect the "fix" is ISO, it does not modify the result value. 

The main difference is we now return FILTER_END as soon as possible, whatever the next entry is. 

Before, if we look in this file (implementation differs from extension to others), the entry may be skipped at line 156, before to be given to addContent/addCommon (where the max keys reached test was). So if the entry which makes the count reached the max keys is followed by skipped entries, we wait to have a valid one before to reurn FILTER_END.

jeremyds:
@DavidPineauScality @MathieuCassagne I am sorry I realized yesterday that I should add this in the basic and the MPU extensions... I realized too that the _maxKeysReached was duplicate. I mutualize it in Extension class and added a test unit for this fix.
I am pushing a version which add some code in this PR you have already reviewed, sorry :S

lib/algos/list/delimiter.js
lamphamsy: Ah yes it is  👍. I did not check the current algo :(

ghost:
@jeremyds should I re-review this ? I realize it's been left to take dust since then, and I wasn't reminded of it.

PR #567, opened 2018-09-14
bf: ZENKO-1144 add support for Redis sorted sets

philipyoo:
These changes are to support use of sorted sets in Redis for crr metrics. Plan is to use hourly sorted sets to track failure metrics and keep a epoch sorted list of failure keys.

Changes in this PR:
- Add wrapper for Redis sorted set methods: ZADD, ZCARD, ZRANGE, ZRANGEBYSCORE, ZREM, ZSCORE
- Add wrapper for Redis methods: SET, EXISTS, TTL
- `StatsModel.getSortedSetCurrentHour` returns array of 24 normalized hourly timestamps
- `StatsModel.addToSortedSet` adds to a sorted set and applies expiry if adding to new sorted set

### DIFF
--- a/lib/metrics/RedisClient.js
+++ b/lib/metrics/RedisClient.js
@@ -204,6 +204,16 @@ class RedisClient {
         return this._client.zrangebyscore(key, min, max, cb);
     }
 
+    /**
+     * get TTL or expiration in seconds
+     * @param {string} key - name of key
+     * @param {function} cb - callback
+     * @return {undefined}
+     */
+    ttl(key, cb) {
+        return this._client.ttl(key, cb);
+    }
+
     clear(cb) {
         return this._client.flushdb(cb);
     }
--- a/lib/metrics/RedisClient.js
+++ b/lib/metrics/RedisClient.js
@@ -69,6 +69,20 @@ class RedisClient {
     }
 
     /**
+     * increment value of a key by a given amount and set a ttl
+     * @param {string} key - key holding the value
+     * @param {number} amount - amount to increase by
+     * @param {number} expiry - expiry in seconds
+     * @param {callback} cb - callback
+     * @return {undefined}
+     */
+    incrbyEx(key, amount, expiry, cb) {
+        return this._client
+            .multi([['incrby', key, amount], ['expire', key, expiry]])
+            .exec(cb);
+    }
+
+    /**
      * decrement value of a key by a given amount
      * @param {string} key - key holding the value
      * @param {number} amount - amount to increase by
@@ -90,17 +104,15 @@ class RedisClient {
     }
 
     /**
-     * increment value of a key by a given amount and set a ttl
-     * @param {string} key - key holding the value
-     * @param {number} amount - amount to increase by
-     * @param {number} expiry - expiry in seconds
-     * @param {callback} cb - callback
+     * Checks if a key exists
+     * @param {string} key - name of key
+     * @param {function} cb - callback
+     *   If cb response returns 0, key does not exist.
+     *   If cb response returns 1, key exists.
      * @return {undefined}
      */
-    incrbyEx(key, amount, expiry, cb) {
-        return this._client
-            .multi([['incrby', key, amount], ['expire', key, expiry]])
-            .exec(cb);
+    exists(key, cb) {
+        return this._client.exists(key, cb);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


philipyoo:
@bert-e wait

lib/metrics/RedisClient.js
```diff
@@ -90,17 +104,25 @@ class RedisClient {
...
+     * set value for key
+     * @param {string} key - name of key to set
      * @param {callback} cb - callback
      * @return {undefined}
      */
-    incrbyEx(key, amount, expiry, cb) {
-        return this._client
-            .multi([['incrby', key, amount], ['expire', key, expiry]])
-            .exec(cb);
+    set(key, cb) {
```
bennettbuchanan: Don't we not need this, since we aren't using failure keys anymore?

lib/metrics/RedisClient.js
```diff
@@ -113,6 +135,93 @@ class RedisClient {
...
+     * @return {undefined}
+     */
+    zrange(key, start, end, cb) {
+        return this._client.zrange(key, start, end, cb);
+    }
+
+    /**
+     * Get range of elements in a sorted set based off score
+     * @param {string} key - name of key
+     * @param {integer} min - min score value (inclusive) (can use "-inf")
```
bennettbuchanan: Let's update this doc to `integer|string` to account for '-inf' values.

lib/metrics/StatsModel.js
```diff
@@ -143,6 +143,76 @@ class StatsModel extends StatsClient {
...
+     * helper method to add element to a sorted set, applying TTL if new set
+     * @param {string} key - name of key
+     * @param {integer} score - score used to order set
+     * @param {string} value - value to store
+     * @param {callback} cb - callback
+     * @return {undefined}
+     */
+    addToSortedSet(key, score, value, cb) {
+        // milliseconds in a day
+        const msInADay = 24 * 60 * 60 * 1000;
```
bennettbuchanan: Minor: for readability, it seems like this should be within the if block in which it is used.

lib/metrics/StatsModel.js
```diff
@@ -143,6 +143,76 @@ class StatsModel extends StatsClient {
...
+                return cb(err);
+            }
+            if (resCode === 0) {
+                const nearestHour = this._normalizeTimestampByHour(new Date());
+                const ttl = msInADay - (Date.now() - nearestHour);
+                const cmds = [
+                    ['zadd', key, score, value],
+                    ['expire', key, ttl],
+                ];
+                return this._redis.batch(cmds, cb);
```
bennettbuchanan: It would be useful to check for the batched command errors (i.e. the array of arrays), so that this method returns an error if any of those sub commands fail as well.

tests/functional/metrics/StatsModel.js
```diff
@@ -243,4 +243,65 @@ describe('StatsModel class', () => {
...
+        const res = statsModel.getSortedSetHours(epoch);
+
+        assert.deepStrictEqual(res, expected);
+    });
+
+    it('should apply TTL on a new sorted set using addToSortedSet', () => {
+        const key = 'a-test-key';
+        const score = 100;
+        const value = 'a-value';
+        // redisClient.ttl(key, (err, res))
```
bennettbuchanan: Extra comment.

lib/metrics/StatsModel.js
philipyoo: sounds good, on success I'm planning on returning only the `zadd` return value to keep both successful returns the same

JianqinWang (approved):

bennettbuchanan (approved):

lib/metrics/StatsModel.js
```diff
@@ -143,6 +143,86 @@ class StatsModel extends StatsClient {
...
+    /**
+     * helper method to add element to a sorted set, applying TTL if new set
+     * @param {string} key - name of key
+     * @param {integer} score - score used to order set
+     * @param {string} value - value to store
+     * @param {callback} cb - callback
+     * @return {undefined}
+     */
+    addToSortedSet(key, score, value, cb) {
+        this._redis.exists(key, (err, resCode) => {
```
rahulreddy, line 198: I am assuming that you are only trying to expire when it's set for the first time and it's 24 hours? Your code will be cleaner if you use `EXPIREAT` which takes a timestamp. Also `EXPIRE` takes value in seconds, the ttl you calculated is in milliseconds. If you use `EXPIREAT`, you can set it everytime and don't have to check if the key exists.

philipyoo: We only want to apply the expire if its brand new. We won't know which hour sorted set we are adding to, so I'll keep it as is except for the ms -> sec issue

PR #571, opened 2018-09-16
bf: ZENKO-1155 add index restriction for mongo find call

alexanderchan-scality:
Fixes the issue in which mongo performs looks through a collection to find an entry with specify version-id


alexanderchan-scality:
@bert-e wait

PR #572, opened 2018-09-16
bf: ZENKO-1155 remove incremental update

alexanderchan-scality:


alexanderchan-scality:
@bert-e wait

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -551,7 +551,9 @@ class MongoClientInterface {
         const { objName, versionId, collectionName } = params;
         if (isUserBucket(collectionName)) {
             if (isUpdate) {
+                const findId = new RegExp(objName);
```
jonathan-gramain: Why the need for a regexp?

alexanderchan-scality: What this code attempts to do is to find objects with a specific version id; in this case, it is attempting to find the master(placeholder), an entry that doesn't  the versioned id appended, and the corresponding version id. So, the regex will looking for entry-ids that match the regex.

vrancurel: yes but there is a chance it translates to a SCAN as well

vrancurel (changes requested):

vrancurel (approved):

jonathan-gramain (approved):

bennettbuchanan (approved):

PR #573, opened 2018-09-20
ft: ZENKO-1186 stalled sanity check

alexanderchan-scality:


alexanderchan-scality:
@bert-e wait

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1277,6 +1277,39 @@ class MongoClientInterface {
...
+                id: '$_id',
+                status: '$value.replicationInfo.status',
+            },
+            'value.last-modified': 1,
+        };
+        return c.aggregate([
+            { $project: reducedFields },
+            { $match: {
+                '_id._id': { $regex: /\0/ },
+                '_id.status': { $eq: 'PENDING' },
```
bennettbuchanan: In a one-to-many scenario, could this be 'PROCESSING' status? Or is it that once it's queued we can assume CRR for all backends will at least be attempted?

alexanderchan-scality: just 'PENDING', I believe. Since we're are looking for entries that aren't touched by backbeat.
@rahulreddy confirm?

rahulreddy: Correct. We only care about `PENDING` entries as we are trying to find out which entries could haven't been queued in Kafka.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1277,6 +1279,39 @@ class MongoClientInterface {
...
+            } },
+        ]).toArray((err, res) => {
+            if (err) {
+                log.debug('Unable to retrieve stalled entries', {
+                    error: err,
+                });
+                return cb(null);
+            }
+            const count = res.filter(data => {
+                const time = data.value['last-modified'] || null;
```
rahulreddy: Is data and value guaranteed to be defined? You may want to check?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1295,13 +1330,18 @@ class MongoClientInterface {
             'value.versionId': { $exists: false },
         };
 
+        const hour = 60 * 60 * 1000;
+        const cmpDate = new Date(Date.now() - hour);
```
rahulreddy: Can you use Date object methods to set the hours? I know that this works but that way is more clean.

rahulreddy (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1277,6 +1279,43 @@ class MongoClientInterface {
...
+            { $match: {
+                '_id.id': { $regex: /\0/ },
+                '_id.status': { $eq: 'PENDING' },
+            } },
+        ]).toArray((err, res) => {
+            if (err) {
+                log.debug('Unable to retrieve stalled entries', {
+                    error: err,
+                });
+                return cb(null);
```
bennettbuchanan, line 1301: Is it intended to not return the error here?

bennettbuchanan (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
bennettbuchanan: As per conversation, this is intended.

PR #575, opened 2018-09-24
bugfix: ZENKO-1165 Allow option to list withscores

bennettbuchanan:
Allow listing sorted sets with scores. 🚀


bennettbuchanan:
@bert-e wait

bennettbuchanan:
Seeing as this is an edge case and we're already far along in the RC process for Zenko 1.1., I’m putting this on hold so that https://scality.atlassian.net/browse/ZENKOIO-11  can take precedence.

PR #588, opened 2018-10-18
ZENKO-833: add ingestion param for BucketInfo

JianqinWang:
The ingestion flag will be sent in the locationConstraint string as `<location name>:ingest`.
The standard location constraint string looks like `us-east-1`. With the ingestion flag, it would look like `us-east-1:ingest`.

### DIFF
--- a/lib/models/BucketInfo.js
+++ b/lib/models/BucketInfo.js
@@ -4,10 +4,9 @@ const uuid = require('uuid/v4');
 const { WebsiteConfiguration } = require('./WebsiteConfiguration');
 const ReplicationConfiguration = require('./ReplicationConfiguration');
 const LifecycleConfiguration = require('./LifecycleConfiguration');
-
 // WHEN UPDATING THIS NUMBER, UPDATE MODELVERSION.MD CHANGELOG
 // MODELVERSION.MD can be found in S3 repo: lib/metadata/ModelVersion.md
-const modelVersion = 9;
+const modelVersion = 10;
 
 class BucketInfo {
     /**
@@ -36,7 +35,8 @@ class BucketInfo {
     * @param {object} versioningConfiguration - versioning configuration
     * @param {string} versioningConfiguration.Status - versioning status
     * @param {object} versioningConfiguration.MfaDelete - versioning mfa delete
-    * @param {string} locationConstraint - locationConstraint for bucket
+    * @param {string} locationConstraint - locationConstraint for bucket that
+    * also includes the ingestion flag
     * @param {WebsiteConfiguration} [websiteConfiguration] - website
     * configuration
     * @param {object[]} [cors] - collection of CORS rules to apply
@@ -54,13 +54,14 @@ class BucketInfo {
     * @param {string} readLocationConstraint - readLocationConstraint for bucket
     * addition for use with lifecycle operations
     * @param {boolean} [isNFS] - whether the bucket is on NFS
+    * @param {object} [ingestionConfig] - object for ingestion status: en/dis
     */
     constructor(name, owner, ownerDisplayName, creationDate,
                 mdBucketModelVersion, acl, transient, deleted,
                 serverSideEncryption, versioningConfiguration,
                 locationConstraint, websiteConfiguration, cors,
                 replicationConfiguration, lifecycleConfiguration, uid,
-                readLocationConstraint, isNFS) {
+                readLocationConstraint, isNFS, ingestionConfig) {
         assert.strictEqual(typeof name, 'string');
         assert.strictEqual(typeof owner, 'string');
         assert.strictEqual(typeof ownerDisplayName, 'string');
@@ -97,6 +98,9 @@ class BucketInfo {
         }
         if (locationConstraint) {
             assert.strictEqual(typeof locationConstraint, 'string');
+        }
+        if (ingestionConfig) {
+            assert.strictEqual(typeof ingestionConfig, 'object');
         }
         if (readLocationConstraint) {
             assert.strictEqual(typeof readLocationConstraint, 'string');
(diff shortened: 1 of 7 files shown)


lib/models/BucketInfo.js
```diff
@@ -97,6 +100,11 @@ class BucketInfo {
         }
         if (locationConstraint) {
             assert.strictEqual(typeof locationConstraint, 'string');
+            ingestion = locationConstraint.split(zenkoSeparator)[1];
+        }
+        if (ingestionConfig) {
+            ingestion = ingestionConfig;
+            assert.strictEqual(typeof ingestion, 'string');
```
philipyoo: when will `locationConstraint` not have the ingestion flag but will have `ingestionConfig`?

lib/models/BucketInfo.js
```diff
@@ -147,14 +155,16 @@ class BucketInfo {
...
+        this._locationConstraint = locationConstraint ?
+            locationConstraint.split(zenkoSeparator)[0] : null;
         this._readLocationConstraint = readLocationConstraint || null;
         this._websiteConfiguration = websiteConfiguration || null;
         this._replicationConfiguration = replicationConfiguration || null;
         this._cors = cors || null;
         this._lifecycleConfiguration = lifecycleConfiguration || null;
         this._uid = uid || uuid();
         this._isNFS = isNFS || null;
+        this._ingestion = ingestion ? { status: 'Enabled' } : null;
```
philipyoo: I think this is just fine, and can't think of any case where on bucket creation, ingestion is not enabled at the start, but maybe just to be safe, use the `setIngestion` method here to set this instance variable.

lib/models/BucketInfo.js
JianqinWang: the ingestion flag would be its own separate value on a bucketInfo model, `locationConstraint` will be the carrier of the ingestion flag but will not store the ingestion value

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1054,6 +1054,7 @@ class MongoClientInterface {
                             isVersioned:
                                 !!bucketInfo.getVersioningConfiguration(),
                             ownerCanonicalId: bucketInfo.getOwner(),
+                            ingestion: bucketInfo.isIngestionEnabled(),
```
rachedbenmustapha: For the management layer's purposes, this should not change when ingestion is paused, what we need here is the nature of the bucket (ingest vs normal), not the current state.

ghost:
There were the following issues with this Pull Request

* Commit: e24a5a14dbdb1ca50f134cebb822ae2c0b51e913
  - ✖ message may not be empty
  - ✖ type may not be empty


You may need to [change the commit messages][ref] to comply with the repository contributing guidelines.

--------

🤖 This comment was generated by [**commitlint[bot]**][repo]. Please report issues [here][issues].

Happy coding!

[ref]: https://help.github.com/articles/changing-a-commit-message/
[repo]: https://github.com/z0al/commitlint-bot
[issues]: https://github.com/z0al/commitlint-bot/issues

lib/storage/metadata/mongoclient/MongoClientInterface.js
JianqinWang: I've added a new function that would check if the bucket is/is not an ingestion bucket, and replaced it here :)

ghost:
There were the following issues with this Pull Request

* Commit: da100967e46eb6ddbb0fdb36615403ac008497cd
  - ✖ message may not be empty
  - ✖ type may not be empty


You may need to [change the commit messages][ref] to comply with the repository contributing guidelines.

--------

🤖 This comment was generated by [**commitlint[bot]**][repo]. Please report issues [here][issues].

Happy coding!

[ref]: https://help.github.com/articles/changing-a-commit-message/
[repo]: https://github.com/z0al/commitlint-bot
[issues]: https://github.com/z0al/commitlint-bot/issues

lib/models/BucketInfo.js
```diff
@@ -147,14 +155,16 @@ class BucketInfo {
...
+        this._locationConstraint = locationConstraint ?
+            locationConstraint.split(zenkoSeparator)[0] : null;
         this._readLocationConstraint = readLocationConstraint || null;
         this._websiteConfiguration = websiteConfiguration || null;
         this._replicationConfiguration = replicationConfiguration || null;
         this._cors = cors || null;
         this._lifecycleConfiguration = lifecycleConfiguration || null;
         this._uid = uid || uuid();
         this._isNFS = isNFS || null;
+        this._ingestion = ingestion ? { Status: 'Enabled' } : null;
```
rahulreddy: Let's keep this lowercased `{ status: enabled }` should be sufficient.

rahulreddy (changes requested):
I would recommend setting 'status: { enabled/disabled }` lowercased. It's easier to manage when the code is either all uppercase or lowercase

lib/models/BucketInfo.js
JianqinWang: I notice that for other configurations in BucketInfo, they have capitalization like this; would it be better to keep this uniform, or is it still better to go with lowercase?

lib/models/BucketInfo.js
```diff
@@ -97,6 +100,11 @@ class BucketInfo {
         }
         if (locationConstraint) {
             assert.strictEqual(typeof locationConstraint, 'string');
+            ingestion = locationConstraint.split(zenkoSeparator)[1];
```
rahulreddy: Re-reviewing this...I don't think we need this here. Location constriant needs to be parsed in the API and the ingestionConfig you send as part of the constructor should be enough to set the bucket for ingestion enabled/disabled.

rahulreddy (changes requested):

philipyoo (approved):

rahulreddy (approved):

PR #598, opened 2018-11-01
feat: passthroughFile service

giacope:
This PR adds the `passthroughFile` "service" which is used for the implementation of the `passthrough-fs daemon`, a.k.a `pfsd`. It can be found here: https://github.com/scality/cloudserver/pull/1609

For the full design of `pfsd`, formerly "fs-data", see https://github.com/scality/Zenko-RFCs/pull/5

### DIFF
--- a/lib/network/rest/utils.js
+++ b/lib/network/rest/utils.js
@@ -1,8 +1,17 @@
 'use strict'; // eslint-disable-line
 
 const errors = require('../../errors');
+const constants = require('../../constants');
+const passthroughPrefixLength = constants.passthroughFileURL.length;
 
 module.exports.explodePath = function explodePath(path) {
+    if (path.startsWith(constants.passthroughFileURL)) {
+        const key = path.slice(passthroughPrefixLength + 1);
+        return {
+            service: constants.passthroughFileURL,
+            key: key.length > 0 ? key : undefined,
+        };
+    }
     const pathMatch = /^(\/[a-zA-Z0-9]+)(\/([0-9a-f]*))?$/.exec(path);
     if (pathMatch) {
         return {
--- a/lib/storage/data/file/DataFileStore.js
+++ b/lib/storage/data/file/DataFileStore.js
@@ -1,6 +1,7 @@
 'use strict'; // eslint-disable-line
 
 const fs = require('fs');
+const path = require('path');
 const crypto = require('crypto');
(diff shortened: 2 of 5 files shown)


giacomoguiulfo:
@bert-e wait

alexanderchan-scality (approved):
lgtm

lib/network/rest/utils.js
```diff
@@ -1,8 +1,15 @@
 'use strict'; // eslint-disable-line
 
 const errors = require('../../errors');
+const constants = require('../../constants');
 
 module.exports.explodePath = function explodePath(path) {
+    if (path.startsWith(constants.fsDataFileURL)) {
+        return {
+            service: constants.fsDataFileURL,
+            key: path.slice(12),
```
rachedbenmustapha: Why 12? A constant that is computed from the prefix length would be clearer

lib/storage/data/file/DataFileStore.js
```diff
@@ -109,6 +113,13 @@ class DataFileStore {
      *   object contents
      */
     getFilePath(key) {
+        if (this.isFs) {
+            const absolute = path.resolve(this.dataPath, key);
+            if (absolute.startsWith(path.resolve(this.dataPath))) {
+                return absolute;
+            }
+            return '';
```
rachedbenmustapha, line 123: Checks would be welcome in the callers so that they can fail with an "invalid argument"-type error instead of failing with an `ENOENT` error that would not immediately make sense to someone trying to understand in good faith (and not trying to exploit) why it's failing.

rachedbenmustapha (changes requested):
The existing `/DataFile` implementation also being filesystem-based, the naming is confusing as to what's the difference between to 2. Something that is actually different and sets the 2 implementations apart is that the new one uses passthrough paths while the other one uses hashing.

This implementation could be named `passhroughFile` and the flag `isPassthrough` and so on. Or at least something different than `fs`.

Also most of the `yarn.lock` changes seem gratuitous, were they intended for this PR?

giacope:
Thanks for the review @RachedBenMustapha. I just addressed all the things you mentioned.

rachedbenmustapha (approved):
LGTM now :) Ideally my naming comment also applies to the cloudserver PR, although I can’t review it right now

PR #601, opened 2018-11-08
ft: ZENKO-1240 ingestion bucket search in mongo

philipyoo:
Add the ability to find ingestion buckets directly from mongo.

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -1366,6 +1366,31 @@ class MongoClientInterface {
             return callback(null, retResult);
         });
     }
+
+    getIngestionBuckets(log, callback) {
+        const m = this.getCollection(METASTORE);
+        // eslint-disable-next-line
+        m.find({
+            _id: {
+                $nin: [PENSIEVE, constants.usersBucket],
+            },
+            'value.ingestion': {
+                $exists: true,
+            },
+        }, {
+            'value.name': true,
+            'value.ingestion': true,
+            'value.locationConstraint': true,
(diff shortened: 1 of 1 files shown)


philipyoo:
@bert-e wait

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1366,6 +1366,31 @@ class MongoClientInterface {
...
+        m.find({
+            _id: {
+                $nin: [PENSIEVE, constants.usersBucket],
+            },
+            'value.ingestion': {
+                $exists: true,
+            },
+        }, {
+            'value.name': true,
+            'value.ingestion': true,
```
JianqinWang, line 1382: right now there are 3 possible values for ingestion: `{ Status: 'Enabled' }`, `{ Status: 'Disabled' }`, `null`

rahulreddy (changes requested):
I would not recommend leaking application logic into abstractions. They have to independent. What you can do is you can extend `MongoClientInterface` in Backbeat directly and write your own method there.

lib/storage/metadata/mongoclient/MongoClientInterface.js
philipyoo: this arg is to indicate the return value should include the BucketInfo `ingestion` field, so unless the value is null, it should return the bucket

philipyoo:
@rahulreddy sounds good, closing and will implement an extension in backbeat

lib/storage/metadata/mongoclient/MongoClientInterface.js
JianqinWang: my bad, misread

PR #604, opened 2018-11-15
feature: ZENKO-732 Lifecycle transition policies

bennettbuchanan:
Add support for lifecycle transition policies.

### DIFF
--- a/lib/models/LifecycleConfiguration.js
+++ b/lib/models/LifecycleConfiguration.js
@@ -3,6 +3,8 @@ const UUID = require('uuid');
 
 const errors = require('../errors');
 
+const MAX_DAYS = 2147483647; // Max 32-bit signed binary integer.
+
 /**
  * Format of xml request:
 
@@ -83,10 +85,13 @@ class LifecycleConfiguration {
     /**
      * Create a Lifecycle Configuration instance
      * @param {string} xml - the parsed xml
+     * @param {object} config - the CloudServer config
      * @return {object} - LifecycleConfiguration instance
      */
-    constructor(xml) {
+    constructor(xml, config) {
         this._parsedXML = xml;
+        this._storageClasses =
+            config.replicationEndpoints.map(endpoint => endpoint.site);
         this._ruleIDs = [];
         this._tagKeys = [];
         this._config = {};
@@ -179,11 +184,6 @@ class LifecycleConfiguration {
      */
     _parseRule(rule) {
         const ruleObj = {};
-        if (rule.Transition || rule.NoncurrentVersionTransition) {
-            ruleObj.error = errors.NotImplemented.customizeDescription(
-                'Transition lifecycle action not yet implemented');
-            return ruleObj;
-        }
         // Either Prefix or Filter must be included, but can be empty string
         if ((!rule.Filter && rule.Filter !== '') &&
         (!rule.Prefix && rule.Prefix !== '')) {
@@ -255,7 +255,7 @@ class LifecycleConfiguration {
         if (Array.isArray(filter)) {
             // if Prefix was included, not Filter, filter will be Prefix array
             // if more than one Prefix is included, we ignore all but the last
-            filterObj.rulePrefix = filter.pop();
+            filterObj.rulePrefix = filter[filter.length - 1];
             return filterObj;
         }
         if (filter.And && (filter.Prefix || filter.Tag) ||
@@ -265,7 +265,7 @@ class LifecycleConfiguration {
             return filterObj;
         }
         if (filter.Prefix) {
-            filterObj.rulePrefix = filter.Prefix.pop();
+            filterObj.rulePrefix = filter.Prefix[filter.Prefix.length - 1];
             return filterObj;
         }
         if (filter.Tag) {
@@ -285,7 +285,7 @@ class LifecycleConfiguration {
                 return filterObj;
             }
             if (andF.Prefix && andF.Prefix.length >= 1) {
-                filterObj.rulePrefix = andF.Prefix.pop();
+                filterObj.rulePrefix = andF.Prefix[andF.Prefix.length - 1];
             }
             const tagObj = this._parseTags(andF.Tag);
             if (tagObj.error) {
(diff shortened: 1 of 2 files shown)


bennettbuchanan:
@bert-e wait

lib/models/LifecycleConfiguration.js
```diff
@@ -3,6 +3,8 @@ const UUID = require('uuid');
 
 const errors = require('../errors');
 
+const MAX_DAYS = 2147483647; // Max 32-bit signed binary integer.
```
jonathan-gramain, line 6: Is this limit mandated by AWS? Otherwise I think it does not have a useful meaning except if actually stored in a 32-bit integer, which does not apply in JS as they are treated as floating point.

If a limit can make sense to avoid users inadvertently entering bad values, maybe a lower one like 100000 could make more sense - rougly 300 years, ought to be enough for anybody (sic).

lib/models/LifecycleConfiguration.js
```diff
@@ -412,6 +412,252 @@ class LifecycleConfiguration {
...
+     * Checks the validity of the given field
+     * @param {object} params - Given function parameters
+     * @param {string} params.days - The value of the field to check
+     * @param {string} params.field - The field name with the value
+     * @param {string} params.ancestor - The immediate ancestor field
+     * @return {object|null} Returns an error object or `null`
+     */
+    _checkDays(params) {
+        const { days, field, ancestor } = params;
+        if (Number.parseInt(days, 10) < 0) {
```
jonathan-gramain: You may use a temporary variable to parse only once.

lib/models/LifecycleConfiguration.js
```diff
@@ -412,6 +412,252 @@ class LifecycleConfiguration {
...
+            const msg = `'${field}' in ${ancestor} action must be nonnegative`;
+            return errors.InvalidArgument.customizeDescription(msg);
+        }
+        if (Number.parseInt(days, 10) > MAX_DAYS) {
+            return errors.MalformedXML;
```
jonathan-gramain: A message can be useful here as well, and the error type should match the negative one (InvalidArgument)

lib/models/LifecycleConfiguration.js
```diff
@@ -412,6 +412,252 @@ class LifecycleConfiguration {
...
+     */
+    _checkStorageClasses(params) {
+        const { usedStorageClasses, storageClass, ancestor, prefix } = params;
+        if (!this._storageClasses.includes(storageClass)) {
+            return errors.MalformedXML;
+        }
+        if (usedStorageClasses.includes(storageClass)) {
+            const msg = `'StorageClass' must be different for '${ancestor}' ` +
+                `actions in same 'Rule' with prefix '${prefix}'`;
+            return errors.InvalidRequest.customizeDescription(msg);
```
jonathan-gramain, line 491: We have three different kind of errors returned (MalformedXML, InvalidArgument, InvalidRequest), could be worth sticking with only one to make things more consistent for the user.

I would avoid InvalidRequest as it's not necessarily triggered by a request, and MalformedXML suits best when the XML structure itself is bad.

lib/models/LifecycleConfiguration.js
```diff
@@ -412,6 +412,252 @@ class LifecycleConfiguration {
...
+     * current rule
+     * @param {string} params.prefix - The prefix of the rule
+     * @return {object|null} Returns an error object or `null`
+     */
+    _checkTimeType(params) {
+        const { usedTimeType, currentTimeType, prefix } = params;
+        if (usedTimeType && usedTimeType !== currentTimeType) {
+            const msg = "Found mixed 'Date' and 'Days' based Transition " +
+                `actions in lifecycle rule for prefix '${prefix}'`;
+            return errors.InvalidRequest.customizeDescription(msg);
```
jonathan-gramain, line 520: Same comment about homogeneizing the error code returned

lib/models/LifecycleConfiguration.js
```diff
@@ -83,10 +85,13 @@ class LifecycleConfiguration {
     /**
      * Create a Lifecycle Configuration instance
      * @param {string} xml - the parsed xml
+     * @param {object} cloudServerConfig - the CloudServer config
      * @return {object} - LifecycleConfiguration instance
      */
-    constructor(xml) {
+    constructor(xml, cloudServerConfig) {
```
jonathan-gramain: At this model level I would rather not expose cloudserver config format directly, but simply provide the list of valid storage locations as an array. That way it becomes more independent from cloudserver details, and also a little bit easier to test.

lib/models/LifecycleConfiguration.js
```diff
@@ -412,6 +412,252 @@ class LifecycleConfiguration {
...
+     */
+    _parseNoncurrentVersionTransition(rule) {
+        const result = { nonCurrentVersionTransition: [] };
+        const usedStorageClasses = [];
+        for (let i = 0; i < rule.NoncurrentVersionTransition.length; i++) {
+            const t = rule.NoncurrentVersionTransition[i]; // Transition object
+            const noncurrentDays = t.NoncurrentDays && t.NoncurrentDays[0];
+            const storageClass = t.StorageClass && t.StorageClass[0];
+            if (noncurrentDays === undefined || storageClass === undefined) {
+                return Object.assign(result, { error: errors.MalformedXML });
```
jonathan-gramain: I don't think we want to return the current list of `nonCurrentVersionTransition` if there's an error, as we will not want to apply any valid rule, so maybe simply return the second argument. May be worth adding an error message as well.

lib/models/LifecycleConfiguration.js
```diff
@@ -412,6 +412,252 @@ class LifecycleConfiguration {
...
+            }
+            let error = this._checkDays({
+                days: noncurrentDays,
+                field: 'NoncurrentDays',
+                ancestor: 'NoncurrentVersionTransition',
+            });
+            if (error) {
+                return Object.assign(result, { error });
```
jonathan-gramain: same here

lib/models/LifecycleConfiguration.js
```diff
@@ -412,6 +412,252 @@ class LifecycleConfiguration {
...
+            }
+            if (days !== undefined && date !== undefined) {
+                return Object.assign(result, { error: errors.MalformedXML });
+            }
+            if (storageClass === undefined) {
+                return Object.assign(result, { error: errors.MalformedXML });
+            }
+            let error = null;
+            if (days !== undefined) {
+                error = this._checkTimeType({
```
jonathan-gramain, line 637: I don't understand the need for this function, since above we check `if (days !== undefined && date !== undefined)`, which looks good enough to me, except we might want to add a specific error message, if I read this correctly?

lib/models/LifecycleConfiguration.js
bennettbuchanan: The limit is from AWS. I imagine it is there to prevent users from sending huge values and taking excessive memory. I'm okay with the limit so I'll tentatively mark this comment as resolved but can change if you feel strongly.

lib/models/LifecycleConfiguration.js
bennettbuchanan: I'll add a message.

lib/models/LifecycleConfiguration.js
bennettbuchanan: The request types are following what AWS returns for those particular errors. I don't feel we need to be beholden to those error types, but that's where it's coming from.

lib/models/LifecycleConfiguration.js
bennettbuchanan: This is to check that 'Days' or 'Date' (but not a combination of the two) is used consistently across the transitions. The check `if (days !== undefined && date !== undefined)` ensures there is not a combination within the transition itself.

lib/models/LifecycleConfiguration.js
```diff
@@ -253,9 +253,7 @@ class LifecycleConfiguration {
         // if no Rule Prefix or Filter, rulePrefix is empty string
         filterObj.rulePrefix = '';
         if (Array.isArray(filter)) {
-            // if Prefix was included, not Filter, filter will be Prefix array
-            // if more than one Prefix is included, we ignore all but the last
-            filterObj.rulePrefix = filter.pop();
+            filterObj.rulePrefix = filter[0];
```
dora-korpar: For removing the `.pop`s, are we going to somehow limit the number of Prefixes someone can include in the configuration? Cos the pops were originally included cos there can be a lot of prefixes, but only the last one in the array should be considered was what I found Amazon functionality to be

lib/models/LifecycleConfiguration.js
```diff
@@ -709,6 +992,40 @@ class LifecycleConfiguration {
                     Action = `<${actionName}>${Days}${Date}${DelMarker}` +
                         `</${actionName}>`;
                 }
+                if (actionName === 'NoncurrentVersionTransition') {
```
dora-korpar, line 997: kind of confusing that these conditionals don't match the format of the others, with setting each XML part to a variable. I don't have a strong preference for either format, but would it be possible/not a massive overhaul to sync transitions up with expiration format?

dora-korpar (commented):
I know the expiration rules were kinda messy, so props for working through them and cleaning them up a bit 👍

lib/models/LifecycleConfiguration.js
bennettbuchanan: I saw that you can include multiple prefixes, but that they were being filtered out by the SDK so that the XML that we receive only includes one prefix. 

I didn't try directly sending the XML though, for example with `curl` or the likes.

Do you know of another case? This can easily be updated to `filter[filter.length -1]`, if so. Mainly I just wanted to avoid altering the array so that I could use the values later on down the line. 😸

lib/models/LifecycleConfiguration.js
bennettbuchanan: I agree. I am aiming to follow the same format by setting each action to a variable. Is this what you were thinking of? https://github.com/scality/Arsenal/pull/604/files/a27abe39bfe360ba430f87bcf9226bdd62225262#diff-d71cc4c205e89f5c3fc949d566f6fd88R1007.

lib/models/LifecycleConfiguration.js
bennettbuchanan: Probably on the safer side to just use `filter[filter.length - 1]`, so I'll plan to update to that instead.

lib/models/LifecycleConfiguration.js
jonathan-gramain: OK, weird but valid point. I think we can derive from what AWS does concerning errors when it makes sense, but no need to do this here and now.

lib/models/LifecycleConfiguration.js
```diff
@@ -412,6 +412,267 @@ class LifecycleConfiguration {
...
+    _parseTransition(rule) {
+        const transition = [];
+        const usedStorageClasses = [];
+        let usedTimeType = null;
+        for (let i = 0; i < rule.Transition.length; i++) {
+            const t = rule.Transition[i]; // Transition object
+            const days = t.Days && Number.parseInt(t.Days[0], 10);
+            const date = t.Date && t.Date[0];
+            const storageClass = t.StorageClass && t.StorageClass[0];
+            if (days === undefined && date === undefined ||
```
jonathan-gramain: To disambiguate, I recommend adding explicit parenthesis around the `&&` checks.

lib/models/LifecycleConfiguration.js
```diff
@@ -709,6 +994,40 @@ class LifecycleConfiguration {
...
+                        );
+                    });
+                    Action = xml.join('');
+                }
+                if (actionName === 'Transition') {
+                    const xml = [];
+                    transition.forEach(transition => {
+                        const { days, date, storageClass } = transition;
+                        let element;
+                        if (days !== undefined || days === 0) {
```
jonathan-gramain: I think you mean `days !== undefined && days !== 0`, could also be more restrictive with `days > 0`

bennettbuchanan: It looks like the comparison to zero is actually redundant here. I'll just remove it altogether because at this point there is no need to check that the number is `>= 0` since the PUT already restricts this.

bennettbuchanan: Nice catch!

jonathan-gramain (approved):

lib/models/LifecycleConfiguration.js
dora-korpar: Woops, looks like I'm too late to look at the link, but I didn't realize there could be more than one transition per rule, so nevermind 😸

dora-korpar (approved):

bennettbuchanan:
@bert-e approve

rahulreddy (approved):
Approving to enable Bert-E

rahulreddy:
Bert-E is a 🔮 now

bennettbuchanan:
Closing and reopening to use the correct Jira issue in the branch.

bennettbuchanan:
Reopened here: https://github.com/scality/Arsenal/pull/620

PR #606, opened 2018-11-15
bf: S3C-1678-ipv6-check

dora-korpar:
Checks whether a system has IPv6 enabled before binding an IP address

### DIFF
--- a/lib/network/http/utils.js
+++ b/lib/network/http/utils.js
@@ -1,5 +1,6 @@
 'use strict'; // eslint-disable-line
 
+const os = require('os');
 const errors = require('../../errors');
 
 /**
@@ -104,6 +105,15 @@ function parseRange(rangeHeader, objectSize) {
     return getByteRangeFromSpec(rangeSpec, objectSize);
 }
 
-module.exports = { parseRangeSpec,
+function checkSupportIPv6() {
+    const niList = os.networkInterfaces();
+    return Object.keys(niList).some(network =>
+        niList[network].some(intfc => intfc.family === 'IPv6'));
+}
+
+module.exports = {
+    parseRangeSpec,
     getByteRangeFromSpec,
-    parseRange };
+    parseRange,
+    checkSupportIPv6,
+};
--- a/lib/network/http/server.js
+++ b/lib/network/http/server.js
@@ -6,6 +6,7 @@ const assert = require('assert');
 const dhparam = require('../../https/dh2048').dhparam;
 const ciphers = require('../../https/ciphers').ciphers;
 const errors = require('../../errors');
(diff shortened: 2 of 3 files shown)


dora-korpar:
@bert-e reset

dora-korpar:
@bert-e reset

alexanderchan-scality (approved):

dora-korpar:
@bert-e reset

lib/network/http/utils.js
```diff
@@ -104,6 +105,20 @@ function parseRange(rangeHeader, objectSize) {
     return getByteRangeFromSpec(rangeSpec, objectSize);
 }
 
-module.exports = { parseRangeSpec,
+function checkSupportIPv6() {
+    const niList = os.networkInterfaces();
+    let supportIPv6 = false;
+    Object.keys(niList).forEach(network => {
```
philipyoo: the `forEach` will overwrite `supportIPv6`, so only the last network will be checked. Maybe another `.some`?

lib/network/http/utils.js
```diff
@@ -104,6 +105,20 @@ function parseRange(rangeHeader, objectSize) {
     return getByteRangeFromSpec(rangeSpec, objectSize);
 }
 
-module.exports = { parseRangeSpec,
+function checkSupportIPv6() {
+    const niList = os.networkInterfaces();
```
philipyoo, line 109: does which network matter?

lib/network/http/utils.js
```diff
@@ -104,6 +105,20 @@ function parseRange(rangeHeader, objectSize) {
...
+    let supportIPv6 = false;
+    Object.keys(niList).forEach(network => {
+        supportIPv6 = niList[network].some(intfc =>
```
jonathan-gramain: It looks like you're only going to consider the latest interface because you're overwriting `supportIPv6` at each loop iteration, probably not what you wanted to do

lib/network/http/utils.js
dora-korpar: nope, the network doesn't matter

lib/network/http/utils.js
dora-korpar: that's a good point

dora-korpar: fixed 😸

lib/network/http/utils.js
dora-korpar: fixed 😸

philipyoo (approved):

dora-korpar:
@bert-e reset

jonathan-gramain (approved):

rahulreddy (approved):

dora-korpar:
@bert-e approve

PR #631, opened 2018-12-18
bugfix(ZENKO-1362): request path encoding

giacope:
This PR will make the `RESTClient` encode the path of its requests and allow the `RESTServer` to decode them.

With this change, PFSD will be able to read request properly when there are extraneous characters (like whitespaces) on the request's path, therefore allowing to read and delete files with such names.

### DIFF
--- a/lib/network/rest/RESTServer.js
+++ b/lib/network/rest/RESTServer.js
@@ -51,7 +51,7 @@ function sendError(res, log, error, optMessage) {
  */
 function parseURL(urlStr, expectKey) {
     const urlObj = url.parse(urlStr);
-    const pathInfo = utils.explodePath(urlObj.path);
+    const pathInfo = utils.explodePath(decodeURI(urlObj.path));
     if ((pathInfo.service !== constants.dataFileURL)
         && (pathInfo.service !== constants.passthroughFileURL)) {
         throw errors.InvalidAction.customizeDescription(
--- a/lib/network/rest/RESTServer.js
+++ b/lib/network/rest/RESTServer.js
@@ -7,7 +7,7 @@ const werelogs = require('werelogs');
 
 const httpServer = require('../http/server');
 const constants = require('../../constants');
-const utils = require('./utils');
+const { parseURL } = require('./utils');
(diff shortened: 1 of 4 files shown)


alexanderchan-scality (approved):

ssalaues (approved):

ploki (approved):

giacope:
@bert-e approve

lib/network/rest/RESTServer.js
```diff
@@ -51,7 +51,7 @@ function sendError(res, log, error, optMessage) {
  */
 function parseURL(urlStr, expectKey) {
```
jonathan-gramain: Would be good to have a couple unit tests for `parseURL` with special characters

giacope: @jonathan-gramain I moved `parseURL` from `RESTServer.js` to `utils.js` and added a new tests file `utils.js` with tests.

tests/unit/network/rest/utils.js
```diff
@@ -0,0 +1,26 @@
...
+const { parseURL } = require('../../../../lib/network/rest/utils');
+
+describe('parseURL function', () => {
+    [
+        {
+            inputUrl: `${constants.passthroughFileURL}/test`,
+            expectedKey: 'test',
+        },
+        {
+            inputUrl: `${constants.passthroughFileURL}/test with spaces`,
```
jonathan-gramain, line 14: Maybe you want the spaces encoded as `%20` or `+` here?

jonathan-gramain (approved):

tests/unit/network/rest/utils.js
giacope: `url.parse()` (called in parseURL) already encodes the whitespaces for us (before decoding).
In other words, both encoding and decoding occur in `parseURL`. We could still provide already encoded URLs. I'll add it as an additional test.

giacope: Done

bennettbuchanan (approved):

PR #632, opened 2018-12-20
feature: ZENKO-1399 Check transitions time gap

bennettbuchanan:
Ensure there is at least 24 hours between transition rules.

### DIFF
--- a/lib/models/LifecycleConfiguration.js
+++ b/lib/models/LifecycleConfiguration.js
@@ -548,6 +548,50 @@ class LifecycleConfiguration {
     }
 
     /**
+     * Ensure that transition rules are at least a day apart from each other.
+     * @param {object} params - Given function parameters
+     * @param {string} [params.days] - The days of the current transition
+     * @param {string} [params.date] - The date of the current transition
+     * @param {string} params.storageClass - The storage class of the current
+     * rule
+     * @param {string} params.rule - The current rule
+     * @return {undefined}
+     */
+    _checkTimeGap(params) {
+        // For testing purposes, we want to allow smaller time intervals.
+        if (process.env.CI === 'true') {
+            return undefined;
+        }
+        const { days, date, storageClass, rule } = params;
+        const invalidTransition = rule.Transition.find(transition => {
+            if (storageClass === transition.StorageClass[0]) {
+                return false;
+            }
+            if (days !== undefined) {
+                return Number.parseInt(transition.Days[0], 10) === days;
+            }
+            if (date !== undefined) {
+                const timestamp = new Date(date).getTime();
+                const compareTimestamp = new Date(transition.Date[0]).getTime();
+                const oneDay = 24 * 60 * 60 * 1000; // Milliseconds in a day.
+                return Math.abs(timestamp - compareTimestamp) < oneDay;
+            }
+            return false;
+        });
+        if (invalidTransition) {
+            const timeType = days !== undefined ? 'Days' : 'Date';
+            const filterMsg = this._getRuleFilter(rule);
+            const compareStorageClass = invalidTransition.StorageClass[0];
+            const msg = `'${timeType}' in the 'Transition' action for ` +
+                `StorageClass '${storageClass}' for ${filterMsg} must be at ` +
+                `least one day apart from ${filterMsg} in the 'Transition' ` +
+                `action for StorageClass '${compareStorageClass}'`;
+            return errors.InvalidArgument.customizeDescription(msg);
+        }
+        return undefined;
+    }
+
+    /**
      * Checks transition time type (i.e. 'Date' or 'Days') only occurs once
      * across transitions and across transitions and expiration policies
      * @param {object} params - Given function parameters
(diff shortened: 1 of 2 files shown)


alexanderchan-scality (approved):

lib/models/LifecycleConfiguration.js
```diff
@@ -547,6 +547,50 @@ class LifecycleConfiguration {
...
+                return Math.abs(timestamp - compareTimestamp) < oneDay;
+            }
+            return undefined;
+        });
+        if (invalidTransition) {
+            const timeType = days !== undefined ? 'Days' : 'Date';
+            const filterMsg = this._getRuleFilter(rule);
+            const compareStorageClass = invalidTransition.StorageClass[0];
+            const msg = `'${timeType}' in the 'Transition' action for ` +
+                `StorageClass '${storageClass}' for ${filterMsg} must be 1 ` +
```
jonathan-gramain: minor suggestion: I would rephrase to "must be different than" or "must be at least one day apart" in place of "must be 1 day more than", as the rule is symmetric to both rules and can be more than one day apart (i.e. one can be one or more days more than the other)

lib/models/LifecycleConfiguration.js
```diff
@@ -547,6 +547,50 @@ class LifecycleConfiguration {
...
+     */
+    _checkTimeGap(params) {
+        // For testing purposes, we want to allow smaller time intervals.
+        if (process.env.CI === 'true') {
+            return undefined;
+        }
+        const { days, date, storageClass, rule } = params;
+        const invalidTransition = rule.Transition.find(transition => {
+            if (storageClass === transition.StorageClass[0]) {
+                return undefined;
```
jonathan-gramain: `false` would be a more appropriate return value

lib/models/LifecycleConfiguration.js
```diff
@@ -547,6 +547,50 @@ class LifecycleConfiguration {
...
+            if (days !== undefined) {
+                return Number.parseInt(transition.Days[0], 10) === days;
+            }
+            if (date !== undefined) {
+                const timestamp = new Date(date).getTime();
+                const compareTimestamp = new Date(transition.Date[0]).getTime();
+                const oneDay = 24 * 60 * 60 * 1000; // Milliseconds in a day.
+                return Math.abs(timestamp - compareTimestamp) < oneDay;
+            }
+            return undefined;
```
jonathan-gramain: same here

tests/unit/models/LifecycleConfiguration.js
```diff
@@ -666,5 +666,135 @@ describe('LifecycleConfiguration', () => {
...
+        it('should return error when transitions have dates less than 24 ' +
+        'hours apart', () => {
+            const params = {
+                rule: {
+                    Prefix: 'prefix',
+                    Transition: [{
+                        Date: ['2019-01-01T00:00:00.000Z'],
+                        StorageClass: ['a'],
+                    }, {
+                        Date: ['2019-01-01T23:59:59.999Z'],
```
jonathan-gramain, line 794: This is *really* on the edge of being valid :wink:

bennettbuchanan: 😆 I had to draw the line somewhere.

lib/models/LifecycleConfiguration.js
```diff
@@ -563,9 +563,11 @@ class LifecycleConfiguration {
             return undefined;
         }
         const { days, date, storageClass, rule } = params;
-        const invalidTransition = rule.Transition.find(transition => {
-            if (storageClass === transition.StorageClass[0]) {
-                return undefined;
+        let compareStorageClass;
+        const invalidTransition = rule.Transition.some(transition => {
+            compareStorageClass = transition.StorageClass[0]
```
jonathan-gramain: minor: you can declare `compareStorageClass` here as a scoped `const`, making it clear it's only valid in this block.

jonathan-gramain (approved):

lib/models/LifecycleConfiguration.js
jonathan-gramain: nevermind, I did not see you use it in the error message

bennettbuchanan: I'm going to change it back to using `.find` (but keeping the boolean return values) to make this more obvious.

bennettbuchanan:
@bert-e approve

lib/models/LifecycleConfiguration.js
```diff
@@ -547,6 +547,50 @@ class LifecycleConfiguration {
...
+        if (process.env.CI === 'true') {
+            return undefined;
+        }
+        const { days, date, storageClass, rule } = params;
+        const invalidTransition = rule.Transition.find(transition => {
+            if (storageClass === transition.StorageClass[0]) {
+                return false;
+            }
+            if (days !== undefined) {
+                return Number.parseInt(transition.Days[0], 10) === days;
```
rahulreddy, line 571: `NaN` may be of a concern here.

rahulreddy (approved):

lib/models/LifecycleConfiguration.js
bennettbuchanan: Okay, I will double check if I can get the lifecycle config to take with `NaN` days. If so, there may be other places where we'd need to add some checks. Thanks for noting.

PR #633, opened 2019-01-04
feat(DataFileStore): add noCache option

giacope:
This PR:

- Adds the noCache option.
- "noCache" will now work for reads where data was already present (i.e. there were no writes before in the current lifetime of the server).


giacope:
Closing in favor of #634 because this is a mainly bugfix.

PR #634, opened 2019-01-04
bugfix(DataFileStore): add noCache option

giacope:
This PR:

- Adds the noCache option.
- "noCache" will now work for reads where data was already present (i.e. there were no writes before in the current lifetime of the server). This fixes the issue of virtual memory ballooning when reading.

### DIFF
--- a/lib/storage/data/file/DataFileStore.js
+++ b/lib/storage/data/file/DataFileStore.js
@@ -6,12 +6,12 @@ const crypto = require('crypto');
 const async = require('async');
 const diskusage = require('diskusage');
 const werelogs = require('werelogs');
-const posixFadvise = require('fcntl');
 
 const errors = require('../../../errors');
 const stringHash = require('../../../stringHash');
 const jsutil = require('../../../jsutil');
 const storageUtils = require('../../utils');
+const releasePageCacheSync = require('./utils');
 
 // The FOLDER_HASH constant refers to the number of base directories
 // used for directory hashing of stored objects.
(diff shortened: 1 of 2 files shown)


lib/storage/data/file/DataFileStore.js
```diff
@@ -45,13 +45,20 @@ class DataFileStore {
...
+     * @param {Boolean} [dataConfig.isPassthrough=false] - If true, enable
+     *   passthrough filesystem I/O
+     * @param {Boolean} [dataConfig.isReadOnly=false] - If true, disable
+     *   put and delete
      */
     constructor(dataConfig, logApi) {
         this.logger = new (logApi || werelogs).Logger('DataFileStore');
         this.dataPath = dataConfig.dataPath;
         this.noSync = dataConfig.noSync || false;
+        this.noCache = dataConfig.noCache || false;
```
jonathan-gramain, line 61: The JSDoc suggests the default value is `true`, here it seems `false` if not set.

lib/storage/data/file/DataFileStore.js
```diff
@@ -178,6 +184,18 @@ class DataFileStore {
                     return cbOnce(null, key);
                 }
                 if (this.noSync) {
+                    /*
+                     * It's is not guaranteed that the Kernel will release page
+                     * caches when this.noSync is true. If you want to ensure
+                     * this behavior, set this.noSync to false.
+                     */
+                    if (this.noCache) {
```
jonathan-gramain, line 192: Now that the code executed post-fsync becomes non-trivial I would recommend creating a helper function to avoid code duplication.

alexanderchan-scality (approved):

lib/storage/data/file/DataFileStore.js
giacope: Good catch, fixed!

lib/storage/data/file/DataFileStore.js
giacope: Added a helper function in `utils.js`.

giacope:
Closing in favor of #635 to target the right branch.

PR #635, opened 2019-01-04
bugfix(DataFileStore): add noCache option

giacope:
Reopening #634 with the right target branch.

This PR:

- Adds the noCache option.
- "noCache" will now work for reads where data was already present (i.e. there were no writes before in the current lifetime of the server). This fixes the issue of virtual memory ballooning when reading.

### DIFF
--- a/lib/storage/data/file/DataFileStore.js
+++ b/lib/storage/data/file/DataFileStore.js
@@ -5,12 +5,12 @@ const crypto = require('crypto');
 const async = require('async');
 const diskusage = require('diskusage');
 const werelogs = require('werelogs');
-const posixFadvise = require('fcntl');
 
 const errors = require('../../../errors');
 const stringHash = require('../../../stringHash');
 const jsutil = require('../../../jsutil');
 const storageUtils = require('../../utils');
+const releasePageCacheSync = require('./utils');
 
 // The FOLDER_HASH constant refers to the number of base directories
 // used for directory hashing of stored objects.
@@ -44,6 +44,8 @@ class DataFileStore {
      *   sync calls that ensure files and directories are fully
      *   written on the physical drive before returning an
      *   answer. Used to speed up unit tests, may have other uses.
+     * @param {Boolean} [dataConfig.noCache=false] - If true, attempt
+     *   to free page caches associated with the managed files
      * @param {werelogs.API} [logApi] - object providing a constructor function
      *                                for the Logger object
      */
@@ -51,6 +53,7 @@ class DataFileStore {
         this.logger = new (logApi || werelogs).Logger('DataFileStore');
         this.dataPath = dataConfig.dataPath;
         this.noSync = dataConfig.noSync || false;
+        this.noCache = dataConfig.noCache || false;
     }
 
     /**
@@ -130,7 +133,6 @@ class DataFileStore {
         log.debug('starting to write data', { method: 'put', key, filePath });
         dataStream.pause();
         fs.open(filePath, 'wx', (err, fd) => {
-            let ret = 0;
             if (err) {
                 log.error('error opening filePath',
                           { method: 'put', key, filePath, error: err });
@@ -151,6 +153,14 @@ class DataFileStore {
                     return cbOnce(null, key);
                 }
                 if (this.noSync) {
+                    /*
+                     * It's not guaranteed that the Kernel will release page
+                     * caches when this.noSync is true. If you want to ensure
+                     * this behavior, set this.noSync to false.
+                     */
+                    if (this.noCache) {
+                        releasePageCacheSync(filePath, fd, log);
+                    }
                     fs.close(fd);
                     return ok();
                 }
@@ -163,10 +173,8 @@ class DataFileStore {
                      * for the pod and can potentially cause the pod
                      * to be killed under memory pressure:
                      */
-                    ret = posixFadvise(fd, 0, size, 4);
-                    if (ret !== 0) {
-                        log.warning(
-                            `error fadv_dontneed ${filePath} returned ${ret}`);
+                    if (this.noCache) {
+                        releasePageCacheSync(filePath, fd, log);
                     }
                     fs.close(fd);
                     if (err) {
@@ -256,7 +264,7 @@ class DataFileStore {
             flags: 'r',
             encoding: null,
             fd: null,
-            autoClose: true,
+            autoClose: false,
         };
         if (byteRange) {
             readStreamOptions.start = byteRange[0];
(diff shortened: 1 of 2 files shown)


giacope:
@bert-e status

lib/storage/data/file/utils.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+function releasePageCacheSync(filePath, fd, log) {
+    function release(filePath, fd, log) {
+        const ret = posixFadvise(fd, 0, 0, 4);
+        if (ret !== 0) {
+            log.warning(
+            `error fadv_dontneed ${filePath} returned ${ret}`);
+        }
+    }
+    if (fd === null) {
```
vrancurel: does this happen ?

giacope: No, but added it as a feature for later use. It can be removed if preferred.

vrancurel: yes please remove if not needed

giacope: done!

giacope:
@bert-e reset

tmacro (approved):

vrancurel (approved):

lib/storage/data/file/DataFileStore.js
```diff
@@ -151,6 +153,14 @@ class DataFileStore {
                     return cbOnce(null, key);
                 }
                 if (this.noSync) {
+                    /*
+                     * It's is not guaranteed that the Kernel will release page
+                     * caches when this.noSync is true. If you want to ensure
+                     * this behavior, set this.noSync to false.
```
rahulreddy, line 159: You mean set `this.noCache` to true, the comment is confusing otherwise?

lib/storage/data/file/DataFileStore.js
```diff
@@ -266,18 +274,30 @@ class DataFileStore {
...
-                      }
-                      log.error('error retrieving file',
-                          { method: 'get', key, filePath,
-                              error: err });
-                      return cbOnce(
-                          errors.InternalError.customizeDescription(
-                              `filesystem read error: ${err.code}`));
-                  })
-                  .on('open', () => { cbOnce(null, rs); });
+                .on('error', err => {
```
rahulreddy, line 277: Do we need a handler for `close` event where the user abruptly closes the connection? It would be ideal to test the case to see that we are not leaking file handles since `autoClose` is now `false`.

lib/storage/data/file/DataFileStore.js
```diff
@@ -266,18 +274,30 @@ class DataFileStore {
...
+                    if (err.code === 'ENOENT') {
+                        return cbOnce(errors.ObjNotFound);
+                    }
+                    log.error('error retrieving file',
+                        { method: 'get', key, filePath,
```
rahulreddy: `method: DataFileStore.get`

lib/storage/data/file/utils.js
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Release free cached pages associated with a file
+ *
+ * @param {String} filePath - absolute path of the associated file
+ * @param {Int} fd - file descriptor of the associated file
+ * @param {werelogs.RequestLogger} log - logging object
+ * @return {undefined}
+ */
+function releasePageCacheSync(filePath, fd, log) {
+    const ret = posixFadvise(fd, 0, 0, 4);
```
rahulreddy, line 12: `posixFadvise` was previously `posixFadvise(fd, 0, size, 4);`. We don't need the set the size param?

rahulreddy (changes requested):

lib/storage/data/file/DataFileStore.js
```diff
@@ -163,10 +173,8 @@ class DataFileStore {
                      * for the pod and can potentially cause the pod
                      * to be killed under memory pressure:
                      */
-                    ret = posixFadvise(fd, 0, size, 4);
-                    if (ret !== 0) {
-                        log.warning(
-                            `error fadv_dontneed ${filePath} returned ${ret}`);
+                    if (this.noCache) {
+                        releasePageCacheSync(filePath, fd, log);
```
bennettbuchanan, line 177: Do we account for the byte `size` here as was done before?

lib/storage/data/file/DataFileStore.js
```diff
@@ -151,6 +153,14 @@ class DataFileStore {
                     return cbOnce(null, key);
                 }
                 if (this.noSync) {
+                    /*
+                     * It's is not guaranteed that the Kernel will release page
```
bennettbuchanan: ~It's is~ It's

lib/storage/data/file/DataFileStore.js
giacope: No, I do mean `this.noSync`. `posix_fadvise` doesn't guarantee to release page caches if the file hasn't been flushed to disk (synced).

lib/storage/data/file/DataFileStore.js
vrancurel: +1

jonathan-gramain (approved):

lib/storage/data/file/utils.js
jonathan-gramain: `man posix_fadvise` says that len 0 affects bytes until the end of the file, I think in this context it's good since we release the cache memory for the whole file.

lib/storage/data/file/DataFileStore.js
jonathan-gramain: see answer above

lib/storage/data/file/DataFileStore.js
giacope: When the `close` event is triggered, the file descriptor is already closed, or what do you mean exactly?

Also, I've tested with both `autoClose` set to `true` or `false`. In both cases, file descriptors remain open when connections are abruptly closed. However, the file descriptors are eventually closed for both cases (after ~2 minutes) Not sure why though...

This behavior seems to be in the nature of streams => https://github.com/nodejs/node/issues/1834

lib/storage/data/file/DataFileStore.js
giacope: Fixed

lib/storage/data/file/DataFileStore.js
giacope: Fixed

giacope:
@bert-e reset

lib/storage/data/file/DataFileStore.js
rahulreddy: Interesting, there could be a timeout emitted by the underlying socket and the fd is closed after that. Since it's not introduced by this PR I think we should follow up with a change where we close the fd on `close` event. It can be used as an attack vector otherwise.

rahulreddy (approved):

lib/storage/data/file/DataFileStore.js
giacope: Yeah. Although I was testing with a small script, and the 'close' event doesn't seem to be triggered at all (even with `autoClose: true`) in this scenario. It will probably have to be handled one layer above (http, so probably in the RestServer).

giacope:
@bert-e force_reset

giacope:
@bert-e reset

giacope:
@bert-e approve

PR #640, opened 2019-01-08
Feature/zenko 1402 move data backends

dora-korpar:
Beginning the work of moving the data backends to Arsenal.
This PR includes the external backend clients, the necessary supporting functions, and the existing unit tests for both of those.
Most files have simply been copied over (with adjusted requirements). The only code changes beyond requirements have been to `getVault.js`, which was copied over from `vault.js` (https://github.com/scality/cloudserver/blob/development/8.1/lib/auth/vault.js) in CloudServer. In order to circumvent the problem of assuming the use of the CloudServer config, the code was wrapped in a function.
The other change is that a config parameter was introduced to several client function, as they too require the CloudServer config.


dora-korpar:
@bert-e reset

JianqinWang (approved):

alexanderchan-scality (approved):

package.json
```diff
@@ -36,6 +39,7 @@
     "socket.io-client": "~2.2.0",
     "utf8": "3.0.0",
     "uuid": "^3.0.1",
+    "vaultclient": "scality/vaultclient#754b6e1",
```
vrancurel: We introduce here a circular dependency: only cloudserver shall be able to have vaultclient in its package.json. The vaultclient lib instance shall be passed from cloudserver to Arsenal (just like sproxydclient and bucketclient)

vrancurel (dismissed):

vrancurel:
where are we now ?

dora-korpar:
The vaultclient dependency was removed, and this is ready to go, there are just a few tests failing in CloudServer that I need to verify before merging this

rahulreddy (approved):

dora-korpar:
@bert-e approve

PR #641, opened 2019-01-11
feature: S3C-1925: KMIP TTLV Codec

ploki:


ploki:
@bert-e reset

ploki:
@bert-e reset

tmacro (approved):

lib/network/kmip/codec/README.md
```diff
@@ -0,0 +1,367 @@
...
+The value `value` is decoded from the payload of the KMIP field. This
+element carries the length information for varying data types.
+
+## Constructing an abstract Message
+
+```javascript
+    const msg = KMIP.Message(content);
+```
+
+The static method `KMIP.Message` instanciates an object of the class
```
alexanderchan-scality:
```suggestion
The static method `KMIP.Message` instantiates an object of the class
```

lib/network/kmip/codec/ttlv.js
```diff
@@ -0,0 +1,434 @@
...
+    }
+    return vec;
+}
+
+function _throwError(logger, msg, data) {
+    logger.error(msg, data);
+    throw Error(msg);
+}
+
+function TTLVCodec() {
```
alexanderchan-scality, line 29: curious as to why `class` isn't used instead

vrancurel:
Nice !

lib/network/kmip/codec/ttlv.js
ploki: It simplifies the code and the scoping of the various encoding/decoding structures which are kept as closures in member functions. I started implementing this using the more modern class construct as well, but in the end this version using a function as a constructor is simpler to understand and maintain.

lib/network/kmip/Message.js
```diff
@@ -0,0 +1,52 @@
...
+    return res;
+}
+
+class Message {
+    /**
+     * Construct a new abstract Message
+     * @param {Object} content - the content  of the message
+     */
+    constructor(content) {
+        if (!new.target) {
```
alexanderchan-scality: minor: this check seems to be redundant since the `new` keyword is required to call the constructor

ploki: Oh yes!

alexanderchan-scality (approved):

lib/network/kmip/Message.js
```diff
@@ -0,0 +1,48 @@
...
+
+const assert = require('assert');
+
+
+function _lookup(decodedTTLV, path) {
+    const xpath = path.split('/');
+    const obj = decodedTTLV;
+    let res = [];
+    for (let current = xpath.shift(); current; current = xpath.shift()) {
+        assert(Array.isArray(obj));
```
rahulreddy: This assertion can be moved outside the iteration

lib/network/kmip/Message.js
```diff
@@ -0,0 +1,48 @@
+'use strict'; // eslint-disable-line
+
+const assert = require('assert');
+
+
+function _lookup(decodedTTLV, path) {
```
rahulreddy, line 6: This method uses recursion. Can we unit test it to ensure that the exit condition is always met?

lib/network/kmip/README.md
```diff
@@ -0,0 +1,103 @@
...
+                KMIP.Enumeration('Query Function', 'Query Operations'),
+                KMIP.Enumeration('Query Function', 'Query Objects'),
+                KMIP.Enumeration('Query Function', 'Query Server Information'),
+                KMIP.Enumeration('Query Function', 'Query Extension Map'),
+            ]),
+        ]),
+    ])
+]);
+
+const encodedMessage = kmip.encodeMessage(logger, message);
```
rahulreddy, line 55: I know this is an example, but `logger` is not defined.

lib/network/kmip/codec/ttlv.js
rahulreddy: My preference still resides with the usage of `class` as they are more readable and easy to maintain in terms of state management. I don't want to push the pattern though, as it's a matter of preference.

rahulreddy (changes requested):
Some minor things that I noticed.

lib/network/kmip/codec/ttlv.js
ploki: Oh my preference still goes to the class construct, I just didn't succeed into articulating a digestible code with a class. IIRC mostly because of the nomenclature which is amended with extensions on a per instance basis (for the purpose of keeping the codec part simple).

lib/network/kmip/Message.js
ploki: oh that's true!

lib/network/kmip/README.md
ploki: it looks like I didn't update the doc after some coding iterations

lib/network/kmip/Message.js
ploki: :hourglass_flowing_sand: Thinking about both an iterative version and a never-exit condition for a test case (my pinkie tells the former is a simpler problem to solve)

rahulreddy: Iterative versions are always preferable over recursion 💯

ploki: Oh actually there is a bug that makes the function prematurely end if there's a leading/trailing or double slash in the search path

lib/network/kmip/README.md
ploki: addressed

rahulreddy (approved):

ploki:
@bert-e reset

ploki:
@bert-e approve

rahulreddy (approved):

PR #655, opened 2019-01-29
feature: ZENKO-1438 Add conditional MD puts

bennettbuchanan:


lib/storage/metadata/mongoclient/utils.js
```diff
@@ -27,4 +28,8 @@ function unserialize(objMD) {
     }
 }
 
-module.exports = { escape, unescape, serialize, unserialize };
+function getTag() {
```
jonathan-gramain: The semantic of `getSomething` is usually to retrieve the value of something that already exists. I would suggest `generateNewTag()` for clarity.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -419,14 +423,18 @@ class MongoClientInterface {
...
+        const update = {
             _id: objName,
+            tag: MongoUtils.getTag(),
             value: objVal,
-        }, {
-            upsert: true,
-        }, err => {
+        };
+        const options = { upsert: true };
+        if (params.condPut && params.condPut.tag) {
```
jonathan-gramain: I think we can name the param `params.cond`, so it's easier to remember if e.g we also implement conditionals for GET etc as we can use the same name.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -451,12 +459,19 @@ class MongoClientInterface {
...
+            query.tag = params.condPut.tag;
+        }
+        c.findOne(query, (err, checkObj) => {
             if (err) {
                 log.error('putObjectVerCase3: mongoDB error finding object');
                 return cb(errors.InternalError);
             }
             const objUpsert = !checkObj;
+            if (objUpsert && params.condPut.tag) {
```
jonathan-gramain: I'm a bit confused about the logic here, can you in a brief code comment explain it? What I understand is, if `findOne` does not return a document (but why `err` is null then), `objUpsert` is true (so we ask for an insert if the object does not exist) but if we asked for a tag check then we don't apply the update (but then, why not returning the `versionId` like done at line 522?)

bennettbuchanan:
Putting this on hold because https://scality.atlassian.net/browse/ZENKOIO-11 is taking precedence in the coming weeks. We can reopen should there be an immediate need for these changes.

PR #666, opened 2019-02-19
S3C-1985: Reduce memory usage of listing algorithms

ghost:
The issue discovered in the field was that an key with a heavy amount of data
could lead to an exhaustion of the available memory for Node.JS javascript
code.
In the case of the buckets handling, that would be MPU objects with an
important number of parts, listed in the "location" field.

A customer case revealed and triggered this corner case, loop-crashing the
processes responsible for the listing of the database, which were using the
Arsenal.algos.list.* extensions.

That issue is hereby fixed by this commit through the trimming of the known
heavy unused data: the "location" field of the bucket's values.
Note that as the code previously didn't need to parse the value before
forwarding it back to the caller, we are now parsing it, removing the unwanted
fields, and re-stringifying it for each selected entry in the listing.

A notable impact of this change is that the CPU usage might go up substantially
while listing the contents of any bucket.
Additionally, and for safety purposes, if the data cannot be parsed and
alterated, it will be returned as-is, at the risk of maintaining the memory
consuming behavior in the case of that corner case; while an error is logged on
stderr.

Fixes S3C-1985

### DIFF
--- a/lib/algos/list/Extension.js
+++ b/lib/algos/list/Extension.js
@@ -2,6 +2,23 @@
 
 const { FILTER_SKIP, SKIP_NONE } = require('./tools');
 
+// Use a heuristic to amortize the cost of JSON
+// serialization/deserialization only on largest metadata where the
+// potential for size reduction is high, considering the bulk of the
+// blob size is due to the "location" field containing a large number
+// of MPU parts.
+//
+// Measured on some standard metadata:
+// - 100 parts -> 9K blob
+// - 2000 parts -> 170K blob
+//
+// Using a 10K threshold should lead to a worst case of about 10M to
+// store a raw listing of 1000 entries, even with some growth
+// multiplication factor due to some internal memory duplication, it
+// should stay within reasonable memory limits.
+
+const TRIM_METADATA_MIN_BLOB_SIZE = 10000;
+
 /**
  *  Base class of listing extensions.
  */
@@ -30,22 +47,28 @@ class Extension {
      *
      * @param {Object} value - The JSON value of a listing item
      *
-     * @return {Object} The value without the filtered-out fields.
+     * @return {Object} The value that may have been trimmed of some
+     * heavy unused fields, or left untouched (depending on size
+     * heuristics)
      */
-    filterContent(value) {
+    trimMetadata(value) {
         let ret = undefined;
-        try {
-            ret = JSON.parse(value);
-            delete ret.location;
-            ret = JSON.stringify(ret);
-        } catch (e) {
-            this.logger.warn('Could not parse Object Metadata while listing',
-                             { err: e.toString() });
+        if (value.length >= TRIM_METADATA_MIN_BLOB_SIZE) {
+            try {
+                ret = JSON.parse(value);
+                delete ret.location;
+                ret = JSON.stringify(ret);
+            } catch (e) {
+                // Prefer returning an unfiltered data rather than
+                // stopping the service in case of parsing failure.
+                // The risk of this approach is a potential
+                // reproduction of MD-692, where too much memory is
+                // used by repd.
+                this.logger.warn(
+                    'Could not parse Object Metadata while listing',
+                    { err: e.toString() });
+            }
         }
-        // Prefer returning an unfiltered data rather than stopping the service
-        // in case of parsing failure.
-        // The risk of this approach is a potential reproduction of MD-692,
-        // where too much memory is used by repd.
         return ret || value;
     }
 
--- a/lib/algos/list/Extension.js
+++ b/lib/algos/list/Extension.js
@@ -24,6 +24,32 @@ class Extension {
     }
 
     /**
+     * Filters-out non-requested optional fields from the value. This function
+     * shall be applied on any value that is to be returned as part of the
+     * result of a listing extension.
+     *
+     * @param {Object} value - The JSON value of a listing item
+     *
+     * @return {Object} The value without the filtered-out fields.
+     */
+    filterContent(value) {
+        let ret = undefined;
+        try {
+            ret = JSON.parse(value);
+            delete ret.location;
+            ret = JSON.stringify(ret);
+        } catch (e) {
+            this.logger.warn('Could not parse Object Metadata while listing',
+                             { err: e.toString() });
+        }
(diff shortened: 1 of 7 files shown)


lib/algos/list/Extension.js
```diff
@@ -23,6 +23,32 @@ class Extension {
...
+    /**
+     * Filters-out non-requested optional fields from the value. This function
+     * shall be applied on any value that is to be returned as part of the
+     * result of a listing extension.
+     *
+     * @param {Object} value - The JSON value of a listing item
+     *
+     * @return {Object} The value without the filtered-out fields.
+     */
+    filterContent(value) {
```
ghost: I was hesitant about adding a code comment explaining the issue here (similarly to the commit message).
I could also add a message hinting to read the commit message associated to the addition of the function.

Please tell me what you think would be best (or do the associated edit yourself, either way :) )

ghost:
@bert-e force_reset

lib/algos/list/delimiter.js
```diff
@@ -61,8 +61,8 @@ class Delimiter extends Extension {
      *                                                   alphabetically ordered
      *                                                   or not
      */
-    constructor(parameters) {
-        super(parameters);
+    constructor(parameters, logger) {
```
ghost, line 66: As described in commit number 2, I had to add this for tests compatibility.
As the loggers are actually provided by cloudserver & metadata, there should be no issue here, and we should have the logger to log in the base Extension's `filterContent()` method.

ghost:
@bert-e force_reset

lib/algos/list/Extension.js
```diff
@@ -23,6 +23,32 @@ class Extension {
...
+        let ret = undefined;
+        try {
+            ret = JSON.parse(value);
+            delete ret.location;
+            ret = JSON.stringify(ret);
+        } catch (e) {
+            this.logger.warn('Could not parse Object Metadata while listing',
```
ghost: If deemed necessary, we could protect against an undefined logger here with a conditional construct.
Wasn't sure it'd be worth it, so just mentionning it here.

lib/algos/list/Extension.js
```diff
@@ -23,6 +23,32 @@ class Extension {
...
+     * result of a listing extension.
+     *
+     * @param {Object} value - The JSON value of a listing item
+     *
+     * @return {Object} The value without the filtered-out fields.
+     */
+    filterContent(value) {
+        let ret = undefined;
+        try {
+            ret = JSON.parse(value);
```
jonathan-gramain: As an improvement attempt to limit the CPU cost I would prefer to have a heuristic condition on the blob size before executing this, so the location field will be trimmed only if the blob is really big. Then we pay the serialize/deserialize cost only when we're certain to gain a significant number of bytes in the result (only if our assumption that the "location" field takes up the bulk of the blob bytes, which in theory should be the case).

This should amortize the cost only to memory-greedy listings and not have an impact on the vast majority of listings, considering the blob size check to be much cheaper than the serialization cycle.

lib/algos/list/Extension.js
jonathan-gramain: minor: renaming to `trimContents` would be more precise I think

jonathan-gramain (commented):
I will attempt the changes I suggested and test on my lab.

jonathan-gramain (approved):
Please someone validates my changes as well

vrancurel (approved):
Ideally we should strip down any sub-document which is larger than a treshold, e.g. in some cases user-metadata or tags can also be > 10K, and therefore we should also strip them down.

jonathan-gramain:
So we would have to extend the `delete` list to those potentially large attributes too when we're in the 10K+ case? Why not, just we would have to do the thorough analysis that it does not break any code that might need them (same kind than has been done for the location field already). Not sure it's worth it for now IMHO but open to discussion.

vrancurel:
> Not sure it's worth it for now IMHO but open to discussion.

This is why I approved the PR :)

vrancurel:
The correct way to implement this would have been acc/to me to have the listing returning only the minimal attributes needed for supporting the S3 listing (e.g. lastModified and maybe others). And do an optional params.projection API typical of document stores where consumers would typically asks for specific fields to be returned. I know it might be useful one day to be able to retrieve sub-documents while listing. I am not aware of an existing use case right now. Also we would have needed to change cloudserver+bucketclient+bucketd+repd which would have been complicated.

This is why also I approved this PR.

jonathan-gramain:
@bert-e reset

lib/algos/list/Extension.js
```diff
@@ -30,22 +47,28 @@ class Extension {
...
+        if (value.length >= TRIM_METADATA_MIN_BLOB_SIZE) {
+            try {
+                ret = JSON.parse(value);
+                delete ret.location;
+                ret = JSON.stringify(ret);
+            } catch (e) {
+                // Prefer returning an unfiltered data rather than
+                // stopping the service in case of parsing failure.
+                // The risk of this approach is a potential
+                // reproduction of MD-692, where too much memory is
```
ghost, line 65: Might need rewording into S3C-1985, though I think the link MD-692 should still be valid.

lib/algos/list/Extension.js
ghost: ok for me

lib/algos/list/Extension.js
ghost: Good point. LGTM, that was one of my worries

rahulreddy (approved):
LGTM. So we are not going ahead with the optional params idea?

ghost:
@rahulreddy Indeed, going from the principle that the field which was an actual issue was never used in CloudServer, as opposed to the suspect one we were looking at (partsLocation) when devising this idea.

ghost:
Given that both Tests on MetaData and CloudServer are green with this patch (branch bugfix/S3C-1985-runTestsWithArsenalPatch , if you're looking for it in the builds listings); I'm considering this patch as OK, given the approval of both @jonathan-gramain and @rahulreddy.

ghost:
@bert-e approve

ghost:
the PR was stuck on the tests of one integration branch which were not properly triggered.
I forced them, to unblock the situation.

PR #691, opened 2019-02-25
ZENKO-1377: update Arsenal mock for ingestion reader tests

JianqinWang:
Needed for  https://github.com/scality/backbeat/pull/578

### DIFF
--- a/lib/testing/mockRes.json
+++ b/lib/testing/mockRes.json
@@ -9,6 +9,10 @@
                 "resType": "objectList",
                 "name": "bucketList"
             },
+            "/default/bucket/bucket1": {
+                "resType": "objectList",
+                "name": "objectList1"
+            },
             "/default/attributes/bucket1": {
                 "resType": "bucketMD",
                 "name": "bucket1"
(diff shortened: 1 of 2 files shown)


rahulreddy (approved):

lib/testing/mockRes.json
```diff
@@ -335,7 +385,7 @@
                                     "READ_ACP": []
                                 },
                                 "key": "",
-                                "location": "",
+                                "location": [],
```
philipyoo, line 388: Not sure, but basing off what I'm seeing in mongo, but location seems to be set to null if none are defined. I'm not sure if an empty array is also set for this if none are defined

JianqinWang: Right now https://github.com/scality/backbeat/blob/development/8.1/lib/clients/backbeat-2017-07-01.api.json has it defined as a list, so even if a null/empty string is passed the BackbeatClient will format the value to an empty list. it's possible the backbeat client api.json file has an error, and the location value is actually a string ?

philipyoo: ah cool beans, ya you're right, if null/empty, I think should just return an empty list, which seems ok to me. The location value seems to be an array though, based on https://github.com/scality/Arsenal/blob/development/8.1/lib/models/ObjectMD.js#L548 so seems good to me 👍

philipyoo (approved):

JianqinWang:
@bert-e approve

PR #697, opened 2019-02-26
ft: ZENKO 1402 move data wrapper

dora-korpar:
Part 2/3 of moving the data backends to Arsenal. This moves the data wrapper and multiple backend gateway, including all the helper functions.


JianqinWang (approved):

lib/storage/data/LocationConstraintParser.js
```diff
@@ -0,0 +1,154 @@
+const https = require('https');
+const http = require('http');
+const url = require('url');
+const AWS = require('aws-sdk');
+const Sproxy = require('sproxydclient');
```
vrancurel, line 5: We should pass the sproxydclient lib as a parameter of DataWrapper to avoid cycling dependencies

vrancurel (dismissed):

lib/storage/data/LocationConstraintParser.js
vrancurel: See how we pass bucketclient in MetadataWrapper

vrancurel (approved):

bennettbuchanan (approved):
Just one comment about the file naming convention. Otherwise, LGTM 🚀.

index.js
```diff
@@ -118,6 +118,10 @@ module.exports = {
             },
         },
         data: {
+            DataWrapper: require('./lib/storage/data/DataWrapper'),
+            MultipleBackendGateway:
+            require('./lib/storage/data/MultipleBackendGateway'),
+            parseLC: require('./lib/storage/data/LocationConstraintParser'),
```
bennettbuchanan, line 124: The convention here seems to be using Pascal case for exporting files with classes of the same name. I'd expect LocationConstraintParser.js to have a class `LocationConstraintParser`, instead it exports a helper method `parseLC` which is a bit confusing.

dora-korpar:
@bert-e approve

index.js
dora-korpar: Yeah, I debated about that, but it would have required a lot of modification of the code both here and in CloudServer, so I thought it wouldn't be worth all the work just to maintain the naming convention.

bennettbuchanan: I guess my suggestion in that case would be to rename the file to parseLC.js, or more explicitly parseLocationConstraint.js.

PR #755, opened 2019-03-08
ft: ZENKO-1560 add getIngestionBuckets to MongoClientInterface

philipyoo:
Add `getIngestionBuckets` to MongoClientInterface and expose through MetadataWrapper

### DIFF
--- a/lib/storage/metadata/MetadataWrapper.js
+++ b/lib/storage/metadata/MetadataWrapper.js
@@ -341,6 +341,15 @@ class MetadataWrapper {
         return this.client.countItems(log, cb);
     }
 
+    getIngestionBuckets(log, cb) {
+        if (this.implName !== 'mongoclient') {
+            log.debug('error getting ingestion buckets', {
+                implName: this.implName });
+            return cb(errors.NotImplemented);
+        }
+        return this.client.getIngestionBuckets(log, cb);
+    }
+
     notifyBucketChange(cb) {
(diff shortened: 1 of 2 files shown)


alexanderchan-scality (approved):

JianqinWang (approved):

lib/storage/metadata/MetadataWrapper.js
```diff
@@ -344,6 +344,10 @@ class MetadataWrapper {
     notifyBucketChange(cb) {
         bucketNotificationHook = cb;
     }
+
```
rahulreddy: I would rather have the method you intended to use here than exporting the client. It defeats the purpose of having a wrapper.

rahulreddy (changes requested):
I would rather have the method you intended to use here than exporting the client. It defeats the purpose of having a wrapper.

philipyoo:
@rahulreddy 
The method is very specific to only ingestion within backbeat and the MetadataWrapper can create different internal clients. I don't think it would fit well on the wrapper either

The alternative was to have a second mongodb instance within backbeats lib/management; one that continues to use the existing MetadataWrapper, and the other that extends the MongoClientInterface (like how we were using it before). Basically 2 separate connections to the same mongo db

rahulreddy (approved):

philipyoo:
@bert-e approve

PR #757, opened 2019-03-12
rf: ZENKO-1566 backbeat route dataPoints by id

philipyoo:
Goes with https://github.com/scality/backbeat/pull/641

Do not pass redis keys to backbeat routes function. Instead
data points should be by common identifier (as strings) and
we can map these data points to their respective redis keys
for a given service

### DIFF
--- a/lib/backbeat/routes.js
+++ b/lib/backbeat/routes.js
@@ -4,13 +4,12 @@
 
 /**
  * The metrics route model.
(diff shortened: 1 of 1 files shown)


rahulreddy (approved):

JianqinWang (approved):

philipyoo:
@bert-e approve

PR #758, opened 2019-03-12
ft: ZENKO-1566 add ingestion metrics

philipyoo:
Goes with https://github.com/scality/backbeat/pull/642

Add ingestion metric routes to backbeat routes.
Needed to add a conditional on response objects to not include
bytes in response.

Includes ingestion: pending, throughput, completions

### DIFF
--- a/lib/backbeat/Metrics.js
+++ b/lib/backbeat/Metrics.js
@@ -130,7 +130,7 @@ class Metrics {
     }
 
     /**
-     * Get replication backlog in ops count and size in bytes
+     * Get backlog in ops count and size in bytes
      * @param {object} details - route details from lib/backbeat/routes.js
      * @param {function} cb - callback(error, data)
      * @param {array} data - optional field providing already fetched data in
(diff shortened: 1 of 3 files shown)


rahulreddy (approved):

JianqinWang (approved):

tmacro (approved):

lib/backbeat/Metrics.js
```diff
@@ -178,6 +178,8 @@ class Metrics {
      * @return {undefined}
      */
     getCompletions(details, cb, data) {
```
bennettbuchanan: I think ideally this can be broken into a scope per extension. That way we can avoid having these various conditions that may only grow in time—for example, we would would be following a similar pattern when introducing lifecycle metrics. I think, at least, the JSDoc could be updated to reflect that it is no longer serving just replicated stats.

philipyoo: Sounds good, so create separate helper methods to handle each extension response case?

philipyoo: I see, like separate entry methods for each extension, and then they can use the same helper methods for finding relevant data? I think this approach would be more straightforward for sure and we can add additional routes in a more clear and cleaner fashion.

I know my post review changes don't address this, but for now, it's more readable than current implementation.

bennettbuchanan: 👍 Thanks for that!

bennettbuchanan (approved):

philipyoo:
@bert-e approve

PR #766, opened 2019-03-25
ft(ZENKO-1616): Update ObjectMD model for azure blob api

tmacro:
Add fields required for Azure Blob API support

### DIFF
--- a/lib/models/ObjectMD.js
+++ b/lib/models/ObjectMD.js
@@ -28,9 +28,14 @@ class ObjectMD {
             } else {
                 this._updateFromParsedJSON(objMd);
             }
+            if (!this._data['creation-time']) {
+                this.setCreationTime(this.getLastModified());
+            }
         } else {
             // set newly-created object md modified time to current time
-            this._data['last-modified'] = new Date().toJSON();
+            const dt = new Date().toJSON();
+            this.setLastModified(dt);
+            this.setCreationTime(dt);
         }
         // set latest md model version now that we ensured
         // backward-compat conversion
@@ -85,6 +90,8 @@ class ObjectMD {
             'content-length': 0,
             'content-type': '',
             'content-md5': '',
+            'content-language': '',
+            'creation-time': undefined,
             // simple/no version. will expand once object versioning is
             // introduced
             'x-amz-version-id': 'null',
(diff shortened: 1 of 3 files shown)


nicolas2bert (approved):

lib/models/ObjectMD.js
```diff
@@ -352,6 +356,46 @@ class ObjectMD {
...
+        this._data['creation-time'] = creationTime;
+        return this;
+    }
+
+    /**
+     * Returns Creation Date
+     *
+     * @return {string} Creation Date
+     */
+    getCreationTime() {
```
jonathan-gramain, line 398: Is there any expectation regarding compatibility with "old" metadata? I.e. do we need e.g. to fallback to `last-modified` if `creation-time` does not exist in the loaded metadata, or is it fine to return `undefined` in such case?

tmacro: Good question, I don't think this has been discussed

JianqinWang: In Mystique (R.I.P) we had `creation-time` fall back to `last-modified`

tmacro: I've added logic to fallback to `last-modified` if `creation-time` is not set

JianqinWang (approved):

tmacro:
@bert-e approve

lib/models/ObjectMD.js
```diff
@@ -910,12 +961,18 @@ class ObjectMD {
...
+     *                         (defaults to x-amz-meta- for backwards
+     *                          compatablility)
      * @return {ObjectMD} itself
      */
     setUserMetadata(metaHeaders) {
         Object.keys(metaHeaders).forEach(key => {
             if (key.startsWith('x-amz-meta-')) {
                 this._data[key] = metaHeaders[key];
+            } else if (key.startsWith('x-ms-meta-')) {
+                const _key = key.replace('x-ms-meta-', 'x-amz-meta-');
```
ploki, line 971: ~is this a rogue header ? ```x-ms-meta-x-ms-meta-plop```~
only the first occurrence is replaced. and the startsWidth() makes sure of this

lib/models/ObjectMD.js
```diff
@@ -910,12 +961,18 @@ class ObjectMD {
      * Set custom meta headers
      *
      * @param {object} metaHeaders - Meta headers
+     * @param {sting} prefix - Prefix used to match meta headers
```
ploki: prefix no longer a parameter

lib/models/ObjectMD.js
```diff
@@ -28,9 +28,14 @@ class ObjectMD {
             } else {
                 this._updateFromParsedJSON(objMd);
             }
+            if (!this.getCreationTime()) {
+                this.setCreationTime(this.getLastModified());
```
ploki, line 32: this block is never executed

ploki (changes requested):

lib/models/ObjectMD.js
```diff
@@ -352,6 +359,50 @@ class ObjectMD {
...
+    }
+
+    /**
+     * Returns Creation Date
+     *
+     * @return {string} Creation Date
+     */
+    getCreationTime() {
+        // If creation-time is not set fallback to LastModified
+        if (!this._data['creation-time']) {
```
ploki, line 400: From my understanding of the constructor code, this condition is never met, or if it is, last-modified is not defined either. Am I wrong?

ploki (approved):

lib/models/ObjectMD.js
```diff
@@ -28,9 +28,14 @@ class ObjectMD {
...
                 this._updateFromParsedJSON(objMd);
             }
+            if (!this._data['creation-time']) {
+                this.setCreationTime(this.getLastModified());
+            }
         } else {
             // set newly-created object md modified time to current time
-            this._data['last-modified'] = new Date().toJSON();
+            const dt = new Date().toJSON();
+            this._data['last-modified'] = dt;
```
nicolas2bert: Is there anything that prevents us from using the setters?

rahulreddy (changes requested):
Please increment Object Model version and document the attributes here https://github.com/scality/Arsenal/blob/development/8.1/lib/constants.js#L37

nicolas2bert (approved):

rahulreddy (approved):

PR #768, opened 2019-03-27
bugfix: ZENKO-1583 Return data in head operation

bennettbuchanan:
Return `data` from head requests to public clouds for access to the response data.

### DIFF
--- a/lib/storage/data/external/AwsClient.js
+++ b/lib/storage/data/external/AwsClient.js
@@ -146,7 +146,7 @@ class AwsClient {
             Bucket: this._awsBucketName,
             Key: key,
             VersionId: dataStoreVersionId,
-        }, err => {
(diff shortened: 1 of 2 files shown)


philipyoo (approved):

dora-korpar:
Is that data going to be used by Backbeat?

rahulreddy (approved):

dora-korpar (approved):

jonathan-gramain (approved):
The change makes sense to me, I'm just a little bit concerned if calls to `head` are included in `waterfall` constructs, as the change would break them by shifting the argument list of the next callback, worth checking where it's called to make sure it's harmless.

tmacro (approved):

bennettbuchanan:
@bert-e approve

PR #769, opened 2019-04-01
bugfix: ZENKO-1681 remove deprecation warnings

rahulreddy:
Removes warning "node:24) [DEP0013] DeprecationWarning: Calling an asynchronous
 function without callback is deprecated."

### DIFF
--- a/lib/storage/data/file/DataFileStore.js
+++ b/lib/storage/data/file/DataFileStore.js
@@ -193,7 +193,17 @@ class DataFileStore {
                     if (this.noCache) {
                         releasePageCacheSync(filePath, fd, log);
                     }
-                    fs.close(fd);
+                    const reqUid = log.getSerializedUids();
+                    fs.close(fd, err => {
+                        if (err) {
+                            this.logger.error('error closing fd after write', {
+                                errorMsg: err.message,
+                                errStack: err.stack,
+                                requestId: reqUid,
+                                key,
+                            });
+                        }
+                    });
                     return ok();
                 }
                 fs.fsync(fd, err => {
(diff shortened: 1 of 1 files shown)


JianqinWang (approved):

lib/storage/data/file/DataFileStore.js
```diff
@@ -193,7 +193,7 @@ class DataFileStore {
                     if (this.noCache) {
                         releasePageCacheSync(filePath, fd, log);
                     }
-                    fs.close(fd);
+                    fs.closeSync(fd);
```
jonathan-gramain: In order not to change the current behavior, I would rather add a dummy callback to the call, not to block the process while the close is happening.

anurag4DSB (approved):

lib/storage/data/file/DataFileStore.js
rahulreddy: I thought that was your intention `to block the process when the close is happening`

lib/storage/data/file/DataFileStore.js
```diff
@@ -193,7 +193,17 @@ class DataFileStore {
                     if (this.noCache) {
                         releasePageCacheSync(filePath, fd, log);
                     }
-                    fs.close(fd);
+                    const reqUid = log.getSerializedUids();
+                    fs.close(fd, err => {
+                        if (err) {
+                            this.logger.error('error closing fd after write', {
```
jonathan-gramain, line 199: Don't you want to do `log.error`?

rahulreddy: Since we are not waiting for callback there will be a chance when end logger is called on `log` and if this call happens after that then an exception will be thrown. So, I am using the module logger for safety and logging the context.

rahulreddy:
@bert-e approve

PR #771, opened 2019-04-04
ft: ZENKO-1585 editable replication group id

philipyoo:
On setup of MongoClientInterface instance, create and save a new replication group id if `infostore` does not already have a stored replication group id. We update the replication group id instance variable in MongoClientInterface. New versions will now use this updated replication group id.

During the initManagement phase for cloudserver and backbeat, we just fetch the replication group id from mongo using the MongoClientInterface and patch configurations.

A potential problem is that during the initManagement phase, if replication group id has not yet been stored in `infostore`, we will be forced to retry initManagement.


philipyoo:
closing - alternative fix in progress

PR #774, opened 2019-04-10
bf: ZENKO-1718 ingestion mongo putObjectVerCase4

philipyoo:
Add a putObjectVerCase to MongoClientInterface for ingestion
use-cases. Remove management of master versions in the
ingestion process and rely on the natural ordering of
objects stored in mongo. Get and Delete operations will
rely on internal MongoClientInterface methods for performing
relevant operations. To do this, we set PHD on master for
each object version ingested.

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -509,6 +509,53 @@ class MongoClientInterface {
     }
 
     /*
+     * In this case the caller provides a versionId and a flag called usePHD.
+     * This function will save the versionId object as is and will set PHD on
+     * master. If a user performs any operation to find IsLatest, the PHD flag
+     * indicates to perform a search on the bucket to find the latest version.
+     * We rely on the natural ordering of versions for listings. We rely on
+     * internal MongoClientInterface operations for master version and repairs.
+     */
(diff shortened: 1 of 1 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -539,9 +588,12 @@ class MongoClientInterface {
         } else if (params && params.versionId === '') {
             return this.putObjectVerCase2(c, bucketName, objName, objVal,
                                           params, log, cb);
-        } else if (params && params.versionId) {
+        } else if (params && params.versionId && !params.usePHD) {
```
bennettbuchanan, line 589: I think it's worth spelling out the param as `usePlaceholder`.

bennettbuchanan (approved):
Will be good to follow with a functional test in CloudServer. LGTM ✅

lib/storage/metadata/mongoclient/MongoClientInterface.js
rahulreddy: It's actually used as PHD throughout our versioning logic. I think its ok.

rahulreddy (approved):

philipyoo:
@bert-e approve

PR #775, opened 2019-04-13
bf: ZENKO 1512 metastore circular json

dora-korpar:
Implements a serializer function for JSON.stringify to process circular JSON, based on this npm package: https://github.com/moll/json-stringify-safe/blob/master/stringify.js

### DIFF
--- a/lib/storage/utils.js
+++ b/lib/storage/utils.js
@@ -32,7 +32,7 @@ function trySetDirSyncFlag(path) {
 
 let loggedWarning = false;
 
-module.exports.setDirSyncFlag = function setDirSyncFlag(path, logger) {
+function setDirSyncFlag(path, logger) {
     const warning =
               'WARNING: Synchronization directory updates are not ' +
               'supported on this platform. Newly written data could ' +
@@ -60,4 +60,45 @@ module.exports.setDirSyncFlag = function setDirSyncFlag(path, logger) {
         }
         loggedWarning = true;
     }
+}
+
+// custom serializer function for JSON stringify to avoid circular JSON error
+// this code is a modified version of json-stringify-safe npm package found at
+// https://github.com/moll/json-stringify-safe/blob/master/stringify.js
+function serializeCircJSON() {
+    const stack = [];
+    const keys = [];
+
+    function cycleReplacer(value) {
+        if (stack[0] === value) {
+            return '[Circular ~]';
+        }
+        return '[Circular ~. ' +
+            `${keys.slice(0, stack.indexOf(value)).join('.')}]`;
+    }
+
+    return function parseJSON(key, value) {
+        if (stack.length > 0) {
+            const thisPos = stack.indexOf(this);
+            if (~thisPos) {
+                stack.splice(thisPos + 1);
+                keys.splice(thisPos, Infinity, key);
+            } else {
+                stack.push(this);
+                keys.push(key);
+            }
+            if (~stack.indexOf(value)) {
+                // eslint-disable-next-line no-param-reassign
+                value = cycleReplacer.call(this, key, value);
+            }
+        } else {
+            stack.push(value);
+        }
+        return value;
+    };
+}
+
+module.exports = {
+    setDirSyncFlag,
+    serializeCircJSON,
 };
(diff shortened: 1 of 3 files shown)


rahulreddy (changes requested):
It's good to do attribution for code that is copied over. You can use this https://meta.stackoverflow.com/questions/253618/if-i-use-so-code-on-my-website-how-should-i-give-attribution for reference.

dora-korpar:
@rahulreddy I added a comment referencing the package with a link to the original

lib/storage/utils.js
```diff
@@ -61,3 +61,39 @@ module.exports.setDirSyncFlag = function setDirSyncFlag(path, logger) {
         loggedWarning = true;
     }
 };
+
+// custom serializer function for JSON stringify to avoid circular JSON error
+// this code is a modified version of json-stringify-safe npm package found at
+// https://github.com/moll/json-stringify-safe/blob/master/stringify.js
+module.exports.serializeCircJSON = function serializeCircJSON() {
```
rahulreddy: One more thing I noticed is that you have multiple module.exports floating in the file. This should have been caught by the linter but I guess not.
```js
function foo() {
 ........
}

function bar() {
.......
}
module.exports = {
foo, bar
}
```
This is the pattern we should be using for exports.

tests/unit/storage/utils.js
```diff
@@ -0,0 +1,24 @@
...
+
+const circularJSON = {
+    key1: 'foo',
+    key2: 1,
+    key3: { key4: 5 },
+};
+circularJSON.key5 = circularJSON;
+
+describe('SerializeCircJSON', () => {
+    it('should stringify circular JSON', () => {
```
rahulreddy, line 11: Can you add a test stringifying a regular json and a test for invalid json?

rahulreddy (changes requested):
Sorry I have two more comments

tests/unit/storage/utils.js
dora-korpar: Added

lib/storage/utils.js
dora-korpar: Fixed

tests/unit/storage/utils.js
```diff
@@ -0,0 +1,48 @@
+const assert = require('assert');
+const { serializeCircJSON } = require('../../../lib/storage/utils');
+
+const jsonObj = {
+    key1: 'foo',
+    key2: 1,
+    key3: { key4: 5 },
+};
+
+describe.only('SerializeCircJSON', () => {
```
philipyoo: ~.only~

philipyoo (approved):
what sorcery is this? 👀

tests/unit/storage/utils.js
```diff
@@ -0,0 +1,48 @@
...
+            assert.deepStrictEqual(parsedJSON.key3, { key4: 5 });
+        } catch (e) {
+            assert.ifError(e);
+        }
+    });
+
+    it('should fail to stringify invalid JSON', () => {
+        try {
+            //eslint-disable-next-line
+            const invalidJSON = { in$valid: value };
```
jonathan-gramain: Why is this object invalid or non convertible to JSON?

tests/unit/storage/utils.js
```diff
@@ -0,0 +1,48 @@
+const assert = require('assert');
+const { serializeCircJSON } = require('../../../lib/storage/utils');
+
+const jsonObj = {
```
jonathan-gramain: No big deal, but the naming can be confusing as this is a javascript object, not a JSON object.

jonathan-gramain (commented):
I'm lacking context on why we have to handle circular references in the first place, as we should deal with object metadata which I would think should be simple flat JS objects. (anyway it's on mem backend so rather low risk but just wondering).

tests/unit/storage/utils.js
dora-korpar: It's invalid because 'value' isn't defined anywhere

dora-korpar:
As to the context, I wasn't able to actually reproduce the error in any testing, and the stack traces from the crashed Cloudservers weren't very forthcoming. It was decided it would be better to just catch the circular json than spend time trying to find the cause since it is just in the mem backend.

rahulreddy (approved):

dora-korpar:
@bert-e approve

PR #779, opened 2019-04-24
ZENKO-1755 HD-97 add support for hdclient

benoit-a:
* import module hdclient
* add support for hdclient in parseLC

Part of support for HDController in Zenko

### DIFF
--- a/lib/storage/data/LocationConstraintParser.js
+++ b/lib/storage/data/LocationConstraintParser.js
@@ -3,6 +3,7 @@ const http = require('http');
 const url = require('url');
 const AWS = require('aws-sdk');
 const Sproxy = require('sproxydclient');
+const Hyperdrive = require('hdclient');
 const HttpsProxyAgent = require('https-proxy-agent');
 
 const constants = require('../../constants');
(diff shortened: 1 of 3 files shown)


rahulreddy (approved):

lib/storage/data/LocationConstraintParser.js
```diff
@@ -25,21 +26,26 @@ function parseLC(config, vault) {
...
+                    // Might be undefined which is ok since there is a default
+                    // set in sproxydclient if chordCos is undefined
+                    chordCos: locationObj.details.connector.sproxyd.chordCos,
+                    // Might also be undefined, but there is a default path set
+                    // in sproxydclient as well
+                    path: locationObj.details.connector.sproxyd.path,
+                    // enable immutable optim for all objects
+                    immutable: true,
+                });
+                clients[location].clientType = 'scality';
```
dora-korpar, line 43: Minor, but this line could be pulled out since it's included in both Sproxyd and Hyperdrive conditionals

JianqinWang (approved):

rachedbenmustapha (approved):

dora-korpar (approved):

lib/storage/data/LocationConstraintParser.js
rahulreddy: May be we should introduce `hyperdrive` as a type. I think back in the day we used `scality` as we didn't envision using any other connector in Scality.
This effort can be on Cloudserver team to make things easier.

bennettbuchanan (approved):

benoit-a:
@bert-e approve

PR #784, opened 2019-05-07
ft(ZENKO-1652): Update MetadataWrapper for List Blobs

tmacro:
Add the ability to pass a custom parser for objects to MetadataWrapper. This is needed as the current parser removes any properties not needed by the S3 response, some of which are required for a Blobserver response. This will allow us to replace the parser for Blobserver and fallback to the existing parser for clients that do not specify a custom parser.


philipyoo (approved):

rahulreddy (approved):

tmacro:
@bert-e approve

PR #790, opened 2019-05-10
bf: ZENKO-1718 putObjectCase4 handle delete marker

philipyoo:
MongoClientInterface.putObjectVerCase4 will save master
as placeholders (isPHD).

The listing algorithm (delimiterMaster) used for listObjects
calls has a filter for delete markers. Since we can have
an isPHD master that is a delete marker, we should mark
them accordingly in its value.


philipyoo:
Closing this for now.
Depending on the resolution of jira#1803, if the delimiterMaster is correct as is, I will need to find a workaround, similar to the change here.

PR #802, opened 2019-05-22
bugfix: S3C-2172 change error message for compatibility

rahulreddy:
When a delete bucket request is sent with an invalid bucket name
the server returns NoSuchBucket instead of InvalidBucketName error
to be compatible with AWS S3.
Note: It will be followed by a functional test in Cloudserver


JianqinWang (approved):

rahulreddy:
@bert-e approve

alexanderchan-scality (approved):

PR #810, opened 2019-06-12
bugfix: S3C-2243 fix check for location type

rahulreddy:
This fixes the check where the logic should be looking at the type
of location instead of the name to leverage batch delete. It also fixes
the format sent to the sproxydclient which expects and object with keys
as an attribute whose value is an array of sproxyd keys.

### DIFF
--- a/lib/storage/data/DataWrapper.js
+++ b/lib/storage/data/DataWrapper.js
@@ -182,19 +182,24 @@ class DataWrapper {
             method: 'batchDelete',
         });
         const keys = [];
+        let backendName = '';
         const shouldBatchDelete = locations.every(l => {
+            // legacy sproxyd location, should fallback to using regular delete
             if (typeof l === 'string') {
-                keys.push(l);
-                return true;
+                return false;
             }
-            if (l.dataStoreName === 'sproxyd') {
-                keys.push(l.key);
+            const { dataStoreName, key } = l;
+            backendName = dataStoreName;
+            const type = this.config.getLocationConstraintType(dataStoreName);
+            // filter out possible `null` created by NFS
+            if (key && type === 'scality') {
+                keys.push(key);
                 return true;
             }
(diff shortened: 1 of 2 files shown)


anurag4DSB (approved):

lib/storage/data/DataWrapper.js
```diff
@@ -182,19 +182,24 @@ class DataWrapper {
             method: 'batchDelete',
         });
         const keys = [];
+        let backendName = '';
         const shouldBatchDelete = locations.every(l => {
+            // legacy sproxyd location, should fallback to using regular delete
             if (typeof l === 'string') {
                 keys.push(l);
```
philipyoo: minor: don't need the push

philipyoo (approved):

lib/storage/data/DataWrapper.js
```diff
@@ -182,19 +182,24 @@ class DataWrapper {
...
-                return true;
+                return false;
             }
-            if (l.dataStoreName === 'sproxyd') {
-                keys.push(l.key);
+            const { dataStoreName, key } = l;
+            backendName = dataStoreName;
+            const type = this.config.getLocationConstraintType(dataStoreName);
+            if (type === 'scality') {
```
alexanderchan-scality: comparing against the cloudserver changes, wouldn't this check require the `key === null` filtering as well?

rahulreddy: Ah I missed it, as I was manually porting the changes. Thanks for pointing it out.

lib/storage/data/DataWrapper.js
rahulreddy: That's true.

alexanderchan-scality (approved):

rahulreddy:
@bert-e approve

PR #812, opened 2019-06-13
OB-1840 Secure MongoDb access

nicolas2bert:
Support mongo username/password in CloudServer as environment variables

Related to: https://github.com/scality/cloudserver/pull/1942

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -87,9 +87,19 @@ function generatePHDVersion(versionId) {
 class MongoClientInterface {
     constructor(params) {
         const { replicaSetHosts, writeConcern, replicaSet, readPreference, path,
-            database, logger, replicationGroupId, config } = params;
-        this.mongoUrl = `mongodb://${replicaSetHosts}/?w=${writeConcern}&` +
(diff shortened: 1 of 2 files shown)


alexanderchan-scality (approved):

ssalaues (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -87,9 +87,17 @@ function generatePHDVersion(versionId) {
...
-            database, logger, replicationGroupId, config } = params;
-        this.mongoUrl = `mongodb://${replicaSetHosts}/?w=${writeConcern}&` +
-            `replicaSet=${replicaSet}&readPreference=${readPreference}`;
+        database, logger, replicationGroupId, authCredentials, config }
+        = params;
+        let cred = '';
+        if (authCredentials &&
+            authCredentials.username &&
+            authCredentials.password) {
+            cred = `${authCredentials.username}:${authCredentials.password}@`;
```
bennettbuchanan: Have we validated and encoded previously for the username/password containing a `@` character or other such special characters? Otherwise it sounds like we'd want to encode that here using [percent encoding](https://tools.ietf.org/html/rfc3986#section-2.1).

bennettbuchanan (approved):

nicolas2bert:
@bert-e approve

PR #813, opened 2019-06-14
bf: ZENKO-1736 count items scan rework

philipyoo:
Count items scan, called by cloudserver reportHandler,
returns metrics by aggregate counts of mongo objects
and buckets. The scan is triggered each hour, but
holds the request hostage.

The change here is to separate the aggregate scan
from the countItems call made by reportHandler.
The scan will instead be called by a kubernetes
cronjob. Results of the scan will be saved in infostore.

Bucket info and bucket count will be collected every time
still and this should not take too long.

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -34,8 +34,8 @@ const METASTORE = '__metastore';
 const INFOSTORE = '__infostore';
 const __UUID = 'uuid';
 const PENSIEVE = 'PENSIEVE';
+const __COUNT_ITEMS = 'countitems';
 const ASYNC_REPAIR_TIMEOUT = 15000;
-const itemScanRefreshDelay = 1000 * 60 * 60; // 1 hour
 const CONNECT_TIMEOUT_MS = 5000;
 
 const initialInstanceID = process.env.INITIAL_INSTANCE_ID;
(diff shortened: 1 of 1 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1035,11 +1034,144 @@ class MongoClientInterface {
...
+                results.bucketList = retBucketInfos;
+                results.buckets = bucketCount;
+                /* eslint-enable */
+                return cb(null, results);
+            });
+        });
+    }
 
-        const res = {
+    scanItemCount(log, cb) {
```
alexanderchan-scality, line 1181: minor: since the scanning is moved to a kubernetes job, `scanItemsCount` and the `this.config` usage can be removed. (can be a second PR)

alexanderchan-scality (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
philipyoo: I kept the scan function in the class and the s3util script will just require the class and invoke this function. I decided to keep the logic in this class to utilize available methods and vars

Ty for the review, after talking with Rahul, to lower changeset, going to leave any removals in later prs

rahulreddy (approved):

philipyoo:
@bert-e approve

PR #820, opened 2019-06-28
fix the design of the putObjectVerCase4

vrancurel:
- Repair master instead of creating PHD.
  - Note that backbeat has to be modified to use
  params.repairMaster instead of params.usePHD.

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -522,49 +522,65 @@ class MongoClientInterface {
     }
 
     /*
-     * In this case the caller provides a versionId and a flag called usePHD.
-     * This function will save the versionId object as is and will set PHD on
-     * master. If a user performs any operation to find IsLatest, the PHD flag
-     * indicates to perform a search on the bucket to find the latest version.
-     * We rely on the natural ordering of versions for listings. We rely on
-     * internal MongoClientInterface operations for master version and repairs.
+     * In this case the caller provides a versionId. We assume that
+     * objVal already contains the destination versionId. We first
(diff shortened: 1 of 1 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -522,49 +522,65 @@ class MongoClientInterface {
...
+                }, err => {
+                    if (err) {
+                        // we accept that the update fails if
+                        // condition is not met, meaning that a more
+                        // recent master was already in place
+                        if (err.code !== 11000) {
+                            log.error(
+                                'putObjectVerCase4: error upserting master',
+                                { error: err.message });
+                            return cb(errors.InternalError);
```
rahulreddy, line 576: Do we want to return `InternalError` in this case because technically it's a known situation not an actual error?

rahulreddy (changes requested):

lib/storage/metadata/mongoclient/MongoClientInterface.js
vrancurel: This update does not return an error if the object is replaced (even if the content is not changed).
The master update may return an error if the condition is not met, hence the check for error 11000.

vrancurel: I mean the InternalError will catch other errors than an actual overwrite (e.g. I/O errors).

rahulreddy: Ok.

rahulreddy (approved):

JianqinWang (approved):

vrancurel:
@bert-e approve

PR #821, opened 2019-07-02
ft: S3C 2276 bucket policy schema

dora-korpar:
Create the bucket policy schema


rahulreddy (changes requested):
Please add unit tests. You can look at the ones that were written for IAM policy.

PR #835, opened 2019-07-25
ft: add awsClient meta-mdonly put

ssalaues:
Allow puts with the header `x-amz-meta-mdonly` to bypass an actual AwsClient put resulting in a purely metadata operation.


ssalaues:
wrong branch name

PR #836, opened 2019-07-25
ft: deprecate client mdonly check

ssalaues:
Deprecates the `x-amz-meta-mdonly` header check in the client since this is now handled by a higher order function here:
https://github.com/scality/cloudserver/pull/2025

### DIFF
--- a/lib/storage/data/external/PfsClient.js
+++ b/lib/storage/data/external/PfsClient.js
@@ -36,8 +36,7 @@ class PfsClient {
 
     put(stream, size, keyContext, reqUids, callback) {
         const log = createLogger(reqUids);
-        if (keyContext.metaHeaders['x-amz-meta-mdonly'] === 'true'
-                || keyContext.isDeleteMarker === true) {
+        if (keyContext.isDeleteMarker === true) {
             const b64 = keyContext.metaHeaders['x-amz-meta-md5chksum'];
             let md5 = null;
             if (b64) {


tmacro (approved):

dora-korpar (approved):

anurag4DSB (approved):

lib/storage/data/DataWrapper.js
```diff
@@ -918,7 +920,23 @@ class DataWrapper {
...
         }
+        if (keyContext.metaHeaders['x-amz-meta-mdonly'] === 'true') {
+            const b64 = keyContext.metaHeaders['x-amz-meta-md5chksum'];
+            let md5 = null;
+            if (b64) {
+                md5 = new Buffer(b64, 'base64').toString('hex');
+            }
+            const dataRetrievalInfo = {
+                key: keyContext.objectKey,
+                dataStoreName: this.client.clients[location],
```
alexanderchan-scality:
```suggestion
                dataStoreName: location,
```

lib/storage/data/DataWrapper.js
```diff
@@ -918,7 +920,24 @@ class DataWrapper {
...
+            keyContext.metaHeaders['x-amz-meta-mdonly'] === 'true') {
+            const b64 = keyContext.metaHeaders['x-amz-meta-md5chksum'];
+            let md5 = null;
+            if (b64) {
+                md5 = new Buffer(b64, 'base64').toString('hex');
+            }
+            const dataRetrievalInfo = {
+                key: keyContext.objectKey,
+                dataStoreName: this.client.clients[location],
+                dataStoreType: this.client.clients[location].clientType,
```
alexanderchan-scality: will need to perform a check to make sure `this.client.client` exists.
`file / mem` clients do not have the `client.client` map

lib/storage/data/DataWrapper.js
ssalaues: Oops yeah you're right

lib/storage/data/DataWrapper.js
ssalaues: Will do, thank you!

ssalaues:
Closing this PR as it is no longer needed. The PFSclient will still need the header check because of how the 0 bytes are handled in the higher function.
scality/cloudserver#2025

PR #840, opened 2019-07-26
improvement: increase the limit of num. of allowed tags

rahulreddy:
This increases the limit of number of allowed tags on an object
from 10 to 50. This is to be inline and retain compatibility with
AWS S3.


JianqinWang (approved):

jonathan-gramain (approved):

rahulreddy:
@bert-e approve

PR #864, opened 2019-08-01
bf: deserialize dots and dollars from oplog

vrancurel:
To allow dots and dollars in tags we serialize them into a unicode
version. We need to properly deserialize them when reading the oplog.

This is a far-from-ideal way of serializing those special characters in MongoDB, especially another way of serializing the tags to strings has been discussed in ZENKO-1928 but we need to be backward compatible with existing deployments.

### DIFF
--- a/lib/storage/metadata/mongoclient/ListRecordStream.js
+++ b/lib/storage/metadata/mongoclient/ListRecordStream.js
@@ -64,6 +64,12 @@ class ListRecordStream extends stream.Readable {
         // always update to most recent uniqID
         this._lastConsumedID = itemObj.h.toString();
 
+        const decode = tags => (
+            JSON.parse(JSON.stringify(tags).
+                       replace(/\uFF0E/g, '.').
+                       replace(/\uFF04/g, '$'))
+        );
+
         // only push to stream unpublished objects
         if (!this._lastSavedID) {
             // process from the first entry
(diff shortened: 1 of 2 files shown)


vrancurel:
Please review carefully as it is impacting the core log reading mechanism !

lib/storage/metadata/mongoclient/ListRecordStream.js
```diff
@@ -91,23 +91,36 @@ class ListRecordStream extends stream.Readable {
         let entry;
         if (itemObj.op === 'i' &&
             itemObj.o && itemObj.o._id) {
+            const value = itemObj.o.value;
+            if (value.tags) {
```
jonathan-gramain, line 101: I would think factorizing this code into a separate function `decodeTags` for reuse, at least across the two cases that need it, is worth it.

jonathan-gramain (approved):

lib/storage/metadata/mongoclient/ListRecordStream.js
jonathan-gramain: Oh and it can be unit tested more easily then ;)

rahulreddy (approved):

tests/unit/storage/metadata/mongoclient/ListRecordStream.spec.js
```diff
@@ -364,4 +364,42 @@ describe('mongoclient.ListRecordStream', () => {
...
+        const logEntry = {
+            h: -42,
+            ts: Timestamp.fromNumber(42),
+            op: 'i',
+            ns: 'some-bucket',
+            o: {
+                _id: 'replicated-key\u000098467518084696999999RG001  19.3',
+                value: {
+                    tags: {
+                        someKey: '\uFF041\uFF0E5',
```
jonathan-gramain: I though you had the issue specifically with tag keys, so `'\uFF041\uFF0E5': 'someValue'` instead

vrancurel: the "tags" object is processed indifferently for keys or values. So you can have dots and dollars in keys too.

jonathan-gramain (approved):

vrancurel:
@bert-e approve

PR #865, opened 2019-08-01
Feature/s3 c 2276 bucket policy model

dora-korpar:
Updates the BucketInfo model and creates the BucketPolicy model

### DIFF
--- a/lib/models/BucketPolicy.js
+++ b/lib/models/BucketPolicy.js
@@ -1,0 +1,127 @@
+const assert = require('assert');
+
+const errors = require('../errors');
+const { validateResourcePolicy } = require('../policy/policyValidator');
+
+/**
+ * Format of json policy:
+ *  {
+ *      "Id": "Policy id",
+ *      "Version": "version date",
+ *      "Statement": [
+ *          {
+ *              "Sid": "Statement id",
+ *              "Effect": "Allow",
+ *              "Principal": "*",
+ *              "Action": "s3:*",
+ *              "Resource": "arn:aws:s3:::examplebucket/bucket2/object"
+ *          },
+ *          {
+ *              "Sid": "Statement id",
+ *              "Effect": "Deny",
+ *              "Principal": {
+ *                  "AWS": ["arn:aws:iam::<account_id>", "different_account_id"]
+ *              },
+ *              "Action": [ "s3:*" ],
+ *              "Resource": [
+ *                  "arn:aws:s3:::examplebucket", "arn:aws:s3:::otherbucket/*"],
+ *              "Condition": {
+ *                 "StringNotLike": {
+ *                      "aws:Referer": [
+ *                          "http://www.example.com/", "http://example.com/*"]
+ *                  }
+ *              }
+ *          }
+ *      ]
+ *  }
+ */
+
+class BucketPolicy {
+    /**
+     * Create a Bucket Policy instance
+     * @param {string} json - the json policy
+     * @return {object} - BucketPolicy instance
+     */
+    constructor(json) {
+        this._json = json;
+        this._policy = {};
+    }
+
+    /**
+     * Get the bucket policy
+     * @return {object} - the bucket policy or error
+     */
+    getBucketPolicy() {
+        const policy = this._getPolicy();
+        return policy;
+    }
+
+    /**
+     * Get the bucket policy array
+     * @return {object} - contains error if policy validation fails
+     */
+    _getPolicy() {
+        if (!this._json || this._json === '') {
+            return { error: errors.MalformedPolicy.customizeDescription(
+                'request json is empty or undefined') };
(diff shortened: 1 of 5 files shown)


lib/models/BucketPolicy.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * Call resource policy schema validation function
+     * @param {object} policy - the bucket policy to validate
+     * @return {undefined}
+     */
+    static validatePolicy(policy) {
+        const validated = validateResourcePolicy(policy);
+        assert.deepStrictEqual(validated, { error: null, valid: true });
```
rahulreddy, line 123: We should not be throwing an assertion at run time. I would suggest returning `{ valid: true }`  and `{ error: <some error> }` for error cases.

rahulreddy (dismissed):

dora-korpar:
@bert-e reset

lib/models/BucketPolicy.js
dora-korpar: Actually, after looking at it again, this function follows the example of LifecycleConfig and ReplicationConfig. This validate function is only used by `BucketInfo` class, so there's no way to really return an error :/

dora-korpar:
@bert-e reset

JianqinWang (approved):

alexanderchan-scality (approved):

tmacro (approved):

dora-korpar:
@bert-e approve

lib/models/BucketPolicy.js
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     * Get the bucket policy array
+     * @return {object} - contains error if policy validation fails
+     */
+    _getPolicyJSON() {
+        const policy = {};
+        if (!this._json || this._json === '') {
+            policy.error = errors.MalformedPolicy.customizeDescription(
+                'request json is empty or undefined');
+            return policy;
```
bennettbuchanan: Can we simply return `{ error: errors.MalformedPolicy... }`? I.e. why first assign to policy here? That seems to apply to all the conditions here.

lib/models/BucketPolicy.js
```diff
@@ -0,0 +1,127 @@
...
+        const valAcRes = this._validateActionResource();
+        if (valAcRes.error) {
+            policy.error = valAcRes.error;
+            return policy;
+        }
+
+        return this._policy;
+    }
+
+    _getStatementArray() {
```
bennettbuchanan: This seems more like a setter as it always returns undefined and sets the internal `this._policy`.

lib/models/BucketPolicy.js
```diff
@@ -0,0 +1,127 @@
...
+     */
+    _validateActionResource() {
+        const obj = {};
+        this._policy.Statement.forEach(s => {
+            const actions = typeof s.Action === 'string' ?
+                [s.Action] : s.Action;
+            const resources = typeof s.Resource === 'string' ?
+                [s.Resource] : s.Resource;
+            const objectAction = actions.some(a => a.includes('Object'));
+            const allAction = actions.some(a => a.includes('*'));
```
bennettbuchanan: I haven't tried it out, but it sounds valid also use wildcards as part of the action name e.g. `s3:Get*` which would then only apply to a subset of actions. I'm wondering because in the AWS docs it shows `"Action": "iam:*AccessKey*"` as valid.

dora-korpar: Oof, you're right about the `s3:Get*` thing, that's a great catch. I didn't realize that was valid. But for the `iam:` action, do you have a link to the example? As far as I can tell, only S3 operations are valid as actions. Even if the `iam` prefix is valid tho, we're still only going to support S3 actions for right now

dora-korpar:
@bert-e wait

rahulreddy (dismissed):

lib/models/BucketPolicy.js
bennettbuchanan: Oh yes, I agree it'd need to start with "s3". It's just that I came across that example [here](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_action.html) using "iam" so figured it could apply similarly.

dora-korpar: aha I see. I think that example page is for IAM/user policies specifically :)

dora-korpar: Ok, I've accounted for an asterisk in the action and added some new tests.

bennettbuchanan (approved):

dora-korpar:
@bert-e approve

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e force_reset

PR #877, opened 2019-08-09
bf: S3C 2276 bucket info should store obj not json

dora-korpar:
The BucketInfo should store bucket policy as object, not a JSON string

### DIFF
--- a/lib/models/BucketPolicy.js
+++ b/lib/models/BucketPolicy.js
@@ -115,11 +115,13 @@ class BucketPolicy {
 
     /**
      * Call resource policy schema validation function
-     * @param {object} policy - the bucket policy to validate
+     * @param {object} policy - the bucket policy object to validate
      * @return {undefined}
      */
     static validatePolicy(policy) {
-        const validated = validateResourcePolicy(policy);
+        // only the BucketInfo constructor calls this function
+        // and BucketInfo will always be passed an object
+        const validated = validateResourcePolicy(JSON.stringify(policy));
         assert.deepStrictEqual(validated, { error: null, valid: true });
     }
 }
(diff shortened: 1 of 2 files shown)


tmacro (approved):

lib/models/BucketInfo.js
```diff
@@ -347,7 +347,7 @@ class BucketInfo {
      * does not have a bucket policy
      */
     getBucketPolicy() {
-        return this._bucketPolicy;
+        return JSON.parse(this._bucketPolicy);
```
alexanderchan-scality: `this._bucketPolicy` might be `undefined` or `any non-json string` because `setBucketPolicy` doesn't have a check for valid values, and `JSON.parse` throws an error for non-JSON strings
```suggestion
        if (this._bucketPolicy === 'non-JSON string') {
            return null;
        }
        return JSON.parse(this._bucketPolicy);
```

minor: why can't this be stored as an object?

dora-korpar: Yeah, that's a good point. Following the format of some of our other classes like LifecycleConfiguration and ReplicationConfiguration, technically, this setter should never be called unless the validation of the, in this case policy, has already happened. It's just that that's done directly on the BucketPolicy class. It's definitely a confusing setup, but it looks like this:
`const bucketPolicy = new BucketPolicy(request.post);`
`const policyObj = bucketPolicy.getBucketPolicy();`
`if (policyObj.error) {`
    `return next(policyObj.error);`
`}`
`bucket.setBucketPolicy(bucketPolicy);`
So the validation happens in the `bucketPolicy.getBucketPolicy()` step.
But you're 100% right, this should just be stored as an object, idk what I was thinking

dora-korpar:
@bert-e reset

bennettbuchanan (approved):

dora-korpar:
@bert-e approve

PR #886, opened 2019-08-21
improv(ZENKO-2068): Improve MongoClientInterface checkHealth

tmacro:


ssalaues (approved):

JianqinWang (approved):

rahulreddy (changes requested):
Why are you not using the `errors.ServiceUnavailable` object? https://github.com/scality/Arsenal/blob/development/8.1/errors/arsenalErrors.json#L334
Not using Arsenal errors would break the contracts we have in the methods elsewhere.

tmacro:
> Why are you not using the `errors.ServiceUnavailable` object? https://github.com/scality/Arsenal/blob/development/8.1/errors/arsenalErrors.json#L334
> Not using Arsenal errors would break the contracts we have in the methods elsewhere.

https://github.com/scality/Arsenal/blob/1c6e56e8eff3dd0f917a965bdcbbc4ab89aa930f/lib/storage/metadata/bucketclient/BucketClientInterface.js#L175 

So according to the comments in  BucketClientInterface the use of a non-standard error is intentional. The consuming logic in https://github.com/scality/cloudserver/blob/f85c3ca4892bbbf992f42ad02cb9783690ce8900/lib/utilities/healthcheckHandler.js#L68 requires the `error` property which an arsenal error does not provide. I've updated the code to read the data from `arsenal.errors` but the custom object must remain

rahulreddy:
> > Why are you not using the `errors.ServiceUnavailable` object? https://github.com/scality/Arsenal/blob/development/8.1/errors/arsenalErrors.json#L334
> > Not using Arsenal errors would break the contracts we have in the methods elsewhere.
> 
> https://github.com/scality/Arsenal/blob/1c6e56e8eff3dd0f917a965bdcbbc4ab89aa930f/lib/storage/metadata/bucketclient/BucketClientInterface.js#L175
> 
> So according to the comments in BucketClientInterface the use of a non-standard error is intentional. The consuming logic in https://github.com/scality/cloudserver/blob/f85c3ca4892bbbf992f42ad02cb9783690ce8900/lib/utilities/healthcheckHandler.js#L68 requires the `error` property which an arsenal error does not provide. I've updated the code to read the data from `arsenal.errors` but the custom object must remain

Yeah it's the method that is returning the object with error property set, but it still returning an Arsenal Error. Also, Metadata was kind of anomaly as it's 500 wasn't really a 500 as we expect the system to be up even when the service is partially degraded. Thanks for the correction.

rahulreddy (approved):

tmacro:
@bert-e approve

PR #899, opened 2019-10-03
bugfix: S3C-1805 Support consecutive hyphens in bucket names

naren-rajendran:
#### Why is this change required? What problem does it solve?
Bucket names with consecutive hyphens were not allowed and the request was responded with error `400 InvalidBucketName`.  Consecutive hyphens in bucket names are allowed in AWS S3, so corrections are made to allow consecutive hyphens in bucket names.
 
Tests are added to check valid combinations of hyphens(-) and dots(.) in bucket names.


bennettbuchanan (approved):

ssalaues (approved):

naren-rajendran:
@bert-e approve

PR #909, opened 2019-11-08
bugfix: use default auth DB in mongo log reader

rachedbenmustapha:
The mongodb log reader included `/local` in its connection string. The path element in mongodb's connection URI is actually the DB authenticate against, not the DB to work on. This lead to authentication failures even though roles, users, passwords etc were correct.


jonathan-gramain (approved):

JianqinWang (approved):

rachedbenmustapha:
@bert-e approve

PR #910, opened 2019-11-15
bf: removing dead code

vrancurel:
The special header handling for metadata only operation
is done in clouserver:lib/api/apiUtils/object/createAndStoreObject.js
This code is actually never reached therefore we remove it.

### DIFF
--- a/lib/storage/data/external/PfsClient.js
+++ b/lib/storage/data/external/PfsClient.js
@@ -36,18 +36,6 @@ class PfsClient {
 
     put(stream, size, keyContext, reqUids, callback) {
         const log = createLogger(reqUids);
-        if (keyContext.metaHeaders['x-amz-meta-mdonly'] === 'true'
-                || keyContext.isDeleteMarker === true) {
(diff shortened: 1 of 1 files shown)


lib/storage/data/external/PfsClient.js
```diff
@@ -36,18 +36,6 @@ class PfsClient {
 
     put(stream, size, keyContext, reqUids, callback) {
         const log = createLogger(reqUids);
-        if (keyContext.metaHeaders['x-amz-meta-mdonly'] === 'true'
```
ssalaues, line 39: The reason why I left this here is to deal with the delete markers and to handle 0-byte objects. This is because the code I added in cloudserver can't differentiate between a 0-byte PUT and an MD only put.

You can see in my original PR to deprecate this code:
https://github.com/scality/Arsenal/pull/836#issuecomment-517067466

ssalaues (changes requested):

vrancurel:
Invalid PR

vrancurel:
This is not dead code but code called in case of CRR RING to RING.

PR #913, opened 2019-11-22
feature: ZENKO-2089 add mongodb conditionals

alexanderchan-scality:
add conditional put/delete operations to allow for correct blobserver behaviors

**update**:
- adds dev dependency [mongodb-memory-server](https://www.npmjs.com/package/mongodb-memory-server)

### DIFF
--- a/lib/storage/metadata/MetadataWrapper.js
+++ b/lib/storage/metadata/MetadataWrapper.js
@@ -356,6 +356,76 @@ class MetadataWrapper {
     notifyBucketChange(cb) {
         bucketNotificationHook = cb;
     }
+
+    close(cb) {
+        if (typeof this.client.close === 'function') {
+            return this.client.close(cb);
+        }
+        return cb();
+    }
+
+    /**
+     * deletes an object that matches the given conditions
+     * @param{string} bucketName -
+     * @param{string} objName -
+     * @param{object} params -
+     * @param{object} params.conditions - conditions object
+     * @param{werelogs.RequestLogger} log -
+     * @param{function} cb -
+     * @returns{undefined}
+     */
+    deleteObjectWithCond(bucketName, objName, params, log, cb) {
+        log.debug('find and delete object from metadata');
+        if (typeof this.client.deleteObjectWithCond !== 'function') {
+            log.debug('backend does not support deletions with conditions', {
+                implName: this.implName,
+            });
+            return cb(errors.NotImplemented);
+        }
+        return this.client.deleteObjectWithCond(bucketName, objName,
+            params, log, err => {
+                if (err) {
+                    log.debug('error from metadata', {
+                        implName: this.implName,
+                    });
+                    return cb(err);
+                }
+                log.debug('object deleted from metadata');
+                return cb();
+            });
+    }
+
+    /**
+     * updates(insert, if missing) an object that matches the given conditions
+     * @param{string} bucketName -
+     * @param{string} objName -
+     * @param{object} objVal -
+     * @param{object} params -
+     * @param{object} params.conditions - conditions object
+     * @param{werelogs.RequestLogger} log -
+     * @param{function} cb -
+     * @returns{undefined}
+     */
+    putObjectWithCond(bucketName, objName, objVal, params, log, cb) {
+        log.debug('find and update object in metadata');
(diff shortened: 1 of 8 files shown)


lib/storage/metadata/MetadataWrapper.js
```diff
@@ -356,6 +356,12 @@ class MetadataWrapper {
     notifyBucketChange(cb) {
         bucketNotificationHook = cb;
     }
+
+    close() {
+        if (this.client.close === 'function') {
```
rachedbenmustapha: Missing `typeof`?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -622,6 +662,10 @@ class MongoClientInterface {
             return this.putObjectVerCase4(c, bucketName, objName, objVal,
                                           params, log, cb);
         }
+        if (params && params.query) {
+            return this.putObjectNoVerConditional(
```
rachedbenmustapha: Is this only the `NoVer` variant because blobserver never supports versioning? What would happen if blobserver were to try updating/deleting a versioned object put by cloudserver?

rachedbenmustapha (changes requested):
This change seems to be breaking some abstraction here, the metadata interface is backend-agnostic. I suggest collecting which filters are going to be useful, and pass them as objects that are part of Arsenal's API and backend-agnostic, then doing the translation in the backend-specific impl (in this case MongoClientInterface).

Also the metadata wrapper should return an internal error if the backend does not support the translation, to avoid unexpected issues.

vrancurel:
> This change seems to be breaking some abstraction here, the metadata interface is backend-agnostic. I suggest collecting which filters are going to be useful, and pass them as objects that are part of Arsenal's API and backend-agnostic, then doing the translation in the backend-specific impl (in this case MongoClientInterface).
> 
> Also the metadata wrapper should return an internal error if the backend does not support the translation, to avoid unexpected issues.

We made the abstraction mistake for search, can we fusion / factorize with search (whom parsing / execution logic lies in CS instead of Arsenal) with those conditional operators ?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1634,6 +1635,103 @@ class MongoClientInterface {
...
+            }
+            if (res.ok !== 1) {
+                log.error('failed to update object', {
+                    method,
+                    error: err.message,
+                });
+                return cb(errors.InternalError);
+            }
+            if (!res.value) {
+                log.debug('object not found...inserting object', {
```
rachedbenmustapha: Since the action is already done, the log should use past tense to avoid confusion (can be surprising when troubleshooting that the DB does not reflect what the log says).

rachedbenmustapha (dismissed):

rachedbenmustapha:
@vrancurel I'm not sure it would be appropriate to merge them. This particular abstraction is fine for programmatic use of the API, JSON is easy to build this way. But we can't expect users to type JSON in the search bar, and using a human-friendly search syntax for API use (especially implemented in javascript, in critical request path such as upsert/conditional writes) would be wasteful.

The search syntax definitely needs to be re-done, and the new parser could definitely translate to this data structure for convenience, but I don't think they are the same abstraction level.

For best impact, the human-friendly syntax would be like lucene/elasticsearch, as many users have used it one way or another, and its basic form is similar enough to a google search.

tests/unit/storage/metadata/mongoclient/utils.js
```diff
@@ -25,3 +28,169 @@ describe('auth credentials', () => {
         assert.strictEqual(credPrefix(creds), 'u%3A:%40p@');
     });
 });
+
+describe.only('translate query object', () => {
```
JianqinWang: :') `.only`

JianqinWang (commented):
other than the `.only` LGTM

JianqinWang (approved):

vrancurel:
I would prefer renaming Query to Cond, @RachedBenMustapha  ?

vrancurel (approved):

rachedbenmustapha (approved):

rachedbenmustapha:
@vrancurel wouldn't these new cond primitives make the versioning code a lot cleaner?

PR #923, opened 2020-01-13
improvement: ZENKO-2278 count items code reorg

alexanderchan-scality:


tmacro (approved):

JianqinWang (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1283,42 +1278,50 @@ class MongoClientInterface {
...
-                    }
-                    return done();
-                });
-            }, err => {
-                if (err) {
-                    return cb(err);
-                }
-                // save to infostore
-                return this.updateCountItems(store, log, err => {
+            let concurrentCursors = CONCURRENT_CURSORS;
```
rachedbenmustapha: This seems to be in the way, why not in the constructor or setup method?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1415,211 +1417,136 @@ class MongoClientInterface {
...
-        const mstFilter = {
-            '_id': { $regex: /^[^\0]+$/ },
-            'value.versionId': { $exists: true },
-        };
-        const verFilter = { _id: { $regex: /\0/ } };
-        const nullFilter = {
-            '_id': { $regex: /^[^\0]+$/ },
-            'value.versionId': { $exists: false },
+        const cursor = c.find({}, {
+            sort: { _id: 1 },
```
rachedbenmustapha: Given that we handle items as they come, is the sort useful? Does it incur a penalty?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1415,211 +1417,136 @@ class MongoClientInterface {
...
-            return cb(null, count);
-        });
+    /**
+     * @param{object} entry -
+     * @param{string} entry._id -
+     * @param{object} entry.value -
+     * @param{Date} cmpDate -
+     * @returns{boolean} stalled -
+     */
+    _processEntryStalled(entry, cmpDate) {
```
rachedbenmustapha: This is a test, so maybe call it `isReplicationEntryStalled`

lib/storage/metadata/mongoclient/MongoClientInterface.js
alexanderchan-scality: you are correct. order doesn't matter in the case of count items. the sort can be removed.

nicolas2bert (approved):

rachedbenmustapha (approved):

alexanderchan-scality:
rebased/squashed

PR #930, opened 2020-02-06
improvement: perform get bucket and object at once

vrancurel:
Instead of doing 2 round trips we perform the 2 queries at once using an
aggregate.

This PR is to check that we pass the end-to-end tests. You can review but we do not want to merge unless we are sure it improves the performance.


vrancurel:
@bert-e wait

PR #931, opened 2020-02-06
improvement: perform get bucket and object at once

vrancurel:
Instead of doing 2 round trips we perform the 2 queries at once using an
aggregate.

This PR is to check that we pass the end-to-end tests. You can review but we do not want to merge unless we are sure it improves the performance.

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -250,25 +250,62 @@ class MongoClientInterface {
     }
 
     getBucketAndObject(bucketName, objName, params, log, cb) {
-        this.getBucketAttributes(bucketName, log, (err, bucket) => {
+        if (params && params.versionId) {
+            // eslint-disable-next-line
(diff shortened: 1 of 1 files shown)


vrancurel:
@bert-e wait

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -250,25 +250,62 @@ class MongoClientInterface {
...
+                    pipeline: [
+                        { $match: { _id: objName } },
+                    ],
+                    as: 'object',
+                },
+            },
+        ], {}, (err, cursor) => {
             if (err) {
                 log.error(
                     'getBucketAttributes: error getting bucket attributes',
```
dora-korpar: not necessary to start benchmark testing, but if we keep this combined approach, this error message should change

vrancurel:
Closing this PR as parallelization https://github.com/scality/Arsenal/pull/932 generates better results.

vrancurel:
Actually sometimes aggregates generate better results too

PR #945, opened 2020-02-24
bugfix: S3C-2604-handle-de-serialize-specific-resource

anurag4DSB:
What does this PR do, and why do we need it?

Handles individual-specific resources for request contexts coming from Policy Evaluation of Vault.

Which issue does this PR fix?
fixes #S3C-2604

Special notes for your reviewers:
This is a quick workaround, we are gonna re-architecture UTAPI eventually.


anurag4DSB:
@bert-e reset

PR #962, opened 2020-02-27
bugfix: ZENKO-2352 send back HTTP 424 when location does not exist

jonathan-gramain:
Send back an HTTP error 424 (Failed Dependency, a WebDAV extension)
instead of HTTP 503 (Service Unavailable), when the backend client
cannot retrieve the location of the data and gets a 404 error from the
server when issuing a HEAD request. This HEAD request is triggered on
the backend when a client issues a GET request on Zenko for an object
stored in a cloud backend. The error returned to the client also
contains a more specific error code "LocationNotFound", which is not
part of AWS standard.

The immediate purpose is to have backbeat not retry on such errors, as
they might arise if e.g. a Zenko bucket is backed by an S3C location,
and the out-of-band updates lag behind, so a user might have deleted
versions on S3C before Zenko got notified of the deletion, thinking
the object is still available.

More generally, even in the hypothetic case where a server-side bug
would have deleted the data, it's better not to have the client retry
as this is a definite failure and the client will just retry vainly
until it times out. Sending back a 4xx error makes this clear to the
client that it should not retry (not until something changes on its
side, like writing the same key again).

The patch applies on all supported backends: AWS, Azure, GCP.

### DIFF
--- a/errors/arsenalErrors.json
+++ b/errors/arsenalErrors.json
@@ -476,6 +476,10 @@
     "code": 409,
     "description": "The bucket you tried to delete has an ongoing multipart upload."
   },
+  "LocationNotFound": {
+    "code": 424,
+    "description": "The object data location does not exist."
+  },
   "_comment": "-------------- Internal project errors --------------",
   "_comment": "----------------------- Vault -----------------------",
   "_comment": "#### formatErrors ####",
--- a/lib/storage/data/external/AwsClient.js
+++ b/lib/storage/data/external/AwsClient.js
@@ -148,21 +148,19 @@ class AwsClient {
             VersionId: dataStoreVersionId,
         }, (err, data) => {
             if (err) {
-                logHelper(log, 'error', 'error heading object ' +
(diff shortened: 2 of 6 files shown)


errors/arsenalErrors.json
```diff
@@ -476,6 +476,10 @@
     "code": 409,
     "description": "The bucket you tried to delete has an ongoing multipart upload."
   },
+  "LocationNotFound": {
+    "code": 424,
+    "description": "the object data location does not exist"
```
naren-rajendran: Just nitpicking..Uppercase T and period!

alexanderchan-scality (approved):

naren-rajendran (approved):

errors/arsenalErrors.json
jonathan-gramain: Right, did not pay enough attention to the format of the other messages!

jonathan-gramain:
@bert-e approve

rahulreddy (approved):

PR #963, opened 2020-03-17
bugfix: update master if version is gte

vrancurel:
This change is a workaround to palliate the fact we do not have micro
version to manage micro changes such as ACLs or tags changed. Indeed in
the AWS S3 specification such changes do not trigger a new version but
update the version (and the master) in place.


naren-rajendran (approved):

rahulreddy (approved):

rahulreddy:
@bert-e bypass_author_approval

PR #970, opened 2020-04-03
Feature: S3C-2729 S3 pre-sign URL expiry customization

ilkescality:

PR #973, opened 2020-04-07
Improvement/S3C-2749 add tests for url duration

ilkescality:
Adds unit tests for customizing S3 pre-signed URL implemented here: https://github.com/scality/Arsenal/pull/970

### DIFF
--- a/tests/unit/auth/v2/queryAuthCheck.js
+++ b/tests/unit/auth/v2/queryAuthCheck.js
@@ -12,58 +12,40 @@ describe('v2: queryAuthCheck', () => {
         { token: undefined, error: false },
         { token: 'invalid-token', error: true },
         { token: 'a'.repeat(128), error: false },
-    ].forEach(test => {
-        it(`test with token(${test.token})`, () => {
-            const request = { method: 'GET' };
-            const data = {
-                SecurityToken: test.token,
-            };
-            const res = queryAuthCheck(request, log, data);
-            if (test.error) {
-                assert.notStrictEqual(res.err, undefined);
-                assert.strictEqual(res.err.InvalidToken, true);
-            } else {
-                assert.notStrictEqual(res.err, undefined);
-                assert.strictEqual(res.err.MissingSecurityHeader, true);
-            }
-        });
-    });
+    ].forEach(test => it(`test with token(${test.token})`, () => {
+        const request = { method: 'GET' };
+        const data = {
+            SecurityToken: test.token,
+        };
+        const res = queryAuthCheck(request, log, data);
+        if (test.error) {
+            assert.notStrictEqual(res.err, undefined);
+            assert.strictEqual(res.err.InvalidToken, true);
+        } else {
+            assert.notStrictEqual(res.err, undefined);
+            assert.strictEqual(res.err.MissingSecurityHeader, true);
+        }
+    }));
 });
 
 describe('v2: queryAuthCheck', () => {
-    let clock;
+    let clock = sinon.useFakeTimers();
 
-    beforeEach(() => {
-        clock = sinon.useFakeTimers();
-    });
-    afterEach(() => {
-        process.env.PRE_SIGN_URL_EXPIRY = 604800000;
-        clock.restore();
-    });
-    it('URL should not expire before 7 days with default expiry', () => {
+    it('S3 URL should not expire before 7 days with default preSignUrlExpiry', () => {
         const currentTime = Date.now() / 1000;
         const expires = currentTime + 604799;  // in seconds
-        const mockRequest = {
-            method: 'GET',
-            url: 'mockurl',
-            query: {
-                Expires: expires,
-            },
-            headers: {
-                'Content-MD5': 'c',
-            },
-        };
-        const data = {
-            Expires: expires,
-            AWSAccessKeyId: 'keyId',
-            Signature: 'sign',
-        };
-        const res = queryAuthCheck(mockRequest, log, data);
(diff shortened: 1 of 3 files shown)


tests/unit/auth/v2/queryAuthCheck.js
```diff
@@ -58,20 +60,50 @@ describe('v2: queryAuthCheck', () => {
...
         const data = { Expires: currentTime };
         const res = queryAuthCheck(request, log, data);
         assert.notStrictEqual(res.err, undefined);
         assert.strictEqual(res.err.AccessDenied, true);
     });
-    it('should raise MissingSecurityHeader with invalid expires param', () => {
-        const request = { method: 'GET', query: { Expires: 'a string' } };
-        const data = { Expires: 'a string' };
+
+    clock.restore();
```
miniscruff: this might need / should be in an afterEach or afterAll, whatever the test framework has I cant remember.

tests/unit/auth/v2/queryAuthCheck.js
```diff
@@ -58,20 +60,50 @@ describe('v2: queryAuthCheck', () => {
...
+            Expires: expires,
+            AWSAccessKeyId: 'keyId',
+            Signature: 'sign',
+        };
+        const res = queryAuthCheck(mockRequest, log, data);
+        assert.notStrictEqual(res.err.AccessDenied, true);
+        assert.notStrictEqual(res.err.RequestTimeTooSkewed, true);
+    });
+
+    clock.tick(604800000);  // take time 604800000ms (7 days) ahead
```
miniscruff: I would put this logic inside the tests that need it

ilkescality: good point, done

tests/unit/auth/v2/queryAuthCheck.js
ilkescality: also done

tests/unit/auth/v2/queryAuthCheck.js
```diff
@@ -31,10 +31,15 @@ describe('v2: queryAuthCheck', () => {
...
-    let clock = sinon.useFakeTimers();
-
-    beforeEach(() => process.env.PRE_SIGN_URL_EXPIRY = 604800000);
+    let clock;
 
+    beforeEach(() => {
+        clock = sinon.useFakeTimers();
+    });
+    afterEach(() => {
+        process.env.PRE_SIGN_URL_EXPIRY = 604800000;
```
miniscruff, line 38: Shouldn't we set this in the beforeEach so it takes affect in each test?

ilkescality: Earlier I put it in beforeEach thinking like that, but essentially the goal was testing it for `undefined` cases as well, so put it inside afterEach so that the first two test cases can test against `undefined` env var.

tests/unit/auth/v2/queryAuthCheck.js
```diff
@@ -11,18 +12,111 @@ describe('v2: queryAuthCheck', () => {
...
-    ].forEach(test => it(`test with token(${test.token})`, () => {
-        const request = { method: 'GET' };
+    ].forEach(test => {
+        it(`test with token(${test.token})`, () => {
+            const request = { method: 'GET' };
+            const data = {
+                SecurityToken: test.token,
+            };
+            const res = queryAuthCheck(request, log, data);
+            if (test.error) {
```
miniscruff: I would advise against conditionals in tests, instead create a test for each condition.

tests/unit/auth/v2/queryAuthCheck.js
```diff
@@ -11,18 +12,111 @@ describe('v2: queryAuthCheck', () => {
...
+            AWSAccessKeyId: 'keyId',
+            Signature: 'sign',
         };
+        const res = queryAuthCheck(mockRequest, log, data);
+        assert.notStrictEqual(res.err.AccessDenied, true);
+        assert.notStrictEqual(res.err.RequestTimeTooSkewed, true);
+    });
+    it('URL should expire after 7 days with default expiry', () => {
+        clock.tick(604800000);  // take time 604800000ms (7 days) ahead
+        const currentTime = Date.now();
```
miniscruff: This might sound kinda lame or cheesy but to call a "currentTime" after setting the time to 7 days ahead makes it a little confusing to me. Maybe like "expiredTime" or just testTime to be more generic for all tests.

tests/unit/auth/v2/queryAuthCheck.js
ilkescality: I agree, although that's not supposed to be a part of my changes - only seems as new addition cause I separated line 16 from line 15 for better readability per style guidelines. 
It seems like the same logic replicated along other places such as: https://github.com/scality/Arsenal/blob/development/7.7/tests/unit/auth/v4/headerAuthCheck.js#L47 ,  https://github.com/scality/Arsenal/blob/development/7.7/tests/unit/auth/v2/headerAuthCheck.js#L21

tests/unit/auth/v2/queryAuthCheck.js
ilkescality: The idea is to emphasize that the URL expires or remains valid at the current time after 7 days, so "expiredTime" could be misleading, but using a generic testTime for all the tests is a good idea.

ilkescality:
Moving to a fresh branch due to corner git issue, closing this, created a new PR: https://github.com/scality/Arsenal/pull/975

PR #975, opened 2020-04-09
improvement/S3C-2749 add tests for url duration

ilkescality:
Adds unit tests for customizing S3 pre-signed URL implemented here: #970

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -72,4 +72,7 @@ module.exports = {
     permittedCapitalizedBuckets: {
         METADATA: true,
     },
+    // Default expiration value of the S3 pre-signed URL duration
+    // 604800 seconds (seven days).
+    defaultPreSignedURLExpiry: 7 * 24 * 60 * 60,
 };
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "eslint-config-scality": "scality/Guidelines#71a059ad",
     "lolex": "1.5.2",
     "mocha": "2.5.3",
+    "sinon": "^9.0.2",
     "temp": "0.8.3"
   },
   "scripts": {
--- a/tests/unit/auth/v2/queryAuthCheck.js
+++ b/tests/unit/auth/v2/queryAuthCheck.js
@@ -1,4 +1,5 @@
 const assert = require('assert');
+const sinon = require('sinon');
 
 const queryAuthCheck =
     require('../../../../lib/auth/v2/queryAuthCheck').check;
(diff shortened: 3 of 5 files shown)


tests/unit/auth/v2/queryAuthCheck.js
```diff
@@ -26,3 +27,94 @@ describe('v2: queryAuthCheck', () => {
...
 });
+
+describe('v2: queryAuthCheck', () => {
+    let clock;
+
+    beforeEach(() => {
+        clock = sinon.useFakeTimers();
+    });
+    afterEach(() => {
+        process.env.PRE_SIGN_URL_EXPIRY = 604800000;
```
naren-rajendran, line 38: I think I missed it in your original PR, could you add the default expiry value to  `/arsenal/lib/constants.js` and use that in `queryAuthCheck.js` and here?  You can include the change in this PR I guess!

package.json
```diff
@@ -28,7 +28,7 @@
     "level": "~5.0.1",
     "level-sublevel": "~6.6.5",
     "node-forge": "^0.7.1",
-    "simple-glob": "^0.1",
+    "simple-glob": "^0.2.0",
```
dora-korpar: Are these package updates needed for the tests? Cos if not they should be in a separate PR

tests/unit/auth/v2/queryAuthCheck.js
jonathan-gramain: If you are going to be adding this, *optionally* you might consider changing the default unit to seconds and convert it to the required unit, it's a more familiar unit and less 0s to count :wink:

jonathan-gramain: Only do this if it's an easy change I don't want you to break all the stack for this :)

tests/unit/auth/v2/queryAuthCheck.js
```diff
@@ -26,3 +27,94 @@ describe('v2: queryAuthCheck', () => {
         }
     }));
 });
+
+describe('v2: queryAuthCheck', () => {
+    let clock;
+
+    beforeEach(() => {
+        clock = sinon.useFakeTimers();
```
jonathan-gramain, line 35: :+1:

package.json
ilkescality: They shouldn't be needed (unless sinon has dependency for their specific version which I don't think is the case), I'll try only leaving the new package and revert `simple-glob` and `eslint-plugin-react` to their original versions.

tests/unit/auth/v2/queryAuthCheck.js
nicolas2bert, line 38: Minor: for better readability, you might want to add a global variable; something like `const SEVEN_HOURS_IN_SEC = 7 * 24 * 60 * 60`;

tests/unit/auth/v2/queryAuthCheck.js
```diff
@@ -26,3 +27,94 @@ describe('v2: queryAuthCheck', () => {
...
+        assert.notStrictEqual(res.err.AccessDenied, true);
+        assert.notStrictEqual(res.err.RequestTimeTooSkewed, true);
+    });
+    it('URL should expire after 7 days with default expiry', () => {
+        clock.tick(604800000);  // take time 604800000ms (7 days) ahead
+        const currentTime = Date.now();
+        const request = { method: 'GET', query: { Expires: currentTime } };
+        const data = { Expires: currentTime };
+        const res = queryAuthCheck(request, log, data);
+        assert.notStrictEqual(res.err, undefined);
```
jonathan-gramain, line 70: minor style: might be slightly more readable to do `assert(res.err)` and it does a truish test, as some callbacks return `null` as the error object (not `undefined`), it can be a more generic way to check that an error occurred.

tests/unit/auth/v2/queryAuthCheck.js
ilkescality: Sure, will try that and see if it breaks anything, so far it doesn't seem like it would, but I'll double-check. :)

jonathan-gramain (approved):

nicolas2bert (approved):

tests/unit/auth/v2/queryAuthCheck.js
naren-rajendran: IMHO, you should, makes it easier to set!

ilkescality: Good idea

naren-rajendran:
@ilkescality I think you should squash your commits.

lib/constants.js
```diff
@@ -72,4 +72,7 @@ module.exports = {
     permittedCapitalizedBuckets: {
         METADATA: true,
     },
+    // Default expiration value of the S3 pre-signed URL duration
+    // 604800 seconds (seven days).
+    defaultExpiry: 7 * 24 * 60 * 60,
```
naren-rajendran: May be a more descriptive name? `defaultPreSignedURLExpiry` instead of `defaultExpiry`!

tests/unit/auth/v2/queryAuthCheck.js
ilkescality: TBH, it bothered me as well :) I thought it could be good to have a helper util such as `nullOrUndefined` or something like that - maybe another to check for even more falsey values depending on the use case... For now, I added additional check for the `null` case to cover more potential scenarios.

naren-rajendran: May be I am wrong, I don't see `queryAuthCheck` returning `err: undefined`!

ilkescality:
@bert-e force_reset

rachedbenmustapha (approved):

PR #977, opened 2020-04-10
improvement: ZENKO-2535 add microVersionId to ObjectMD

jonathan-gramain:
Add a new microVersionId field that is a hex-encoded field of 64
bits randomly generated.

It can be useful to:

- force updates in MongoDB when no other metadata changes, by updating
  the microVersionId field

- manage concurrent updates, by adding a condition on the
  microVersionId and updating it for each metadata update to perform

In order for the change to be less intrusive, it is an optional field:
if ObjectMD.updateMicroVersionId() is not called by the client, the
metadata will not contain a microVersionId field. Clients will call
this function when needed on a case-by-case basis.

### DIFF
--- a/lib/models/ObjectMD.js
+++ b/lib/models/ObjectMD.js
@@ -1,3 +1,5 @@
+const crypto = require('crypto');
+
 const constants = require('../constants');
 const VersionIDUtils = require('../versioning/VersionID');
 
@@ -1026,6 +1028,40 @@ class ObjectMD {
     overrideMetadataValues(headers) {
         Object.assign(this._data, headers);
         return this;
+    }
+
+    /**
+     * Create or update the microVersionId field
+     *
+     * This field can be used to force an update in MongoDB. This can
+     * be needed in the following cases:
+     *
+     * - in case no other metadata field changes
+     *
+     * - to detect a change when fields change but object version does
+     *   not change e.g. when ingesting a putObjectTagging coming from
+     *   S3C to Zenko
+     *
+     * - to manage conflicts during concurrent updates, using
+     *   conditions on the microVersionId field.
+     *
+     * It's a field of 16 hexadecimal characters randomly generated
+     *
+     * @return {ObjectMD} itself
+     */
+    updateMicroVersionId() {
+        this._data.microVersionId = crypto.randomBytes(8).toString('hex');
+        return this;
+    }
+
+    /**
+     * Get the microVersionId field, or null if not set
+     *
+     * @return {string|null} the microVersionId field if exists, or
+     * {null} if it does not exist
+     */
+    getMicroVersionId() {
+        return this._data.microVersionId || null;
     }
 
     /**
--- a/tests/unit/models/ObjectMD.js
+++ b/tests/unit/models/ObjectMD.js
@@ -257,6 +257,25 @@ describe('ObjectMD class setters/getters', () => {
         md.clearMetadataValues();
         assert.strictEqual(md.getUserMetadata(), undefined);
     });
+
+    it('ObjectMD::microVersionId unset', () => {
+        assert.strictEqual(md.getMicroVersionId(), null);
+    });
+
+    it('ObjectMD::microVersionId set', () => {
+        const generatedIds = new Set();
+        for (let i = 0; i < 100; ++i) {
+            md.updateMicroVersionId();
+            generatedIds.add(md.getMicroVersionId());
+        }
+        // all generated IDs should be different
+        assert.strictEqual(generatedIds.size, 100);
+        generatedIds.forEach(key => {
+            // length is always 16 in hex because leading 0s are
+            // also encoded in the 8-byte random buffer.
+            assert.strictEqual(key.length, 16);
+        });
+    });
 });
 
 describe('ObjectMD import from stored blob', () => {


lib/models/ObjectMD.js
```diff
@@ -1028,6 +1030,33 @@ class ObjectMD {
...
+     * no other metadata field changes) or to manage conflicts during
+     * concurrent updates, using conditions on the microVersionId
+     * field.
+     *
+     * It's a base64-encoded field of 64 bits randomly generated
+     *
+     * @return {ObjectMD} itself
+     */
+    updateMicroVersionId() {
+        this._data.microVersionId = crypto.randomBytes(8).toString('base64');
```
vrancurel: I would prefer 'hex' than 'base64' because it is more readable (although a bit longer).

vrancurel (changes requested):

lib/models/ObjectMD.js
jonathan-gramain: not convinced, but ok, I don't mind changing it.

jonathan-gramain: done

tests/unit/models/ObjectMD.js
```diff
@@ -257,6 +257,23 @@ describe('ObjectMD class setters/getters', () => {
...
         assert.strictEqual(md.getUserMetadata(), undefined);
     });
+
+    it('ObjectMD::microVersionId', () => {
+        assert.strictEqual(md.getMicroVersionId(), null);
+
+        const generatedIds = {};
+        for (let i = 0; i < 100; ++i) {
+            md.updateMicroVersionId();
+            generatedIds[md.getMicroVersionId()] = 1;
```
miniscruff: If all you care about is the key here you can use a Set instead of an Object

tests/unit/models/ObjectMD.js
```diff
@@ -257,6 +257,23 @@ describe('ObjectMD class setters/getters', () => {
...
+        }
+        // all generated IDs should be different
+        assert.strictEqual(Object.keys(generatedIds).length, 100);
+        Object.keys(generatedIds).forEach(key => {
+            // length is always 16 in hex because leading 0s are
+            // also encoded in the 8-byte random buffer.
+            assert.strictEqual(key.length, 16);
```
miniscruff, line 276: Instead of comparing lengths, you could mock the crypto call to ensure it is calling it with specific params

tests/unit/models/ObjectMD.js
```diff
@@ -257,6 +257,23 @@ describe('ObjectMD class setters/getters', () => {
...
+    it('ObjectMD::microVersionId', () => {
+        assert.strictEqual(md.getMicroVersionId(), null);
+
+        const generatedIds = {};
+        for (let i = 0; i < 100; ++i) {
+            md.updateMicroVersionId();
+            generatedIds[md.getMicroVersionId()] = 1;
+        }
+        // all generated IDs should be different
+        assert.strictEqual(Object.keys(generatedIds).length, 100);
```
miniscruff: Although unlikely this does have a chance to fail due to getting the same random 8 bytes out of 100, I am sure the odds are incredibly tiny but I would also argue that this doesn't really validate our code much either. Simply mocking the random call will have the same affect here.

tests/unit/models/ObjectMD.js
```diff
@@ -257,6 +257,23 @@ describe('ObjectMD class setters/getters', () => {
         md.clearMetadataValues();
         assert.strictEqual(md.getUserMetadata(), undefined);
     });
+
+    it('ObjectMD::microVersionId', () => {
+        assert.strictEqual(md.getMicroVersionId(), null);
```
miniscruff, line 262: This line looks like a separate test, that it defaults to none and that you can update it.

lib/models/ObjectMD.js
```diff
@@ -1028,6 +1030,33 @@ class ObjectMD {
...
+        return this;
+    }
+
+    /**
+     * Get the microVersionId field, or null if not set
+     *
+     * @return {string|null} the microVersionId field if exists, or
+     * {null} if it does not exist
+     */
+    getMicroVersionId() {
```
miniscruff, line 1063: Way outside the scope of this change but javascript now has get/set methods for classes https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set that would do wonders here.

tests/unit/models/ObjectMD.js
jonathan-gramain: The idea of this test is to ensure that we get a different ID every time it's called, it's more of a smoke test than anything else.

I'm not convinced mocking the random call would add much value, since there is also value in running the actual random code that is being used in production, and it's possible to call it within a unit test.

tests/unit/models/ObjectMD.js
jonathan-gramain: see answer below

tests/unit/models/ObjectMD.js
miniscruff: Basically by mocking an existing method we piggyback on all the testing and validation of that method ( or anything really but here its a method ). `crypto.randomBytes(8)` is going to be a very well tested method, simply by mocking and asserting we called that method is plenty of value to me.

Creating a smoke test, or length check or any repeated test here is going to be much less exhaustive when compared to what the crypto package already has to go through... probably. So I would argue being lazy here has the most value.

tl;dr having a test that says we call the crypto package is pretty convincing to me its well tested.

lib/models/ObjectMD.js
miniscruff: Going to resolve this now since there isnt really any action to take here, but will keep it in mind if any refactoring time is allocated.

miniscruff: @rahulreddy fyi

vrancurel (approved):

lib/models/ObjectMD.js
```diff
@@ -1028,6 +1030,33 @@ class ObjectMD {
         return this;
     }
 
+    /**
+     * Create or update the microVersionId field
+     *
+     * This field can be used to force an update in MongoDB (in case
+     * no other metadata field changes) or to manage conflicts during
+     * concurrent updates, using conditions on the microVersionId
+     * field.
```
vrancurel: Would be interesting to mention that it is also useful to detect a change when fields change but object version does not change (e.g. when ingesting a putObjectTagging coming from S3C to Zenko).

vrancurel: ... And bring clean active-active semantics for S3 calls that are not clean (like putObjectTagging for instance).

tests/unit/models/ObjectMD.js
```diff
@@ -257,6 +257,25 @@ describe('ObjectMD class setters/getters', () => {
...
+
+    it('ObjectMD::microVersionId set', () => {
+        const generatedIds = new Set();
+        for (let i = 0; i < 100; ++i) {
+            md.updateMicroVersionId();
+            generatedIds.add(md.getMicroVersionId());
+        }
+        // all generated IDs should be different
+        assert.strictEqual(generatedIds.size, 100);
+        generatedIds.forEach(key => {
```
naren-rajendran, line 273: Could checking length of `microVersionId` be a separate test? It will be cleaner!

jonathan-gramain: IMHO this is a good place to check that the length of all entries is good as well as we just generated the set of version IDs. I don't think it deserves a test on its own.

naren-rajendran (approved):

lib/models/ObjectMD.js
jonathan-gramain: done

jonathan-gramain:
@bert-e approve

rahulreddy (approved):

PR #980, opened 2020-04-14
Correct UTF-16 surrogates URI encoding

pepijnve:
BF: Correctly encode high-order codepoints in awsURIencode

awsURIencode iterates over and encodes UTF-16 values one by one. As a consequence high order codepoints that are split into two surrogates in UTF-16 get encoded as two separate UTF-8 sequences that encode the codepoint value of each surrogate.
The correct way to encode this type of codepoint is to encode them as a single UTF-8 sequence representing the codepoint value of the recombined surrogates.

Signed-off-by: Pepijn Van Eeckhoudt <pepijn.vaneeckhoudt@datadobi.com>

### DIFF
--- a/lib/auth/v4/awsURIencode.js
+++ b/lib/auth/v4/awsURIencode.js
@@ -36,7 +36,7 @@ function awsURIencode(input, encodeSlash, noEncodeStar) {
     const encSlash = encodeSlash === undefined ? true : encodeSlash;
     let encoded = '';
     for (let i = 0; i < input.length; i++) {
-        const ch = input.charAt(i);
+        let ch = input.charAt(i);
         if ((ch >= 'A' && ch <= 'Z') ||
             (ch >= 'a' && ch <= 'z') ||
             (ch >= '0' && ch <= '9') ||
@@ -50,6 +50,20 @@ function awsURIencode(input, encodeSlash, noEncodeStar) {
         } else if (ch === '*') {
             encoded = encoded.concat(noEncodeStar ? '*' : '%2A');
         } else {
+            if (ch >= '\uD800' && ch <= '\uDBFF') {
+                // If this character is a high surrogate peek the next character
+                // and join it with this one if the next character is a low
+                // surrogate.
+                // Otherwise the encoded URI will contain the two surrogates as
+                // two distinct UTF-8 sequences which is not valid UTF-8.
+                if (i + 1 < input.length) {
+                    const ch2 = input.charAt(i + 1);
+                    if (ch2 >= '\uDC00' && ch2 <= '\uDFFF') {
+                        i++;
+                        ch += ch2;
+                    }
+                }
+            }
             encoded = encoded.concat(_toHexUTF8(ch));
         }
     }
--- a/tests/unit/auth/v4/awsURIencode.js
+++ b/tests/unit/auth/v4/awsURIencode.js
@@ -53,4 +53,12 @@ describe('should URIencode in accordance with AWS rules', () => {
         const actualOutput = awsURIencode(input);
         assert.strictEqual(actualOutput, expectedOutput);
     });
+
+    it('should encode codepoints that use surrogate pairs in UTF-16 as a ' +
+        'single UTF-8 sequence', () => {
+        const input = '/s3amazonaws.com/I-like-🌮s';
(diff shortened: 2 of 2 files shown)


lib/auth/v4/awsURIencode.js
```diff
@@ -36,7 +36,7 @@ function awsURIencode(input, encodeSlash, noEncodeStar) {
     const encSlash = encodeSlash === undefined ? true : encodeSlash;
     let encoded = '';
     for (let i = 0; i < input.length; i++) {
-        const ch = input.charAt(i);
+        var ch = input.charAt(i);
```
jonathan-gramain: I'd rather use `let` as this is a local variable.

jonathan-gramain (approved):
Looks good to me, thanks for your contribution @pepijnve!

pepijnve:
@jonathan-gramain I was testing this a bit more on my end and noticed there are still some issue in key listing. The logic that filters out out-of-order keys at (amongst other places) https://github.com/scality/Arsenal/blob/development/8.1/lib/algos/list/delimiter.js#L170 is using `<=` which compares using UTF-16 sort order. That can cause certain keys to go missing in certain situations. I have a fix ready for that as well consisting of a codepoint based String comparison function which I'll submit in a second PR tomorrow. Still need to write some test cases for that.

miniscruff (commented):
This code looks fine to me, but I feel like there is probably a javascript built-in or common package that will handle all this encoding for us, I am new but this just seems like something javascript should have. I found: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI which seems promising.

lib/auth/v4/awsURIencode.js
pepijnve: Are you ok with me force pushing this change or would you prefer individual commits and squash when merging?

pepijnve:
@miniscruff that looks even better indeed. I'm happy to change the patch to use encodeURI instead if you prefer.

lib/auth/v4/awsURIencode.js
jonathan-gramain: Yes, force push is good

jonathan-gramain:
@pepijnve go ahead if you wish to replace this function call by a standard `encodeURI` call if this solves the issues you're facing it's even better.

FYI, we might need to open a new PR on our side eventually to merge the fix because we're using a merge bot internally that has some requirements on branches, but the commit history will be kept.

pepijnve:
@jonathan-gramain in retrospect it might not be worth it. Looking at the [AWS SDK](https://github.com/aws/aws-sdk-java/blob/7b1e5b87b0bf03456df9e77716b14731adf9a7a7/aws-java-sdk-core/src/main/java/com/amazonaws/util/SdkHttpUtils.java#L66), you have to do some hoop jumping even if you use an existing URI encoder. The implementation you have at the moment is a close match to the example one in the [v4 signing documenation](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html#w78aab9c23c19b8); might be better to leave it as is at the moment. FWIW, the example implementation in the docs has the same problem as what we're fixing here.

rahulreddy:
@pepijnve You are right about the hoop jumping in AWS V4 which is why we had to write our custom encoding method at that time looking at AWS' docs.

rahulreddy:
@pepijnve  Sorry since our internal CI is not equipped with handling users outside of the org, I took the patch and ran it in the CI. Good news is that all test stages are passing except the linter stage. These are the three linter errors being thrown
```
lib/auth/v4/awsURIencode.js
  55:1   error  Line 55 exceeds the maximum line length of 80  max-len
  56:1   error  Line 56 exceeds the maximum line length of 80  max-len
  59:47  error  Infix operators must be spaced                 space-infix-ops

/tests/unit/auth/v4/awsURIencode.js
  57:1  error  Line 57 exceeds the maximum line length of 80  max-len
```
You can run npm run lint to check/verify these errors in console. You can fix them, amend and force push, I will merge the PR.

pepijnve:
@rahulreddy thanks for the pointers. I'll get that sorted out tomorrow.

pepijnve:
@rahulreddy should be all good now. I'll already run the linter on and make fixes in PR #986 as well.

rahulreddy:
@pepijnve I will be merging your changes here https://github.com/scality/Arsenal/pull/1001

PR #981, opened 2020-04-14
bugfix: Remove tag regex to allow utf8 characters

miniscruff:
Simply removing the regex check should allow utf8 characters. I added some tests as there were none.


miniscruff:
@bert-e reset

miniscruff:
recreating to rename branch

PR #982, opened 2020-04-14
bugfix: Remove tag regex to allow utf8 characters

miniscruff:
Remove regex check, add tests and update error message to be more specific

### DIFF
--- a/tests/unit/s3middleware/tagging.js
+++ b/tests/unit/s3middleware/tagging.js
@@ -1,0 +1,21 @@
+const assert = require('assert');
+const { _validator } = require('../../../lib/s3middleware/tagging');
+
+describe('tagging validator', () => {
+    it('validates keys and values are less than 128 and 256', () => {
+        const result = _validator.validateKeyValue('hey', 'you guys');
+        assert.strictEqual(result, true);
+    });
+    it('returns error for keys greater than 128', () => {
+        const result = _validator.validateKeyValue('Y'.repeat(200), 'you guys');
+        assert(result instanceof Error);
+    });
+    it('returns error for values greater than 256', () => {
+        const result = _validator.validateKeyValue('X', 'Z'.repeat(300));
+        assert(result instanceof Error);
+    });
+    it('allows any utf8 string in keys and values', () => {
+        const result = _validator.validateKeyValue('あいう', '😀😄');
+        assert.strictEqual(result, true);
+    });
+});
--- a/lib/s3middleware/tagging.js
+++ b/lib/s3middleware/tagging.js
@@ -3,8 +3,6 @@ const { parseString } = require('xml2js');
 const errors = require('../errors');
 const escapeForXml = require('./escapeForXml');
 
-const tagRegex = new RegExp(/[^a-zA-Z0-9 +-=._:/]/g);
-
 const errorInvalidArgument = errors.InvalidArgument
   .customizeDescription('The header \'x-amz-tagging\' shall be ' +
   'encoded as UTF-8 then URLEncoded URL query parameters without ' +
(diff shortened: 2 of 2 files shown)


tests/unit/s3middleware/tagging.js
```diff
@@ -0,0 +1,21 @@
...
+        const result = _validator.validateKeyValue('hey', 'you guys');
+        assert.strictEqual(result, true);
+    });
+    it('returns error for keys greater than 128', () => {
+        const result = _validator.validateKeyValue('Y'.repeat(200), 'you guys');
+        assert(result instanceof Error);
+    });
+    it('returns error for values greater than 256', () => {
+        const result = _validator.validateKeyValue('X', 'Z'.repeat(300));
+        assert(result instanceof Error);
```
ilkescality, line 15: 👍

tests/unit/s3middleware/tagging.js
```diff
@@ -0,0 +1,21 @@
...
+    });
+    it('allows any utf8 string in keys and values', () => {
+        const result = _validator.validateKeyValue('あいう', '😀😄');
```
ilkescality, line 18: 👍

ilkescality (approved):

jonathan-gramain (approved):
The fix looks good to me, I recommend adding the ticket number in the commit summary e.g. after `bugfix:`, it can help for backtracking changes later.

miniscruff:
> The fix looks good to me, I recommend adding the ticket number in the commit summary e.g. after `bugfix:`, it can help for backtracking changes later.

@jonathan-gramain 

As part of the one-liner? Like:
`bugfix: S3C-2668 allow utf8 encoded keys and tags`

Or as part of the summary on a new line?

Was referencing https://github.com/scality/Guidelines/blob/development/8.1/CONTRIBUTING.md of course and it seemed like the ticket was inferred from the branch name but can of course adjust.

Edit: Branch names would only help if the merge strategy was "Merge Commit" I think. Should be able to grep for that in one of the git outputs I would imagine...

jonathan-gramain:
@miniscruff yes, as part of the one-liner is good.

The bots are happy as long as the branch name contains the ticket number (and the ticket fixVersion and PR target branch match as well), but having the ticket number to the commit message is useful for future reference when browsing the history.

tmacro (approved):

tests/unit/s3middleware/tagging.js
```diff
@@ -0,0 +1,21 @@
...
+const { _validator } = require('../../../lib/s3middleware/tagging');
+
+describe('tagging validator', () => {
+    it('validates keys and values are less than 128 and 256', () => {
+        const result = _validator.validateKeyValue('hey', 'you guys');
+        assert.strictEqual(result, true);
+    });
+    it('returns error for keys greater than 128', () => {
+        const result = _validator.validateKeyValue('Y'.repeat(200), 'you guys');
+        assert(result instanceof Error);
```
rachedbenmustapha, line 11: There is `assert.throws()` for this case.

miniscruff: the validation doesn't actually throw an error though it just returns one. ( I commented on this in slack about it ) It would fail CI/CD if it threw one.

rachedbenmustapha: I didn't know if this API was already widely used or not. If it's already used everywhere, no point in changing, but otherwise it's a suboptimal API to just return errors as if the call were asynchronous while it's not.

miniscruff: I agree but it felt a bit out of scope for my task, it also sounds like a backward-incompatible change we would have to fix whatever does use it. And at the very least the method is being used in S3C when putting new tags on objects.

rachedbenmustapha (approved):

PR #986, opened 2020-04-15
Do not filter keys that use codepoints outside the BMP when listing

pepijnve:
This PR is related to #980.

When an upper bound (lt or lte) is not provided, level-sublevel uses a default upperbound of `<= 0xFFFF`. This incorrectly filters out any keys using codepoints outside the BMP because the keys in leveldb are UTF-8 encoded and leveldb uses binary comparison when comparing keys.

A similar issue exists in Arsenal itself where correct key ordering is verified in an extra filter on the keys during listing.

This PR resolves the first issue by avoiding the usage of the default level-sublevel filter.
The second issue is resolved by using a custom string comparison function that compare codepoint values instead of UTF-16 chars. This is effectively the same as UTF-8 binary ordering.

I'm not sure if the default filter avoidance is in the correct place. Currently it's applied to all backend that use the list algorithms; perhaps this should be pushed down to file backend specific code?

### DIFF
--- a/lib/algos/list/tools.js
+++ b/lib/algos/list/tools.js
@@ -3,6 +3,7 @@ const SKIP_NONE = undefined; // to be inline with the values of NextMarker
 const FILTER_ACCEPT = 1;
 const FILTER_SKIP = 0;
 const FILTER_END = -1;
+const UNICODE_MAX = String.fromCodePoint(0x10FFFF);
 
 /**
  * This function check if number is valid
@@ -38,4 +39,5 @@ module.exports = {
     FILTER_END,
     FILTER_SKIP,
     FILTER_ACCEPT,
+    UNICODE_MAX
 };
--- a/lib/algos/list/tools.js
+++ b/lib/algos/list/tools.js
@@ -37,16 +37,17 @@ function inc(str) {
  * @function
  * @param {String} s1 the first string to compare
  * @param {String} s2 the second string to compare
- * @return {number} -1, 0, or 1 if s1 is less than, equal or greater than s2 respectively
+ * @return {number} -1, 0, or 1 if s1 is less than, equal or greater than s2
+ *  respectively
  */
 function utf8Compare(s1, s2) {
     const l1 = s1.length;
     const l2 = s2.length;
-    const l = Math.min(l1, l2)
+    const l = Math.min(l1, l2);
 
     for (let i = 0; i < l; i++) {
-        let cp1 = s1.codePointAt(i);
-        let cp2 = s2.codePointAt(i);
+        const cp1 = s1.codePointAt(i);
+        const cp2 = s2.codePointAt(i);
 
         if (cp1 < cp2) {
             return -1;
@@ -59,13 +60,7 @@ function utf8Compare(s1, s2) {
         }
     }
 
-    if (l1 < l2) {
-        return -1;
-    } else if (l1 > l2) {
-        return 1;
-    } else {
-        return 0;
-    }
+    return Math.sign(l1 - l2);
 }
 
 module.exports = {
@@ -76,5 +71,5 @@ module.exports = {
     FILTER_END,
     FILTER_SKIP,
     FILTER_ACCEPT,
-    UNICODE_MAX
+    UNICODE_MAX,
 };
--- a/lib/algos/list/tools.js
+++ b/lib/algos/list/tools.js
@@ -32,9 +32,46 @@ function inc(str) {
             String.fromCharCode(str.charCodeAt(str.length - 1) + 1)) : str;
 }
 
+/*
+ * Compares two Strings, s1 and s2, using UTF-8 lexicographic ordering.
+ * @function
+ * @param {String} s1 the first string to compare
+ * @param {String} s2 the second string to compare
+ * @return {number} -1, 0, or 1 if s1 is less than, equal or greater than s2 respectively
+ */
+function utf8Compare(s1, s2) {
+    const l1 = s1.length;
+    const l2 = s2.length;
+    const l = Math.min(l1, l2)
+
+    for (let i = 0; i < l; i++) {
+        let cp1 = s1.codePointAt(i);
+        let cp2 = s2.codePointAt(i);
+
+        if (cp1 < cp2) {
+            return -1;
+        } else if (cp1 > cp2) {
+            return 1;
+        }
+
+        if (cp1 > 0xFFFF) {
+            i++;
+        }
+    }
+
+    if (l1 < l2) {
+        return -1;
+    } else if (l1 > l2) {
+        return 1;
+    } else {
+        return 0;
+    }
+}
+
 module.exports = {
     checkLimit,
     inc,
+    utf8Compare,
     SKIP_NONE,
     FILTER_END,
     FILTER_SKIP,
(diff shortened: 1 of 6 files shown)


lib/algos/list/tools.js
```diff
@@ -31,11 +32,49 @@ function inc(str) {
...
 }
 
+/*
+ * Compares two Strings, s1 and s2, using UTF-8 lexicographic ordering.
+ * @function
+ * @param {String} s1 the first string to compare
+ * @param {String} s2 the second string to compare
+ * @return {number} -1, 0, or 1 if s1 is less than, equal or greater than s2 respectively
+ */
+function utf8Compare(s1, s2) {
```
miniscruff, line 43: Hi, me again. This seems very similar to what localeCompare probably does, have you tried using that? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

pepijnve: If I understand it correctly localeCompare does locale sensitive collation. That's not what you want here; it needs to be locale independent UTF-8 binary order.

pepijnve: I had another look at the description on MDN and the two pointers to the relevant Ecma specifications. There's a lot of room for 'implementation dependent behaviour' which makes localeCompare unsuitable for this usage. Amazon is fairly vague on this, but https://docs.aws.amazon.com/AmazonS3/latest/dev/ListingKeysUsingAPIs.html does state
> List results are always returned in UTF-8 binary order.

miniscruff: It seems like the optional arguments will have this type of comparison but I can not find the exact one as I don't fully understand the problem.

pepijnve: You're going to go down a rabbit hole here 😄

Looking at V8 you end up at https://github.com/v8/v8/blob/4b9b23521e6fd42373ebbcb20ebe03bf445494f9/src/builtins/builtins-string.cc#L135
which, assuming ICU support has been compiled in, delegates to
https://github.com/v8/v8/blob/4b9b23521e6fd42373ebbcb20ebe03bf445494f9/src/objects/intl-objects.cc#L947
which in turn creates an ICU collator and then compares using that.

Making some assumption here, but you're most likely going get a Unicode collator that uses the system's default locale if you do not specify any options. You can find the details of what that is exactly at https://www.unicode.org/reports/tr10/#Main_Algorithm. In a nutshell this is sorting for humans taking language/culture specific details into account. That's not what you want here.
I had hoped to be able to find a locale code or an option value that says 'I want codepoint order', but as far as I can tell that does not exist.

UTF-8 binary order means encoding both strings as UTF-8 and comparing byte-wise. Thanks to the way UTF-8 is constructed this is identical to the ordering you get when you take the codepoint values of the string and compare those integer-wise.

You might be wondering why the default UTF-16 comparison isn't good enough. Best way I can explain that is with an example.
Take codepoints [48](http://www.fileformat.info/info/unicode/char/0030/index.htm), [65104](https://www.fileformat.info/info/unicode/char/fe50/index.htm) and [129648](http://www.fileformat.info/info/unicode/char/1fa70/index.htm). That's 0,﹐and 🩰 respectively.
Sorting in codepoint order you get them in exactly that order.

If you encode in UTF-16 you get
```
0x0030
0xFE50
0xD83E 0xDE70
```
respectively. UTF-16 binary sorting compares each 16-bit value one by one so that would return the string in the order [0, 🩰, ﹐].

Encode in UTF-8 and you get
```
0x30
0xEF 0xB9 0x90
0xF0 0x9F 0xA9 0xB0
```
respectively. UTF-8 binary sorting compares each 8-bit value one by one and then you get [0,﹐,🩰] which is what you want.

This is by design in UTF-8. Binary sort order for UTF-8 is identical to codepoint order.

miniscruff: Yea that's fair, I had assumed it was something similar to that affect but was hoping there was something built in as this seems like it would be easy and common enough. I am new to scality so I won't be able to merge this but its got my approval if that means anything.

lib/algos/list/tools.js
```diff
@@ -31,11 +32,49 @@ function inc(str) {
...
+    const l1 = s1.length;
+    const l2 = s2.length;
+    const l = Math.min(l1, l2)
+
+    for (let i = 0; i < l; i++) {
```
miniscruff, line 48: Side note, I was trying to find a way to convert this to more functional javascript with .map and .filter but JS doesn't seem to have a nice zip method like python does to loop over two iterables. Something along the lines of
```javascript
const c1 = Array.from(s1).map(c => c.codePointAt(0)).filter(c => c <= 0xffff);
const c2 = Array.from(s2).map(c => c.codePointAt(0)).filter(c => c <= 0xffff);
// but you would still have to loop over the indecies anyway
// plus having to go from string to array is kinda lame
```

Sure wish you could map over a string in javascript... seems a bit weird you can not.
Anyway, this was just me messing around.

lib/algos/list/tools.js
```diff
@@ -31,11 +32,49 @@ function inc(str) {
...
+        } else if (cp1 > cp2) {
+            return 1;
+        }
+
+        if (cp1 > 0xFFFF) {
+            i++;
+        }
+    }
+
+    if (l1 < l2) {
```
miniscruff: This if, else if, else block can just be `return Math.sign(l1 - l2);` ( I am pretty sure anyway )
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/sign

lib/algos/list/tools.js
```diff
@@ -31,11 +32,49 @@ function inc(str) {
...
+ * @param {String} s2 the second string to compare
+ * @return {number} -1, 0, or 1 if s1 is less than, equal or greater than s2 respectively
+ */
+function utf8Compare(s1, s2) {
+    const l1 = s1.length;
+    const l2 = s2.length;
+    const l = Math.min(l1, l2)
+
+    for (let i = 0; i < l; i++) {
+        let cp1 = s1.codePointAt(i);
```
miniscruff: You could declare these as `const`s

lib/algos/list/tools.js
pepijnve: I tried to make this thing a bit more elegant as well, but I reckoned it could be rather performance critical so I went for an imperative, zero-allocation style in the end. Not pretty, but it does the job.

lib/algos/list/tools.js
pepijnve: Sure, I'm open to suggestions here. Not sure what the best practice is TBH. I haven't touched Javascript for 5-6 years and back then var was all you had.

lib/algos/list/tools.js
miniscruff: Yea was thinking the same

lib/algos/list/tools.js
miniscruff: const is for anything that doesn't get reassigned ( pushing to a const array is fine ). Let is for values that do change, such as `i` here. Var is basically unused at this point and considered bad practice. Eslint might fail you for it depending on settings I think.

rahulreddy:
@pepijnve I will ask the team to review this, one more review I will merge it in the branch. You can squash your commits after that 😃

ilkescality (approved):
LGTM

PR #988, opened 2020-04-22
bugfix: S3C-2726 remove some default attributes from ObjectMD

jonathan-gramain:
Remove "nullVersionId", "isNull" and "isDeleteMarker" default values
from ObjectMD model, instead of the previous '' (empty string) default
value that was incorrect and could cause an issue by misinterpreting
the empty "nullVersionId" as an actual null version ID.

The original issue as reported in the Jira ticket is tested in a new integration test case: https://github.com/scality/Integration/pull/731


lib/models/ObjectMD.js
```diff
@@ -106,10 +106,12 @@ class ObjectMD {
             },
             'key': '',
             'location': null,
-            'isNull': '',
-            'nullVersionId': '',
-            'isDeleteMarker': '',
-            'versionId': undefined, // If no versionId, it should be undefined
+            // versionId, isNull, nullVersionId and isDeleteMarker
+            // should be undefined when not set explicitly
+            'isNull': undefined,
```
nicolas2bert, line 111: 'isNull' and 'isDeleteMarker' are booleans and could be set to false by default?

jonathan-gramain: I wanted to stay consistent with what cloudserver writes when it puts a new object, which has no such attributes in the metadata.

nicolas2bert (approved):

naren-rajendran (approved):

rahulreddy (commented):
The reason these attributes are an empty string is because `undefined` ones don't make it past `JSON.stringify`. If there are going to be conditional attributes that may break some parts of the code where the checks are not done properly, right?

jonathan-gramain:
Some parts of the code might break, that's why we need to check that all components behave properly with this change. I don't think it would be related to JSON.stringify() which just ignores the undefined properties:
```
> const obj = { a: 42, b: undefined }
undefined
> JSON.stringify(obj)
'{"a":42}'
```

jonathan-gramain:
@rahulreddy if you mean by this that the attributes will not be present in the metadata if they are undefined, yes, and it's already the case since those attributes are not set by Cloudserver when they are not needed today (Cloudserver does not use ObjectMD for most parts).

ilkescality (approved):

tmacro (approved):

rahulreddy (approved):

jonathan-gramain:
@bert-e approve

PR #995, opened 2020-05-01
ft: S3C-2789 put objlock bucketinfo update

dora-korpar:
The necessary updates to BucketInfo and supporting methods for PutObjectLockConfiguration API. Full ObjectLockConfiguration model to follow.
Also updates Model Version of BucketInfo model.
To be merged after https://github.com/scality/Arsenal/pull/992

PR #996, opened 2020-05-02
Feature/s3 c 2789 put object lock configuration

dora-korpar:
ObjectLockConfiguration model
To be merged after https://github.com/scality/Arsenal/pull/995

### DIFF
--- a/lib/models/ObjectLockConfiguration.js
+++ b/lib/models/ObjectLockConfiguration.js
@@ -1,5 +1,7 @@
 const assert = require('assert');
 
+const errors = require('../errors');
+
 /**
  * Format of xml request:
  *
@@ -9,7 +11,7 @@ const assert = require('assert');
  *          <DefaultRetention>
  *              <Mode>GOVERNANCE|COMPLIANCE</Mode>
  *              <Days>1</Days>
- *              <Years>1</Years>z
+ *              <Years>1</Years>
  *          </DefaultRetention>
  *      </Rule>
  * </ObjectLockConfiguration>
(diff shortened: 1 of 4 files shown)


lib/models/ObjectLockConfiguration.js
```diff
@@ -0,0 +1,183 @@
...
+/**
+ * Format of xml request:
+ *
+ * <ObjectLockConfiguration>
+ *      <ObjectLockEnabled>Enabled</ObjectLockEnabled>
+ *      <Rule>
+ *          <DefaultRetention>
+ *              <Mode>GOVERNANCE|COMPLIANCE</Mode>
+ *              <Days>1</Days>
+ *              <Years>1</Years>z
```
ilkescality: Noticed an extra `z` after the last `>`

lib/models/ObjectLockConfiguration.js
```diff
@@ -0,0 +1,183 @@
...
+            validConfig.error = validMode.error;
+            return validConfig;
+        }
+        const validTime = this._parseTime(drArray[0]);
+        if (validTime.error) {
+            validConfig.error = validTime.error;
+            return validConfig;
+        }
+
+        this._config.rule = {};
```
ilkescality, line 178: I think in addition to the `rule`, `objectLockEnabled` property could be added to the config as well (as it seems to be a part of the `objectLockConfiguration` type per AWS specs), but also, as I wrote the getBucketObjectLock in a way that it returns the response in the following format anyway (by adding ObjectLockEnabled property if object lock is enabled):
```
{
    "ObjectLockConfiguration": {
        "ObjectLockEnabled": "Enabled",
        "Rule": {
            "DefaultRetention": {
                "Mode": "COMPLIANCE",
                "Days": 50
            }
        }
    }
}
```
so I couldn't be sure if it would necessarily needed to be added, think we should be good either way.

lib/models/ObjectLockConfiguration.js
```diff
@@ -0,0 +1,183 @@
...
+        this._config.rule.mode = validMode.mode;
+        this._config.rule[validTime.timeType] = validTime.timeValue;
+        return validConfig;
+    }
+
+    /**
+     * Validate the bucket metadata lifecycle configuration structure and
```
ilkescality: Is `lifecycle` meant to be `object lock` ?

lib/models/ObjectLockConfiguration.js
```diff
@@ -0,0 +1,183 @@
...
+     * value types
+     * @param {object} config - The lifecycle configuration to validate
```
ilkescality: same here for: lifecycle -> object lock

ilkescality (changes requested):
I noticed a missed edge-case; when a `PUT / objectLockConfiguration` request with a retention of 0 days or 0 years made, AWS responds with an invalid argument error: `An error occurred (InvalidArgument) when calling the PutObjectLockConfiguration operation: Default retention period must be a positive integer value.` 
But current implementation puts the object lock configuration with 0 or negative retention values with success.

In order to cover the zero or negative retention value case, could we add a new conditional into the _parseObjectLockConfig method to check for 0-day or 0-year retention values (and additional corresponding unit tests to cover these scenarios) ?

lib/models/ObjectLockConfiguration.js
ilkescality: Nevermind about this one; we’re actually already storing this exactly as `ObjectLockEnabled` in the bucket info already. :)

lib/models/ObjectLockConfiguration.js
```diff
@@ -0,0 +1,188 @@
...
+            validMode.error = errors.MalformedXML.customizeDescription(
+                'request xml does not contain Mode');
+            return validMode;
+        }
+        if (!expectedModes.includes(mode[0])) {
+            validMode.error = errors.MalformedXML.customizeDescription(
+                'Mode request xml must be one of "GOVERNANCE", "COMPLIANCE"');
+            return validMode;
+        }
+        validMode.mode = mode[0];
```
jonathan-gramain, line 76: Could be worth checking also that `mode` is a single-element array, if that makes sense as it seems to be the only supported case.

lib/models/ObjectLockConfiguration.js
```diff
@@ -0,0 +1,188 @@
...
+                'request xml does not contain Days or Years');
+            return validTime;
+        }
+        const timeValue = Number.parseInt(dr[timeType][0], 10);
+        if (Number.isNaN(timeValue)) {
+            validTime.error = errors.MalformedXML.customizeDescription(
+                'request xml does not contain valid retention period');
+            return validTime;
+        }
+        if (timeValue < 1) {
```
jonathan-gramain, line 109: Does the retention have a higher bound? If so, could be checked here, by experience there is usually a higher bound defined somewhere even if it's the language-defined integer limit, which could be checked anyway, to avoid silent integer-wrapping issues that would be bad if the user is not aware.

lib/models/ObjectLockConfiguration.js
```diff
@@ -0,0 +1,188 @@
...
+     */
+    _parseTime(dr) {
+        const validTime = {};
+        if (dr.Days && dr.Years) {
+            validTime.error = errors.MalformedXML.customizeDescription(
+                'request xml contains both Days and Years');
+            return validTime;
+        }
+        const timeType = dr.Days ? 'Days' : 'Years';
+        if (!dr[timeType] || !dr[timeType][0] || dr[timeType][0] === '') {
```
jonathan-gramain: Same comment than above about checking whether there are extra items in `dr[timeType]`, and about the superfluous check `dr[timeType][0] === ''`

lib/models/ObjectLockConfiguration.js
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+     * Check that mode is valid
+     * @param {array} mode - array containing mode value
+     * @return {object} - contains error if parsing failed
+     */
+    _parseMode(mode) {
+        const validMode = {};
+        const expectedModes = ['GOVERNANCE', 'COMPLIANCE'];
+        if (!mode || !mode[0] || mode[0] === '') {
```
jonathan-gramain: I think the check `mode[0] === ''` is superfluous because `!mode[0]` would return `true` in this case.

lib/models/ObjectLockConfiguration.js
```diff
@@ -0,0 +1,188 @@
...
+     */
+    _parseObjectLockConfig() {
+        const validConfig = {};
+        if (!this._parsedXml || this._parsedXml === '') {
+            validConfig.error = errors.MalformedXML.customizeDescription(
+                'request xml is undefined or empty');
+            return validConfig;
+        }
+        const objectLockConfig = this._parsedXml.ObjectLockConfiguration;
+        if (!objectLockConfig && objectLockConfig !== '') {
```
jonathan-gramain: I think this need parenthesis like: `if (!(objectLockConfig && objectLockConfig !== ''))` or should be written as the previous check with `||`

lib/models/ObjectLockConfiguration.js
```diff
@@ -0,0 +1,188 @@
...
+            validConfig.error = errors.MalformedXML.customizeDescription(
+                'request xml does not include ObjectLockConfiguration');
+            return validConfig;
+        }
+        const objectLockEnabled = objectLockConfig.ObjectLockEnabled;
+        if (!objectLockEnabled || objectLockEnabled[0] !== 'Enabled') {
```
jonathan-gramain, line 143: isn't `['Disabled']` a valid case that should be accepted?

tests/unit/models/ObjectLockConfiguration.js
```diff
@@ -0,0 +1,168 @@
...
+    }
+    return `<Rule><DefaultRetention>${mode}${time}</DefaultRetention></Rule>`;
+}
+
+function generateXml(testParams) {
+    const Enabled = testParams.key === 'ObjectLockEnabled' ?
+        `<ObjectLockEnabled>${testParams.value}</ObjectLockEnabled>` :
+        '<ObjectLockEnabled>Enabled</ObjectLockEnabled>';
+    const Rule = generateRule(testParams);
+    const ObjectLock = testParams.key === 'ObjectLockConfiguration' ? '' :
```
jonathan-gramain, line 41: I'm confused by this line, what if the key was `EmptyObjectLockConfiguration` for example?

lib/models/ObjectLockConfiguration.js
jonathan-gramain: Which I'm thinking must come from the XML parser if a user specifies multiple mode tags.

lib/models/ObjectLockConfiguration.js
dora-korpar: Nope. Object lock can't be disabled. It's kind of a redundant field in the config 🤷‍♀️

tests/unit/models/ObjectLockConfiguration.js
dora-korpar: it's just following the same pattern for setting up the object lock config to test like the rest of the keys. The only difference for `ObjectLockConfiguration` is that its only error condition is if it's empty

lib/models/ObjectLockConfiguration.js
jonathan-gramain: You forgot this one :wink:

jonathan-gramain (approved):

ilkescality (approved):

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e approve

PR #1002, opened 2020-05-11
bugfix: MD-712 tooling class to merge two sorted streams

jonathan-gramain:
Create class MergeStream to merge two readable sorted stream into one
readable stream, providing a comparison function.

This class is used to implement listing in bucket versioning key
format v1, that requires listing master keys and version keys
synchronously.


jonathan-gramain:
Replaced by https://github.com/scality/Arsenal/pull/1003

PR #1009, opened 2020-05-12
ft: S3C-2790 get object lock configuration

ilkescality:
This PR includes the route, custom error, helper `getConfigXML` method, and the unit tests for the `getConfigXML` method for the `bucketGetObjectLock` API.

Related to https://github.com/scality/cloudserver/pull/2535

### DIFF
--- a/lib/models/ObjectLockConfiguration.js
+++ b/lib/models/ObjectLockConfiguration.js
@@ -199,6 +199,38 @@ class ObjectLockConfiguration {
         }
     }
 
+    /**
+     * Get the XML representation of the configuration object
+     * @param {object} config - The bucket object lock configuration
+     * @return {string} - The XML representation of the configuration
+     */
+    static getConfigXML(config) {
+        // object lock is enabled on the bucket but object lock configuration
+        // not set
+        if (config.rule === undefined) {
+            return '<?xml version="1.0" encoding="UTF-8"?>' +
+                '<ObjectLockConfiguration ' +
+                'xmlns="http://s3.amazonaws.com/doc/2006-03-01/">' +
+                '<ObjectLockEnabled>Enabled</ObjectLockEnabled>' +
+                '</ObjectLockConfiguration>';
+        }
+        const { days, years, mode } = config.rule;
+        const Mode = `<Mode>${mode}</Mode>`;
+        const Days = days !== undefined ? `<Days>${days}</Days>` : '';
+        const Years = years !== undefined ? `<Years>${years}</Years>` : '';
+        const Time = Days || Years;
+        return '<?xml version="1.0" encoding="UTF-8"?>' +
+            '<ObjectLockConfiguration ' +
+            'xmlns="http://s3.amazonaws.com/doc/2006-03-01/">' +
+            '<ObjectLockEnabled>Enabled</ObjectLockEnabled>' +
+            '<Rule>' +
+            '<DefaultRetention>' +
+            `${Mode}` +
+            `${Time}` +
+            '</DefaultRetention>' +
+            '</Rule>' +
+            '</ObjectLockConfiguration>';
+    }
 }
 
 module.exports = ObjectLockConfiguration;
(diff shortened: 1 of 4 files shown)


nicolas2bert (approved):

lib/models/ObjectLockConfiguration.js
```diff
@@ -199,6 +199,38 @@ class ObjectLockConfiguration {
         }
     }
 
+    /**
+     * Get the XML representation of the configuration object
+     * @param {object} config - The bucket object lock configuration
+     * @return {string} - The XML representation of the configuration
+     */
+    static getConfigXML(config) {
```
jonathan-gramain, line 207: Could be good to have some basic tests for this function

jonathan-gramain: I just saw there are some in cloudserver. Could it make more sense to have the unit tests moved to arsenal? At least those that do not have a dependency on cloudserver.

jonathan-gramain: Since it's a model class, it should be easy to test part of it as unit tests, the functional tests that use the S3 API make sense too but it's good to have some local unit tests as well.

ilkescality: Yeah, like you mentioned it’s indirectly getting unit tested in cloudserver, also I realized we don’t have unit tests for any of the similar static class methods under other configuration models.
I still wanted to add some unit tests instead of going along with the current pattern though, yet, it hasn’t been as easy as I imagined due to the limitations of testing a static class method by calling it on the instance of the class (which I need to be doing for the unit test). 

As an alternative, I made the method non-static to be able to call it on the instance of the class (as static method calls are made directly on the class and are not callable on instances of the class) and to be able to unit test it, but suggestions for a better approach are welcome.

jonathan-gramain: You should be able to test static methods by calling them like `ObjectLockConfiguration.getConfigXML(...)`, I think? Or maybe I'm missing your point.

ilkescality: I think it should work by calling it on the class, but for the unit test, I needed to call it on the instance of the class which is when it becomes tricky. But I didn't necessarily have to call it on the instance in the first place, in fact, for the unit test, probably it should suffice to just unit-test the method alone by calling it on the class. I'll do that.

jonathan-gramain: Yes that's what I mean, you can test this method separately as a class method test, and have other instance method tests.

dora-korpar (approved):

ilkescality:
@bert-e reset

lib/models/ObjectLockConfiguration.js
ilkescality: @jonathan-gramain tests added, can you take another look ?

tests/unit/models/ObjectLockConfiguration.js
```diff
@@ -178,3 +246,14 @@ describe('ObjectLockConfiguration class getValidatedObjectLockConfiguration',
...
     });
 });
+
+describe('ObjectLockConfiguration class getConfigXML', () => {
+    passTestsGetConfigXML.forEach(test => {
+        const { config, description, expectedXml } = test;
+        it(`should return correct XML ${description}`, done => {
+            const responseXml = ObjectLockConfiguration.getConfigXML(config);
+            assert.strictEqual(responseXml, expectedXml);
+            done();
```
jonathan-gramain: minor point: since it's synchronous you don't need a `done` function, you can declare it without parameter as `() => {...}` and not call `done()` at the end.

jonathan-gramain (approved):

tests/unit/models/ObjectLockConfiguration.js
ilkescality: you're right; updated 👍

ilkescality:
@bert-e reset

ilkescality:
@bert-e force-reset

ilkescality:
@bert-e force_reset

ilkescality:
@bert-e reset

ilkescality:
@bert-e approve

PR #1017, opened 2020-05-18
bugfix: S3C-2899 support vFormat v1 for MPU listing

jonathan-gramain:
Support listing MPUs stored with versioning key format v1

### DIFF
--- a/lib/algos/list/MPU.js
+++ b/lib/algos/list/MPU.js
@@ -1,10 +1,10 @@
 'use strict'; // eslint-disable-line strict
 
-const errors = require('../../errors');
-const { inc, checkLimit, FILTER_END, FILTER_ACCEPT } = require('./tools');
+const { inc, checkLimit, listingParamsMasterKeysV0ToV1,
+        FILTER_END, FILTER_ACCEPT } = require('./tools');
 const DEFAULT_MAX_KEYS = 1000;
 const VSConst = require('../../versioning/constants').VersioningConstants;
-const { BucketVersioningKeyFormat } = VSConst;
+const { DbPrefixes, BucketVersioningKeyFormat } = VSConst;
 
 function numberDefault(num, defaultNum) {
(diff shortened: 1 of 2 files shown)


tests/utils/performListing.js
```diff
@@ -1,5 +1,9 @@
-module.exports = function performListing(data, Extension, params, logger) {
-    const listing = new Extension(params, logger);
+const assert = require('assert');
+
+module.exports = function performListing(data, Extension, params, logger, vFormat) {
+    const listing = new Extension(params, logger, vFormat);
+    const mdParams = listing.genMDParams();
+    assert(typeof mdParams === 'object');
```
miniscruff: I have been wondering this for a while now and just want to confirm. Assets like these are not removed during any sort of run or precompile step right? In python, asserts are removed in some cases of running so are not recommended as a means of runtime checks.

miniscruff (approved):

tests/utils/performListing.js
jonathan-gramain: Yes, the asserts are kept in the production code.

Generally speaking our guidelines discourage from using them in running code (i.e. not tests) because of the risk of provoking exceptions instead of proper error management. I personally think that if wisely used they can be beneficial, but we have to make sure we don't use them in place of error management.

miniscruff: Sounds good, I would have to see what these look like when actually running to see how I liked or disliked them. Coming from python these sorts of assertions are either required as part of general processing due to type conditions or just completely ignored with the expectation that an error is raised by design.

jonathan-gramain:
@bert-e reset

tests/utils/performListing.js
jonathan-gramain: I just changed this into a `assert.strictEqual` since this is more on par with the guidelines.

alexanderchan-scality (approved):

jonathan-gramain:
@bert-e approve

PR #1018, opened 2020-05-19
bugfix: S3C-2899 helper for v1 genMDParams() of master keys listing

jonathan-gramain:
New helper function to convert listing params from v0 to v1, when a
listing of master keys is requested. This logic is shared between
DelimiterMaster and MPU listing, hence a shared helper is useful.

Also, update the test function performListing to prepare for v1
testing of listing algos, by adding the vFormat parameter.

**Note**: I rebased on an eslint commit hash change to benefit from the longer line length allowed.

### DIFF
--- a/lib/algos/list/tools.js
+++ b/lib/algos/list/tools.js
@@ -1,3 +1,5 @@
+const { DbPrefixes } = require('../../versioning/constants').VersioningConstants;
+
 // constants for extensions
 const SKIP_NONE = undefined; // to be inline with the values of NextMarker
 const FILTER_ACCEPT = 1;
@@ -31,9 +33,36 @@ function inc(str) {
             String.fromCharCode(str.charCodeAt(str.length - 1) + 1)) : str;
 }
 
+/**
+ * Transform listing parameters for v0 versioning key format to make
+ * it compatible with v1 format
+ *
+ * @param {object} v0params - listing parameters for v0 format
+ * @return {object} - listing parameters for v1 format
+ */
+function listingParamsMasterKeysV0ToV1(v0params) {
+    const v1params = Object.assign({}, v0params);
+    if (v0params.gt !== undefined) {
+        v1params.gt = `${DbPrefixes.Master}${v0params.gt}`;
+    } else if (v0params.gte !== undefined) {
+        v1params.gte = `${DbPrefixes.Master}${v0params.gte}`;
+    } else {
+        v1params.gte = DbPrefixes.Master;
+    }
+    if (v0params.lt !== undefined) {
+        v1params.lt = `${DbPrefixes.Master}${v0params.lt}`;
+    } else if (v0params.lte !== undefined) {
+        v1params.lte = `${DbPrefixes.Master}${v0params.lte}`;
+    } else {
+        v1params.lt = inc(DbPrefixes.Master); // stop after the last master key
+    }
+    return v1params;
+}
+
 module.exports = {
     checkLimit,
     inc,
+    listingParamsMasterKeysV0ToV1,
     SKIP_NONE,
     FILTER_END,
     FILTER_SKIP,
--- a/tests/unit/algos/list/tools.js
+++ b/tests/unit/algos/list/tools.js
@@ -2,7 +2,10 @@
 
 const assert = require('assert');
 
-const checkLimit = require('../../../../lib/algos/list/tools').checkLimit;
+const { checkLimit, inc, listingParamsMasterKeysV0ToV1 } =
+      require('../../../../lib/algos/list/tools');
+const VSConst = require('../../../../lib/versioning/constants').VersioningConstants;
+const { DbPrefixes } = VSConst;
 
 describe('checkLimit function', () => {
     const tests = [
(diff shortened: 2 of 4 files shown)


lib/algos/list/tools.js
```diff
@@ -31,9 +34,29 @@ function inc(str) {
             String.fromCharCode(str.charCodeAt(str.length - 1) + 1)) : str;
 }
 
+function listingParamsMasterKeysV0ToV1(v0params) {
+    const v1params = Object.assign({}, v0params);
```
miniscruff, line 44: Just a preference thing but the spread operator should work here.
`const v1Params = {...v0params};` less typing but same result.

lib/algos/list/tools.js
```diff
@@ -31,9 +34,29 @@ function inc(str) {
...
+    if (v0params.gt !== undefined) {
```
miniscruff, line 45: Since you copied the v0 data here, it might be good to use v1params instead of v0 across the board.

lib/algos/list/tools.js
```diff
@@ -1,3 +1,6 @@
+const VSConst = require('../../versioning/constants').VersioningConstants;
```
miniscruff: Could combine these two lines into one, I didn't see VSConst referenced anywhere.

miniscruff (approved):
Just preference ideas

lib/algos/list/tools.js
jonathan-gramain: The idea is to check against what is provided from the input, because we are going to transform an input into an output, so I think `v0params.gt` makes sense.

lib/algos/list/tools.js
jonathan-gramain: Yes it's just to fetch the DbPrefixes. I copied this pattern from other files that do this, but indeed I could make it a single line.

jonathan-gramain:
I'm also going to add some JSDoc to the helper.

lib/algos/list/tools.js
jonathan-gramain: Our linter config does not like this construct, so will leave it as is.

miniscruff: Oh, might have an older version of eslint. Oh well.

lib/algos/list/tools.js
```diff
@@ -31,9 +33,36 @@ function inc(str) {
...
+        v1params.gte = `${DbPrefixes.Master}${v0params.gte}`;
+    } else {
+        v1params.gte = DbPrefixes.Master;
+    }
+    if (v0params.lt !== undefined) {
+        v1params.lt = `${DbPrefixes.Master}${v0params.lt}`;
+    } else if (v0params.lte !== undefined) {
+        v1params.lte = `${DbPrefixes.Master}${v0params.lte}`;
+    } else {
+        v1params.lt = inc(DbPrefixes.Master); // stop after the last master key
```
miniscruff, line 57: I just want to make sure that it is intended that if lt and lte are missign we set lt but if gt and gte are missing we set gte. And not that they should both be *t or *te together.

lib/algos/list/tools.js
miniscruff: Right, just commenting that since you copy the input straight into the output on the first line. We could just check the output across the board since they are the same. Or just set the output to an empty object and merge the two at the end. Would have identical results. But just a preference thing here.

tests/unit/algos/list/tools.js
```diff
@@ -23,3 +26,79 @@ describe('checkLimit function', () => {
...
+                gte: DbPrefixes.Master,
+                lt: inc(DbPrefixes.Master),
+            },
+        },
+        {
+            v0params: {
+                gt: 'foo/bar',
+            },
+            v1params: {
+                gt: `${DbPrefixes.Master}foo/bar`,
```
miniscruff, line 44: kinda feels to be like there should be a slash before foo for `${Master}/foo/bar` but that might just be me.

lib/algos/list/tools.js
jonathan-gramain: Yes, those are the maximum boundaries if none is provided in v0 (so `gte` and `lt` indeed)

tests/unit/algos/list/tools.js
jonathan-gramain: No slash, `DbPrefixes.Master` is only a db key prefix, it does not need a delimiter.

jonathan-gramain:
@bert-e reset

jonathan-gramain:
@bert-e reset

ilkescality:
LGTM - except for one question about your note on eslint; what's the way to keep it like this moving forward ? 
That got me curious as I got the linter errors for those 2 lines and was wondering how to keep it like this (without adding `/* eslint-disable max-len */`, `/* eslint-enable max-len */` in `delimiterVersions.js` which is basically what I'm doing locally to get around for now).

ilkescality (approved):

jonathan-gramain:
@ilkescality I bet what you need is just `yarn install`, maybe `rm -rf node_modules` first

jonathan-gramain:
@bert-e approve

PR #1028, opened 2020-05-19
bugfix: S3C-2899 support v1 in Delimiter, DelimiterMaster

jonathan-gramain:
The two listing methods Delimiter and DelimiterMaster now support v1
versioning key format in addition to v0.

Modify the listing algo classes to support buckets in v1 versioning
key format, in addition to v0.

Enhance existing unit tests to check the result of getMDParams()

### DIFF
--- a/lib/algos/list/delimiterMaster.js
+++ b/lib/algos/list/delimiterMaster.js
@@ -1,6 +1,5 @@
 'use strict'; // eslint-disable-line strict
 
-const errors = require('../../errors');
 const Delimiter = require('./delimiter').Delimiter;
 const Version = require('../../versioning/Version').Version;
 const VSConst = require('../../versioning/constants').VersioningConstants;
@@ -8,6 +7,7 @@ const { BucketVersioningKeyFormat } = VSConst;
 const { FILTER_ACCEPT, FILTER_SKIP, SKIP_NONE } = require('./tools');
 
 const VID_SEP = VSConst.VersionId.Separator;
+const { DbPrefixes } = VSConst;
 
 /**
  * Handle object listing with parameters. This extends the base class Delimiter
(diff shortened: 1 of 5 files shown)


miniscruff (approved):

lib/algos/list/delimiterMaster.js
```diff
@@ -120,7 +135,24 @@ class DelimiterMaster extends Delimiter {
...
 
+    filterV1(obj) {
+        // Filtering master keys in v1 is simply listing the master
+        // keys, as the state of version keys do not change the
+        // result, so we can use Delimiter method directly.
+        return super.filter(obj);
+    }
+
     skipping() {
+        if (this.vFormat === BucketVersioningKeyFormat.v0) {
```
miniscruff: Just want to mention the idea of having each version be either a class or an object instead of putting a bunch of these if's around. Could have just one if in a factory method somewhere then everything else is standard.
Mabye a base Bucket Version class if there is a lot of similarities.
Alternatively, an object that maps the v0 and v1 methods around in one place.
Something along the lines of: 
```javascript
const versionRouter = {
  [BucketVersioningKeyFormat.v0]: {
    skipping: this.skippingV0,
    other: this.otherv0,
  },
  [BucketVersioningKeyFormat.v1]: { etc }
}

// then map through this:
versionRouter[this.vFormat].skipping()
// make wrap it with a check the version exists if required
```

Anyway, just ideas.

tests/unit/algos/list/tools.js
```diff
@@ -23,3 +26,79 @@ describe('checkLimit function', () => {
...
+                limit: 5,
+            },
+            v1params: {
+                gte: `${DbPrefixes.Master}baz/qux`,
+                lte: `${DbPrefixes.Master}foo/bar`,
+                limit: 5,
+            },
+        },
+    ];
+    testCases.forEach(testCase => {
```
miniscruff: I am pretty sure you can reduce some typing here by doing this:
```javascript
testCases.forEach(({v0Params, v1Params}) => {
    // can then just use v0Params instead of testCase.v0Params
}
```

miniscruff (approved):

lib/algos/list/delimiterMaster.js
jonathan-gramain: It sounds better indeed since there is a common pattern, especially after the migration states are introduced (v0 to v1)

tests/unit/algos/list/tools.js
jonathan-gramain: Ah yeah, did not think about using object deconstruction here

lib/algos/list/delimiterMaster.js
jonathan-gramain: updated

tests/unit/algos/list/tools.js
jonathan-gramain: updated

jonathan-gramain:
@bert-e reset

jonathan-gramain:
@bert-e reset

alexanderchan-scality (approved):

jonathan-gramain:
@bert-e approve

PR #1031, opened 2020-05-19
bugfix: S3C-2899 implement v1 format for DelimiterVersions listing

jonathan-gramain:
Implement the v1 versioning key format for DelimiterVersions listing
method, in addition to v0.

Enhance existing unit tests to check the result of getMDParams()

### DIFF
--- a/lib/algos/list/delimiterVersions.js
+++ b/lib/algos/list/delimiterVersions.js
@@ -1,6 +1,5 @@
 'use strict'; // eslint-disable-line strict
 
-const errors = require('../../errors');
 const Delimiter = require('./delimiter').Delimiter;
 const Version = require('../../versioning/Version').Version;
 const VSConst = require('../../versioning/constants').VersioningConstants;
@@ -8,7 +7,7 @@ const { inc, FILTER_END, FILTER_ACCEPT, FILTER_SKIP, SKIP_NONE } =
     require('./tools');
 
 const VID_SEP = VSConst.VersionId.Separator;
-const { BucketVersioningKeyFormat } = VSConst;
+const { DbPrefixes, BucketVersioningKeyFormat } = VSConst;
 
 /**
  * Handle object listing with parameters
@@ -34,13 +33,19 @@ class DelimiterVersions extends Delimiter {
         // listing results
         this.NextMarker = parameters.keyMarker;
         this.NextVersionIdMarker = undefined;
-    }
 
-    genMDParams() {
-        if (this.vFormat === BucketVersioningKeyFormat.v0) {
-            return this.genMDParamsV0();
-        }
-        throw errors.NotImplemented;
+        Object.assign(this, {
+            [BucketVersioningKeyFormat.v0]: {
+                genMDParams: this.genMDParamsV0,
+                filter: this.filterV0,
+                skipping: this.skippingV0,
+            },
+            [BucketVersioningKeyFormat.v1]: {
+                genMDParams: this.genMDParamsV1,
+                filter: this.filterV1,
+                skipping: this.skippingV1,
+            },
+        }[this.vFormat]);
     }
 
     genMDParamsV0() {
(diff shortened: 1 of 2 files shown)


lib/algos/list/delimiterVersions.js
```diff
@@ -66,6 +69,59 @@ class DelimiterVersions extends Delimiter {
         return params;
     }
 
+    genMDParamsV1() {
+        // return an array of two listing params sets to ask for
+        // synchronized listing of M and V ranges
+        const params = [{}, {}];
+        if (this.parameters.prefix) {
+            params[0].gte = `${DbPrefixes.Master}${this.parameters.prefix}`;
```
miniscruff: If you are just concatenating two strings would a+b be faster? Or maybe more readable? Feels like the `${}` might add a little unnecessary noise here.
```
`${DbPrefixes.Master}${this.parameters.prefix}`; // or
DbPrefixes.Master + this.parameters.prefix;
```

Definitely doesn't matter just thought I would bring up the comparison.

lib/algos/list/delimiterVersions.js
```diff
@@ -66,6 +69,59 @@ class DelimiterVersions extends Delimiter {
...
+        }
+        if (this.parameters.keyMarker) {
+            if (params[1].gte <= `${DbPrefixes.Version}${this.parameters.keyMarker}`) {
+                delete params[0].gte;
+                delete params[1].gte;
+                params[0].gt = `${DbPrefixes.Master}${inc(this.parameters.keyMarker + VID_SEP)}`;
+                if (this.parameters.versionIdMarker) {
+                    // versionIdMarker should always come with keyMarker
+                    // but may not be the other way around
+                    params[1].gt = `${DbPrefixes.Version}${this.parameters.keyMarker}` +
```
miniscruff: Similar to my other comment about adding two strings, here it could be:
`[DbPrefixes.Version, this.parameters.keyMarker, VID_SEP, this.parameters.versionIdMarker].join("");`
Probably just as long but might make it more clear its just joining 4 parameters. "Might" being a keyword there :).

lib/algos/list/delimiterVersions.js
```diff
@@ -66,6 +69,59 @@ class DelimiterVersions extends Delimiter {
...
+     *   * -1 if obj1 < obj2
+     *   * 1 if obj1 > obj2
+     */
+    compareObjects(obj1, obj2) {
+        const key1 = obj1.key.slice(DbPrefixes.Master.length);
+        const key2 = obj2.key.slice(DbPrefixes.Version.length);
+        if (key1 < key2) {
+            return -1;
+        }
+        return 1;
```
miniscruff: I assume they can not be equal then? You could also ternary this since it's pretty short.

miniscruff (approved):

lib/algos/list/delimiterVersions.js
jonathan-gramain: Yes that could work, I tend to use this pattern when it comes to building strings, also because our beloved linter does not like when we concatenate both string templates and regular strings together, it asks for only string templates or only regular strings (but here it would be happy I think with what you suggest).

miniscruff: Right if you are combining a string and variable or combination it makes sense. But 2+ variables together I think should be a + or join. String templates are great for combinations of strings and variables though.

lib/algos/list/delimiterVersions.js
jonathan-gramain: Yeah likely, that was part of the original code so did not change it, but could be worth simplifying some parts of it.

lib/algos/list/delimiterVersions.js
jonathan-gramain: Indeed, key1 always have a `M` prefix and key2 a `V` prefix, so are never equal.

jonathan-gramain: I changed the variable names and JSDoc to make it clearer, and use a ternary.

lib/algos/list/delimiterVersions.js
jonathan-gramain: updated

lib/algos/list/delimiterVersions.js
jonathan-gramain: updated

jonathan-gramain:
@bert-e reset

jonathan-gramain:
@miniscruff I also implemented what you suggested in #1028 about having an object for mapping format version numbers to method names. It should also (slightly) improve performance by avoiding an explicit comparison on the version number at each call, explicitly for the `filter()` method which is called for every entry to list.

jonathan-gramain:
@bert-e reset

jonathan-gramain:
@bert-e reset

lib/algos/list/delimiterVersions.js
```diff
@@ -34,13 +33,19 @@ class DelimiterVersions extends Delimiter {
...
         this.NextMarker = parameters.keyMarker;
         this.NextVersionIdMarker = undefined;
-    }
 
-    genMDParams() {
-        if (this.vFormat === BucketVersioningKeyFormat.v0) {
-            return this.genMDParamsV0();
-        }
-        throw errors.NotImplemented;
+        Object.assign(this, {
```
miniscruff, line 37: :sparkles:

alexanderchan-scality (approved):

jonathan-gramain:
@bert-e approve

PR #1049, opened 2020-05-21
ft: S3C 2787 put object retention

dora-korpar:
Adds necessary changes for Put Object Retention API.
Includes new route, new ObjectMD field, and tests.

### DIFF
--- a/lib/s3middleware/objectRetention.js
+++ b/lib/s3middleware/objectRetention.js
@@ -1,0 +1,135 @@
+const { parseString } = require('xml2js');
+
+const constants = require('../constants');
+const errors = require('../errors');
+
+/*
+    Format of xml request:
+    <ObjectRetention>
+      <Mode>COMPLIANCE|GOVERNANCE</Mode>
+      <RetainUntilDate>2020-05-20T04:58:45.413000Z</RetainUntilDate>
+    </ObjectRetention>
+*/
+
+/**
+ * validateMode - validate retention mode
+ * @param {array} mode - parsed xml mode array
+ * @return {object} - contains mode or error
+ */
+function validateMode(mode) {
+    const modeObj = {};
+    const expectedModes = new Set(['GOVERNANCE', 'COMPLIANCE']);
+    if (!mode || !mode[0]) {
+        modeObj.error = errors.MalformedXML.customizeDescription(
+            'request xml does not contain Mode');
+        return modeObj;
+    }
+    if (mode.length > 1) {
+        modeObj.error = errors.MalformedXML.customizeDescription(
+            'request xml contains more than one Mode');
+        return modeObj;
+    }
+    if (!expectedModes.has(mode[0])) {
+        modeObj.error = errors.MalformedXML.customizeDescription(
+            'Mode request xml must be one of "GOVERNANCE", "COMPLIANCE"');
+        return modeObj;
+    }
+    modeObj.mode = mode[0];
+    return modeObj;
+}
+
+/**
+ * validateRetainDate - validate retain until date
+ * @param {array} retainDate - parsed xml retention date array
+ * @return {object} - contains retain until date or error
+ */
+function validateRetainDate(retainDate) {
+    const dateObj = {};
+    if (!retainDate || !retainDate[0]) {
+        dateObj.error = errors.MalformedXML.customizeDescription(
+            'request xml does not contain RetainUntilDate');
+        return dateObj;
+    }
+    if (!constants.iso8601Regex.test(retainDate[0])) {
+        dateObj.error = errors.InvalidRequest.customizeDescription(
+            'RetainUntilDate timestamp must be ISO-8601 format');
+        return dateObj;
+    }
+    const date = new Date(retainDate[0]);
+    if (date < Date.now()) {
+        dateObj.error = errors.InvalidRequest.customizeDescription(
+            'RetainUntilDate must be in the future');
+        return dateObj;
+    }
+    dateObj.date = retainDate[0];
+    return dateObj;
+}
+
+/**
+ * validate retention - validate retention xml
+ * @param {object} parsedXml - parsed retention xml object
+ * @return {object} - contains retention information on success,
+ * error on failure
+ */
+function validateRetention(parsedXml) {
+    const retentionObj = {};
+    if (!parsedXml) {
+        retentionObj.error = errors.MalformedXML.customizeDescription(
+            'request xml is undefined or empty');
+        return retentionObj;
+    }
+    const retention = parsedXml.ObjectRetention;
+    if (!retention) {
+        retentionObj.error = errors.MalformedXML.customizeDescription(
+            'request xml does not contain ObjectRetention');
+        return retentionObj;
+    }
+    const modeObj = validateMode(retention.Mode);
+    if (modeObj.error) {
+        retentionObj.error = modeObj.error;
+        return retentionObj;
+    }
+    const dateObj = validateRetainDate(retention.RetainUntilDate);
+    if (dateObj.error) {
+        retentionObj.error = dateObj.error;
+        return retentionObj;
+    }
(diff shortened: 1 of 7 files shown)


dora-korpar:
@bert-e reset

dora-korpar:
@bert-e reset

miniscruff:
I left some similar comments on another PR, going to wait for those to resolve/be commented on before repeating myself.

tmacro (approved):

lib/s3middleware/objectRetention.js
```diff
@@ -0,0 +1,125 @@
...
+
+/*
+    Format of xml request:
+    <Retention>
+      <Mode>COMPLIANCE|GOVERNANCE</Mode>
+      <RetainUntilDate>2020-05-20T04:58:45.413000Z</RetainUntilDate>
+    </Retention>
+*/
+
+function validateMode(mode) {
```
miniscruff, line 19: Should we comment on missing jsdocs?

lib/s3middleware/objectRetention.js
```diff
@@ -0,0 +1,125 @@
...
+    const validMode = {};
+    const expectedModes = ['GOVERNANCE', 'COMPLIANCE'];
```
miniscruff: Set's are better for pure lookup collections

lib/s3middleware/objectRetention.js
```diff
@@ -0,0 +1,125 @@
...
+}
+
+function validateRetainDate(retainDate) {
+    const validRetainDate = {};
+    if (!retainDate || !retainDate[0]) {
+        validRetainDate.error = errors.MalformedXML.customizeDescription(
+            'request xml does not contain RetainUntilDate');
+        return validRetainDate;
+    }
+    if (!/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(retainDate[0])) {
```
miniscruff: Although this looks right, I can't help but think javascript has a built-in date checker. Or something close. Might need a plugin but even ISO-8601 I believe has a few variants or optional components that we would need to check for.

lib/s3middleware/objectRetention.js
```diff
@@ -0,0 +1,125 @@
...
+/*
+    Format of xml request:
+    <Retention>
+      <Mode>COMPLIANCE|GOVERNANCE</Mode>
+      <RetainUntilDate>2020-05-20T04:58:45.413000Z</RetainUntilDate>
+    </Retention>
+*/
+
+function validateMode(mode) {
+    const validMode = {};
```
miniscruff: Calling these variables valid* is a bit misleading as sometimes they are just errors. Maybe a more generic term like "results" or something that doesn't imply a state?

lib/s3middleware/objectRetention.js
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * validate retention - validate retention xml
+ * @param {object} parsedXml - parsed retention xml object
+ * @return {object} - contains retention information on success,
+ * error on failure
+ */
+function validateRetention(parsedXml) {
+    const validRetention = {};
+    if (!parsedXml || parsedXml === '') {
```
miniscruff: I think empty string is counted in the first check

lib/s3middleware/objectRetention.js
```diff
@@ -0,0 +1,125 @@
...
+    return validRetention;
+}
+
+/**
+ * parseRetentionXml - Parse and validate xml body, returning callback with
+ * object retentionInfo: { mode: <value>, retainUntilDate: <value> }
+ * @param {string} xml - xml body to parse and validate
+ * @param {object} log - Werelogs logger
+ * @param {function} cb - callback to server
+ * @return {Error|object} - returns object with retention information on
```
miniscruff: Nothing is returned here

ilkescality (approved):
LGTM - I got comment on my PR that instead of directly adding the fields to the constructor (ObjectMD) by default, using only the setter could be beneficial to avoid adding unnecessary optional fields for all objects even when object lock is not enabled. Not sure if we'd want similar approach here too, but just in case.

lib/s3middleware/objectRetention.js
miniscruff: hey @dora-korpar you marked this as resolved but I don't see any change or response mentioning why the code should be kept.

dora-korpar:
@bert-e reset

lib/s3middleware/objectRetention.js
jonathan-gramain: If there's no standard function to check it, it could also likely be a reusable constant regexp.

lib/s3middleware/objectRetention.js
```diff
@@ -0,0 +1,134 @@
...
+function validateRetainDate(retainDate) {
+    const dateObj = {};
+    if (!retainDate || !retainDate[0]) {
+        dateObj.error = errors.MalformedXML.customizeDescription(
+            'request xml does not contain RetainUntilDate');
+        return dateObj;
+    }
+    if (!/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(retainDate[0])) {
+        dateObj.error = errors.InvalidRequest.customizeDescription(
+            'retain until date timestamp must be ISO-8601 format');
```
jonathan-gramain: Could be better to reference the attribute name: `RetainUntilDate must be ...`

lib/s3middleware/objectRetention.js
```diff
@@ -0,0 +1,134 @@
...
+        return dateObj;
+    }
+    const date = new Date(retainDate[0]);
+    if (date < Date.now()) {
+        dateObj.error = errors.InvalidRequest.customizeDescription(
+            'retain until date must be in the future');
```
jonathan-gramain: same

lib/s3middleware/objectRetention.js
```diff
@@ -0,0 +1,134 @@
...
+ * error on failure
+ */
+function validateRetention(parsedXml) {
+    const retentionObj = {};
+    if (!parsedXml) {
+        retentionObj.error = errors.MalformedXML.customizeDescription(
+            'request xml is undefined or empty');
+        return retentionObj;
+    }
+    const retention = parsedXml.ObjectRetention;
```
jonathan-gramain, line 81: I'm confused to see `ObjectRetention` here but the description on top of file says `<Retention>...`, as well as the following error messages. Might be correct but I'm a bit puzzled.

dora-korpar: Should be ObjectRetention. I changed it in the code, but forgot to change it in the comments 👍

dora-korpar:
@jonathan-gramain not sure why I can't reply to your comment about the regex, but there's no reliable, native way I could find to check dates in Javascript. There is a module I could use called MomentJS, which I've used in Cloudserver for date comparisons. Do you think it would be worth it to introduce a new module into Arsenal as well? Or just keep the regex?

jonathan-gramain:
@dora-korpar IMHO no need to introduce a module just for this, but we could still have this regexp put in some arsenal source file holding various constants (something like constants.js probably), with a proper name to know how/when it can be reused, because I'm thinking that there could be more than one place where validation of such date format would be needed in the future or even in existing code. Alternatively we can make a helper function in a shared utility place to validate such date.

jonathan-gramain (approved):

miniscruff:
@jonathan-gramain @dora-korpar Have you tried Date.parse? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse

dora-korpar:
@miniscruff we decided to stay away from using the Date object since it can behave unpredictably

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e approve

PR #1052, opened 2020-05-26
feature: S3C-2944 put object legal hold

ilkescality:
- Adds setter and getter to get/set object's legal hold status.
- Adds helper methods to be consumed by PUT/GET Object legal hold APIs and the corresponding unit tests.
- Adds the route to `objectPutLegalHold`.

### DIFF
--- a/lib/models/ObjectMD.js
+++ b/lib/models/ObjectMD.js
@@ -884,6 +884,24 @@ class ObjectMD {
     }
 
     /**
+     * Set object legal hold status
+     * @param {boolean} legalHold - true if legal hold is 'ON' false if 'OFF'
+     * @return {ObjectMD} itself
+     */
+    setLegalHold(legalHold) {
+        this._data.legalHold = legalHold || false;
+        return this;
+    }
+
+    /**
+     * Get object legal hold status
+     * @return {boolean} legal hold status
+     */
+    getLegalHold() {
+        return this._data.legalHold || false;
+    }
+
+    /**
      * Returns metadata object
      *
      * @return {object} metadata object
@@ -891,6 +909,7 @@ class ObjectMD {
     getValue() {
         return this._data;
     }
+
 }
 
 module.exports = ObjectMD;
--- a/lib/s3middleware/objectLegalHold.js
+++ b/lib/s3middleware/objectLegalHold.js
@@ -1,0 +1,112 @@
+const { parseString } = require('xml2js');
+const errors = require('../errors');
+
+/*
+    Format of the xml request:
+    <LegalHold>
+        <Status>ON|OFF</Status>
+    </LegalHold>
+*/
+
+/**
+ * @param {string[]} status - legal hold status parsed from xml to be validated
+ * @return {Error|object} - legal hold status or error
+ */
+function _validateStatus(status) {
+    const validatedStatus = {};
+    const expectedValues = new Set(['OFF', 'ON']);
+    if (!status || status[0] === '') {
+        validatedStatus.error = errors.MalformedXML.customizeDescription(
+            'request xml does not contain Status');
+        return validatedStatus;
+    }
+    if (status.length > 1) {
+        validatedStatus.error = errors.MalformedXML.customizeDescription(
+            'request xml contains more than one Status');
+        return validatedStatus;
+    }
+    if (!expectedValues.has(status[0])) {
+        validatedStatus.error = errors.MalformedXML.customizeDescription(
+            'Status request xml must be one of "ON", "OFF"');
+        return validatedStatus;
+    }
+    validatedStatus.status = status[0];
+    return validatedStatus;
+}
+
+/**
+ * validate legal hold - validates legal hold xml
+ * @param {object} parsedXml - parsed legal hold xml object
+ * @return {object} - object with status or error
+ */
+function _validateLegalHold(parsedXml) {
+    const validatedLegalHold = {};
+    if (!parsedXml) {
+        validatedLegalHold.error = errors.MalformedXML.customizeDescription(
+            'request xml is undefined or empty');
+        return validatedLegalHold;
+    }
+    if (!parsedXml.LegalHold) {
+        validatedLegalHold.error = errors.MalformedXML.customizeDescription(
+            'request xml does not contain LegalHold');
+        return validatedLegalHold;
+    }
+    const validatedStatus = _validateStatus(parsedXml.LegalHold.Status);
+    if (validatedStatus.error) {
+        validatedLegalHold.error = validatedStatus.error;
+        return validatedLegalHold;
+    }
+    validatedLegalHold.status = validatedStatus.status;
+    return validatedLegalHold;
+}
+
+/**
+ * parse object legal hold - parse and validate xml body
+ * @param {string} xml - xml body to parse and validate
+ * @param {object} log - werelogs logger
+ * @param {function} cb - callback to server
+ * @return {undefined} - calls callback with legal hold status or error
+ */
+function parseLegalHoldXml(xml, log, cb) {
+    parseString(xml, (err, result) => {
+        if (err) {
+            log.debug('xml parsing failed', {
+                error: { message: err.message },
+                method: 'parseLegalHoldXml',
+                xml,
+            });
+            return cb(errors.MalformedXML);
+        }
+        const validatedLegalHold = _validateLegalHold(result);
+        const validatedLegalHoldStatus = validatedLegalHold.status === 'ON';
+        if (validatedLegalHold.error) {
+            log.debug('legal hold validation failed', {
+                error: { message: validatedLegalHold.error.message },
+                method: 'parseLegalHoldXml',
+                xml,
+            });
+            return cb(validatedLegalHold.error);
+        }
+        return cb(null, validatedLegalHoldStatus);
+    });
+}
+
+/**
+ * convert to xml - generates legal hold xml
+ * @param {boolean} legalHold - true if legal hold is on else false
+ * @return {string} - returns legal hold xml
+ */
+function convertToXml(legalHold) {
+    const xml = `<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+        <LegalHold>
+        <Status>
+        ${legalHold ? 'ON' : 'OFF'}
+        </Status>
+        </LegalHold>`;
+    return xml;
+}
+
+module.exports = {
+    convertToXml,
+    parseLegalHoldXml,
+};
--- a/lib/policyEvaluator/RequestContext.js
+++ b/lib/policyEvaluator/RequestContext.js
@@ -54,6 +54,7 @@ const _actionMap = {
     objectPut: 's3:PutObject',
     objectPutACL: 's3:PutObjectAcl',
     objectPutACLVersion: 's3:PutObjectVersionAcl',
+    objectPutLegalHold: 's3:PutObjectLegalHold',
     objectPutPart: 's3:PutObject',
     objectPutTagging: 's3:PutObjectTagging',
     objectPutTaggingVersion: 's3:PutObjectVersionTagging',
--- a/lib/s3routes/routes/routePUT.js
+++ b/lib/s3routes/routes/routePUT.js
@@ -16,7 +16,6 @@ function routePUT(request, response, api, log, statsClient) {
             return routesUtils.responseNoBody(
               errors.BadRequest, null, response, null, log);
         }
-
         // PUT bucket ACL
         if (request.query.acl !== undefined) {
             api.callApiMethod('bucketPutACL', request, response, log,
@@ -87,8 +86,8 @@ function routePUT(request, response, api, log, statsClient) {
                 });
         }
     } else {
-        // PUT object, PUT object ACL, PUT object multipart or
-        // PUT object copy
+        // PUT object, PUT object ACL, PUT object multipart,
+        // PUT object copy or PUT object legal hold
         // if content-md5 is not present in the headers, try to
         // parse content-md5 from meta headers
 
@@ -146,6 +145,13 @@ function routePUT(request, response, api, log, statsClient) {
                     return routesUtils.responseNoBody(err, resHeaders,
                         response, 200, log);
                 });
+        } else if (request.query['legal-hold'] !== undefined) {
+            api.callApiMethod('objectPutLegalHold', request, response, log,
+                (err, resHeaders) => {
+                    routesUtils.statsReport500(err, statsClient);
+                    return routesUtils.responseNoBody(err, resHeaders,
+                        response, 200, log);
+                });
         } else if (request.query.tagging !== undefined) {
             api.callApiMethod('objectPutTagging', request, response, log,
                 (err, resHeaders) => {
@@ -174,7 +180,6 @@ function routePUT(request, response, api, log, statsClient) {
             log.end().addDefaultFields({
                 contentLength: request.parsedContentLength,
             });
-
             api.callApiMethod('objectPut', request, response, log,
             (err, resHeaders) => {
                 routesUtils.statsReport500(err, statsClient);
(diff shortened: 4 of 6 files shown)


lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,117 @@
...
+    });
+}
+
+/**
+ * parse object legal hold - Parse and validate xml body, returning callback with
+ * object legalHold: { status: ON | OFF }
+ * @param {object} legalHold - valid legal hold object
+ * @return {object|Error} - returns object with legal hold status on success,
+ * error on failure
+ */
```
miniscruff, line 98: might be worth mentioning backtick strings can be multiline without any sort of escaping or adding.

ilkescality: good point; updated with backticks.

lib/models/ObjectMD.js
```diff
@@ -891,6 +895,25 @@ class ObjectMD {
     getValue() {
         return this._data;
     }
+
```
jonathan-gramain: minor readability comment: the getter/setter could be moved just before `getValue` since it would be grouped with the other getters/setters (`getValue` is not really a getter but returns the whole object metadata).

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * validate legal hold - validate legal hold xml
+ * @param {object} parsedXml - parsed legal hold xml object
+ * @return {object} - contains legal hold information on success,
+ * error on failure
+ */
+function _validateLegalHold(parsedXml) {
+    const validatedLegalHold = {};
+    if (!parsedXml || parsedXml === '') {
```
jonathan-gramain: `if (!parsedXml)` would be enough, as `parsedXml === ''` implies `!parsedXml`.

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,117 @@
...
+        validatedLegalHold.error = errors.MalformedXML.customizeDescription(
+            'request xml is undefined or empty');
+        return validatedLegalHold;
+    }
+    const legalHold = parsedXml.LegalHold;
+    if (!legalHold || legalHold === '') {
```
jonathan-gramain: same

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,117 @@
...
+                method: 'parseLegalHoldXml',
+            });
+            log.debug('invalid xml', { xml });
+            return cb(errors.MalformedXML);
+        }
+        const validatedLegalHold = _validateLegalHold(result);
+        if (validatedLegalHold.error) {
+            log.debug('legal hold validation failed', {
+                error: validatedLegalHold.error,
+                method: 'validateLegalHold',
```
jonathan-gramain: Technically the method should be `parseLegalHoldXml`

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,117 @@
...
+ * @param {object} log - Werelogs logger
+ * @param {function} cb - callback to server
+ * @return {Error|object} - returns object with legal hold status on success,
+ * error on failure
+ */
+function parseLegalHoldXml(xml, log, cb) {
+    parseString(xml, (err, result) => {
+        if (err) {
+            log.trace('xml parsing failed', {
+                error: err,
```
jonathan-gramain: In order to cope with dynamically generated error messages, it's better to use: `error: { message: err.message }`, otherwise the error might not appear in the log in some cases.

Some people like to dump the backtrace as well (`err.stack`) although I personally think that it's seldom useful (mostly because of the async nature of most calls) and clutters the logs too much.

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,117 @@
...
+                method: 'parseLegalHoldXml',
+            });
+            log.debug('invalid xml', { xml });
+            return cb(errors.MalformedXML);
+        }
+        const validatedLegalHold = _validateLegalHold(result);
+        if (validatedLegalHold.error) {
+            log.debug('legal hold validation failed', {
+                error: validatedLegalHold.error,
```
jonathan-gramain: same: `error: { message: validatedLegalHold.error.message }`

lib/models/ObjectMD.js
ilkescality: done

lib/s3middleware/objectLegalHold.js
ilkescality: removed redundant 👍

lib/s3middleware/objectLegalHold.js
ilkescality: Yeah `parseLegalHoldXml` makes more sense here; updated.

lib/s3middleware/objectLegalHold.js
ilkescality: That's a good point.

jonathan-gramain (approved):

lib/models/ObjectMD.js
```diff
@@ -124,6 +124,10 @@ class ObjectMD {
                 dataStoreVersionId: '',
             },
             'dataStoreName': '',
+            // legalHold status can be 'ON' or 'OFF'
+            'legalHold': {
+                status: '',
```
rahulreddy: I think it would be beneficial if this is a boolean instead of values `ON`/`OFF` unless there is a value in having three states - disabled, ON or OFF. This will make a evaluation logic cleaner as you have to check on every object deletion if legal hold is set or not. For writing back to the client this can be converted to ON/OFF depending on the boolean value, so `'legalHold': true` or `legalHold: false` is sufficient.
Another thing I remember is that, we should not add `legalHold` to the constructor as it would add unnecessary optional fields for all objects even when object lock is not enabled. Using the setter and the getter should do the job.

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * validate status - validate status xml
+ * @param {object} status - parsed status object
+ * @return {object} - contains validated status object on success,
+ * error on failure
+ */
+function _validateStatus(status) {
+    const validatedStatus = {};
+    const expectedValues = ['OFF', 'ON'];
```
rahulreddy: Have you checked if `aws-cli` or `aws-sdk` allows lowecase/mixed-case usage of `ON/OFF`?

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,117 @@
...
+ * error on failure
+ */
+function parseLegalHoldXml(xml, log, cb) {
+    parseString(xml, (err, result) => {
+        if (err) {
+            log.trace('xml parsing failed', {
+                error: { message: err.message },
+                method: 'parseLegalHoldXml',
+            });
+            log.debug('invalid xml', { xml });
```
rahulreddy: We don't need both logs here, `debug` or `trace` should be sufficient.

rahulreddy (changes requested):

lib/models/ObjectMD.js
ilkescality: I iterated with that approach as well, I agree that it'd be cleaner and we don't (internally) need to have three states such as disabled, ON or OFF. Also makes sense to not to add the legalHold to the constructor by default as well - updating legalHold to be the type boolean as an optional field.

lib/s3middleware/objectLegalHold.js
ilkescality: It doesn't allow; I checked by passing  `On`, `off`, `Off`, `on` as the Status via the SDK and got 
```'MalformedXML: Status request xml must be one of "OFF", "ON"\n    at Request.extractError```

**Review by anurag4DSB — APPROVED** [review]

File: lib/models/ObjectMD.js [hunk 11]
```diff
@@ -883,6 +883,24 @@ class ObjectMD {
         return this;
     }
 
+    /**
+     * Set object legal hold status
+     * @param {boolean} legalHold - true if legal hold is 'ON' false if 'OFF'
+     * @return {ObjectMD} itself
+     */
+    setLegalHold(legalHold) {
```
miniscruff, line 891: I know its a bit late but javascript supports getters and setters now.

ilkescality: That's right, `set legalhol(val)` `get legalHold()`works as well; I thought it might worth to keep it consistent across the ObjectMD so just went with the flow.

rahulreddy (approved):

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,111 @@
...
+    Format of the xml request:
+    <LegalHold>
+        <Status>ON|OFF</Status>
+    </LegalHold>
+*/
+
+/**
+ * validate status - validates status xml
+ * @param {object} status - parsed status object
+ * @return {boolean} - true or false on success, error on failure
```
miniscruff: This doesn't seem right to me, generally speaking, functions should return one type. A method name like validate status to me seems like it would simply throw or return an error but not return anything. Maybe this is more acceptable in javascript than other languages but by returning two different types you force callees to conditional check by type. Where as if you just threw the error it would be a standard try/catch. Or maybe if you returned another object or split this method in two.

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,111 @@
...
+/*
+    Format of the xml request:
+    <LegalHold>
+        <Status>ON|OFF</Status>
+    </LegalHold>
+*/
+
+/**
+ * validate status - validates status xml
+ * @param {object} status - parsed status object
```
miniscruff: Based on the work in the method, I am not sure this is an object. It looks more like a string array with 1 item.

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,111 @@
...
+
+/*
+    Format of the xml request:
+    <LegalHold>
+        <Status>ON|OFF</Status>
+    </LegalHold>
+*/
+
+/**
+ * validate status - validates status xml
```
miniscruff: Not a very descriptive method summary.

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,111 @@
...
+ * @param {object} status - parsed status object
+ * @return {boolean} - true or false on success, error on failure
+ */
+function _validateStatus(status) {
+    const validatedStatus = {};
+    const expectedValues = ['OFF', 'ON'];
```
miniscruff: I would declare these once and reuse it, also Set is generally preferred for checking if an item exists in some collection as lookups are faster.

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,111 @@
...
+            'request xml is undefined or empty');
+        return validatedLegalHold;
+    }
+    if (!parsedXml.LegalHold) {
+        validatedLegalHold.error = errors.MalformedXML.customizeDescription(
+            'request xml does not contain LegalHold');
+        return validatedLegalHold;
+    }
+    const validatedStatus = _validateStatus(parsedXml.LegalHold.Status);
+    if (validatedStatus.error) {
```
miniscruff, line 55: Does this not error if validatedStatus returned a bool? Like bool has no property? Or JS is just like "meh whatever"?

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,111 @@
...
+    validatedLegalHold.status = validatedStatus;
+    return validatedLegalHold;
+}
+
+/**
+ * parse object legal hold - parse and validate xml body
+ * @param {string} xml - xml body to parse and validate
+ * @param {object} log - werelogs logger
+ * @param {function} cb - callback to server
+ * @return {function} - callback with legal hold boolean on success,
```
miniscruff: no function is returned here, not sure if the method is wrong or the comment.

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,111 @@
...
+    });
+}
+
+/**
+ * convert to xml - generates legal hold xml
+ * @param {boolean} legalHold - true if legal hold is on else false
+ * @return {string} - returns legal hold xml
+ */
+function convertToXml(legalHold) {
+    const status = legalHold ? 'ON' : 'OFF';
```
miniscruff: This can all be inlined and returned as one expression. Also might want to add a few more new lines to XML to be more human readable.

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,111 @@
...
+        return cb(null, validatedLegalHold.status);
+    });
+}
+
+/**
+ * convert to xml - generates legal hold xml
+ * @param {boolean} legalHold - true if legal hold is on else false
+ * @return {string} - returns legal hold xml
+ */
+function convertToXml(legalHold) {
```
miniscruff, line 99: Side note, I only see this used in the tests, not sure if that is intentional but test only code should be in test files.

miniscruff (changes requested):

lib/s3middleware/objectLegalHold.js
ilkescality: That makes sense. I tweaked it so that it always returns an object; either with error or with status.

lib/s3middleware/objectLegalHold.js
ilkescality: added `[]`

lib/s3middleware/objectLegalHold.js
ilkescality: As far as I tested it JS was just like "meh whatever" :) but I still tweaked my implementation a bit so it can never be bool anymore; as it's still better not to check property of boolean anyway.

lib/s3middleware/objectLegalHold.js
ilkescality: It's not test code; going to be consumed by cloudserver. I updated PR description.

lib/s3middleware/objectLegalHold.js
```diff
@@ -67,7 +65,7 @@ function _validateLegalHold(parsedXml) {
  * @param {string} xml - xml body to parse and validate
  * @param {object} log - werelogs logger
  * @param {function} cb - callback to server
- * @return {function} - callback with legal hold boolean on success,
+ * @return {Error|object} - calls callback with legal hold status on success,
```
miniscruff: I still think the @return here is incorrect as nothing is actually returned we just call the callback.

lib/s3middleware/objectLegalHold.js
miniscruff: ok, that is fine I just hadn't seen it used.

lib/s3middleware/objectLegalHold.js
```diff
@@ -0,0 +1,113 @@
...
+
+/*
+    Format of the xml request:
+    <LegalHold>
+        <Status>ON|OFF</Status>
+    </LegalHold>
+*/
+
+/**
+ * @param {object[]} status - legal hold status parsed from xml to be validated
```
miniscruff: is this an object or a string?

ilkescality: it's a string inside array (object supposed to be string)

lib/s3middleware/objectLegalHold.js
ilkescality: Yeah seems like across similar methods returning callback it's mostly listed as function so I updated to function. Let me know if you have any other suggestion instead of that. (I thought of `undefined` also but it didn't seem quite right either.)

miniscruff: I am fairly certain it is `undefined`, you can add a test case or adjust an existing test case to see what is returned. But if you do not return anything in JS you will inherently get undefined as a return value. Most languages do this as far as I know.

miniscruff (approved):

lib/s3middleware/objectLegalHold.js
ilkescality: Updated - didn't realize `undefined` was what you meant initially.

PR #1060, opened 2020-05-28
bugfix: S3C-2899 handle MergeStream.destroy()

jonathan-gramain:
Make sure MergeStream destroys the non-ended input streams when
destroy() is called

### DIFF
--- a/lib/algos/stream/MergeStream.js
+++ b/lib/algos/stream/MergeStream.js
@@ -32,6 +32,15 @@ class MergeStream extends stream.Readable {
         }
     }
 
+    _destroy(err, callback) {
+        for (let i = 0; i < 2; ++i) {
+            if (!this._streamEof[i]) {
+                this._streams[i].destroy();
+            }
+        }
+        callback();
+    }
+
     _onItem(myStream, myItem, myIndex, otherIndex) {
         this._peekItems[myIndex] = myItem;
         const otherItem = this._peekItems[otherIndex];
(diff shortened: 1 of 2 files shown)


tests/unit/algos/stream/MergeStream.spec.js
```diff
@@ -39,17 +47,18 @@ function readAll(stream, usePauseResume, cb) {
     stream.once('error', err => cb(err));
 }
 
+function compareInt(a, b) {
```
miniscruff, line 50: I think this can just be `return Math.sign(a - b);`

lib/algos/stream/MergeStream.js
```diff
@@ -32,6 +32,15 @@ class MergeStream extends stream.Readable {
         }
     }
 
+    _destroy(err, callback) {
+        for (let i = 0; i < 2; ++i) {
```
miniscruff, line 36: is 2 here all the items? This could use more functional code like:
```
this._streamEof.filter(x => x).forEach(stream => stream.destory())
```

Not sure where everyone stands on functional programming or not, I like it but also know it has limitations.

jonathan-gramain: In this case I'm checking `streamEof[i]` but destroying `stream[i]`, so it's useful to have the index.

jonathan-gramain:
@bert-e reset

lib/algos/stream/MergeStream.js
miniscruff: gotcha, I haven't quite gotten multi-list/array looping in javascript. In python we have zip but nothing quite like it in JS as far as I know.

miniscruff (approved):

jonathan-gramain:
@bert-e reset

tmacro (approved):

ilkescality (approved):

jonathan-gramain:
@bert-e approve

PR #1076, opened 2020-06-01
bugfix: S3C-2945 fix get legal hold route

ilkescality:
Fixes the route for GET Object legal hold API.

### DIFF
--- a/lib/s3routes/routes/routeGET.js
+++ b/lib/s3routes/routes/routeGET.js
@@ -103,16 +103,14 @@ function routerGET(request, response, api, log, statsClient, dataRetrievalFn) {
                     return routesUtils.responseXMLBody(err, xml, response, log,
                         corsHeaders);
                 });
-        } else if (request.query.tagging !== undefined) {
-            // GET object Tagging
-            api.callApiMethod('objectGetTagging', request, response, log,
+        } else if (request.query['legal-hold'] !== undefined) {
+            api.callApiMethod('objectGetLegalHold', request, response, log,
                 (err, xml, corsHeaders) => {
                     routesUtils.statsReport500(err, statsClient);
                     return routesUtils.responseXMLBody(err, xml, response, log,
                         corsHeaders);
                 });
         } else if (request.query.tagging !== undefined) {
-            // GET object legal hold
             api.callApiMethod('objectGetTagging', request, response, log,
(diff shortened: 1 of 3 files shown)


jonathan-gramain (approved):

miniscruff:
Bug fixes should have test case changes to accompany the fix.

miniscruff (changes requested):
Bugfix should have test case(s) to go with the fix.

naren-rajendran (approved):

lib/s3routes/routes/routeGET.js
```diff
@@ -103,16 +103,16 @@ function routerGET(request, response, api, log, statsClient, dataRetrievalFn) {
                     return routesUtils.responseXMLBody(err, xml, response, log,
                         corsHeaders);
                 });
-        } else if (request.query.tagging !== undefined) {
-            // GET object Tagging
-            api.callApiMethod('objectGetTagging', request, response, log,
+        } else if (request.query['legal-hold'] !== undefined) {
+            // GET object legal hold
```
miniscruff: this comment is probably not required given the if condition and call api method name handle this already

tests/unit/s3middleware/objectLegalHold.js
```diff
@@ -1,11 +1,13 @@
 const assert = require('assert');
-const { parseLegalHoldXml } = require('../../../lib/s3middleware/objectLegalHold');
+const { convertToXml, parseLegalHoldXml } =
+    require('../../../lib/s3middleware/objectLegalHold');
 const DummyRequestLogger = require('../helpers').DummyRequestLogger;
 
 const log = new DummyRequestLogger();
 
 function generateXml(status) {
```
miniscruff: Since you only use this in one describe block, if you move this into the describe block it will be closer to its use cases.

miniscruff (approved):

miniscruff:
@ilkescality once reviews start coming in can you use normal commits to address comments and not force commits? It is common to view commits to see the diffs but with only a single commit it is not possible to do that.

tests/unit/s3middleware/objectLegalHold.js
ilkescality: It is used in both, but moved bit closer.

dora-korpar (approved):

PR #1079, opened 2020-06-02
bf: do not sort when searching

vrancurel:
We change the default behavior of search by removing the sort directive.
Indeed some search queries may use other customer created index and
therefore cannot use the primary key _id, so it has to wait for
all results client side and sort them in memory which is highly
inefficient. Note that the sort might be useful when searching
so a long term fix shall have the sort optional.

### DIFF
--- a/lib/storage/metadata/mongoclient/readStream.js
+++ b/lib/storage/metadata/mongoclient/readStream.js
@@ -63,9 +63,13 @@ class MongoReadStream extends Readable {
             Object.assign(query, searchOptions);
         }
 
-        this._cursor = c.find(query).sort({
-            _id: options.reverse ? -1 : 1,
-        });
+        this._cursor = c.find(query);
+
+        if (!searchOptions) {
+            this._cursor.sort({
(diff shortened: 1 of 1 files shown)


vrancurel:
@bert-e help

vrancurel:
@bert-e wait

lib/storage/metadata/mongoclient/readStream.js
```diff
@@ -63,9 +63,13 @@ class MongoReadStream extends Readable {
             Object.assign(query, searchOptions);
         }
 
-        this._cursor = c.find(query).sort({
-            _id: options.reverse ? -1 : 1,
-        });
+        this._cursor = c.find(query);
+
+        if (!searchOptions) {
```
jonathan-gramain, line 68: I assume we sort when there are no search options provided because we can consider that sorting by `_id` is always efficient? Could deserve a comment since it does not look obvious.

jonathan-gramain (commented):
Could you add the ticket number in the commit message?

PR #1099, opened 2020-06-15
improvement: S3C-3044-add-audit-log-fields

anurag4DSB:
This PR adds audit log fields for Cloudserver without adding an extra log line

Extra fields:
accountDisplayName: account name
iamDisplayName: username

### DIFF
--- a/lib/auth/Vault.js
+++ b/lib/auth/Vault.js
@@ -24,6 +24,12 @@ function vaultSignatureCb(err, authInfo, log, callback, streamingV4Params) {
     const info = authInfo.message.body;
     const userInfo = new AuthInfo(info.userInfo);
     const authorizationResults = info.authorizationResults;
+    const auditLog = { accountDisplayName: userInfo.getAccountDisplayName() };
+    const iamDisplayName = userInfo.getIAMdisplayName();
+    if (iamDisplayName) {
+        auditLog.IAMdisplayName = iamDisplayName;
+    }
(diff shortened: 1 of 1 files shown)


ilkescality (approved):

lib/auth/Vault.js
```diff
@@ -24,6 +24,12 @@ function vaultSignatureCb(err, authInfo, log, callback, streamingV4Params) {
     const info = authInfo.message.body;
     const userInfo = new AuthInfo(info.userInfo);
     const authorizationResults = info.authorizationResults;
+    const auditLog = { accountDisplayName: userInfo.getAccountDisplayName() };
+    const iamDisplayName = userInfo.getIAMdisplayName();
+    if (iamDisplayName) {
```
miniscruff, line 29: wondering if it would be beneficial to list the display name even if it is null, for logging purposes. Not sure how helpful that would be though.

miniscruff (approved):

lib/auth/Vault.js
anurag4DSB: Good point! The requirement for the customer is only username, I added account name cuz I thought that was important. Adding user name as null will just extra add extra processing for the of leef log converter.  We already log too much in info :) 
So that’s why I didn’t add it .

naren-rajendran (approved):

jonathan-gramain (approved):

PR #1126, opened 2020-06-24
bugfix: S3C-2987 helper to get stream data as a JSON payload

jonathan-gramain:
Add a new helper function to get data from a stream as a JSON payload,
optionally validated against a joi schema.

Note: uses async/await, so updated the scality/Guidelines dependency
to please the linter

### DIFF
--- a/index.js
+++ b/index.js
@@ -121,4 +121,7 @@ module.exports = {
     pensieve: {
         credentialUtils: require('./lib/executables/pensieveCreds/utils'),
     },
+    stream: {
+        readJSONStreamObject: require('./lib/stream/readJSONStreamObject'),
+    },
 };
--- a/lib/stream/readJSONStreamObject.js
+++ b/lib/stream/readJSONStreamObject.js
@@ -1,0 +1,38 @@
+const errors = require('../errors');
+
+/**
+ * read a JSON object from a stream returned as a javascript object,
+ * handle errors.
+ *
+ * @param {stream.Readable} s - Readable stream
+ * @param {@hapi/joi} [joiSchema] - optional validation schema for the JSON object
+ * @return {Promise} a Promise resolved with the parsed JSON object as a result
+ */
+async function readJSONStreamObject(s, joiSchema) {
+    return new Promise((resolve, reject) => {
+        const contentsChunks = [];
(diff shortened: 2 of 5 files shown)


lib/stream/utils.js
```diff
@@ -0,0 +1,42 @@
...
+    return new Promise((resolve, reject) => {
+        const contentsChunks = [];
+        s.on('data', chunk => {
+            contentsChunks.push(chunk);
+        });
+        s.on('end', () => {
+            const contents = contentsChunks.join('');
+            try {
+                const parsedContents = JSON.parse(contents);
+                let validatedContents;
```
miniscruff: Could just resolve(value) and resolve(parsedContents) instead of creating this variable. Doesn't matter though, either way.

miniscruff (approved):

rahulreddy (commented):
One thing I suggest is instead of utils holding a couple of stream methods, export each method directly with its own file.
Instead of
-` lib/stream/utils.js`
```js
module.exports = {
    readJSONStreamObject,
};
```

use 
`+lib/stream/readJSONStreamObject.js`
```js
module.exports = readJSONStreamObject;
```

It's easy to discover and use the methods this way.

naren-rajendran (approved):

jonathan-gramain:
@bert-e reset

jonathan-gramain:
@bert-e reset

miniscruff:
> One thing I suggest is instead of utils holding a couple of stream methods, export each method directly with its own file.
> Instead of
> -` lib/stream/utils.js`
> 
> ```js
> module.exports = {
>     readJSONStreamObject,
> };
> ```
> 
> use
> `+lib/stream/readJSONStreamObject.js`
> 
> ```js
> module.exports = readJSONStreamObject;
> ```
> 
> It's easy to discover and use the methods this way.

An alternative is a flat structure used by standard packages like python, go, or node. Most imports require no more than one path.
So something like:
`const { streamToObj } = require('stream')`

Even if the files are split up using index.js to manipulate and simplify the imports can be convenient. Very different from super nested structures from Java and C++ ( I think ).

jonathan-gramain:
@bert-e approve

PR #1213, opened 2020-07-15
bf: search with aggregate instead of find

vrancurel:
The search was using find().sort() and was disrupting user defined
search queries and custom indexes. The sort() is needed to implement
a stateless paging system. The combo of user defined query and sort is
now implemented with a 2 stage aggregate on server side.
We always limit the execution time maxTimeMs to 5mn (tunable by an
environment variable).

Note that it the number of concurrent search queries is for now not limited (and it should). We know that the aggregate will put an additional burden on the primary, so micro-sharding shall be implemented to divide and conquer. But this is out of scope of this PR so far.

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -10,6 +10,7 @@
  * We use proper atomic operations when needed.
  */
 const async = require('async');
+const crypto = require('crypto');
 
 const constants = require('../../../constants');
 
@@ -39,6 +40,10 @@ const CONNECT_TIMEOUT_MS = 5000;
 // MongoDB default
 const SOCKET_TIMEOUT_MS = 360000;
 const CONCURRENT_CURSORS = 10;
+// Search
+const MAX_TIME_MS = 300000;
+const TEMP_SEARCH_PREFIX = '__temp_search';
+const SEARCH_PREFIX = '__search';
 
 const initialInstanceID = process.env.INITIAL_INSTANCE_ID;
 
@@ -109,6 +114,11 @@ class MongoClientInterface {
             !Number.isNaN(process.env.CONCURRENT_CURSORS))
             ? Number.parseInt(process.env.CONCURRENT_CURSORS, 10)
             : CONCURRENT_CURSORS;
+
+        this.maxTimeMs = (process.env.MAX_TIME_MS &&
+            !Number.isNaN(process.env.MAX_TIME_MS))
+            ? Number.parseInt(process.env.MAX_TIME_MS, 10)
+            : MAX_TIME_MS;
     }
 
     setup(cb) {
@@ -939,9 +949,8 @@ class MongoClientInterface {
                                       params, log, cb);
     }
 
-    internalListObject(bucketName, params, extension, log, cb) {
-        const c = this.getCollection(bucketName);
-        const stream = new MongoReadStream(c, params, params.mongifiedSearch);
+    internalListObject(c, params, extension, log, cb) {
+        const stream = new MongoReadStream(c, params);
         const skip = new Skip({
             extension,
             gte: params.gte,
@@ -963,7 +972,7 @@ class MongoClientInterface {
             newParams.gte = range;
 
             // then continue listing the next key range
-            this.internalListObject(bucketName, newParams, extension, log, cb);
+            this.internalListObject(c, newParams, extension, log, cb);
         });
 
         stream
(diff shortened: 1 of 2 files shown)


rahulreddy (commented):
LGTM
Were you able to measure the impact (cpu, elapsed_ms, index usage etc) of using this aggregate?

lib/storage/metadata/mongoclient/readStream.js
```diff
@@ -60,12 +60,21 @@ class MongoReadStream extends Readable {
         }
 
         if (searchOptions) {
-            Object.assign(query, searchOptions);
+            this._cursor = c.aggregate([
+                { $match: searchOptions }, // user query
+                { $match: query }, // for paging
```
alexanderchan-scality: was wondering if the pipeline would benefit from first filtering by the indexed `_id`, then followed by the `{ $match: searchOptions}`, which can possibly involve a full collection scan

vrancurel: What we want is first to do the user defined query: searchOptions, then save it on disk (potentially larger results set than 100MB), then sort this results set.

vrancurel: But I think we need to add a `$out` stage into a temporary collection and obtain a cursor on this temp collection instead.
Later on a sweeper shall delete temp collections.

vrancurel:
I did not find a way to $out in a separate namespace. So we will need to modify the oplog to filter out __search collections.

vrancurel:
The other potential problem, which maybe is not a problem is that there is a race between concurrent identical queries, maybe one is not finished and still appears, so the second one will iterate over a partial result.

In order to mitigate this, it is possible to generate a temp name and do a renameCollection() when the aggregate.toArray() is terminated. The drawback becomes then to have the work being done twice.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -993,22 +1001,98 @@ class MongoClientInterface {
...
+            { $match: searchOptions }, // user query
+            { $sort: { _id: 1 } }, // needed for paging
+            { $out: tempCollection }, // a job will clean it up
+        ],
+            {
+                maxTimeMs: this.maxTimeMs, // always limit
+                allowDiskUse: true, // stage large queries on disk
+            },
+            null);
+        _cursor.toArray(err => {
```
alexanderchan-scality, line 1031: this can possibly lead to out of memory if the aggregate results becomes too large

vrancurel:
We should also include the date of the query for making the cache hash.

lib/storage/metadata/mongoclient/MongoClientInterface.js
vrancurel: No the result is empty because there is a `$out`.

vrancurel: But one problem of this approach is that it returns nothing until the query is done. This is weird from the API client perspective. Any suggestion ?

rahulreddy: If it's streaming the output, may be use a combined approach of streaming 1000 documents back to the client and starting a background async job that writes to the temporary collection
https://mongodb.github.io/node-mongodb-native/3.6/reference/cursors/#stream-api
It sounds complex and dirty, ideally I would delegate the task to some other worker but I can't see how.

vrancurel: Yes we could to basically the user-query in an aggregate AND in a find() at the same time. But it will execute the query twice... just to avoid sessions, that's a bit too much...

vrancurel: Actually the $sort here is also killing the query plan, and the good news is ... no need to do the $sort because in the $out collection the keys will already be sorted...

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -993,22 +1002,125 @@ class MongoClientInterface {
...
+                    }
+                    log.info('doSearch: done', {
+                        searchCollection,
+                    });
+                    return undefined;
+                });
+            // fallthrough
+            // eslint-disable-next-line
+            params.searchOptions = searchOptions;
+            return this.internalListObject(
```
vrancurel: For the first page it reverts to regular search (as before).

lib/storage/metadata/mongoclient/MongoClientInterface.js
vrancurel: Need to also limit it.

lib/storage/metadata/mongoclient/MongoClientInterface.js
vrancurel: Need to also limit it.

PR #1223, opened 2020-07-27
bf: S3C-1286 increase bucketname length

dora-korpar:
Increases bucketname length to 255 characters for 7.4 and above

### DIFF
--- a/lib/s3routes/routesUtils.js
+++ b/lib/s3routes/routesUtils.js
@@ -886,7 +886,7 @@ const routesUtils = {
         // eslint-disable-next-line no-useless-escape
         const dnsRegex = new RegExp(/^[a-z0-9]+((\.|\-+)[a-z0-9]+)*$/);
         // Must be at least 3 and no more than 63 characters long.
-        if (bucketname.length < 3 || bucketname.length > 63) {
+        if (bucketname.length < 3 || bucketname.length > 255) {
             return false;
         }
         // Certain prefixes may be reserved, for example for shadow buckets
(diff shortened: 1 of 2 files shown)


dora-korpar:
@bert-e reset

miniscruff (approved):

lib/s3routes/routesUtils.js
```diff
@@ -886,7 +886,7 @@ const routesUtils = {
         // eslint-disable-next-line no-useless-escape
         const dnsRegex = new RegExp(/^[a-z0-9]+((\.|\-+)[a-z0-9]+)*$/);
         // Must be at least 3 and no more than 63 characters long.
```
jonathan-gramain, line 888: The comment does not reflect the changed value anymore

rachedbenmustapha (commented):
Copy-paste from my comment in JIRA:

What’s the rationale for becoming incompatible with AWS S3 and incompatible with standard DNS?

According to [their docs](https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html) the 255 size limit was a 1-region glitch and was since deprecated, coincidentally 4 days before this ticket was opened:

> “Before March 1, 2018, buckets created in the US East (N. Virginia) Region could have names that were up to 255 characters long and included uppercase letters and underscores. Beginning March 1, 2018, new buckets in US East (N. Virginia) must conform to the same rules applied in all other Regions.”

rahulreddy:
Good catch @RachedBenMustapha !

damienverine-scality:
Can you close this PR as we closed the Jira ticket as "Won't Fix" ?

PR #1231, opened 2020-08-08
ft: S3C-2797 bucket notification model

dora-korpar:
Adds bucket notification model

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -78,4 +78,14 @@ module.exports = {
     // Regex for ISO-8601 formatted date
     shortIso8601Regex: /\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z/,
     longIso8601Regex: /\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/,
+    supportedNotificationEvents: new Set([
+        's3:ObjectCreated:*',
+        's3:ObjectCreated:Put',
+        's3:ObjectCreated:Copy',
+        's3:ObjectCreated:CompleteMultipartUpload',
+        's3:ObjectRemoved:*',
+        's3:ObjectRemoved:Delete',
+        's3:ObjectRemoved:DeleteMarkerCreated',
+    ]),
+    notificationArnPrefix: 'arn:scality:bucketnotif',
 };
--- a/lib/models/NotificationConfiguration.js
+++ b/lib/models/NotificationConfiguration.js
@@ -1,0 +1,250 @@
+const UUID = require('uuid');
+
+const {
+    supportedNotificationEvents,
+    notificationArnPrefix,
+} = require('../constants');
+const errors = require('../errors');
+
+/**
+ * Format of xml request:
+ *
+ * <ONotificationConfiguration>
+ *      <QueueConfiguration>
+ *          <Event>array</Event>
+ *          <Filter>
+ *              <S3Key>
+ *                  <FilterRule>
+ *                      <Name>string</Name>
+ *                      <Value>string</Value>
+ *                  </FilterRule>
+ *              </S3Key>
+ *          </Filter>
+ *          <Id>string</Id>
+ *          <Queue>string</Queue>
+ *      </QueueConfiguration>
+ * </NotificationConfiguration>
+ */
+
+ /**
+  * Format of config:
+  *
+  * config = {
+  *     queueConfig: [
+  *         {
+  *         events: array,
+  *         queueArn: string,
+  *         filterRules: [
+  *             {
+  *                 name: string,
+  *                 value: string
+  *             },
+  *             {
+  *                 name: string,
+  *                 value:string
+  *             },
+  *         ],
+  *         id: string
+  *         }
+  *     ]
+  * }
+  */
+
+class NotificationConfiguration {
+    /**
+     * Create a Notification Configuration instance
+     * @param {string} xml - parsed configuration xml
+     * @return {object} - NotificationConfiguration instance
+     */
+    constructor(xml) {
+        this._parsedXml = xml;
+        this._config = {};
+        this._ids = new Set([]);
+    }
+
+    /**
+     * Get notification configuration
+     * @return {object} - contains error if parsing failed
+     */
+    getValidatedNotificationConfiguration() {
+        const validationError = this._parseNotificationConfig();
+        if (validationError) {
+            this._config.error = validationError;
+        }
+        return this._config;
+    }
+
+    /**
+     * Check that notification configuration is valid
+     * @return {error | undefined} - error if parsing failed, else undefined
+     */
+    _parseNotificationConfig() {
+        if (!this._parsedXml || this._parsedXml === '') {
+            return errors.MalformedXML.customizeDescription(
+                'request xml is undefined or empty');
+        }
+        const notificationConfig = this._parsedXml.NotificationConfiguration;
+        if (!notificationConfig || notificationConfig === '') {
+            return errors.MalformedXML.customizeDescription(
+                'request xml does not include NotificationConfiguration');
+        }
+        const queueConfig = notificationConfig.QueueConfiguration;
+        if (!queueConfig || !queueConfig[0] || queueConfig[0] === '') {
+            return errors.MalformedXML.customizeDescription(
+                'request xml does not include QueueConfiguration');
+        }
+        this._config.queueConfig = [];
+        let parseError;
+        for (let i = 0; i < queueConfig.length; i++) {
+            const eventObj = this._parseEvents(queueConfig[i].Event);
+            const filterObj = this._parseFilter(queueConfig[i].Filter);
+            const idObj = this._parseId(queueConfig[i].Id);
+            const arnObj = this._parseArn(queueConfig[i].QueueArn);
+
+            if (eventObj.error) {
+                parseError = eventObj.error;
+                this._config = {};
+                break;
+            }
+            if (filterObj.error) {
+                parseError = filterObj.error;
+                this._config = {};
+                break;
+            }
+            if (idObj.error) {
+                parseError = idObj.error;
+                this._config = {};
+                break;
+            }
+            if (arnObj.error) {
+                parseError = arnObj.error;
+                this._config = {};
+                break;
+            }
+            this._config.queueConfig.push({
+                events: eventObj.events,
+                queueArn: arnObj.arn,
+                id: idObj.id,
+                filterRules: filterObj.filterRules,
+            });
+        }
+        return parseError;
+    }
+
+    /**
+     * Check that events array is valid
+     * @param {array} events - event array
+     * @return {object} - contains error if parsing failed or events array
+     */
+    _parseEvents(events) {
+        const eventsObj = {
+            events: [],
+        };
+        if (!events || !events[0]) {
+            eventsObj.error = errors.MalformedXML.customizeDescription(
+                'each queue configuration must contain an event');
+            return eventsObj;
+        }
+        events.forEach(e => {
+            if (!supportedNotificationEvents.has(e)) {
+                eventsObj.error = errors.MalformedXML.customizeDescription(
+                    'event array contains invalid or unsupported event');
+            } else {
+                eventsObj.events.push(e);
+            }
+        });
+        return eventsObj;
+    }
+
+    /**
+     * Check that filter array is valid
+     * @param {array} filter - filter array
+     * @return {object} - contains error if parsing failed or filter array
+     */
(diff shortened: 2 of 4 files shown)


dora-korpar:
@bert-e approve

dora-korpar:
@bert-e wait

lib/constants.js
```diff
@@ -78,4 +78,14 @@ module.exports = {
     // Regex for ISO-8601 formatted date
     shortIso8601Regex: /\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z/,
     longIso8601Regex: /\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/,
+    bucketNotifSupportedEvents: [
```
rahulreddy: Something like `supportedNotificationEvents` is more readable.

lib/constants.js
```diff
@@ -78,4 +78,14 @@ module.exports = {
...
+        's3:ObjectCreated:*',
+        's3:ObjectCreated:Put',
+        's3:ObjectCreated:Post',
```
rahulreddy: We don't support `POST object` API on Cloudsever, right? So we don't support this event.

naren-rajendran (approved):

rahulreddy (changes requested):

tmacro (approved):

lib/models/NotificationConfiguration.js
```diff
@@ -0,0 +1,245 @@
...
+        let parseError;
+        queueConfig.forEach(c => {
+            const eventObj = this._parseEvents(c.Event);
+            const filterObj = this._parseFilter(c.Filter);
+            const idObj = this._parseId(c.Id);
+            const arnObj = this._parseArn(c.QueueArn);
+
+            if (eventObj.error) {
+                parseError = eventObj.error;
+                this._config = {};
```
miniscruff, line 106: now that I am in the proper PR...
Resetting `_config` here will be ignored given that `_config` is updated in every successful loop.
So if there was an error followed by success, it was reset then pushed again.
This would only work if on any error you broke out of the forEach but thats not possible unless you use a for loop.

lib/models/NotificationConfiguration.js
```diff
@@ -0,0 +1,245 @@
...
+        } else {
+            idObj.id = id[0];
+        }
+        // Each ID in a list of rules must be unique.
+        if (this._ids.includes(idObj.id)) {
+            idObj.error = errors.InvalidRequest.customizeDescription(
+                'queue configuration ID must be unique');
+            return idObj;
+        }
+        this._ids.push(idObj.id);
```
miniscruff: `_ids` is pushed to as a side effect of this function, if it matters.

lib/models/NotificationConfiguration.js
```diff
@@ -0,0 +1,245 @@
...
+        }
+        if (!id || !id[0] || id[0] === '') {
+            // id is optional property, so create one if not provided or is ''
+            // We generate 48-character alphanumeric, unique id for rule
+            idObj.id = Buffer.from(UUID.v4()).toString('base64');
+        } else {
+            idObj.id = id[0];
+        }
+        // Each ID in a list of rules must be unique.
+        if (this._ids.includes(idObj.id)) {
```
miniscruff: If `this._ids` is only used to check for uniqueness I would use a Set

lib/models/NotificationConfiguration.js
```diff
@@ -0,0 +1,245 @@
...
+        if (!arn || !arn[0]) {
+            arnObj.error = errors.MalformedXML.customizeDescription(
+                'each queue configuration must contain a queue arn');
+            return arnObj;
+        }
+        const splitArn = arn[0].split(':');
+        if (splitArn.length !== 6 ||
+            !(splitArn[0] === 'arn') ||
+            !(splitArn[1] === 'scality') ||
+            !(splitArn[2] === 'bucketnotif')) {
```
miniscruff: Should use !== instead of !(===)
```suggestion
            splitArn[0] !== 'arn' ||
            splitArn[1] !== 'scality' ||
            splitArn[2] !== 'bucketnotif') {
```

lib/models/NotificationConfiguration.js
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /**
+     * Check that id string is valid
+     * @param {string} id - id string (optional)
+     * @return {object} - contains error if parsing failed or id
+     */
+    _parseId(id) {
+        const idObj = {};
+        idObj.propName = 'ruleID';
```
miniscruff: Can initialize in one.
```suggestion
        const idObj = { propName: 'ruleID' };
```

lib/models/NotificationConfiguration.js
```diff
@@ -0,0 +1,245 @@
...
+     * Check that arn string is valid
+     * @param {string} arn - queue arn
+     * @return {object} - contains error if parsing failed or queue arn
+     */
+    _parseArn(arn) {
+        const arnObj = {};
+        if (!arn || !arn[0]) {
+            arnObj.error = errors.MalformedXML.customizeDescription(
+                'each queue configuration must contain a queue arn');
+            return arnObj;
```
miniscruff: Instead of creating an empty object, adding one attribute and then returning it. You can just return a new object with that one attribute inline.
```suggestion
        if (!arn || !arn[0]) {
            return { error: errors.MalformedXML.customizeDescription(
                    'each queue configuration must contain a queue arn')
            }
```

Can be applied to multiple cases if you want, if not I won't comment on the others.

tests/unit/models/NotificationConfiguration.js
```diff
@@ -0,0 +1,216 @@
...
+    cb();
+}
+
+function generateEvent(testParams) {
+    const event = [];
+    if (testParams.key === 'Event') {
+        if (Array.isArray(testParams.value)) {
+            testParams.value.forEach(v => {
+                event.push(`${event}<Event>${v}</Event>`);
+            });
```
miniscruff, line 21: It might look better if you used map instead of forEach

```suggestion
            event.push(
                ...testParams.value.map(v => `${event}<Event>${v}</Event>`)
            );
```

tests/unit/models/NotificationConfiguration.js
```diff
@@ -0,0 +1,216 @@
...
+        }
+    } else {
+        event.push('<Event>s3:ObjectCreated:*</Event>');
+    }
+    return event.join('');
+}
+
+function generateFilter(testParams) {
+    let filter = '';
+    if (testParams.key === 'Filter') {
```
miniscruff, line 33: I would move these into functions where `.key` is the key on an object and the value is the function. Something like:

```javascript
const filterMapper = {
    "Filter": test => `<Filter>...`,
    "S3Key: test => `you get the idea`,
}

function generateFilter(testParams) {
    return filterMapper[testParams.key](testParams)
}
```

tests/unit/models/NotificationConfiguration.js
```diff
@@ -0,0 +1,216 @@
...
+        } else {
+            filter = `<Filter><S3Key><FilterRule>${testParams.value}` +
+                '</FilterRule></S3Key></Filter>';
+        }
+    }
+    return filter;
+}
+
+function generateXml(testParams) {
+    const id = testParams.key === 'Id' ? `<Id>${testParams.value}</Id>` : '<Id>queue-id</Id>';
```
miniscruff, line 54: Could create another object map or maybe reuse the other one.

lib/models/NotificationConfiguration.js
miniscruff: You could also slice and compare, something like
`splitArn.slice(3) == ['arn', 'scality', 'bucketnotif']`

lib/constants.js
```diff
@@ -78,4 +78,13 @@ module.exports = {
     // Regex for ISO-8601 formatted date
     shortIso8601Regex: /\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z/,
     longIso8601Regex: /\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/,
+    supportedNotificationEvents: [
```
miniscruff: If this is purely used as a lookup I would use a Set

lib/models/NotificationConfiguration.js
dora-korpar: not gonna worry about this one

lib/models/NotificationConfiguration.js
miniscruff: ooh something I forgot to say is that I think `_ids` could be a little more specific, maybe like `_existingIds`. I thought the side effect of adding to `_ids` was a bigger deal until I realized its just used for this.

lib/models/NotificationConfiguration.js
dora-korpar: Yeah, I revamped this. After discussing with Naren, we don't need such individualized checks cos there's only one valid arn

lib/models/NotificationConfiguration.js
dora-korpar: I went through and changed it to your suggestion where it made sense

tests/unit/models/NotificationConfiguration.js
dora-korpar: gonna leave the test file as is for the sake of time

lib/models/NotificationConfiguration.js
```diff
@@ -0,0 +1,251 @@
...
+                'request xml does not include NotificationConfiguration');
+        }
+        const queueConfig = notificationConfig.QueueConfiguration;
+        if (!queueConfig || !queueConfig[0] || queueConfig[0] === '') {
+            return errors.MalformedXML.customizeDescription(
+                'request xml does not include QueueConfiguration');
+        }
+        this._config.queueConfig = [];
+        let parseError;
+        for (let i = 0; i < queueConfig.length; i++) {
```
miniscruff, line 98: You can use `for ... of` syntax here. Makes it easier than using `[i]` each time.

tests/unit/models/NotificationConfiguration.js
miniscruff: ok, will have to put it on the "come back" list.

lib/models/NotificationConfiguration.js
```diff
@@ -0,0 +1,251 @@
...
+                return { error: errors.MalformedXML.customizeDescription(
+                    'if included, queue configuration filter must contain S3Key') };
+            }
+            const filterRules = filter[0].S3Key[0];
+            if (!filterRules.FilterRule || !filterRules.FilterRule[0]) {
+                return { error: errors.MalformedXML.customizeDescription(
+                    'if included, queue configuration filter must contain a rule') };
+            }
+            const ruleArray = filterRules.FilterRule;
+            for (let i = 0; i < ruleArray.length; i++) {
```
miniscruff: Another good case for `for ... of` loop.

lib/models/NotificationConfiguration.js
```diff
@@ -0,0 +1,251 @@
...
+    /**
+     * Check that filter array is valid
+     * @param {array} filter - filter array
+     * @return {object} - contains error if parsing failed or filter array
+     */
+    _parseFilter(filter) {
+        const filterObj = {
+            filterRules: [],
+        };
+        if (filter && filter[0]) {
```
miniscruff: Could reduce tabbing by turning this into a guard clause. Something like:
```javascript
        if (!filter || !filter[0]) {
            return filterObj;
       }
```

lib/models/NotificationConfiguration.js
```diff
@@ -0,0 +1,251 @@
...
+                    || !ruleArray[i].Value
+                    || !ruleArray[i].Value[0]) {
+                    filterObj.error = errors.MalformedXML.customizeDescription(
+                        'each included filter must contain a name and value');
+                    break;
+                }
+                if (!['Prefix', 'Suffix'].includes(ruleArray[i].Name[0])) {
+                    filterObj.error = errors.MalformedXML.customizeDescription(
+                        'filter Name must be one of Prefix or Suffix');
+                    break;
```
miniscruff: These two could be `return { error : ... }` instead of setting and breaking.

miniscruff (approved):
Going to approve as it looks good, my minor comments are stylistic at most. Maybe a few more but not a big deal.

rahulreddy (approved):

dora-korpar:
@bert-e reset

jonathan-gramain (approved):
I did not review the code but approving to unblock PR

dora-korpar:
@bert-e approve

PR #1235, opened 2020-08-12
Feature/s3 c 2798 get bucket notification

dora-korpar:
Adds `getConfigXML` static to NotificationConfiguration class for `getBucketNotification` api

### DIFF
--- a/lib/models/NotificationConfiguration.js
+++ b/lib/models/NotificationConfiguration.js
@@ -245,6 +245,40 @@ class NotificationConfiguration {
         // remaining 3 parts of arn are evaluated in cloudserver
         return { arn: arn[0] };
     }
+
+    /**
+     * Get XML representation of notification configuration object
+     * @param {object} config - notification configuration object
+     * @return {string} - XML representation of config
+     */
+    static getConfigXML(config) {
+        const xmlArray = [];
+        if (config) {
+            config.queueConfig.forEach(c => {
+                xmlArray.push('<QueueConfiguration>');
+                xmlArray.push(`<Id>${c.id}</Id>`);
+                xmlArray.push(`<QueueArn>${c.queueArn}</QueueArn>`);
+                c.events.forEach(e => {
+                    xmlArray.push(`<Event>${e}</Event>`);
+                });
+                if (c.filterRules) {
+                    xmlArray.push('<Filter><S3Key>');
+                    c.filterRules.forEach(r => {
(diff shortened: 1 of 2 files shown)


dora-korpar:
@bert-e approve

dora-korpar:
@bert-e wait

lib/models/NotificationConfiguration.js
```diff
@@ -0,0 +1,275 @@
...
+            } else {
+                this._config.queueConfig.push({
+                    events: eventObj.events,
+                    queueArn: arnObj.arn,
+                    id: idObj.id,
+                    filterRules: filterObj.filterRules,
+                });
+            }
+        });
+        return parseError;
```
miniscruff: The logic here will override errors as you move through, additionally even if something does have an error `_config` will be changed in the process. Not a fan of the methods having (in my opinion hidden) side effects either.

dora-korpar: If there are any errors, `_config` won't be changed. And yeah, the errors will overwrite each other if there are multiple, but since we're trying to return more specific error messages, it doesn't really make sense to combine them, and it doesn't really matter which one gets returned first

miniscruff: > If there are any errors, `_config` won't be changed. 

Per iteration in the loop yes, but ( and I am not sure this matters ) if the first was valid and the second item in the list was an error then `_config` would be changed and an error returned.

dora-korpar: oh, you're totally right, didn't even think of that. i'll reset `this._config` if there's an error

miniscruff: I believe `_id` is also changed as part of the `_parseId` method but don't take my word for it.

dora-korpar: I'll take a look. Any changes I make to this class are going to be in this PR tho: https://github.com/scality/Arsenal/pull/1231
The diff is in this PR because currently this branch is based off of that one until it gets merged.

miniscruff: ok no worries, feel free to point me somewhere else if I do that again.

naren-rajendran (approved):

tmacro (approved):

jonathan-gramain (approved):
I did not review the code but approving to unblock PR

dora-korpar:
@bert-e reset

PR #1248, opened 2020-08-25
bugfix: ZENKO 2702 Use replica set config instead of rs0

miniscruff:
We already had the replica set value in the config and was loading it but was using a fixed `rs0` in place of the real value.

### DIFF
--- a/lib/storage/metadata/mongoclient/LogConsumer.js
+++ b/lib/storage/metadata/mongoclient/LogConsumer.js
@@ -17,9 +17,10 @@ class LogConsumer {
      * @param {string} logger - logger
      */
     constructor(mongoConfig, logger) {
-        const { authCredentials, replicaSetHosts, database } = mongoConfig;
+        const { authCredentials, replicaSetHosts, replicaSet, database } = mongoConfig;
         const cred = MongoUtils.credPrefix(authCredentials);
         this._mongoUrl = `mongodb://${cred}${replicaSetHosts}/`;
+        this._replicaSet = replicaSet;
         this._logger = logger;
         this._oplogNsRegExp = new RegExp(`^${database}\\.`);
         // oplog collection
@@ -36,7 +37,7 @@ class LogConsumer {
     */
     connectMongo(done) {
         MongoClient.connect(this._mongoUrl, {
-            replicaSet: 'rs0',
+            replicaSet: this._replicaSet,
             useNewUrlParser: true,
         },
         (err, client) => {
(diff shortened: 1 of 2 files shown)


alexanderchan-scality (approved):

nicolas2bert (approved):

anurag4DSB (changes requested):
Shouldn't we create development/8.2 in Arsenal?

We are planning to concentrate on only Zenko v2.x
In the future if we have a bug in Zenko v1.x and we have to update arsenal in cloud server, this might break things.

I checked the ticket and it's fix version is only 2.x so I am guessing there is nothing planned for cloudsevrer dev/8.1.

anurag4DSB (approved):

anurag4DSB:
Okay approved, after checking cloud server config and your PR summary :D. 
But I think we should still make a development/8.2 for 2.x
what do you think?
CC @jonathan-gramain

miniscruff:
> Shouldn't we create development/8.2 in Arsenal?
> 
> We are planning to concentrate on only Zenko v2.x
> In the future if we have a bug in Zenko v1.x and we have to update arsenal in cloud server, this might break things.
> 
> I checked the ticket and it's fix version is only 2.x so I am guessing there is nothing planned for cloudsevrer dev/8.1.

The versioning of our packages are a mystery to me, I picked 8.1 as it was the default and @rahulreddy said that was fine over slack.

anurag4DSB:
> > Shouldn't we create development/8.2 in Arsenal?
> 
> > 
> 
> > We are planning to concentrate on only Zenko v2.x
> 
> > In the future if we have a bug in Zenko v1.x and we have to update arsenal in cloud server, this might break things.
> 
> > 
> 
> > I checked the ticket and it's fix version is only 2.x so I am guessing there is nothing planned for cloudsevrer dev/8.1.
> 
> 
> 
> The versioning of our packages are a mystery to me, I picked 8.1 as it was the default and @rahulreddy said that was fine over slack.

This PR is good for 8.1. We do not have any breaking changes till now. I was considering the features we will add in the future specific to 2.x.

ilkescality (approved):

rahulreddy:
If you ask me I would prefer 9.0 but I don't want to break RING 9 convention. Since 8.1 didn't have any breaking changes, we kept it as-is. Let me ask the question in the channel and see what everyone thinks.

rahulreddy (approved):

jonathan-gramain (approved):

miniscruff:
@bert-e status

miniscruff:
@bert-e approve

miniscruff:
@bert-e help

PR #1254, opened 2020-09-01
bf: S3C-3303 empty notif config ok

dora-korpar:
Allow empty notification configuration as a means to delete existing notification configuration.

### DIFF
--- a/lib/models/NotificationConfiguration.js
+++ b/lib/models/NotificationConfiguration.js
@@ -77,7 +77,7 @@ class NotificationConfiguration {
 
     /**
      * Check that notification configuration is valid
-     * @return {error | undefined} - error if parsing failed, else undefined
+     * @return {error | null} - error if parsing failed, else undefined
      */
     _parseNotificationConfig() {
         if (!this._parsedXml || this._parsedXml === '') {
@@ -90,9 +90,9 @@ class NotificationConfiguration {
                 'request xml does not include NotificationConfiguration');
         }
         const queueConfig = notificationConfig.QueueConfiguration;
-        if (!queueConfig || !queueConfig[0] || queueConfig[0] === '') {
-            return errors.MalformedXML.customizeDescription(
-                'request xml does not include QueueConfiguration');
+        if (!queueConfig || !queueConfig[0]) {
+            // if undefined or empty QueueConfiguration, notif configuration is deleted
+            return null;
         }
         this._config.queueConfig = [];
         let parseError;
@@ -210,7 +210,7 @@ class NotificationConfiguration {
                 'queue configuration ID is greater than 255 characters long') };
         }
         let validId;
-        if (!id || !id[0] || id[0] === '') {
+        if (!id || !id[0]) {
             // id is optional property, so create one if not provided or is ''
             // We generate 48-character alphanumeric, unique id for rule
             validId = Buffer.from(UUID.v4()).toString('base64');
@@ -254,7 +254,7 @@ class NotificationConfiguration {
      */
     static getConfigXML(config) {
         const xmlArray = [];
-        if (config) {
+        if (config && config.queueConfig) {
             config.queueConfig.forEach(c => {
                 xmlArray.push('<QueueConfiguration>');
                 xmlArray.push(`<Id>${c.id}</Id>`);
@@ -289,14 +289,14 @@ class NotificationConfiguration {
      */
     static validateConfig(config) {
         assert.strictEqual(typeof config, 'object');
-        assert.strictEqual(Array.isArray(config.queueConfig), true);
+        if (!config.queueConfig) {
+            return;
+        }
         config.queueConfig.forEach(q => {
             const { events, queueArn, filterRules, id } = q;
-            assert.strictEqual(Array.isArray(events), true);
             events.forEach(e => assert.strictEqual(typeof e, 'string'));
             assert.strictEqual(typeof queueArn, 'string');
             if (filterRules) {
-                assert.strictEqual(Array.isArray(filterRules), true);
                 filterRules.forEach(f => {
                     assert.strictEqual(typeof f.name, 'string');
                     assert.strictEqual(typeof f.value, 'string');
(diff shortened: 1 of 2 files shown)


lib/models/NotificationConfiguration.js
```diff
@@ -289,21 +289,23 @@ class NotificationConfiguration {
...
-                filterRules.forEach(f => {
-                    assert.strictEqual(typeof f.name, 'string');
-                    assert.strictEqual(typeof f.value, 'string');
-                });
-            }
-            assert.strictEqual(typeof id, 'string');
-        });
+        if (config.queueConfig) {
+            assert.strictEqual(Array.isArray(config.queueConfig), true);
+            config.queueConfig.forEach(q => {
```
miniscruff: Not sure this matters, but I just want to compare two assertions and mention that even without asserting something you are effectively writing assertions in certain cases.
These two lines contain two asserts first the strict equal and second that queueConfig contains a forEach function.
The reason for this is that assert just takes some values or conditions and throws an exception if some condition is not met. Similarly trying to call a function on a variable that has no function of that name will also throw an exception because that condition was not met.

I also want to compare the output of these two assertions as sometimes not asserting can give more information in logs than writing an assert.

The first creates an exception that basically says `assertion error: false !== true` if the queue config is not an array. A message that is not too helpful without the line number that gives us the assertion.
The second "hidden" assert gives an error similar to `Uncaught TypeError: config.queueConfig.forEach is not a function`. In my opinion, the second error gives more valuable debug information without needing extra verbose assert messages. And because any error results in an invalid configuration it should work identically.

tl;dr the assert is not required as the forEach method is only found on arrays anyway.

lib/models/NotificationConfiguration.js
```diff
@@ -289,21 +289,23 @@ class NotificationConfiguration {
...
-            if (filterRules) {
-                assert.strictEqual(Array.isArray(filterRules), true);
-                filterRules.forEach(f => {
-                    assert.strictEqual(typeof f.name, 'string');
-                    assert.strictEqual(typeof f.value, 'string');
-                });
-            }
-            assert.strictEqual(typeof id, 'string');
-        });
+        if (config.queueConfig) {
```
miniscruff: Regardless of my other comment, I would turn this into a guard clause and early return over the extra tab.

lib/models/NotificationConfiguration.js
dora-korpar: i removed all the array asserts before `.forEach` uses

lib/models/NotificationConfiguration.js
dora-korpar: the way that function is used doesn't allow for it to have a return, so there's not really a way to make it a guard clause

dora-korpar:
reset

lib/models/NotificationConfiguration.js
miniscruff: Does this not work?
```javascript
if (!config.queueConfig) {
    return
}
```

dora-korpar: i mean yeah, but it's extra lines vs an extra tab. i prefer an extra tab. we can leave it up to jonathan if he wants to enforce a style

jonathan-gramain: Personally I tend to prefer `return` for error cases to return early from the whole processing, and use a block when the flow logic becomes more complex. But I don't really mind in the end as long as the indentation level does not go up too much.

lib/models/NotificationConfiguration.js
```diff
@@ -91,8 +91,8 @@ class NotificationConfiguration {
         }
         const queueConfig = notificationConfig.QueueConfiguration;
         if (!queueConfig || !queueConfig[0] || queueConfig[0] === '') {
```
jonathan-gramain: Not changed in this PR, but the last test is superfluous (covered by the 2nd, as an empty string is false-ish)

lib/models/NotificationConfiguration.js
```diff
@@ -254,7 +254,7 @@ class NotificationConfiguration {
      */
     static getConfigXML(config) {
         const xmlArray = [];
-        if (config) {
+        if (config && Object.keys(config).length > 0) {
```
jonathan-gramain: Why not checking instead `if (config && config.queueConfig)`?

dora-korpar: cos I copied the logic over from cloudserver where it made more sense. good catch

lib/models/NotificationConfiguration.js
miniscruff: For me when writing javascript, especially when using callbacks, the reduced indentation levels is usually preferred over the one saved line. The opening and close brace are required regardless, it is just the return line that is added with guard clauses.

miniscruff (approved):

dora-korpar:
@bert-e reset

lib/models/NotificationConfiguration.js
```diff
@@ -289,21 +289,20 @@ class NotificationConfiguration {
      */
     static validateConfig(config) {
         assert.strictEqual(typeof config, 'object');
-        assert.strictEqual(Array.isArray(config.queueConfig), true);
+        if (!config.queueConfig) return;
```
jonathan-gramain: Can you add `{` and `}` and return on new line to conform to style guidelines

dora-korpar:
@bert-e reset

lib/models/NotificationConfiguration.js
```diff
@@ -289,21 +289,22 @@ class NotificationConfiguration {
...
             if (filterRules) {
-                assert.strictEqual(Array.isArray(filterRules), true);
                 filterRules.forEach(f => {
                     assert.strictEqual(typeof f.name, 'string');
                     assert.strictEqual(typeof f.value, 'string');
                 });
             }
             assert.strictEqual(typeof id, 'string');
         });
+        return;
```
jonathan-gramain, line 307: I'm surprised this is needed, if it's linter complaint, fair enough

jonathan-gramain (approved):

lib/models/NotificationConfiguration.js
dora-korpar: yeah, just for the linter

dora-korpar:
@bert-e approve

PR #1257, opened 2020-09-02
bf: ZENKO-2768 encode tags properly

vrancurel:
Even if >= Mongo3.6 supports . (dot) and $ (dollar) in property names, not all drivers support it (as per their doc -- see ZENKO ticket).

Escaping is problematic for values as some users directly query MongoDB and expect their values with a specific format.

Decision to keep the escaping for the property names (in order to match the AWS spec allowing . and $ in tag names). Note we decided to keep tags as properties and not key/values in an array for users to build their own compound indexes on them.

Since this is not backward compatible, this will require a straightforward migration script for existing deployment using tags. 

- Force LogReader to use MongoUtils.unescape().
- Change MongoUtils escape/unescape to encode only the property names
and not the values.
- Add a unit test to check that the escape/unescape works.

### DIFF
--- a/lib/storage/metadata/mongoclient/ListRecordStream.js
+++ b/lib/storage/metadata/mongoclient/ListRecordStream.js
@@ -1,4 +1,5 @@
 const stream = require('stream');
+const unescape = require('./utils').unescape;
 
 /**
  * @class ListRecordStream
@@ -64,12 +65,6 @@ class ListRecordStream extends stream.Readable {
         // always update to most recent uniqID
         this._lastConsumedID = itemObj.h.toString();
 
-        const decode = tags => (
-            JSON.parse(JSON.stringify(tags).
-                       replace(/\uFF0E/g, '.').
-                       replace(/\uFF04/g, '$'))
-        );
-
         // only push to stream unpublished objects
         if (!this._lastSavedID) {
             // process from the first entry
@@ -99,7 +94,7 @@ class ListRecordStream extends stream.Readable {
             itemObj.o && itemObj.o._id) {
             const value = itemObj.o.value;
             if (value && value.tags) {
-                value.tags = decode(value.tags);
+                value.tags = unescape(value.tags);
             }
             entry = {
                 type: 'put',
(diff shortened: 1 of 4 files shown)


lib/storage/metadata/mongoclient/utils.js
```diff
@@ -5,15 +5,29 @@ const {
...
+    const _obj = {};
+    for (const prop in obj) {
+        if (Object.prototype.hasOwnProperty.call(obj, prop)) {
+            const _prop = prop.
+                replace(/\$/g, '\uFF04').
+                replace(/\./g, '\uFF0E');
+            _obj[_prop] = obj[prop];
+        }
+    }
+    return _obj;
```
alexanderchan-scality: `Object.keys` would skip the enumeration of inherited properties.
```suggestion
    const _obj = {};
    Object.keys(obj).forEach(prop => {
        const _prop = prop.
            replace(/\$/g, '\uFF04').
            replace(/\./g, '\uFF0E');
        _obj[_prop] = obj[prop];
    });
    
    return _obj;
```

alexanderchan-scality (approved):

lib/storage/metadata/mongoclient/ListRecordStream.js
```diff
@@ -65,9 +66,7 @@ class ListRecordStream extends stream.Readable {
         this._lastConsumedID = itemObj.h.toString();
 
         const decode = tags => (
-            JSON.parse(JSON.stringify(tags).
-                       replace(/\uFF0E/g, '.').
-                       replace(/\uFF04/g, '$'))
+            MongoUtils.unescape(tags)
         );
```
miniscruff: If decode is just shorthand for unescape you do not need the extra function.

```javascript
const decode = MongoUtils.unescape
```

miniscruff (approved):

lib/storage/metadata/mongoclient/ListRecordStream.js
vrancurel: Can even get shorter:
``
const unescape = require('./utils').unescape;
``

jonathan-gramain (approved):

rahulreddy (approved):

PR #1266, opened 2020-09-22
Feature/s3 c 1801 policy tag condition keys

dora-korpar:
In order for the policy condition keys s3:ExistingObjectTag, s3:RequestObjectTagKey, and s3:RequestObjectTagKeys to be evaluated, Arsenal requires information from CloudServer.
As such, the policy review process, if it includes any of these condition keys, will be as follows:

Review policy as normal, but include information of whether any tag condition keys are in the policy
Return to CloudServer, which will check whether a tag condition key was present and gather the necessary tag information, if needed
Return to Arsenal with tag info for another complete policy evaluation, including tag condition keys
Return auth results to Cloudserver and continue with request
I don't believe my solution is necessarily the most elegant, but it is the simplest I could come up with to change as little of the existing code as possible. Any feedback on logic or style or anything would be greatly appreciated! ~Tests in progress.~ Tests included :)

### DIFF
--- a/lib/policyEvaluator/evaluator.js
+++ b/lib/policyEvaluator/evaluator.js
@@ -6,6 +6,7 @@ const conditions = require('./utils/conditions.js');
 const findConditionKey = conditions.findConditionKey;
 const convertConditionOperator = conditions.convertConditionOperator;
 const checkArnMatch = require('./utils/checkArnMatch.js');
+const { transformTagKeyValue } = require('./utils/objectTags');
 
 const evaluators = {};
 
@@ -16,6 +17,7 @@ const operatorsWithVariables = ['StringEquals', 'StringNotEquals',
 const operatorsWithNegation = ['StringNotEquals',
     'StringNotEqualsIgnoreCase', 'StringNotLike', 'ArnNotEquals',
     'ArnNotLike', 'NumericNotEquals'];
+const tagConditions = new Set(['s3:ExistingObjectTag', 's3:RequestObjectTagKey', 's3:RequestObjectTagKeys']);
 
 
 /**
@@ -96,20 +98,27 @@ evaluators.isActionApplicable = (requestAction, statementAction, log) => {
  * @param {RequestContext} requestContext - info about request
  * @param {Object} statementCondition - Condition statement from policy
  * @param {Object} log - logger
- * @return {boolean} true if meet conditions, false if not
+ * @return {Object} contains whether conditions are allowed and whether they
+ * contain any tag condition keys
  */
 evaluators.meetConditions = (requestContext, statementCondition, log) => {
     // The Condition portion of a policy is an object with different
     // operators as keys
+    const conditionEval = {};
     const operators = Object.keys(statementCondition);
     const length = operators.length;
     for (let i = 0; i < length; i++) {
         const operator = operators[i];
+        const hasPrefix = operator.includes(':');
         const hasIfExistsCondition = operator.endsWith('IfExists');
-        // If has "IfExists" added to operator name, find operator name
-        // without "IfExists"
-        const bareOperator = hasIfExistsCondition ? operator.slice(0, -8) :
+        // If has "IfExists" added to operator name, or operator has "ForAnyValue" or
+        // "For All Values" prefix, find operator name without "IfExists" or prefix
+        let bareOperator = hasIfExistsCondition ? operator.slice(0, -8) :
             operator;
+        let prefix;
+        if (hasPrefix) {
+            [prefix, bareOperator] = bareOperator.split(':');
+        }
         const operatorCanHaveVariables =
             operatorsWithVariables.indexOf(bareOperator) > -1;
         const isNegationOperator =
@@ -118,6 +127,9 @@ evaluators.meetConditions = (requestContext, statementCondition, log) => {
         // Note: this should be the actual operator name, not the bareOperator
         const conditionsWithSameOperator = statementCondition[operator];
         const conditionKeys = Object.keys(conditionsWithSameOperator);
+        if (conditionKeys.some(key => tagConditions.has(key)) && !requestContext.getNeedTagEval()) {
+            conditionEval.tagConditions = true;
+        }
         const conditionKeysLength = conditionKeys.length;
         for (let j = 0; j < conditionKeysLength; j++) {
             const key = conditionKeys[j];
@@ -130,14 +142,18 @@ evaluators.meetConditions = (requestContext, statementCondition, log) => {
                 value = value.map(item =>
                     substituteVariables(item, requestContext));
             }
+            // if condition key is RequestObjectTag or ExistingObjectTag,
+            // tag key is included in condition key and needs to be
+            // moved to value for evaluation, otherwise key/value are unchanged
+            const [transformedKey, transformedValue] = transformTagKeyValue(key, value);
             // Pull key using requestContext
             // TODO: If applicable to S3, handle policy set operations
             // where a keyBasedOnRequestContext returns multiple values and
             // condition has "ForAnyValue" or "ForAllValues".
             // (see http://docs.aws.amazon.com/IAM/latest/UserGuide/
             // reference_policies_multi-value-conditions.html)
-            const keyBasedOnRequestContext =
-                findConditionKey(key, requestContext);
+            let keyBasedOnRequestContext =
+                findConditionKey(transformedKey, requestContext);
             // Handle IfExists and negation operators
             if ((keyBasedOnRequestContext === undefined ||
                 keyBasedOnRequestContext === null) &&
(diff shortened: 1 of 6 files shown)


lib/policyEvaluator/evaluator.js
```diff
@@ -16,6 +17,7 @@ const operatorsWithVariables = ['StringEquals', 'StringNotEquals',
 const operatorsWithNegation = ['StringNotEquals',
     'StringNotEqualsIgnoreCase', 'StringNotLike', 'ArnNotEquals',
     'ArnNotLike', 'NumericNotEquals'];
+const tagConditions = ['s3:ExistingObjectTag', 's3:RequestObjectTagKey', 's3:RequestObjectTagKeys'];
```
miniscruff: I would use a Set here as its only used as a contains check ( includes in JS's case )

lib/policyEvaluator/evaluator.js
```diff
@@ -118,6 +127,9 @@ evaluators.meetConditions = (requestContext, statementCondition, log) => {
         // Note: this should be the actual operator name, not the bareOperator
         const conditionsWithSameOperator = statementCondition[operator];
         const conditionKeys = Object.keys(conditionsWithSameOperator);
+        if (conditionKeys.some(key => tagConditions.includes(key)) && !requestContext.getNeedTagEval()) {
```
miniscruff: I found out there is a second optional parameter to some for `thisArg` so you could do this.
```suggestion
        if (conditionKeys.some(tagConditions.has, tagConditions) && !requestContext.getNeedTagEval()) {
```

If you like it anyway, it's arguably not much if any better.

lib/policyEvaluator/utils/conditions.js
```diff
@@ -232,12 +252,21 @@ conditions.convertConditionOperator = operator => {
...
+        StringLike: function stringLike(key, value, prefix) {
+            function policyValRegex(testKey) {
+                return value.some(item => {
+                    const wildItem = handleWildcards(item);
+                    const wildRegEx = new RegExp(wildItem);
+                    return wildRegEx.test(testKey);
+                });
+            }
+            if (prefix === 'ForAnyValue') {
+                return key.some(k => policyValRegex(k));
```
miniscruff: should just be able to pass in the method here directly.
```suggestion
                return key.some(policyValRegex);
```

lib/policyEvaluator/evaluator.js
```diff
@@ -96,20 +98,27 @@ evaluators.isActionApplicable = (requestAction, statementAction, log) => {
...
+ * contain any tag condition keys
  */
 evaluators.meetConditions = (requestContext, statementCondition, log) => {
     // The Condition portion of a policy is an object with different
     // operators as keys
+    const conditionEval = {};
     const operators = Object.keys(statementCondition);
     const length = operators.length;
     for (let i = 0; i < length; i++) {
         const operator = operators[i];
```
miniscruff, line 111: Could use Object.entires here instead of keys + for i. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

miniscruff (commented):
I don't understand the reasoning here much but had a few JS comments.

lib/policyEvaluator/evaluator.js
dora-korpar: oh yeah, i always forget about sets

anurag4DSB (approved):
the logic looks pretty straight forwards and simple.
I agree with comments by @miniscruff

lib/policyEvaluator/evaluator.js
dora-korpar: yeah, that's kind of confusing to me readability-wise, so i'll keep the arrow function

lib/policyEvaluator/evaluator.js
dora-korpar: that would be more of a refactor, so gonna leave it for now. but noted 👍

lib/policyEvaluator/evaluator.js
miniscruff: Haha yea it is a bit of a stretch for sure.

lib/policyEvaluator/utils/objectTags.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Gets array of tag key names from request tag query string
+ * @param {string} tagQuery - request tags in query string format
+ * @return {array} array of tag key names
+ */
+function getTagKeys(tagQuery) {
+    const tagsArray = tagQuery.split('&');
+    const keysArray = tagsArray.map(tag => tag.split('=')[0]);
+    return keysArray;
```
miniscruff: As a single block.

```suggestion
    return tagQuery.split('&')
        .map(tag => tag.split('=')[0])
```

miniscruff (approved):

lib/policyEvaluator/evaluator.js
```diff
@@ -96,20 +98,27 @@ evaluators.isActionApplicable = (requestAction, statementAction, log) => {
...
+        const hasPrefix = operator.startsWith('ForAnyValue') || operator.startsWith('ForAllValues');
```
jonathan-gramain: Isn't checking for the presence of `:` a better way to check if there is a prefix? Since this is what we use later to extract `bareOperator`

lib/policyEvaluator/principal.js
```diff
@@ -35,7 +35,8 @@ class Principal {
                 // In case of anonymous NotPrincipal, this will neutral everyone
                 return 'Neutral';
             }
-            if (!Principal._evaluateCondition(params, statement)) {
+            const conditionEval = Principal._evaluateCondition(params, statement);
+            if (!conditionEval || conditionEval.allow === false) {
```
jonathan-gramain, line 39: Could write `!conditionEval.allow` (it would also catch any result that is not explicitly allowed)`

lib/policyEvaluator/utils/objectTags.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Removes tag key value from condition key and adds it to value if needed
+ * @param {string} key - condition key
+ * @param {string} value - condition value
+ * @return {array} key/value pair to use
+ */
+function transformTagKeyValue(key, value) {
+    if (!key.includes('/')) {
```
jonathan-gramain: It might be a good idea not to include any pattern that we do not explicitly recognize and support and keep the default behavior for those. To do this, we could check explicitly each recognized prefix pattern `s3:ExistingObjectTag/` etc. instead of assuming that because there is a `/` it is a tag pattern, which looks somewhat hazardous to me.

tests/unit/policyEvaluator.js
```diff
@@ -1161,6 +1161,74 @@ describe('policyEvaluator', () => {
                 };
                 check(requestContext, rcModifiers, policy, 'Neutral');
             });
+
+            it('should allow with StringEquals operator and ExistingObjectTag ' +
```
jonathan-gramain, line 1165: I'd suggest adding a couple failing tests for unsupported patterns like `s3:FooBar/tagKey` (see earlier comment)

lib/policyEvaluator/principal.js
dora-korpar: I had that originally and the check was failing. Not sure why, but this seems to work

lib/policyEvaluator/utils/objectTags.js
dora-korpar: I was trying to keep the function generic so that it could be used in the future if we support other condition keys that use `/`, and there's no condition key that contains a `/` that wouldn't require transformation, but I'll change it if you think it might introduce problems.

tests/unit/policyEvaluator.js
dora-korpar: oh should've read this comment before responding to the last one. makes sense now 👍

lib/policyEvaluator/utils/objectTags.js
dora-korpar: just for the record, I added the test suggested below, and it does get transformed through this function, but because of the way condition keys are evaluated, it doesn't give permission. I personally don't think there's any harm in that sort of error handling, but will defer to your wisdom :)

jonathan-gramain: If the transformation that the function is doing makes sense for all cases where `key` contains a `/`, it's ok for me.

lib/policyEvaluator/principal.js
jonathan-gramain: `conditionEval.allow === false` implies `!conditionEval.allow` (but not the other way) so it should make this test broader, but it should not fail with `!conditionEval.allow` where it succeeds with `conditionEval.allow === false`. Can you check this to understand exactly what is happening? I think it's important since it looks like a critical piece of code to me.

lib/policyEvaluator/evaluator.js
dora-korpar: lol, this seems like the opposite logic from the `transformTagKeyValue` comment, but I'll change it

jonathan-gramain (approved):

dora-korpar:
@bert-e reset

dora-korpar:
@bert-e approve

PR #1293, opened 2020-11-04
bugfix: ZENKO-2905 avoid error callback call in external AWS GET

jonathan-gramain:
Don't call the callback a second time on stream error during the execution of a GET from AWS external backend. This in turn would break the logic up in the code and cause a crash of a worker.

- test: ZENKO-2905 unit test for external backend streamed GET

Add a new unit test that does a GET on external backends with a range. For this, introduced mocking for the backend functions
getObject() (AWS style) and getBlobToStream (Azure style) that return bytes from a virtual 1GB object.

### DIFF
--- a/lib/storage/data/external/AwsClient.js
+++ b/lib/storage/data/external/AwsClient.js
@@ -180,19 +180,14 @@ class AwsClient {
         });
         const stream = request.createReadStream().on('error', err => {
             let logLevel;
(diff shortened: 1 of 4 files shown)


dora-korpar (approved):
lgtm 👍

ilkescality (approved):

PR #1303, opened 2020-11-18
ft: ZENKO-2968 enable sharding on bucket creation

vrancurel:
Enable sharding on bucket creation (experimental - open for comments)

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -91,11 +91,15 @@ function generatePHDVersion(versionId) {
 class MongoClientInterface {
     constructor(params) {
         const { replicaSetHosts, writeConcern, replicaSet, readPreference, path,
-            database, logger, replicationGroupId, authCredentials,
+                database, enableSharding, logger,
+                replicationGroupId, authCredentials,
             isLocationTransient } = params;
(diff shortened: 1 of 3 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -95,7 +95,8 @@ class MongoClientInterface {
             isLocationTransient } = params;
         const cred = MongoUtils.credPrefix(authCredentials);
         this.mongoUrl = `mongodb://${cred}${replicaSetHosts}/` +
-            `?w=${writeConcern}&replicaSet=${replicaSet}` +
+            `?w=${writeConcern}` +
+            `&${replicaSet !== undefined ? 'replicaSet=' : ''}${replicaSet}` +
```
alexanderchan-scality: this creates a malformed mongourl if `replicaSet` is undefined/null/empty.

```suggestion
            `${!!replicaSet ? '&replicaSet='+replicaSet : ''}` +
```

vrancurel: yes right! thanks

vrancurel:
This code is experimental, e.g. I assume if replicaset === undefined we want sharding, we probably want to propagate an explicit config param for enabling sharding.

vrancurel:
Also we need to manage changestreams for the LogConsumer

PR #1315, opened 2020-12-03
bf: S3C-3425 extract client ip from header if available

ilkescality:


lib/policyEvaluator/requestUtils.js
```diff
@@ -18,14 +28,17 @@ function getClientIp(request, s3config) {
...
             clientIp)) {
             const ipFromHeader
-            // eslint-disable-next-line operator-linebreak
                 = request.headers[requestConfig.extractClientIPFromHeader];
             if (ipFromHeader && ipFromHeader.trim().length) {
-                return ipFromHeader.split(',')[0].trim();
+                return _extractIpFromHeader(ipFromHeader);
             }
         }
+    } else {
```
NicolasT, line 36: Doesn't this now allow *any* client to forge/fake the client IP used to validate a request by simply setting this `x-forwarded-for` header in the HTTP request it sends? :thinking:

rahulreddy: That's very true, it was also the old behavior unfortunately. I think the logic should be extended such that header parsing happens only for whitelisted IP addresses, with the nodes in the inventory added by default.
@ilkescality I recommend going back to the original implementation we talked about. I can help sometime this week.

ilkescality: @NicolasT Yes, it does allow that indeed, thanks for pointing this out. I will switch to another approach to prevent this. 
@rahulreddy I agree, I will go back to the original implementation we talked about then. I will pause running the Integration tests for this fix and will proceed retrying to test the Federation fix.

ilkescality:
Closing as the new fix will be on Federation.

PR #1340, opened 2021-02-04
bugfix: S3C-3904: better-s3-action-logs

alexandre-merle:
Introduce a map meant to override default
actionMap values for S3, will be used in logs
to monitor the s3 actions instead of the iam
permissions needed for that action

### DIFF
--- a/lib/policyEvaluator/utils/actionMaps.js
+++ b/lib/policyEvaluator/utils/actionMaps.js
@@ -29,6 +29,12 @@ const sharedActionMap = {
     objectPutTagging: 's3:PutObjectTagging',
 };
 
+// Meant to override shared action map values with fakes name for monitoring of s3
+const sharedMonitoringMap = {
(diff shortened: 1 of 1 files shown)


miniscruff (approved):

lib/policyEvaluator/utils/actionMaps.js
```diff
@@ -29,6 +29,12 @@ const sharedActionMap = {
     objectPutTagging: 's3:PutObjectTagging',
 };
 
+// Meant to override shared action map values with fakes name for moniroting of s3
```
jonathan-gramain: moni~ro~toring

jonathan-gramain (approved):

lib/policyEvaluator/utils/actionMaps.js
alexandre-merle: :)

alexandre-merle:
@bert-e reset

PR #1366, opened 2021-03-17
ft: S3C-4172 custom filter

vrancurel:
Perform an optional filter on customAttributes sub-object with filterKey
and filterKeyStartWith optional parameters on basic filter.

### DIFF
--- a/lib/algos/list/basic.js
+++ b/lib/algos/list/basic.js
@@ -2,7 +2,7 @@
 
 const Extension = require('./Extension').default;
 
-const { checkLimit, FILTER_END, FILTER_ACCEPT } = require('./tools');
+const { checkLimit, FILTER_END, FILTER_ACCEPT, FILTER_SKIP } = require('./tools');
 const DEFAULT_MAX_KEYS = 10000;
 
 /**
@@ -21,6 +21,8 @@ class List extends Extension {
         this.res = [];
         if (parameters) {
             this.maxKeys = checkLimit(parameters.maxKeys, DEFAULT_MAX_KEYS);
+            this.filterKey = parameters.filterKey;
+            this.filterKeyStartsWith = parameters.filterKeyStartsWith;
         } else {
             this.maxKeys = DEFAULT_MAX_KEYS;
         }
@@ -45,6 +47,43 @@ class List extends Extension {
     }
 
     /**
+     * Filters customAttributes sub-object if present
+     *
+     * @param {String} value - The JSON value of a listing item
+     *
+     * @return {Boolean} Returns true if matches, else false.
+     */
+    customFilter(value) {
+        let _value;
+        try {
+            _value = JSON.parse(value);
+        } catch (e) {
+            // Prefer returning an unfiltered data rather than
+            // stopping the service in case of parsing failure.
+            // The risk of this approach is a potential
+            // reproduction of MD-692, where too much memory is
+            // used by repd.
+            this.logger.warn(
+                'Could not parse Object Metadata while listing',
+                { err: e.toString() });
+            return false;
+        }
+        if (_value.customAttributes !== undefined) {
+            for (const key of Object.keys(_value.customAttributes)) {
+                if (this.filterKey !== undefined &&
+                    key === this.filterKey) {
+                    return true;
+                }
+                if (this.filterKeyStartsWith !== undefined &&
+                    key.startsWith(this.filterKeyStartsWith)) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    /**
      *  Function apply on each element
      *  Just add it to the array
      *  @param {Object} elem - The data from the database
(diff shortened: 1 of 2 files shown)


anurag4DSB (approved):

ilkescality (approved):

saitharunthodupunuru (approved):

tests/unit/algos/list/basic.js
```diff
@@ -65,4 +65,50 @@ describe('Basic listing algorithm', () => {
                                    Basic, { maxKeys: 1 }, logger);
         assert.deepStrictEqual(res, ['key1']);
     });
+
+    it('Shall support filtering custom attributes', () => {
```
miniscruff: I would split the three asserts into separate tests

lib/algos/list/basic.js
```diff
@@ -56,6 +102,13 @@ class List extends Extension {
         if (this.keys >= this.maxKeys) {
             return FILTER_END;
         }
+        if (this.filterKey !== undefined ||
+            this.filterKeyStartsWith !== undefined) {
+            if (typeof elem === 'object' &&
+                !this.customFilter(elem.value)) {
```
miniscruff: I would merge these two ifs

lib/algos/list/basic.js
```diff
@@ -44,6 +46,50 @@ class List extends Extension {
...
 
+    /**
+     * Filters customAttributes sub-object if present
+     *
+     * @param {String} value - The JSON value of a listing item
+     *
+     * @return {Boolean} Returns false if the value is filtered out, else true.
+     */
+    customFilter(value) {
+        let ret = false;
```
miniscruff: You could remove this variable and just `return true` in the for loops

lib/algos/list/basic.js
```diff
@@ -44,6 +46,50 @@ class List extends Extension {
...
+        try {
+            const _value = JSON.parse(value);
```
miniscruff: I would only try/catch the JSON parse as that is the explicit error we return later on

lib/algos/list/basic.js
```diff
@@ -44,6 +46,50 @@ class List extends Extension {
...
+            if (_value.customAttributes !== undefined) {
+                const keys = Object.keys(_value.customAttributes);
+                let i;
+                for (i = 0; i < keys.length; i++) {
+                    const key = keys[i];
```
miniscruff: You can use for... of here
```suggestion
                for (const key of Object.keys(_value.customAttributes)) {
```

lib/algos/list/basic.js
```diff
@@ -44,6 +46,50 @@ class List extends Extension {
...
+                    if (this.filterKey !== undefined) {
+                        if (key === this.filterKey) {
```
miniscruff: Would merge ifs

lib/algos/list/basic.js
```diff
@@ -44,6 +46,50 @@ class List extends Extension {
...
+                            ret = true;
+                            break;
+                        }
+                    } else if (this.filterKeyStartsWith !== undefined) {
+                        const patternLength = this.filterKeyStartsWith.length;
+                        const keyPrefix = key.substring(0, patternLength);
+                        if (keyPrefix === this.filterKeyStartsWith) {
```
miniscruff: Can use startsWith here
```suggestion
                        if (keyPrefix.startsWith(this.filterKeyStartsWith)) {
```
( might be the other way around not sure )

tests/unit/algos/list/basic.js
```diff
@@ -65,4 +65,50 @@ describe('Basic listing algorithm', () => {
...
+        const attr1 = {
```
jonathan-gramain: Since the tests embed a different set of parameters and are not really parameterized, you might as well set the attribute values close to the tested calls. Otherwise I would have suggested a `forEach()` loop to parameterized but I think it's not applicable here.

lib/algos/list/basic.js
```diff
@@ -44,6 +46,50 @@ class List extends Extension {
...
+                const keys = Object.keys(_value.customAttributes);
+                let i;
+                for (i = 0; i < keys.length; i++) {
+                    const key = keys[i];
+                    if (this.filterKey !== undefined) {
+                        if (key === this.filterKey) {
+                            ret = true;
+                            break;
+                        }
+                    } else if (this.filterKeyStartsWith !== undefined) {
```
jonathan-gramain: is the `filterKeyStartsWith` exclusive with `filterKey`? If so, looks good, otherwise need to remove the `else`.

lib/algos/list/basic.js
jonathan-gramain: Actually I think you mean `if (key.startsWith) ...`

miniscruff: Yea that one... 😆

tests/unit/algos/list/basic.js
vrancurel: not applicable

jonathan-gramain (approved):

PR #1403, opened 2021-04-12
bf S3C-4239 log consumer callback error fix

naren-rajendran:
A guard is added to ensure that the callback is called only once in the event of an error while reading records in log consumer.


naren-rajendran:
Still working on finding the exact reason as to why callbacks are fired multiple times and couldn't pinpoint the log entry or the type of entry that causes the issue.

naren-rajendran:
@bert-e reset

jonathan-gramain (approved):

miniscruff (approved):

naren-rajendran:
@bert-e approve

PR #1416, opened 2021-04-20
Feature/zenko 3368/add chain backend

alexanderchan-scality:
* reorganized `in_memory` backend files
* adds auth `ChainBackend`:
  * allows for authentication to be done using multiple auth backends or vault clients
  * authenticates in order of client/backend list provided until one succeeds


lib/auth/backends/ChainBackend.js
```diff
@@ -0,0 +1,194 @@
...
+    /*
+     * apply task to all clients
+     */
+    _forEachClient(task, cb) {
+        async.map(this._clients, task, cb);
+    }
+
+    verifySignatureV2(stringToSign, signatureFromRequest, accessKey, options, callback) {
+        this._tryEachClient((client, done) => {
+            if (typeof client.verifySignatureV2 !== 'function') {
```
rachedbenmustapha: This is a pretty expensive check to do at each invocation. I’d check for all backends in the chain in the constructor and raise an exception if any of the functions is missing (which should happen close to never).

lib/auth/backends/ChainBackend.js
```diff
@@ -0,0 +1,194 @@
...
+            }
+            return callback(null, {
+                message: {
+                    body: ChainBackend._mergePolicies(res),
+                },
+            });
+        });
+    }
+
+    healthCheck(reqUid, callback) {
```
rachedbenmustapha: Is the difference in case intentional? (healthCheck vs healthcheck)

rachedbenmustapha (approved):

lib/auth/backends/ChainBackend.js
alexanderchan-scality: good catch. this is a typo will fix, as there is a fallback, the tests I have done were false positive passes

naren-rajendran (approved):

lib/auth/backends/ChainBackend.js
```diff
@@ -0,0 +1,188 @@
...
+        this._forEachClient((client, done) => {
+            if (typeof client.healthcheck !== 'function') {
+                return done(null, { code: 200, message: 'OK' });
+            }
+            return client.healthcheck(reqUid, done);
+        }, (err, res) => {
+            if (err) {
+                return callback(err);
+            }
+            return callback(null, res);
```
rachedbenmustapha: Seems fine in that if any backend fails the check, an error will be returned, but doesn't this change the shape of `res`?

lib/auth/backends/ChainBackend.js
```diff
@@ -0,0 +1,188 @@
...
+
+    static _mergeObjects(objectResponses) {
+        return objectResponses.reduce(
+            (retObj, resObj) => Object.assign(retObj, resObj.message.body),
+            {});
+    }
+
+    getCanonicalIds(emailAddresses, options, callback) {
+        this._forEachClient((client, done) => {
+            if (typeof client.getCanonicalIds !== 'function') {
```
rachedbenmustapha: This and `getEmailAddresses` would also benefit from being checked once only. If it's expected that one backend doesn't implement them (which I don't think is the case), then just checking truthiness would be enough.

rachedbenmustapha (commented):
We don't have time to perform performance testing unfortunately, but this change will definitely need to be evaluated for performance.

lib/auth/backends/ChainBackend.js
alexanderchan-scality: What I had in mind was to just returning the error in the case of that  the healthcheck fails
after taking a step back and having a better look at this, res is even more important in the case of an error.

lib/auth/backends/ChainBackend.js
alexanderchan-scality: yes. You're right. If a client/backend doesn't implement all these methods, it should be assumed to be invalid.

lib/auth/backends/ChainBackend.js
```diff
@@ -0,0 +1,187 @@
...
+            }
+            return callback(null, {
+                message: {
+                    body: ChainBackend._mergePolicies(res),
+                },
+            });
+        });
+    }
+
+    healthcheck(reqUid, callback) {
```
alexanderchan-scality, line 167: @naren-scality @RachedBenMustapha updated the healthcheck to test all clients and not exit on early error.

rachedbenmustapha (approved):

PR #1419, opened 2021-04-26
feature: S3C-4073: backport probe server

alexandre-merle:
Backporting probe server which will be use to setting health server
check and metrics gathering

### DIFF
--- a/lib/network/probe/HealthProbeServer.js
+++ b/lib/network/probe/HealthProbeServer.js
@@ -1,0 +1,99 @@
+const httpServer = require('../http/server');
+const werelogs = require('werelogs');
+const errors = require('../../errors');
+const ZenkoMetrics = require('../../metrics/ZenkoMetrics');
+
+function sendError(res, log, error, optMessage) {
+    res.writeHead(error.code);
+    let message;
+    if (optMessage) {
+        message = optMessage;
+    } else {
+        message = error.description || '';
+    }
+    log.debug('sending back error response', { httpCode: error.code,
+        errorType: error.message,
+        error: message });
+    res.end(`${JSON.stringify({ errorType: error.message,
+        errorMessage: message })}\n`);
+}
+
+function sendSuccess(res, log, msg) {
+    res.writeHead(200);
+    log.debug('replying with success');
+    const message = msg || 'OK';
+    res.end(message);
+}
+
+function checkStub(log) { // eslint-disable-line
+    return true;
+}
+
+class HealthProbeServer extends httpServer {
+    constructor(params) {
+        const logging = new werelogs.Logger('HealthProbeServer');
+        super(params.port, logging);
+        this.logging = logging;
+        this._metrics = params.metrics || ZenkoMetrics;
+        this.setBindAddress(params.bindAddress || 'localhost');
+        // hooking our request processing function by calling the
+        // parent's method for that
+        this.onRequest(this._onRequest);
+        this._reqHandlers = {
+            '/_/health/liveness': this._onLiveness.bind(this),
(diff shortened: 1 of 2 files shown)


lib/network/probe/HealthProbeServer.js
```diff
@@ -0,0 +1,97 @@
...
+        super(params.port, logging);
+        this.logging = logging;
+        this.setBindAddress(params.bindAddress || 'localhost');
+        // hooking our request processing function by calling the
+        // parent's method for that
+        this.onRequest(this._onRequest);
+        this._reqHandlers = {
+            '/_/health/liveness': this._onLiveness.bind(this),
+            '/_/health/readiness': this._onReadiness.bind(this),
+            '/_/monitoring/metrics': this._onMetrics.bind(this),
```
miniscruff, line 45: I know this is a straight backport but I think we want to use `/_/metrics` instead moving forward, so we can add both for now.

lib/network/probe/HealthProbeServer.js
```diff
@@ -0,0 +1,97 @@
...
+            sendError(res, log, errors.ServiceUnavailable);
+        }
+    }
+
+    // expose metrics to Prometheus
+    _onMetrics(req, res) {
+        res.writeHead(200, {
+            'Content-Type': ZenkoMetrics.asPrometheusContentType(),
+        });
+        res.end(ZenkoMetrics.asPrometheus());
```
miniscruff: I think hardcoding ZenkoMetrics here might cause an issue in backbeat's two processes needing different metrics.

lib/network/probe/HealthProbeServer.js
alexandre-merle: route added

lib/network/probe/HealthProbeServer.js
alexandre-merle: could be now override with constructor params

alexandre-merle:
@bert-e reset

alexandre-merle:
ping

miniscruff (approved):

lib/network/probe/HealthProbeServer.js
```diff
@@ -0,0 +1,99 @@
+const httpServer = require('../http/server');
```
miniscruff, line 1: I left a comment on your backbeat PR about returning errors instead of a generic "Service Unavailable" response to help with diagnosing. I wanted to also mention the lack of JsDocs and there is a bit of repetition in here we could also refactor if we wanted to. Alternatively, we could create a more generic `ProbeServer` for all the simple probelike requests ( liveness, ready, metrics, startup, etc ) and slowly transition.

But of course, we could just ship it... :boat:

lib/network/probe/HealthProbeServer.js
```diff
@@ -0,0 +1,99 @@
...
+        // parent's method for that
+        this.onRequest(this._onRequest);
+        this._reqHandlers = {
+            '/_/health/liveness': this._onLiveness.bind(this),
+            '/_/health/readiness': this._onReadiness.bind(this),
+            '/_/monitoring/metrics': this._onMetrics.bind(this),
+            '/_/metrics': this._onMetrics.bind(this),
+        };
+        this._livenessCheck = params.livenessCheck || checkStub;
+        this._readinessCheck = params.readinessCheck || checkStub;
```
miniscruff, line 49: Thinking about this more and re-reading k8s docs I feel that these should not default to "an existing route that returns 200 OK" and just not exist if not defined.

miniscruff:
This was handled by creating a new ProbeServer in a different PR

PR #1428, opened 2021-04-30
feature: S3C-4073 Add new probe server

miniscruff:
* JsDocs for arsenal error
* ProbeServer as a replacement to HealthProbeServer

### DIFF
--- a/lib/network/http/server.js
+++ b/lib/network/http/server.js
@@ -8,6 +8,7 @@ const ciphers = require('../../https/ciphers').ciphers;
 const errors = require('../../errors');
 const { checkSupportIPv6 } = require('./utils');
 
+
 class Server {
 
     /**
@@ -186,7 +187,7 @@ class Server {
      * Function called when no handler specified in the server
      *
      * @param {http.IncomingMessage|https.IncomingMessage} req - Request object
-     * @param {http.ServerResponse|https.ServerResponse} res - Response object
+     * @param {http.ServerResponse} res - Response object
      * @return {undefined}
      */
     _noHandlerCb(req, res) {
@@ -203,8 +204,8 @@ class Server {
     /**
      * Function called when request received
      *
-     * @param {http.IncomingMessage|https.IncomingMessage} req - Request object
-     * @param {http.ServerResponse|https.ServerResponse} res - Response object
+     * @param {http.IncomingMessage} req - Request object
+     * @param {http.ServerResponse} res - Response object
      * @return {undefined}
      */
     _onRequest(req, res) {
@@ -375,7 +376,7 @@ class Server {
      * where the value is not 100-continue
      *
      * @param {http.IncomingMessage|https.IncomingMessage} req - Request object
-     * @param {http.ServerResponse|https.ServerResponse} res - Response object
+     * @param {http.ServerResponse} res - Response object
      * @return {undefined}
      */
     _onCheckExpectation(req, res) {
@@ -387,7 +388,7 @@ class Server {
      * is received
      *
      * @param {http.IncomingMessage|https.IncomingMessage} req - Request object
-     * @param {http.ServerResponse|https.ServerResponse} res - Response object
+     * @param {http.ServerResponse} res - Response object
      * @return {undefined}
      */
     _onCheckContinue(req, res) {
--- a/lib/network/probe/ProbeServer.js
+++ b/lib/network/probe/ProbeServer.js
@@ -1,0 +1,109 @@
+const httpServer = require('../http/server');
+const werelogs = require('werelogs');
+const errors = require('../../errors');
+
+const DEFAULT_LIVE_ROUTE = '/_/live';
+const DEFAULT_READY_ROUTE = '/_/live';
+const DEFAULT_METRICS_ROUTE = '/_/metrics';
+
+/**
+ * ProbeDelegate is used to determine if a probe is successful or
+ * if any errors are present.
+ * If everything is working as intended, it is a no-op.
+ * Otherwise, return a string representing what is failing.
+ * @callback ProbeDelegate
+ * @param { import('http').ServerResponse } res - HTTP response for writing
+ * @param {werelogs.Logger} log - Werelogs instance for logging if you choose to
+ * @return {(string|undefined)} String representing issues to report. An empty
+ * string or undefined is used to represent no issues.
+ */
+
+/**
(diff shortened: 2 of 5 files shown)


tests/unit/network/probe/ProbeServer.spec.js
```diff
@@ -0,0 +1,105 @@
...
+    }).end();
+}
+
+describe('network.probe.ProbeServer', () => {
+    /** @type {ProbeServer} */
+    let server;
+
+    beforeEach(done => {
+        server = new ProbeServer({ port: 4042 });
+        server._cbOnListening = done;
```
naren-rajendran, line 29: what does this do?

tests/unit/errors.js
```diff
@@ -26,4 +26,17 @@ describe('Errors: ', () => {
...
+        const fakeRes = {
+            writeHead: code => assert.strictEqual(code, 404),
+            end: msg => {
+                assert.strictEqual(
+                    msg,
+                    errors.NoSuchEntity.toString(),
+                );
+            },
+        };
+        errors.NoSuchEntity.writeResponse(fakeRes);
```
naren-rajendran, line 40: Neat fake response 👍🏽

tests/unit/network/probe/ProbeServer.spec.js
```diff
@@ -0,0 +1,105 @@
...
+
+    it('error on bad method', done => {
+        makeRequest('POST', '/unused', (err, res) => {
+            assert.ifError(err);
+            assert.strictEqual(res.statusCode, 405);
+            done();
+        });
+    });
+
+    it('error on missing request', done => {
```
naren-rajendran: missing request?

tests/unit/network/probe/ProbeServer.spec.js
miniscruff: I think this is a callback used for when the server is up and listening, I believe

tests/unit/network/probe/ProbeServer.spec.js
miniscruff: Like it is missing the /unused request endpoint.
Edit: Could maybe use a better name...

lib/network/http/server.js
```diff
@@ -203,8 +204,8 @@ class Server {
     /**
      * Function called when request received
      *
-     * @param {http.IncomingMessage|https.IncomingMessage} req - Request object
-     * @param {http.ServerResponse|https.ServerResponse} res - Response object
+     * @param {http.IncomingMessage} req - Request object
+     * @param {http.ServerResponse} res - Response object
```
anurag4DSB, line 208: Is it just standard practice to write jsdoc or we do not support receiving HTTPs requests?

anurag4DSB: Same at other places.

lib/network/probe/ProbeServer.js
```diff
@@ -0,0 +1,109 @@
...
+const DEFAULT_LIVE_ROUTE = '/_/live';
+const DEFAULT_READY_ROUTE = '/_/live';
+const DEFAULT_METRICS_ROUTE = '/_/metrics';
+
+/**
+ * ProbeDelegate is used to determine if a probe is successful or
+ * if any errors are present.
+ * If everything is working as intended, it is a no-op.
+ * Otherwise, return a string representing what is failing.
+ * @callback ProbeDelegate
```
anurag4DSB, line 14: How is ProbeDelegate a callback?

anurag4DSB (approved):
Just curious about some things.
But otherwise looks good to me.

lib/network/probe/ProbeServer.js
```diff
@@ -0,0 +1,109 @@
...
+    _onRequest(req, res) {
+        const log = this.logging.newRequestLogger();
+        log.debug('request received', { method: req.method, url: req.url });
+
+        if (req.method !== 'GET') {
+            errors.MethodNotAllowed.writeResponse(res);
+            return;
+        }
+
+        if (!(req.url in this._handlers)) {
```
tmacro: Functionally the same, but `if (!this._handlers.includes(req.url))` avoids the nested `()`

tests/unit/network/probe/ProbeServer.spec.js
naren-rajendran: invalid route?

lib/network/http/server.js
miniscruff: There is no https request, I looked it up, https server uses the same response and request as http.

lib/network/probe/ProbeServer.js
miniscruff: There is not jsdoc for delegates so I used callback here as the closest but also included the @returns to specify you can return values like a normal delegate.

lib/network/probe/ProbeServer.js
miniscruff: Objects do not have an includes method, I can do:
```suggestion
        if (!this._handlers.hasOwnProperty(req.url)) {
```

lib/network/probe/ProbeServer.js
miniscruff: https://jsdoc.app/tags-callback.html says you can specify the return value, seems JsDoc does not differentiate between a callback ( no return value ) and a delegate ( one or more values ) and just has one group. But we can at least use naming to separate them.

lib/network/probe/ProbeServer.js
miniscruff: I could use a Map instead and then use the `has` method. Would prevent any weird stuff from the built-in object methods and be more aligned with it being a map and not an Object with methods and things.

naren-rajendran (approved):

lib/network/probe/ProbeServer.js
miniscruff: @tmacro switched to a Map, let me know your thoughts

lib/network/probe/ProbeServer.js
anurag4DSB: This reminds me of our discussion of specifying `@returns {undefined}` or `@returns {callback}`.
I think this info completes that loop :)

lib/network/probe/ProbeServer.js
tmacro: TIL about Map :+1:

miniscruff: ooh Maps are super great, would recommend.

PR #1436, opened 2021-05-05
impr(S3C-4336): Add BucketInfoModelVersion.md from cloudserver

tmacro:


BucketInfoModelVersion.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Usage
+
+Used to determine which splitter to use ( < 2 means old splitter)
+
+## Model version 3
+
+### Properties Added
+
+```
```
anurag4DSB:
```suggestion
```javascript
```

anurag4DSB (approved):

BucketInfoModelVersion.md
```diff
@@ -0,0 +1,82 @@
+# BucketInfo Model Version History
```
jonathan-gramain: Can you create a "documentation" directory and put this file in there? Just so it looks more tidy :broom: 

Or put it in lib/models/ directly, close to the code, e.g. I already see lib/storage/metadata/mongoclient/Mongoclient.md.

miniscruff: I am a huge fan of sphinx `versionAdded` label and think it would work great here.
Turns out JsDoc has this here https://jsdoc.app/tags-since.html.

```javascript
this._acl = aclInstance;
this._name = name;
this._owner = owner;
this._ownerDisplayName = ownerDisplayName;
this._creationDate = creationDate;
/** JSDOC information here
 * @since v2
 * @type {number}
 */
this._mdBucketModelVersion = mdBucketModelVersion || 0
//etc
```

Could probably remove the markdown entirely and just list it in the javascript too.

vrancurel:
Can we create a doc folder at the root of the project ?

saitharunthodupunuru (approved):

rahulreddy:
This is a much bigger problem than we anticipated. I suggest we confine the scope of this PR to only a move of the doc and nothing more. I will create a follow up discussion to think about making it better.

PR #1478, opened 2021-06-08
ft: S3C-4456 tiny version IDs

vrancurel:
For some applications like OSE.

### DIFF
--- a/lib/versioning/VersionID.js
+++ b/lib/versioning/VersionID.js
@@ -6,6 +6,10 @@
 // - rep_group_id           07 bytes        replication group identifier
 // - other_information      arbitrary       user input, such as a unique string
 
+const base62Integer = require('base62');
+const BASE62 = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ';
+const base62String = require('base-x')(BASE62);
+
 // the lengths of the components in bytes
 const LENGTH_TS = 14; // timestamp: epoch in ms
 const LENGTH_SEQ = 6; // position in ms slot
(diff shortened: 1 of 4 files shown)


tests/unit/versioning/VersionID.js
```diff
@@ -39,4 +39,48 @@ describe('test generating versionIds', () => {
...
+    });
+
+    it('should encode and decode versionIds', () => {
+        const encoded = vids.map(vid => VID.tinyEncode(vid));
+        const decoded = encoded.map(vid => VID.tinyDecode(vid));
+        assert.strictEqual(vids.length, count);
+        assert.deepStrictEqual(vids, decoded);
+    });
+
+    it('shall encode/decode tiny versionIds', () => {
```
jonathan-gramain: Could say in the description that it is an error case

tests/unit/versioning/VersionID.js
```diff
@@ -39,4 +39,48 @@ describe('test generating versionIds', () => {
         assert.strictEqual(vids.length, count);
         assert.deepStrictEqual(vids, decoded);
     });
+
+    it('simple tiny version test', () => {
```
jonathan-gramain, line 43: Do you have other test cases available? E.g. with version IDs of different length, I believe they can vary by one to a few characters in length.

lib/versioning/VersionID.js
```diff
@@ -152,4 +156,73 @@ function decode(str) {
...
+    const part2 = Number(str.substring(half, total));
+    const part3 = Buffer.from(str.substring(total));
+    const enc1 = base62Integer.encode(part1);
+    const enc2 = base62Integer.encode(part2);
+    const enc3 = base62String.encode(part3);
+    return base62Integer.encode(enc1.length) +
+        enc1 +
+        base62Integer.encode(enc2.length) +
+        enc2 +
+        base62Integer.encode(enc3.length) +
```
jonathan-gramain: You may not need to include `enc3.length` since it is the last field, it must end at the end of the whole encoded string

lib/versioning/VersionID.js
```diff
@@ -152,4 +156,73 @@ function decode(str) {
...
+ *
+ * @param {string} str - the encoded tinyVersionId to decode
+ * @return {(string|Error)} - the decoded versionId or an error
+ */
+function tinyDecode(str) {
+    try {
+        let start = 0;
+        const enc1LenStr = str.substring(start, 1);
+        const enc1Len = base62Integer.decode(enc1LenStr);
+        if (enc1Len > 6) {
```
jonathan-gramain: This check makes me think that if you pad the encoded `enc1` with leading 0s (or the equivalent base62 character) to reach 6 chars, you don't need to embed the length, and similarly for `enc2`. Not sure of this but in case `enc1` is usually 6 chars, you can save one char.

lib/versioning/VersionID.js
vrancurel: If I don't encode the length, I cannot detect if there is garbage at the end of the provided version ID, so the decoder will attempt to decode garbage (last error test case).

lib/versioning/VersionID.js
vrancurel: Indeed.

tests/unit/versioning/VersionID.js
```diff
@@ -39,4 +39,52 @@ describe('test generating versionIds', () => {
...
+        const vid = '98376906954349999999RG001  145.20.5';
+        const tinyVersionId = VID.tinyEncode(vid);
+        const orig = VID.tinyDecode(tinyVersionId);
+        assert.strictEqual(vid, orig);
```
miniscruff: I am generally pretty skeptical of using two methods to check equality here. This does not guarantee they are right only that they are the same. Had a production failure once because our JSON tool had an issue that both the reader and writer had and the only test case was like this. This may not apply here as I have no written any complicated encoding myself but just my two cents.

vrancurel: There is a test below which tests a bunch of random strings.

lib/versioning/VersionID.js
```diff
@@ -152,4 +156,61 @@ function decode(str) {
...
+        let start = 0;
+        const enc1 = str.substring(start, start + TV_EPAD.length);
+        const orig1 = base62Integer.decode(enc1);
+        start += TV_EPAD.length;
+        const enc2 = str.substring(start, start + TV_EPAD.length);
+        const orig2 = base62Integer.decode(enc2);
+        start += TV_EPAD.length;
+        const enc3 = str.substring(start);
+        const orig3 = base62String.decode(enc3);
+        return (TV_DPAD + orig1.toString()).slice(-TV_DPAD.length) +
```
saitharunthodupunuru, line 207: Why don't we slice `TV_DPAD.length` on `enc1` or `enc2` before decoding to make sure the decode was done on same encoded string

saitharunthodupunuru (changes requested):

tests/unit/versioning/VersionID.js
miniscruff: Right, that takes a bunch of random strings, encodes them, and then decodes them back for assertions.
But that middle part is not tested at all, again this may not be something that is generally tested in these scenarios I don't know. But there are no one-way tests like:

```javascript
assert.strictEqual(Vid.tinyEncode('some random string'), 'some constant encoded result');
```

Then of course a separate one way test the other way.

vrancurel: ok

vrancurel:
reopening as S3C-4552

PR #1480, opened 2021-06-09
ft: S3C-4552 tiny version IDs

vrancurel:
Will be enabled on new buckets only.

### DIFF
--- a/lib/versioning/VersionID.js
+++ b/lib/versioning/VersionID.js
@@ -6,6 +6,10 @@
 // - rep_group_id           07 bytes        replication group identifier
 // - other_information      arbitrary       user input, such as a unique string
 
+const base62Integer = require('base62');
+const BASE62 = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ';
+const base62String = require('base-x')(BASE62);
+
 // the lengths of the components in bytes
 const LENGTH_TS = 14; // timestamp: epoch in ms
 const LENGTH_SEQ = 6; // position in ms slot
@@ -127,7 +131,7 @@ function generateVersionId(info, replicationGroupId) {
  * @param {string} str - the versionId to encode
  * @return {string} - the encoded versionId
  */
-function encode(str) {
+function hexEncode(str) {
     return Buffer.from(str, 'utf8').toString('hex');
 }
 
@@ -138,7 +142,7 @@ function encode(str) {
  * @param {string} str - the encoded versionId to decode
  * @return {(string|Error)} - the decoded versionId or an error
  */
-function decode(str) {
+function hexDecode(str) {
     try {
         const result = Buffer.from(str, 'hex').toString('utf8');
         if (result === '') {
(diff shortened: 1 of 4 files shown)


jonathan-gramain (approved):

saitharunthodupunuru (approved):
LGTM

vrancurel:
I need to rename encode()/decode() in legacyEncode()/legacyDecode() and write a wrapper encode(tiny=true)/decode(tiny=true)

lib/versioning/VersionID.js
```diff
@@ -6,6 +6,10 @@
 // - rep_group_id           07 bytes        replication group identifier
 // - other_information      arbitrary       user input, such as a unique string
 
+const base62Integer = require('base62');
+const BASE62 = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ';
+const base62String = require('base-x')(BASE62);
+
 // the lengths of the components in bytes
 const LENGTH_TS = 14; // timestamp: epoch in ms
 const LENGTH_SEQ = 6; // position in ms slot
```
saitharunthodupunuru, line 15: Is it a wise move have these values fixed?

lib/versioning/VersionID.js
```diff
@@ -152,4 +156,94 @@ function decode(str) {
...
+        (TV_EPAD + enc2).slice(-TV_EPAD.length) +
+        enc3;
+}
+
+/**
+ * Decode a tinyVersionId. May return an error if the input string is
+ * invalid hex string or results in an invalid value.
+ *
+ * @param {string} str - the encoded tinyVersionId to decode
+ * @return {(string|Error)} - the decoded versionId or an error
```
miniscruff, line 194: I am not sure returning a string or error is the best case as handling it is a bit hard.
I think just using @throws is fine?

```javascript
const decoded = tinyDecode('blah');
// have to check type of returned value
if (decoded typeof string) {
   // use decoded?
} else {
    // handle the error?
}
```

lib/versioning/VersionID.js
```diff
@@ -138,7 +142,7 @@ function encode(str) {
  * @param {string} str - the encoded versionId to decode
  * @return {(string|Error)} - the decoded versionId or an error
  */
-function decode(str) {
+function legacyDecode(str) {
```
miniscruff: instead of calling this legacy ( which just describes the time in which it was created ) calling it like hexEncode/Decode which describes what it is doing.

lib/versioning/VersionID.js
miniscruff: Note I realize now that it matches the other form so in keeping it consistent I guess it is fine even if not ideal.

lib/versioning/VersionID.js
```diff
@@ -152,4 +156,94 @@ function decode(str) {
...
+        return err;
+    }
+}
+
+/**
+ * Encode a versionId to obscure internal information contained
+ * in a version ID.
+ *
+ * @param {string} str - the versionId to encode
+ * @param {boolean} [tiny=false] - use the tiny format (optional)
```
miniscruff: could potentially switch to string/enum but it may be overkill. I only mention it cause adding a new default parameter is backwards compatible, but switching a parameter from bool to something else is not.

lib/versioning/VersionID.js
vrancurel: Yes it is to keep it consistent with the legacy routine.

lib/versioning/VersionID.js
miniscruff: Another comment here is that the brackets around tiny=false is ESLints optional format so the (optional) text at the end is redundant.

vrancurel: You are right, maybe I will use 2 exported constants VID_HEX=0 and VID_BASE62=1.
I will rename "tiny" to "base62" as well.

lib/versioning/VersionID.js
```diff
@@ -152,4 +156,98 @@ function decode(str) {
...
+
+const ENC_TYPE_HEX = 0; // legacy (large) encoding
+const ENC_TYPE_BASE62 = 1; // new (tiny) encoding
+
+/**
+ * Encode a versionId to obscure internal information contained
+ * in a version ID.
+ *
+ * @param {string} str - the versionId to encode
+ * @param {boolean} [encType=ENC_TYPE_HEX] - encode format
```
miniscruff: I think this is a number now not a boolean

vrancurel:
So `decode()` is sometimes called in Cloudserver w/o context (e.g. before we get the bucket MD) so instead of taking the risk of changing the logic, we can just detect the format in the function. For now a heuristic.

miniscruff (approved):

anurag4DSB (approved):

saitharunthodupunuru (approved):

PR #1490, opened 2021-06-25
S3C-4618: backport HealthProbeServer

alexanderchan-scality:


alexanderchan-scality:
opt to use generic probe server instead

PR #1510, opened 2021-07-15
improvement: S3C-4312 encryption info in ObjectMDLocation.setDataLocation

jonathan-gramain:
Support setting encryption info in ObjectMDLocation with the method
setDataLocation(), used by backbeat to set the new target location
before writing metadata on the target.

### DIFF
--- a/lib/models/ObjectMDLocation.js
+++ b/lib/models/ObjectMDLocation.js
@@ -22,6 +22,8 @@ class ObjectMDLocation {
             size: locationObj.size,
             dataStoreName: locationObj.dataStoreName,
             dataStoreETag: locationObj.dataStoreETag,
+            cryptoScheme: locationObj.cryptoScheme,
+            cipheredDataKey: locationObj.cipheredDataKey,
         };
     }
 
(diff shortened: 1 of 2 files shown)


lib/models/ObjectMDLocation.js
```diff
@@ -33,9 +33,25 @@ class ObjectMDLocation {
...
+     * is encrypted: the base64-encoded ciphered data key
+     * @return {ObjectMDLocation} return this
+     */
     setDataLocation(location) {
         this._data.key = location.key;
         this._data.dataStoreName = location.dataStoreName;
+        if (location.cryptoScheme) {
+            this._data.cryptoScheme = location.cryptoScheme;
+            this._data.cipheredDataKey = location.cipheredDataKey;
+        }
```
miniscruff: Could you do this all with Object.assign?
```suggestion
        this._data = Object.assign(this._data, location);
```

miniscruff (approved):

alexanderchan-scality (approved):

lib/models/ObjectMDLocation.js
```diff
@@ -33,9 +33,20 @@ class ObjectMDLocation {
...
+     * @param {number} [location.cryptoScheme] - if location data is
+     * encrypted: the encryption scheme version
+     * @param {string} [location.cipheredDataKey] - if location data
+     * is encrypted: the base64-encoded ciphered data key
+     * @return {ObjectMDLocation} return this
+     */
     setDataLocation(location) {
-        this._data.key = location.key;
-        this._data.dataStoreName = location.dataStoreName;
+        this._data = Object.assign(this._data, location);
```
jonathan-gramain: @miniscruff I realized that neither the old code nor the change to use `Object.assign` are correct: indeed, we need to *overwrite* any encryption field with the one provided in the new location, i.e. the difference is that if the encryption params do not exist in the `location` parameter, they must be erased from `this._data.` I saw this in the Zenko tests that contain ObjectMDLocation tests, I'm considering backporting them to this 7.10 branch as well.

miniscruff: oh ok, that makes sense. Will re-review when it is ready.

miniscruff (approved):
Re-approving

alexanderchan-scality (approved):

PR #1516, opened 2021-07-21
improvement: S3C-4312 fix ObjectMDLocation.setDataLocation()

jonathan-gramain:
Fix ObjectMDLocation.setDataLocation() behavior when cryptoScheme and
cipheredDataKey location params are undefined: instead of setting the
attributes as undefined, remove the attributes.

The previous situation made some backbeat tests fail due to those
attributes existing, and it's cleaner this way.

### DIFF
--- a/lib/models/ObjectMDLocation.js
+++ b/lib/models/ObjectMDLocation.js
@@ -14,6 +14,10 @@ class ObjectMDLocation {
      * @param {string} locationObj.dataStoreName - type of data store
      * @param {string} locationObj.dataStoreETag - internal ETag of
      *   data part
+     * @param {number} [location.cryptoScheme] - if location data is
+     * encrypted: the encryption scheme version
+     * @param {string} [location.cipheredDataKey] - if location data
+     * is encrypted: the base64-encoded ciphered data key
      */
     constructor(locationObj) {
         this._data = {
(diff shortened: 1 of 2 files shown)


lib/models/ObjectMDLocation.js
```diff
@@ -54,7 +60,11 @@ class ObjectMDLocation {
             'cryptoScheme',
             'cipheredDataKey',
         ].forEach(attrName => {
-            this._data[attrName] = location[attrName];
+            if (location[attrName] !== undefined) {
```
alexanderchan-scality, line 63: should the check be against `null` or `undefined` instead?  ~(or any `falsy` value)~ (`falsy` check may be incorrect as scheme is a number type)

jonathan-gramain: `undefined` is what we want here, we want to check if the location does not contain such attribute, where this test would return `true`. If it is `null` we would want to copy the `null` value and not delete it.

alexanderchan-scality (approved):

lib/models/ObjectMDLocation.js
miniscruff: I always used `hasOwnProperty` for these checks but I don't see that used anywhere here. It is a very similar check but it also handles if that property is itself undefined.
```js
> a
{ a: 5, b: undefined }
> a['b'] !== undefined
false
> a.hasOwnProperty('b')
true
```

Idk if this changes the logic, just figured I would mention it.

jonathan-gramain: Right, I could have used it here. In case it would come with an existing attribute with `undefined` value for some reason (bug in code somewhere else, whatever) I think it's good to clean it up though.

miniscruff (approved):

PR #1518, opened 2021-07-23
bugfix: ARSN-4 rework KMIP connection handling

jonathan-gramain:
Rework KMIP connection handling to catch all errors, including before the connection is established, and return the error to each pending command response.

In particular, setup the 'error' listener (also 'data' and 'end' listeners) as soon as the TLS client socket is created instead of
waiting for the connection to be established to set the listeners.


jonathan-gramain:
ping

tcarmet:
ping

jonathan-gramain:
ping

tmacro (approved):

mrepop (approved):

PR #1523, opened 2021-07-28
Bugfix/arsn 6/reverse arsn 3

scality-gdoumergue:
The patches for S3C-656 and ARSN-3 made more harm than good, because nginx doesn't like touching the Expect HTTP header. Reverting.


anurag4DSB (approved):

rdebay-scality (approved):

rdebay-scality:
bert-e ?

PR #1545, opened 2021-08-16
feature: ARSN-12-introduce-cond-put-op

anurag4DSB:
Introduced a way to add conditions.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -40,6 +40,7 @@ class IndexTransaction {
         this.operations = [];
         this.db = db;
         this.closed = false;
+        this.conditions = [];
     }
 
     /**
@@ -119,6 +120,35 @@ class IndexTransaction {
     }
 
     /**
+     * Adds a condition for the transaction
+     *
+     * @argument {object} condition an object with the following attributes:
+     *                    {
+     *                      <condition>: the object key
+     *                    }
+     *                    example: { notExists: 'key1' }
+     *
+     * @throws {Error} an error described by the following properties
+     *                 - pushOnCommittedTransaction if already committed
+     *                 - missingCondition if the condition is empty
+     *
+     * @returns {undefined}
+     */
+    addCondition(condition) {
+        if (this.closed) {
+            throw propError('pushOnCommittedTransaction',
+                            'can not add conditions to already committed transaction');
+        }
+        if (condition === undefined || Object.keys(condition).length === 0) {
+            throw propError('missingCondition', 'missing condition for conditional put');
+        }
+        if (typeof (condition.notExists) !== 'string') {
+            throw propError('unsupportedConditionalOperation', 'missing key or supported condition');
+        }
+        this.conditions.push(condition);
+    }
+
+    /**
      * Applies the queued updates in this transaction atomically.
      *
      * @argument {function} cb function to be called when the commit
@@ -138,6 +168,7 @@ class IndexTransaction {
         }
 
         this.closed = true;
+        writeOptions.conditions = this.conditions;
 
         // The array-of-operations variant of the `batch` method
         // allows passing options such has `sync: true` whereas the
--- a/tests/unit/db.js
+++ b/tests/unit/db.js
@@ -1,14 +1,22 @@
 'use strict';// eslint-disable-line strict
 
 const assert = require('assert');
+const async = require('async');
 
 const leveldb = require('level');
 const temp = require('temp');
 temp.track();
 
 const db = require('../../index').db;
+const errors = require('../../lib/errors');
 
 const IndexTransaction = db.IndexTransaction;
+const key1 = 'key1';
+const key2 = 'key2';
+const key3 = 'key3';
+const value1 = 'value1';
+const value2 = 'value2';
+const value3 = 'value3';
 
 function createDb() {
     const indexPath = temp.mkdirSync();
@@ -37,6 +45,45 @@ function checkValueNotInDb(db, k, done) {
 
         return done();
     });
+}
+
+function checkKeyNotExistsInDB(db, key, cb) {
+    return db.get(key, (err, value) => {
+        if (err && !err.notFound) {
+            return cb(err);
+        }
+        if (value) {
+            return cb(errors.PreconditionFailed);
+        }
+        return cb();
+    });
+}
+
+class ConditionalLevelDB {
+    constructor() {
+        this.db = createDb();
+    }
+
+    batch(operations, writeOptions, cb) {
+        return async.eachLimit(writeOptions.conditions, 10, (cond, asyncCallback) => {
+            switch (true) {
+            case ('notExists' in cond):
+                checkKeyNotExistsInDB(this.db, cond.notExists, asyncCallback);
+                break;
+            default:
+                asyncCallback(new Error('unsupported conditional operation'));
+            }
+        }, err => {
+            if (err) {
+                return cb(err);
+            }
+            return this.db.batch(operations, writeOptions, cb);
+        });
+    }
+
+    get client() {
+        return this.db;
+    }
 }
 
 describe('IndexTransaction', () => {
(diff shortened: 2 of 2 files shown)


jonathan-gramain (changes requested):
I generally agree with the principle of doing a conditional operation, but I think it should be changed a bit on the semantic level:

- the condition(s) should apply to the whole batch instead of on any single batch entry, since it is a global condition that will be checked before applying or not the whole batch (e.g. in some other use case there might be an existence check on a key that does not appear in the batch at all)
- hence instead of a new API call `condPut`, we could have something like `addCondition(condObject)`which takes an object specifying a condition to check that must be true for the whole batch to be applied (and there may be more than one condition with a repeated call, in which case all conditions must be fulfilled, even though we may only support a single condition for now)
- I also think that re-using the HTTP semantics may not be the best fit here and we might better come up with our own scheme (for example, the HTTP `If*` conditional header specs can check timestamps etc. and has a specific implicit semantic for what is matched like etag etc. which does not apply in the general case of database precondition checks).

For example, I am thinking something like that could be a good fit for the relatively simple need of today while staying extensible enough for future use cases:
```
tx.put('some-key', 'some-value');
tx.put('some-other-key', 'some-other-value');
tx.addCondition({ notExists: 'some-key-that-should-not-exist' });
```
And for now we only have to support the `notExists` condition on the backend, by doing a database GET on the specified key, which result (exists or `notFound`) conditions the batch execution.

anurag4DSB:
> I generally agree with the principle of doing a conditional operation, but I think it should be changed a bit on the semantic level...

Changed the approach in squash commit: https://github.com/scality/Arsenal/pull/1545/commits#:~:text=%5Bsquash%5D%20post%20review

mrepop (approved):

lib/db.js
```diff
@@ -118,6 +118,39 @@ class IndexTransaction {
...
+     */
+    addCondition(condition) {
+        if (this.closed) {
+            throw propError('pushOnCommittedTransaction',
+                            'can not add ops to already committed transaction');
+        }
+        if (condition === undefined || Object.keys(condition).length === 0) {
+            throw propError('missingCondition', 'missing condition for conditional put');
+        }
+        if (!('notExists' in condition)) {
```
naren-rajendran: Looks like, a better conditional check with `in` operator is, `if ('notExists' in condition === false) {}`, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in

jonathan-gramain: Why is it better?

jonathan-gramain: Also we might merge those two tests into a single `if (typeof condition.notExists !== 'string') { /*error*/ }` (for example)

lib/db.js
```diff
@@ -118,6 +118,39 @@ class IndexTransaction {
...
+            throw propError('unsupportedConditionalOperation', 'unsupported conditional operation');
+        }
+        if (condition.notExists === undefined) {
+            throw propError('missingKey', 'missing key for conditional operation');
+        }
+        this.operations.push(condition);
```
jonathan-gramain: Let's create another object `this.conditions` for conditions: since we are committing a batch, it does not make sense to keep the order of condition checks, as we will always:

1. check all conditions
2. apply the batch (if all conditions are ok)

It will also simplify the check loop (no need to do `splice` etc)

lib/db.js
anurag4DSB: Thanks! your way much better and readable.
changed in [`35b801e` (#1545)](https://github.com/scality/Arsenal/pull/1545/commits/35b801e107f803d621aa54b80d46fca2b74a950a)

tests/unit/db.js
```diff
@@ -39,6 +47,47 @@ function checkValueNotInDb(db, k, done) {
     });
 }
 
+function checkKeyInDB(db, key, cb) {
+    return db.get(key, (err, value) => {
+        if (value || (err && !err.stack.includes('Error: NotFound:'))) {
```
jonathan-gramain: Don't we have an attribute `err.notFound` for this? LevelDB returns one I'm pretty sure, so if `ConditionalLevelDB` forwards the error as is it should be there I believe.

tests/unit/db.js
```diff
@@ -317,4 +366,114 @@ describe('IndexTransaction', () => {
...
+        transaction.addCondition({ notExists: 'key1' });
+        transaction.commit(err => {
+            if (err) {
+                return done(err);
+            }
+            return checkValueInDb(client, key1, value1, done);
+        });
+    });
+
+    it('should not allow any op in a batch operation with notExits condition if key exists', done => {
```
jonathan-gramain:
```suggestion
    it('should not allow any op in a batch operation with notExists condition if key exists', done => {
```

tests/unit/db.js
```diff
@@ -39,6 +47,47 @@ function checkValueNotInDb(db, k, done) {
     });
 }
 
+function checkKeyInDB(db, key, cb) {
```
jonathan-gramain: It's confusing to have a check function that returns no error whether the object exists or not (if my understanding is correct): I suggest splitting into `checkKeyExistsInDB` and `checkKeyNotExistsInDB`, and using the appropriate one wherever it makes sense.

jonathan-gramain (changes requested):

lib/db.js
naren-rajendran: better to read I guess

tests/unit/db.js
```diff
@@ -39,6 +47,47 @@ function checkValueNotInDb(db, k, done) {
...
+                next => {
+                    operations.splice(index, 1);
+                    return next();
+                },
+            ], asyncCallback);
+        }, err => {
+            if (err) {
+                return cb(err);
+            }
+            return this.db.batch(operations, writeOptions, cb);
```
jonathan-gramain, line 80: It should then be ok to pass the `this.conditions` array inside the `writeOptions`, e.g. in `writeOptions.conditions`.

tests/unit/db.js
anurag4DSB: I thought the same.
From `db.get(key, (err, value) => {` I get an error which is a string. So I figured that's what LevelDB returns(console.log('this is the error from levelDB:', err):
```
this is the error from levelDB: [Error: NotFound: ]
```

lib/db.js
anurag4DSB: Merged the tests in review commit [`2583686` (#1545)](https://github.com/scality/Arsenal/pull/1545/commits/25836868e54ed1bcb3a5f1ee23dc43fd5f004844)

lib/db.js
anurag4DSB: Thanks. changed in in review commit [`2583686` (#1545)](https://github.com/scality/Arsenal/pull/1545/commits/25836868e54ed1bcb3a5f1ee23dc43fd5f004844)

tests/unit/db.js
anurag4DSB: changed in review commit [`2583686` (#1545)](https://github.com/scality/Arsenal/pull/1545/commits/25836868e54ed1bcb3a5f1ee23dc43fd5f004844)

tests/unit/db.js
anurag4DSB: renamed the function in review commit [`2583686` (#1545)](https://github.com/scality/Arsenal/pull/1545/commits/25836868e54ed1bcb3a5f1ee23dc43fd5f004844)

tests/unit/db.js
anurag4DSB: used it :) 
in review commit [`2583686` (#1545)](https://github.com/scality/Arsenal/pull/1545/commits/25836868e54ed1bcb3a5f1ee23dc43fd5f004844)

tests/unit/db.js
jonathan-gramain: It should not be a string. Did you check it e.g. with `typeof err`? Note that `console.log` does not show everything from the original error object, it just gives a simplified representation of it.

lib/db.js
```diff
@@ -138,11 +168,12 @@ class IndexTransaction {
         }
 
         this.closed = true;
+        writeOptions.conditions = this.conditions;
 
         // The array-of-operations variant of the `batch` method
         // allows passing options such has `sync: true` whereas the
         // chained form does not.
-        return this.db.batch(this.operations, writeOptions, cb);
+        return this.db.batch(this.operations, writeOptions, cb, this.conditions);
```
jonathan-gramain: I believe this change is not needed as `this.conditions` is passed via `writeOptions`

anurag4DSB: Removed. Was left over,

tests/unit/db.js
anurag4DSB: thanks for pointing me in the right direction over Slack.
It was the DB instance issue.

tests/unit/db.js
```diff
@@ -39,6 +47,42 @@ function checkValueNotInDb(db, k, done) {
     });
 }
 
+function checkKeyNotExistsInDB(db, key, cb) {
+    return db.get(key, (err, value) => {
+        if (value || (err && !err.notFound)) {
```
jonathan-gramain: I would rather separate the `err && !err.notFound` test and return the original error then, because if any other error happens on the DB we would rather not treat it as an "object exists" status. But it's minor as it should not affect the test ability to catch errors, so not blocking.

tests/unit/db.js
```diff
@@ -39,6 +47,42 @@ function checkValueNotInDb(db, k, done) {
...
+            return cb(errors.EntityAlreadyExists);
```
jonathan-gramain: Can you change this error to `PreconditionFailed` instead? I am going to use this error code in metadata, because we are dealing with generic condition checking hence we should return a generic "condition check fail" kind of error. It carries less information to the client, but the upside is that the client need not be aware of the various kinds of error codes that we would return eventually with more condition types, and if needed we may still add a small description of which condition(s) fail and attach it to the error object, but no need to add this description for now as we are just dealing with a single condition check.

jonathan-gramain (approved):

anurag4DSB:
Asked for re-review from @mrepop as a lot has changed.

mrepop (approved):

PR #1579, opened 2021-09-23
ARSN-20 migrate to github actions

tcarmet:


.github/workflows/pre-merge.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - name: run lint
+      run: yarn run --silent lint -- --max-warnings 0
+    - name: run lint_md
+      run: yarn run --silent lint_md
+    - name: run test
+      run: yarn run --silent test
+    - name: run ft_test
+      run: yarn run ft_test
+    - name: run executables tests
+      run: yarn install && yarn test
```
miniscruff: this looks duplicated unless there is some difference between `yarn install` and `yarn install --frozen-lockfile` which there shouldn't be.

.github/workflows/pre-merge.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - name: run lint yml
+      run: yarn run --silent lint_yml
+    - name: run lint
+      run: yarn run --silent lint -- --max-warnings 0
+    - name: run lint_md
+      run: yarn run --silent lint_md
+    - name: run test
+      run: yarn run --silent test
+    - name: run ft_test
+      run: yarn run ft_test
```
miniscruff:
```suggestion
      run: yarn ft_test
```

.github/workflows/pre-merge.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - name: install dependencies
+      run: yarn install --frozen-lockfile
+    - name: run lint yml
+      run: yarn run --silent lint_yml
+    - name: run lint
+      run: yarn run --silent lint -- --max-warnings 0
+    - name: run lint_md
+      run: yarn run --silent lint_md
+    - name: run test
+      run: yarn run --silent test
```
miniscruff:
```suggestion
      run: yarn --silent test
```

.github/workflows/pre-merge.yaml
```diff
@@ -0,0 +1,53 @@
...
+        node-version: '10'
+        cache: 'yarn'
+    - name: install dependencies
+      run: yarn install --frozen-lockfile
+    - name: run lint yml
+      run: yarn run --silent lint_yml
+    - name: run lint
+      run: yarn run --silent lint -- --max-warnings 0
+    - name: run lint_md
+      run: yarn run --silent lint_md
```
miniscruff:
```suggestion
      run: yarn --silent lint_md
```

.github/workflows/pre-merge.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - uses: actions/setup-node@v2
+      with:
+        node-version: '10'
+        cache: 'yarn'
+    - name: install dependencies
+      run: yarn install --frozen-lockfile
+    - name: run lint yml
+      run: yarn run --silent lint_yml
+    - name: run lint
+      run: yarn run --silent lint -- --max-warnings 0
```
miniscruff:
```suggestion
      run: yarn --silent lint -- --max-warnings 0
```

.github/workflows/pre-merge.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - name: Checkout
+      uses: actions/checkout@v2
+    - uses: actions/setup-node@v2
+      with:
+        node-version: '10'
+        cache: 'yarn'
+    - name: install dependencies
+      run: yarn install --frozen-lockfile
+    - name: run lint yml
+      run: yarn run --silent lint_yml
```
miniscruff:
```suggestion
      run: yarn --silent lint_yml
```

miniscruff (changes requested):

.github/workflows/pre-merge.yaml
tcarmet: wow, I've copied paste what we had in eve, this means that we have been running duplicated install since forever :sweat_smile:

tcarmet: actually, no, I don't think they are duplicated as the yarn install is running in a different directory with a different `package.json` I believe we should leave it this way.

miniscruff: oh yea I see the working-directory thing now

.github/workflows/pre-merge.yaml
```diff
@@ -0,0 +1,53 @@
...
+on:
+  push:
+    branches:
+    - 'user/**'
+    - 'feature/**'
+    - 'improvement/**'
+    - 'bugfix/**'
+    - 'w/**'
+    - 'q/**'
+    - 'dependabot/**'
```
miniscruff: are there branches we don't want to run on? this seems pretty inclusive.

tcarmet: :thinking: only one I can think is `development/**`

Should we change it to something like this?
```yaml
  push:
    branches-ignore:
    - 'development/**'
```

miniscruff: we could but I don't see much of an issue with running these tests on development as well, I think the confusing-ish part is that we call it the "pre-merge" pipeline when I think just calling it "tests" or "validations" or something similar to that effect. The fact that it happens in pre-merge is cool but irrelevant to what is actually happening. And we can run these pre, post or whenever. tl;dr name pipelines for what it does not when it happens

tcarmet: You're right, I've renamed the pipeline

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,50 @@
...
+      run: yarn install --frozen-lockfile
+    - name: run lint yml
+      run: yarn --silent lint_yml
+    - name: run lint
+      run: yarn --silent lint -- --max-warnings 0
+    - name: run lint_md
+      run: yarn --silent lint_md
+    - name: run test
+      run: yarn --silent test
+    - name: run ft_test
```
miniscruff: This is definitely a bike shedding thing but I think it looks better in logs when the name of sections say what is going on and not just the command.

```suggestion
    - name: lint yaml
      run: yarn --silent lint_yml
    - name: lint javascript
      run: yarn --silent lint -- --max-warnings 0
    - name: lint markdown
      run: yarn --silent lint_md
    - name: run unit tests
      run: yarn --silent test
    - name: run functional tests
```

miniscruff (approved):
ok as long as CI passes it is good for me

alexanderchan-scality (approved):

PR #1586, opened 2021-09-29
ARSN-22 - Fix bug in `onPutObjectTagging` and `getObjectTagging`

lavigne958:
do not use the version stored in the location object,
use the object version located in the object metadata.

quick fix on the way, use actual string instead of object
when building object key (for the bucket part).


lavigne958:
ping

lavigne958:
close this PR, need to re-work it. this is not the correct fix.

PR #1587, opened 2021-09-29
Bugfix/ARSN-22 fix put delete tagging

lavigne958:
when sending command (put|delete)-object-tagging AWS interface receives a bucket as argument and
when creating the key should only send the name of the bucket and not the whole bucket object.

Added a test on put/delete tagging methods.

### DIFF
--- a/tests/unit/storage/data/DummyService.js
+++ b/tests/unit/storage/data/DummyService.js
@@ -3,6 +3,8 @@ const { EventEmitter } = require('events');
 
 const DummyObjectStream = require('./DummyObjectStream');
 const { parseRange } = require('../../../../lib/network/http/utils');
+const assert = require('assert');
+const errors = require('../../../../lib/errors');
 
 const OBJECT_SIZE = 1024 * 1024 * 1024;
 
@@ -131,6 +133,38 @@ class DummyService {
               Math.min(Number.parseInt(rangeEnd, 10), OBJECT_SIZE - 1) : OBJECT_SIZE - 1;
         const objStream = new DummyObjectStream(firstByte, lastByte - firstByte + 1);
         objStream.pipe(writeStream);
+        return callback();
+    }
+    putObjectTagging(tagParams, callback) {
+        if (tagParams.Key === 'externalBackendTestBucket/externalBackendMissingKey') {
+            const err = errors.NoSuchKey;
+            return callback(err);
+        }
+
+        const keys = Object.keys(tagParams);
+        assert(keys.length > 0);
+        assert(tagParams.Tagging.TagSet.length > 0);
+        tagParams.Tagging.TagSet.forEach(tag => {
+            assert(tag.Key.length > 0);
+            assert(tag.Value.length > 0);
+        });
+
+        if (tagParams.VersionId) {
+            assert.strictEqual(tagParams.VersionId, 'latestversion');
+        }
+
+        return callback();
+    }
+    deleteObjectTagging(tagParams, callback) {
+        if (tagParams.Key === 'externalBackendTestBucket/externalBackendMissingKey') {
+            const err = errors.NoSuchKey;
+            return callback(err);
+        }
+
+        if (tagParams.VersionId) {
+            assert.strictEqual(tagParams.VersionId, 'latestversion');
+        }
+
         return callback();
     }
     // To-Do: add tests for other methods
--- a/tests/unit/storage/data/external/ExternalClients.js
+++ b/tests/unit/storage/data/external/ExternalClients.js
@@ -1,4 +1,5 @@
 const assert = require('assert');
+const async = require('async');
 const stream = require('stream');
 
 const AwsClient = require('../../../../../lib/storage/data/external/AwsClient');
@@ -7,6 +8,7 @@ const AzureClient =
     require('../../../../../lib/storage/data/external/AzureClient');
 const DummyService = require('../DummyService');
 const { DummyRequestLogger } = require('../../../helpers');
+const BucketInfo = require('../../../../../lib/models/BucketInfo');
 
 const backendClients = [
     {
@@ -154,6 +156,68 @@ describe('external backend clients', () => {
                     });
             });
         });
+
+        if (backend.config.type !== 'azure') {
+            it(`${backend.name} should set tags and then delete it`, done => {
+                const key = 'externalBackendTestKey';
+                const bucketData = {
+                    _name: 'externalBackendTestBucket',
+                    _owner: 'abcdef0123456789',
+                    _ownerDisplayName: 'UnitTestOwner',
+                    _creationDate: '2021-10-05T08:59:12.546Z',
+                };
+                const bucket = BucketInfo.fromObj(bucketData);
+                const objectMd = {
+                    tags: {
+                        Key1: 'value_1',
+                        Key2: 'value_2',
+                    },
+                    location: [
+                        {
+                            dataStoreVersionId: 'latestversion',
+                        },
+                    ],
+                };
+                async.series([
+                    next => testClient.objectPutTagging(key.key, bucket, objectMd, log, next),
+                    next => testClient.objectDeleteTagging(key.Key, bucket, objectMd, log, next),
+                ], done);
+            });
+
+            it(`${backend.name} should fail to set tag on missing key`, done => {
+                const key = 'externalBackendMissingKey';
+                const bucketData = {
+                    _name: 'externalBackendTestBucket',
+                    _owner: 'abcdef0123456789',
+                    _ownerDisplayName: 'UnitTestOwner',
+                    _creationDate: '2021-10-05T08:59:12.546Z',
+                };
+                const bucket = BucketInfo.fromObj(bucketData);
+                const objectMD = {
+                    tags: {
+                        Key1: 'value_1',
+                    },
+                    location: [
+                        {
+                            dataStoreVersionId: 'latestversion',
+                        },
+                    ],
+                };
+                async.series(
+                    [
+                        next => testClient.objectPutTagging(key, bucket, objectMD, log, (err) => {
+                            assert(err.ServiceUnavailable);
+                            next();
+                        }),
+                        next => testClient.objectDeleteTagging(key, bucket, objectMD, log, (err) => {
+                            assert(err.ServiceUnavailable);
+                            next();
+                        }),
+                    ],
+                    done,
+                );
+            });
+        }
         // To-Do: test the other external client methods (delete, createMPU ...)
     });
 });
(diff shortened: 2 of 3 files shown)


tests/unit/storage/data/DummyService.js
```diff
@@ -133,6 +134,42 @@ class DummyService {
...
+            err.message = 'NoSuchKey';
+            return callback(err);
+        }
+
+        const keys = Object.keys(tagParams);
+        assert.ok(keys.length > 0);
+        assert.ok(tagParams.Tagging.TagSet.length > 0);
+        tagParams.Tagging.TagSet.forEach(tag => {
+            assert.ok(tag.Key.length > 0);
+            assert.ok(tag.Value.length > 0);
```
miniscruff: You can just use assert
```suggestion
        assert(keys.length > 0);
        assert(tagParams.Tagging.TagSet.length > 0);
        tagParams.Tagging.TagSet.forEach(tag => {
            assert(tag.Key.length > 0);
            assert(tag.Value.length > 0);
```

Although I think it should be possible to know exactly how much of each is included and use `assert.strictEqual(x.length, y);`

tests/unit/storage/data/DummyService.js
```diff
@@ -133,6 +134,42 @@ class DummyService {
...
+
+        if (tagParams.VersionId) {
+            assert.ok(tagParams.VersionId === 'latestversion');
+        }
+
+        return callback();
+    }
+    deleteObjectTagging(tagParams, callback) {
+        if (tagParams.Key === 'externalBackendTestBucket/externalBackendMissingKey') {
+            const err = new Error();
```
miniscruff: You should be able to use errors.NoSuchKey for this, look at errors/arsenalErrors.json

tests/unit/storage/data/DummyService.js
```diff
@@ -133,6 +134,42 @@ class DummyService {
...
+            err.code = 404;
+            err.message = 'NoSuchKey';
+            return callback(err);
+        }
+
+        if (tagParams.VersionId) {
+            assert.ok(tagParams.VersionId === 'latestversion');
```
miniscruff:
```suggestion
            assert.strictEqual(tagParams.VersionId, 'latestversion');
```

tests/unit/storage/data/external/ExternalClients.js
```diff
@@ -48,6 +48,16 @@ const backendClients = [
 ];
 const log = new DummyRequestLogger();
 
+class DummyBucket {
```
miniscruff: I think we have other dummy classes around here but a sinon.spy could also work, I believe I added it to arsenal not too long ago.

tests/unit/storage/data/external/ExternalClients.js
```diff
@@ -154,6 +164,45 @@ describe('external backend clients', () => {
...
+                );
+                testClient.objectDeleteTagging(
+                    key,
+                    bucket,
+                    objectMd,
+                    log,
+                    (err) => {
+                        assert.ifError(err);
+                        done();
+                    },
```
miniscruff: You can pass in done as the callback and it will do the same thing, https://mochajs.org/#asynchronous-code

miniscruff (changes requested):

tests/unit/storage/data/external/ExternalClients.js
```diff
@@ -154,6 +164,45 @@ describe('external backend clients', () => {
                     });
             });
         });
+
+        if (backend.config.type !== 'azure') {
+            it(`${backend.name} objectPutTagging() then objectDeleteTagging() should set tags then delete it`, done => {
```
naren-rajendran: To keep it simple,
`it('${backend.name} should set tags and then delete it', done => {});`

tests/unit/storage/data/external/ExternalClients.js
naren-rajendran: Yep, `sinon` sounds good here!

tests/unit/storage/data/external/ExternalClients.js
```diff
@@ -154,6 +164,45 @@ describe('external backend clients', () => {
...
+                        Key1: 'value_1',
+                        Key2: 'value_2',
+                    },
+                    location: [
+                        {
+                            dataStoreVersionId: 'latestversion',
+                        },
+                    ],
+                };
+                testClient.objectPutTagging(
```
naren-rajendran:
```Suggestion
async.series([
    next => testClient.objectPutTagging(key, bucket, objectMd, log, next),
    next => testClient.objectDeleteTagging(key, bucket, objectMd, log, next),
], done);
```
Use async maybe?

naren-rajendran: If you decide to use `async`, you will have to add its reference on top!

tests/unit/storage/data/DummyService.js
```diff
@@ -3,6 +3,8 @@ const { EventEmitter } = require('events');
 
 const DummyObjectStream = require('./DummyObjectStream');
 const { parseRange } = require('../../../../lib/network/http/utils');
+const assert = require('assert');
+const errors = require('../../../../errors/arsenalErrors.json');
```
miniscruff: I don't think we import the json here, you should import lib/errors which will turn the JSON into objects. It is a little confusing.

tests/unit/storage/data/external/ExternalClients.js
```diff
@@ -154,6 +156,62 @@ describe('external backend clients', () => {
                     });
             });
         });
+
+        if (backend.config.type !== 'azure') {
+            it(`${backend.name} shoudl set tags and then delete it`, done => {
```
miniscruff:
```suggestion
            it(`${backend.name} should set tags and then delete it`, done => {
```

tests/unit/storage/data/DummyService.js
lavigne958: I need it for line 140 and line 160 to set the error to `NoSuchKey`.

lavigne958: My bad I understand what you mean, ok I will update it and set the error to an actual `ArsenalError` object

lavigne958: done

tests/unit/storage/data/external/ExternalClients.js
```diff
@@ -154,6 +156,62 @@ describe('external backend clients', () => {
...
+                    location: [
+                        {
+                            dataStoreVersionId: 'latestversion',
+                        },
+                    ],
+                };
+                async.series(
+                    [
+                        next => testClient.objectPutTagging(key, bucket, objectMD, log, (err) => {
+                            assert(err);
```
miniscruff: You can technically remove this assert as it is included in the next one.

miniscruff (approved):

tests/unit/storage/data/external/ExternalClients.js
naren-rajendran:
```suggestion
```

tests/unit/storage/data/external/ExternalClients.js
```diff
@@ -154,6 +156,62 @@ describe('external backend clients', () => {
...
+                            assert(err.ServiceUnavailable);
+                            next();
+                        }),
+                        next => testClient.objectDeleteTagging(key, bucket, objectMD, log, (err) => {
+                            assert(err);
```
naren-rajendran:
```suggestion
```

tests/unit/storage/data/external/ExternalClients.js
lavigne958: I am not too sure, here the first `assert` checks that `err` is not `undefined` or `null`. then the second `assert` check that `err.ServiceUnavailable` is not `undefined` or `null`.
If i remove the first `assert` and `err` is `null` then trying to reach a property of a `null` object would fail.

lavigne958:
I'll squash all the commits when build is green and review is OK.

tests/unit/storage/data/external/ExternalClients.js
miniscruff: If you try and access a property that is undefined you get an exception which is a test failure so that is fine.
Basically: `assert.strictEqual(a.b.c.d.e.myProperty, 5)` is effectively testing:

```js
assert.notStrictEqual(a, undefined);
assert.notStrictEqual(a.b, undefined);
assert.notStrictEqual(a.b.c, undefined);
assert.notStrictEqual(a.b.c.d, undefined);
assert.notStrictEqual(a.b.c.d.e, undefined);
assert.strictEqual(a.b.c.d.e.myProperty, 5);
```

Because for myProperty to equal 5 all the parent attributes must also exist. So one assertion here is fine.

lavigne958: Alright, I will update the test and remove the asserts but I maintain that raising an exception due to trying to access the property of an `undefined` object is different that raising an assert because the object is undefined. but this is very very minor differences.

naren-rajendran (approved):

tests/unit/storage/data/external/ExternalClients.js
```diff
@@ -154,6 +156,60 @@ describe('external backend clients', () => {
...
+                };
+                async.series([
+                    next => testClient.objectPutTagging(key.key, bucket, objectMd, log, next),
+                    next => testClient.objectDeleteTagging(key.Key, bucket, objectMd, log, next),
+                ], done);
+            });
+
+            it(`${backend.name} should fail to set tag on missing key`, done => {
+                const key = 'externalBackendMissingKey';
+                const bucket = {
```
nicolas2bert: I would use the `BucketInfo` class here to make it closer to reality. https://github.com/scality/Arsenal/blob/development/8.1/lib/models/BucketInfo.js#L16
Might be using its static `fromObj` method?

lavigne958:
squashed all commits.

tests/unit/storage/data/external/ExternalClients.js
lavigne958: done.

nicolas2bert (approved):

tests/unit/storage/data/DummyService.js
```diff
@@ -133,6 +135,38 @@ class DummyService {
...
+        const keys = Object.keys(tagParams);
+        assert(keys.length > 0);
+        assert(tagParams.Tagging.TagSet.length > 0);
+        tagParams.Tagging.TagSet.forEach(tag => {
+            assert(tag.Key.length > 0);
+            assert(tag.Value.length > 0);
+        });
+
+        if (tagParams.VersionId) {
+            assert.strictEqual(tagParams.VersionId, 'latestversion');
```
nicolas2bert, line 153: Why do you make this assertion here?

lavigne958: to make sure we send the versionId, in case one day we remove it (like I was about to do before you came to me)

PR #1591, opened 2021-10-12
improvement: ARSN-25 update version to 8.1.5

naren-rajendran:
Update version to 8.1.5

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=8"
   },
-  "version": "8.1.3",
+  "version": "8.1.5",
   "description": "Common utilities for the S3 project components",
   "main": "index.js",
   "repository": {


naren-rajendran:
@bert-e reset

miniscruff (approved):
I think we should update the package json and then create a git tag that way if you npm install it has the same version.

naren-rajendran:
> I think we should update the package json and then create a git tag that way if you npm install it has the same version.

Yes, I missed it.

nicolas2bert (approved):

tmacro (approved):

package.json
```diff
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=8"
   },
-  "version": "8.1.3",
+  "version": "8.1.5",
```
jonathan-gramain, line 6: Should it be bumped to 8.1.6, since the tag 8.1.5 already exists?

naren-rajendran: I removed the one created earlier.

PR #1602, opened 2021-10-26
Bugfix/arsn 31 invalid query params

rahulreddy:
-  test: test for invalid type for encoding strings 
-  bugfix: ARSN-31 return empty string for invalid requests

### DIFF
--- a/lib/auth/v4/awsURIencode.js
+++ b/lib/auth/v4/awsURIencode.js
@@ -35,6 +35,13 @@ function _toHexUTF8(char) {
 function awsURIencode(input, encodeSlash, noEncodeStar) {
     const encSlash = encodeSlash === undefined ? true : encodeSlash;
     let encoded = '';
+    /**
+     * Duplicate query params are not suppported by AWS S3 APIs. These params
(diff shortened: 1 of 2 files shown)


lib/auth/v4/awsURIencode.js
```diff
@@ -35,6 +35,13 @@ function _toHexUTF8(char) {
 function awsURIencode(input, encodeSlash, noEncodeStar) {
     const encSlash = encodeSlash === undefined ? true : encodeSlash;
     let encoded = '';
+    /**
+     * Duplicate query params are not suppported by AWS S3 APIs. These params
+     * are parsed as Arrays by Node.js HTTP parser which breaks this method
+    */
+    if (typeof input !== 'string') {
```
miniscruff, line 42: could move this to the top of the function as it just returns an empty string either way.

miniscruff (approved):

rahulreddy:
I ran the test against AWS S3, it ignores query params that it does not recognize

alexanderchan-scality (approved):

PR #1623, opened 2021-11-05
Improvement/arsn 39 support OIDC principals and conditions

ghost:
We want to update Arsenal to support:

- The use of a hostname as a trust policy principal, 
- The use of OIDC Conditions in trust policies.

... as per the Data Services Roles design.

### 1. The Federated OIDC Principal

The [Data Services Roles design](https://github.com/scality/citadel/blob/development/1.0/docs/design/data-services-roles.md) introduces three new trust policies, policies and roles for the StorageManager, StorageAccountOwner and DataConsumer.

In the design, [we define these trust policy/policy/role triplets](https://github.com/scality/citadel/blob/development/1.0/docs/design/data-services-roles.md#fine-grained-policy-and-role-editor), and Trust Policies contain two fields, i.e. `"Principal": {"Federated": "keycloak-http.artesca-auth.svc.cluster.local"},` and ` "StringEquals": {"keycloak:groups": "accountName::StorageAccountOwner"},` that, if we wanted to use them in Vault now, would cause an Arsenal error at startup time:
```js
ValidationError: bad accountSeeds config:  trust policy is invalid: Syntax errors in policy.
```
To illustrate, this error was generated within the [pre-merge](https://eve.devsca.com/github/scality/vault/#/builders/19/builds/1901/steps/10/logs/stdio) of the [Improvement/vault 108 seed dc and sao roles in vault](https://github.com/scality/Vault/pull/1773) PR, because we are waiting for this PR to be merged first.

Investigating the error will reveal that the only Federated Principal type that is accepted by Arsenal prior to this PR is a SAML with the validation regex `^arn:aws:iam::[0-9]{12}:saml-provider/[\w._-]{1,128}`, preventing Vault from defining a Federated Principal type with a hostname as the value, as required by the design.
AWS background for Principals: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html

### 2. The Condition
Second, when we call `getRolesForWebIdentity` or `assumeRoleWithWebIdentity`, we want to evaluate policies to verify that the user is allowed to assume a given role. In this sense, we want to allow the condition `"StringEquals": {"keycloak:groups": "accountName::StorageManager"},` to be evaluated. However, Arsenal does not know this type of condition, since the conditions are fixed and listed in [`lib/policyEvaluator/utils/conditions.js`](https://github.com/scality/Arsenal/blob/development/8.1/lib/policyEvaluator/utils/conditions.js). So this PR adds a new entry to [link the associated condition dynamically from the policy evaluation request](https://github.com/scality/Vault/pull/1776/files#diff-bd63757779a3c9b8567154fdec6d06bc71f596bc96d14de2c7c1b63739fcd3c1R781) when performing `getRolesForWebIdentity` or `assumeRoleWithWebIdentity`.
This PR hence introduces a new mapping to let Arsenal know about this `keycloak:groups` condition.

Commit 1: improvement: ARSN-39 support OIDC principals
I introduce here a new OIDC pattern that validates hostnames to support trust policies with `Federated` web identity user principals.

Commit 2: improvement: ARSN-39 support keycloak groups condition
Here, I map the new condition `keycloak:groups` introduced by the data services roles design.

### DIFF
--- a/lib/policyEvaluator/utils/conditions.js
+++ b/lib/policyEvaluator/utils/conditions.js
@@ -146,6 +146,7 @@ conditions.findConditionKey = (key, requestContext) => {
     map.set('s3:ObjLocationConstraint',
         headers['x-amz-meta-scal-location-constraint']);
     map.set('sts:ExternalId', requestContext.getRequesterExternalId());
+    map.set('keycloak:groups', requesterInfo.keycloakGroup);
     map.set('iam:PolicyArn', requestContext.getPolicyArn());
     // s3:ExistingObjectTag - Used to check that existing object tag has
     // specific tag key and value. Extraction of correct tag key is done in CloudServer.
--- a/lib/policyEvaluator/utils/conditions.js
+++ b/lib/policyEvaluator/utils/conditions.js
@@ -147,6 +147,7 @@ conditions.findConditionKey = (key, requestContext) => {
         headers['x-amz-meta-scal-location-constraint']);
     map.set('sts:ExternalId', requestContext.getRequesterExternalId());
     map.set('keycloak:groups', requesterInfo.keycloakGroup);
+    map.set('keycloak:roles', requesterInfo.keycloakRole);
     map.set('iam:PolicyArn', requestContext.getPolicyArn());
     // s3:ExistingObjectTag - Used to check that existing object tag has
     // specific tag key and value. Extraction of correct tag key is done in CloudServer.
--- a/lib/policy/userPolicySchema.json
+++ b/lib/policy/userPolicySchema.json
@@ -38,6 +38,10 @@
             "type": "string",
             "pattern": "^arn:aws:iam::[0-9]{12}:saml-provider/[\\w._-]{1,128}$"
         },
+        "principalFederatedOidcIdp": {
+            "type": "string",
+            "pattern": "^(?:http(s)?:\/\/)?[\\w.-]+(?:\\.[\\w\\.-]+)+[\\w\\-\\._~:/?#[\\]@!\\$&'\\(\\)\\*\\+,;=.]+$"
+        },
         "principalAWSItem": {
             "type": "object",
             "properties": {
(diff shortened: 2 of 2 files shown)


lib/policyEvaluator/utils/conditions.js
```diff
@@ -146,6 +146,7 @@ conditions.findConditionKey = (key, requestContext) => {
     map.set('s3:ObjLocationConstraint',
         headers['x-amz-meta-scal-location-constraint']);
     map.set('sts:ExternalId', requestContext.getRequesterExternalId());
+    map.set('keycloak:groups', requesterInfo.keycloakGroup);
```
ghost, line 149: I'm not sure I understand what this does exactly?

ghost (commented):
So for example, what is a OIDC Condition?
Why do we need these to fit requirements?
What is a Federated web identity user principal and how/why is it used here?

lib/policyEvaluator/utils/conditions.js
ghost: Short answer: as Arsenal is fixing the conditions keys internatlly, we want to add a new entry to support the `keycloak:groups` condition. This update will let Arsenal map the `keycloakGroup` field of the request context we create when we cant to evaluate a policy with the actual condition name, so, when the policy is being evaluated, the condition will be evaluated properly.

For more details, please refer to the original (updated) PR message.

XinLi0207 (approved):

lib/policyEvaluator/utils/conditions.js
ghost: Added another condition mapping for the keycloak role. Even if we don't use it (depends on https://github.com/scality/citadel/pull/71), it will allow us to have this functionality in the future, if needed.

ghost (approved):

ghost:
ping

lib/policyEvaluator/utils/conditions.js
ghost: Merging the PR as this is a blocker. I will reopen a new PR to remove the keycloak:roles condition if needed later.

PR #1647, opened 2021-12-07
RING-36074 Upgrade Node from 10 to 16

dimitri-bourreau:
This PR aims to upgrade Node from 10 to 16.


dimitri-bourreau:
Oops, wrong ticket number. I'll do another PR for ARSN-21, not RING-36074.

PR #1680, opened 2022-01-20
feature: ARSN-54 revert node 16

miniscruff:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "arsenal",
   "engines": {
-    "node": ">=16"
+    "node": ">=10 <12"
```
miniscruff: @nicolas2bert @naren-rajendran what do you think of limiting the node version until we are safe with 16?

nicolas2bert: Let's revert the `engines` section as it was ("node": ">=6.9.5") until we have a clear defined supported node version (ie >= 16).
We would also need to specify the supported version in the respective project `README.md` .

miniscruff: Are you sure we should remove it entirely? That would effectively mean Arsenal supports any node version.

naren-rajendran: I like the idea of locking the supporting node version, until we fix the issues.

nicolas2bert: I meant to revert the change without changing the engine for now. 
We would update it once we have a clear defined supported node version. But if you feel like adding supported versions, go for it.

miniscruff: After seeing vault work on node 16 I will just use >=10 for now. One moment.

nicolas2bert:
Not reverting the commit would not be an issue?
 
Cloudserver 8.1 still runs on Node 10 (until upgrade)
Cloudserver 8.1 will use the Node 16 compatible Arsenal version (with the close event that does not exist anymore)
So, here https://github.com/scality/Arsenal/pull/1663/files#diff-8d2652b35175eaaa7cee1011b7478e2f406f0ffee0bdd89e16efc99a7c741994L262 if the socket closes unexpectedly the file descriptor won't be closed and the data file won't be removed.


So, we might want to push all the Node 16 changes at the "same" time.
Another solution would be to add the close event back to Arsenal dev/8.1 until Cloudserver runs on Node 16.

miniscruff:
> Not reverting the commit would not be an issue?
> 
> Cloudserver 8.1 still runs on Node 10 (until upgrade) Cloudserver 8.1 will use the Node 16 compatible Arsenal version (with the close event that does not exist anymore) So, here https://github.com/scality/Arsenal/pull/1663/files#diff-8d2652b35175eaaa7cee1011b7478e2f406f0ffee0bdd89e16efc99a7c741994L262 if the socket closes unexpectedly the file descriptor won't be closed and the data file won't be removed.
> 
> So, we might want to push all the Node 16 changes at the "same" time. Another solution would be to add the close event back to Arsenal dev/8.1 until Cloudserver runs on Node 16.

I reverted the close event on the 8.1 integration branch. But I did not revert one of the other fixes as it was just a general fix. See this one: https://github.com/scality/Arsenal/pull/1663/files#diff-9d1eef5329ffe7aa6427ae1c4fe53213acb34b454ecbe49477c7af9252e2165dR44

I think the plan would be to create a new branch now in Arsenal with the node 16 stuff, and continue making these branches until we can organize a merge party ( :balloon: :tada: )

nicolas2bert:
Haha! Sounds great! I like it!!!

nicolas2bert (approved):

naren-rajendran (approved):

PR #1691, opened 2022-01-28
bf: ARSN-57 log correct client ip

naren-rajendran:
Log correct client ip.

### DIFF
--- a/lib/s3routes/routes.js
+++ b/lib/s3routes/routes.js
@@ -69,6 +69,7 @@ function checkBucketAndKey(bucketName, objectKey, method, reqQuery,
     return undefined;
 }
 
+// TODO: ARSN-59 remove assertions or restrict it to dev environment only.
 function checkTypes(req, res, params, logger, s3config) {
     assert.strictEqual(typeof req, 'object',
         'bad routes param: req must be an object');
--- a/lib/s3routes/routes.js
+++ b/lib/s3routes/routes.js
@@ -10,6 +10,8 @@ const routeOPTIONS = require('./routes/routeOPTIONS');
 const routesUtils = require('./routesUtils');
 const routeWebsite = require('./routes/routeWebsite');
 
+const requestUtils = require('../../lib/policyEvaluator/requestUtils');
+
 const routeMap = {
     GET: routeGET,
     PUT: routePUT,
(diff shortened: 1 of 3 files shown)


lib/s3routes/routes.js
```diff
@@ -134,9 +136,10 @@ function checkTypes(req, res, params, logger) {
  *  values for whether queries are supported
  * @param {function} params.dataRetrievalFn - function to retrieve data
  * @param {RequestLogger} logger - werelogs logger instance
+ * @param {String} [s3config] - s3 configuration
```
miniscruff, line 143: Is this actually optional? It looks required to me.

lib/s3routes/routes.js
```diff
@@ -134,9 +136,10 @@ function checkTypes(req, res, params, logger) {
...
  * @returns {undefined}
  */
-function routes(req, res, params, logger) {
+function routes(req, res, params, logger, s3config) {
     checkTypes(req, res, params, logger);
```
miniscruff, line 147: Should you add the new value to the check types method?

miniscruff (changes requested):
Should there be a test as well?

lib/s3routes/routes.js
naren-rajendran: Yes optional. Not sure if this being used elsewhere.

lib/s3routes/routes.js
naren-rajendran: It is optional, I will add type check if it is required.

naren-rajendran:
> Should there be a test as well?

Unit tests are already available for the function being used. I can add functional test only when arsenal is merged and the updated version is used in cloudserver.

lib/s3routes/routes.js
naren-rajendran: I added a check now.

naren-rajendran:
@bert-e reset

nicolas2bert (approved):

miniscruff (approved):

miniscruff (approved):

jonathan-gramain (approved):
LGTM, but I would have expected this fix to be targeting 7.4, since it's still our current LTS and in active maintenance.

alexanderchan-scality (approved):

naren-rajendran:
Opening a new PR with development/7.4 as base.

PR #1700, opened 2022-02-02
feature: ARSN-58-add-MD-actionMap

XinLi0207:
1. add  `metadataSearch: 's3:MetadataSearch'` actionMap( action name is different from what defined in the [design](https://github.com/scality/citadel/blob/development/1.0/docs/design/metadata-search.md), which is `mdsearch`. Through observation, I found it's not common to have abbreviation in the action name, so wanted to change to `metadataSearch` to make it more accord with naming convention )
2. add this logic in routeGet.js: if query.search is not null, call metadataSearch
3. bump version


lib/policyEvaluator/utils/actionMaps.js
```diff
@@ -30,6 +30,7 @@ const sharedActionMap = {
     bypassGovernanceRetention: 's3:BypassGovernanceRetention',
     listMultipartUploads: 's3:ListBucketMultipartUploads',
     listParts: 's3:ListMultipartUploadParts',
+    metadataSearch: 's3:MetadataSearch',
```
ghost, line 33: In the design we have `s3:mdsearch`, is it expected to use this format now?

ghost (commented):
Quick question before approving

lib/policyEvaluator/utils/actionMaps.js
XinLi0207: wanted to ask your suggestion on this, feel like in action names it's not common to have abbreviation, `s3: MDSearch` seems a little bit weird here
@anurag4DSB  didn't think it through when we decided on this name, can we change the action name defined in the design?

ghost: IMO as long as all implementations are compliant with this name and the design is updated in a quick PR, that's ok

KillianG (approved):

ghost (approved):

PR #1710, opened 2022-02-08
ARSN-62: include session token in v4 signature calculation

rachedbenmustapha:
Adds STS session token to requests and generated AWSV4 signature when required. This will help vaultclient perform admin actions against Vault using assumed role credentials.

### DIFF
--- a/lib/auth/auth.js
+++ b/lib/auth/auth.js
@@ -154,10 +154,12 @@ function doAuth(request, log, cb, awsService, requestContexts) {
  * @param {string} secretKeyValue - the secretKey
  * @param {string} awsService - Aws service related
  * @param {sting} [proxyPath] - path that gets proxied by reverse proxy
+ * @param {string} [sessionToken] - security token if the access/secret keys
+ *                                are temporary credentials from STS
  * @return {undefined}
  */
 function generateV4Headers(request, data, accessKey, secretKeyValue,
-                           awsService, proxyPath) {
+                           awsService, proxyPath, sessionToken) {
     Object.assign(request, { headers: {} });
     const amzDate = convertUTCtoISO8601(Date.now());
     // get date without time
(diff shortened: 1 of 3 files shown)


tests/unit/auth/v4/generateV4Headers.js
```diff
@@ -0,0 +1,57 @@
...
+        const req = http.request({ headers });
+
+        generateV4Headers(req, data, 'accessKey', 'secretKey', 'iam', null, token);
+
+        try {
+            assert.deepStrictEqual(req.getHeader('x-amz-security-token'), token);
+            return done();
+        } catch (err) {
+            return done(err);
+        }
```
miniscruff:
```suggestion
        assert.doesNotThrow(
            assert.deepStrictEqual(req.getHeader('x-amz-security-token'), token)
        );
```

Could use does not throw in place of a try catch, removes the need for done as well.

lib/auth/auth.js
```diff
@@ -154,10 +154,12 @@ function doAuth(request, log, cb, awsService, requestContexts) {
  * @param {string} secretKeyValue - the secretKey
  * @param {string} awsService - Aws service related
  * @param {sting} [proxyPath] - path that gets proxied by reverse proxy
+ * @param {string} sessionToken - security token if the access/secret keys
```
miniscruff: This looks optional
```suggestion
 * @param {string} [sessionToken] - security token if the access/secret keys
```

naren-rajendran (approved):

tests/unit/auth/v4/generateV4Headers.js
rachedbenmustapha: I started without `done` but the test suite was hanging, not sure why as it should work, but settled on this.

lib/auth/auth.js
rachedbenmustapha: updated, thanks!

tests/unit/auth/v4/generateV4Headers.js
rachedbenmustapha: @miniscruff actually it just came back to me, `assert.doesNotThrow` will throw an `AssertionError` if the code to assert on throws an exception (be it `AssertionError` or anything else) so it would be back to square one.

miniscruff (approved):

PR #1714, opened 2022-02-11
ft: ARSN-64 sorted set routines

vrancurel:
Large sorted sets management routines implemented with array dichotomies.
No particular suitable external module was found.

An Arsenal tag is not needed at this stage.

### DIFF
--- a/lib/algos/set/ArrayUtils.js
+++ b/lib/algos/set/ArrayUtils.js
@@ -1,0 +1,87 @@
+function indexOf(arr, value) {
+    if (!arr.length) {
+        return -1;
+    }
+    let lo = 0;
+    let hi = arr.length - 1;
+
+    while (hi - lo > 1) {
+        const i = lo + ((hi - lo) >> 1);
+        if (arr[i] > value) {
+            hi = i;
+        } else {
+            lo = i;
+        }
+    }
+    if (arr[lo] === value) {
+        return lo;
+    }
+    if (arr[hi] === value) {
+        return hi;
+    }
+    return -1;
+}
+
+function indexAtOrBelow(arr, value) {
+    let i;
+    let lo;
+    let hi;
+
+    if (!arr.length || arr[0] > value) {
+        return -1;
+    }
+    if (arr[arr.length - 1] <= value) {
+        return arr.length - 1;
+    }
+
+    lo = 0;
+    hi = arr.length - 1;
+
+    while (hi - lo > 1) {
+        i = lo + ((hi - lo) >> 1);
+        if (arr[i] > value) {
+            hi = i;
+        } else {
+            lo = i;
+        }
+    }
+
+    return lo;
+}
+
+/*
+ * perform symmetric diff in O(m + n)
+ */
+function symDiff(k1, k2, v1, v2, cb) {
+    let i = 0;
+    let j = 0;
+    const n = k1.length;
+    const m = k2.length;
+
+    while (i < n && j < m) {
+        if (k1[i] < k2[j]) {
+            cb(v1[i]);
+            i++;
+        } else if (k2[j] < k1[i]) {
+            cb(v2[j]);
+            j++;
+        } else {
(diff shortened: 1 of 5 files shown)


tests/unit/set/SortedSet.js
```diff
@@ -0,0 +1,16 @@
+const assert = require('assert');
+const SortedSet = require('../../../lib/algos/set/SortedSet');
+
+describe('SortedSet', () => {
+    it('basic', () => {
+        const set = new SortedSet();
+        set.set('foo', 'bar');
+        assert(set.isSet('foo'));
+        assert(!set.isSet('foo2'));
+        assert(set.get('foo') === 'bar');
```
miniscruff: assert.strictEquals

miniscruff (approved):

tests/unit/algos/set/ArrayUtils.js
```diff
@@ -0,0 +1,12 @@
+const assert = require('assert');
+const symDiff = require('../../../../lib/algos/set/ArrayUtils').symDiff;
+
+describe('ArrayUtils', () => {
```
jonathan-gramain, line 5:
```suggestion
describe('ArrayUtils.symDiff', () => {
```

lib/algos/set/ArrayUtils.js
```diff
@@ -0,0 +1,87 @@
+function indexOf(arr, value) {
```
jonathan-gramain, line 1: might be worth unit testing this function thoroughly (correctness, and in case it has corner cases causing infinite loops etc.) if not completely correct it could be a source of nasty bugs later on

lib/algos/set/ArrayUtils.js
```diff
@@ -0,0 +1,87 @@
...
+    if (arr[lo] === value) {
+        return lo;
+    }
+    if (arr[hi] === value) {
+        return hi;
+    }
+    return -1;
+}
+
+function indexAtOrBelow(arr, value) {
```
jonathan-gramain, line 25: same here

tests/unit/algos/set/ArrayUtils.js
jonathan-gramain: please ignore this if you're going to add more tests for the other helpers (but then add their name in the test cases)

lib/algos/set/ArrayUtils.js
vrancurel: done

tests/unit/algos/set/ArrayUtils.js
```diff
@@ -0,0 +1,55 @@
...
+        }
+        arr.sort();
+        const refMap2 = new Map([...refMap.entries()].sort());
+        let i = 0;
+        for (const key of refMap2.keys()) {
+            assert.strictEqual(indexAtOrBelow(arr, key), i++);
+        }
+    });
+
+    it('shall find symmetric difference', () => {
```
jonathan-gramain, line 82:
```suggestion
    it('symDiff shall find symmetric difference', () => {
```

tests/unit/algos/set/ArrayUtils.js
```diff
@@ -0,0 +1,55 @@
+const assert = require('assert');
+const { indexOf, indexAtOrBelow, symDiff } = require('../../../../lib/algos/set/ArrayUtils');
+const crypto = require('crypto');
+
+describe('ArrayUtils', () => {
+    it('indexOf', () => {
+        const numOps = 10000;
```
jonathan-gramain, line 24: This test is definitely useful, I'd also add a 0, 1 and 2-item array test for both `indexOf` and `indexAtOrBelow`

jonathan-gramain (commented):
I would expect the target branch to be `development/7.10` since the RING epic is targeted at 8.5.3.

Other than that + a few tests suggestion, good for me

jonathan-gramain (approved):

PR #1723, opened 2022-02-18
ft: ARSN-84 Switch to Jest

ghivert:
Switch the test runner to Jest, and upgrade ESLint.
All files are committed in order to comply with the new ESLint.
The test runner is now Jest and the code coverage is Jest + Istanbul.
All new test files should be written as `[filename].spec.js` or `[filename].spec.ts`.

PR #1729, opened 2022-02-23
Improvement: ARSN-88 add MongoClientInterface tests

welansari:


jbertran:
Come on Bert-E work your magic :point_up:

tests/functional/metadata/mongodb/withCond.js
```diff
@@ -28,20 +28,25 @@ const mongoserver = new MongoMemoryReplSet({
...
-                    writeConcern: 'majority',
-                    replicaSet: 'rs0',
-                    readPreference: 'primary',
-                    database: dbName,
-                },
-            };
-            metadata = new MetadataWrapper(implName, opts, null, logger);
-            metadata.setup(done);
-        });
+        async.series([
```
jbertran: Could we just have a sequence of `await` operations?

tests/functional/metadata/mongodb/putObject.spec.js
```diff
@@ -0,0 +1,442 @@
...
+    return `${DbPrefixes.Master}${key}`;
+}
+
+function formatVersionKeyV0(key, versionId) {
+    return `${key}${VID_SEP}${versionId}`;
+}
+
+function formatVersionKeyV1(key, versionId) {
+    return `${DbPrefixes.Version}${formatVersionKeyV0(key, versionId)}`;
+}
```
jbertran: I was surprised these are not available in source :shrug: maybe there's some factorization we can do after this PR

tests/functional/metadata/mongodb/withCond.spec.js
```diff
@@ -16,7 +16,7 @@ const bucketName = 'testbucket';
 const mongoserver = new MongoMemoryReplSet({
     debug: false,
     instanceOpts: [
-        { port: 27018 },
+        { port: 27022 },
```
jbertran, line 19: What's the issue with having the same port everywhere? You could put a little bit more details on the rationale behind this change in the commit message :slightly_smiling_face:

jbertran (approved):
Nice work!

For the style-related changes: a `yarn lint --fix` should take care of everything.

tests/functional/metadata/mongodb/withCond.spec.js
welansari: CI fails when i use the same port everywhere (with the error can't connect to mongodb, connection refused), and when i checked the logs i noticed that test outputs overlap each other which left me thinking that maybe the tests are ran in parallel, not sure it is the case though

tests/functional/metadata/mongodb/putObject.spec.js
welansari: yep i agree

tests/functional/metadata/mongodb/withCond.spec.js
jbertran: Could try initializing `mongoserver` inside the `beforeAll`. If that doesn't work, we can merge as is - this is a pattern that's present in a lot of places in this code IIRC so it can be taken care of by a refactor pass at some point.

tests/functional/metadata/mongodb/putObject.spec.js
```diff
@@ -0,0 +1,442 @@
...
+            versionId: 'null',
+            updated: false,
+        };
+        const params = {
+            versioning: null,
+            versionId: null,
+            repairMaster: null,
+        };
+        async.waterfall([
+            next => metadata.putObjectMD(bucketName, objName, objVal, params, logger, err => {
```
francoisferrand: Could we not just use 'next' as last argument?

```suggestion
            next => metadata.putObjectMD(bucketName, objName, objVal, params, logger, next)
```

tests/functional/metadata/mongodb/putObject.spec.js
```diff
@@ -0,0 +1,442 @@
...
+                    if (err) {
+                        return next(err);
+                    }
+                    return next();
+                });
+            },
+            // object metadata must be updated
+            next => getObject(objName, null, BUCKET_VFORMAT_V0, (err, object) => {
+                assert.deepStrictEqual(err, null);
+                assert.strictEqual(object.key, OBJECT_NAME);
```
francoisferrand, line 180: we should eiher check against `objName`, or remove that variable and use `OBJECT_NAME` directly everywhere.

```suggestion
                assert.strictEqual(object.key, objName);
```

tests/functional/metadata/mongodb/putObject.spec.js
```diff
@@ -0,0 +1,442 @@
...
+        async.waterfall([
+            next => metadata.putObjectMD(bucketName, objName, objVal, params, logger, err => {
+                if (err) {
+                    return next(err);
+                }
+                return next();
+            }),
+            next => {
+                objVal.updated = true;
+                metadata.putObjectMD(bucketName, objName, objVal, params, logger, err => {
```
francoisferrand: same here

tests/functional/metadata/mongodb/putObject.spec.js
```diff
@@ -0,0 +1,442 @@
...
+    let metadata;
+    let collection;
+
+    function getObject(objName, versionId, vFormat, cb) {
+        let key = null;
+        if (versionId) {
+            key = formatVersionKey(objName, versionId, vFormat);
+        } else {
+            key = formatMasterKey(objName, vFormat);
+        }
```
francoisferrand: Not sure about this (and all the associated functions) : for _tests_, we don't care about hard-coding values, and the goal is to make the test easy to read (and thus identify an issue, if it fails).

In my experience, it is thus generally better to avoid writing "generic" code to generate some values, and instead just clearly and explicitly hard-code the "target" value in the test (with no need to scroll to try to understand what is happening)

e.g. in this case build "manually" the key in each test case, something like this:

```javascript
it('Should put a new non versionned object', done => {
   async.waterfall([
            next => metadata.putObjectMD(bucketName, "foo", objVal, params, logger, next),
                                                   // ^^^ hard-coded name here
            next => getObject("v1-M-foo", err => { .... })
                            // ^^^^^^^^ and resulting key here
```

tests/functional/metadata/mongodb/putObject.spec.js
```diff
@@ -0,0 +1,442 @@
...
+        let versionId = null;
+        async.waterfall([
+            // We first create a new version and master
+            next => metadata.putObjectMD(bucketName, objName, objVal, params, logger, err => {
+                if (err) {
+                    return next(err);
+                }
+                return next();
+            }),
+            next => getObject(objName, null, BUCKET_VFORMAT_V0, (err, object) => {
```
francoisferrand: this has already been tested in previus test : no need to complexify this test to re-check it.

tests/functional/metadata/mongodb/putObject.spec.js
```diff
@@ -0,0 +1,442 @@
...
+            }),
+            // The second put shouldn't create a new version
+            next => getObjectCount((err, count) => {
+                assert.deepStrictEqual(err, null);
+                assert.strictEqual(count, 2);
+                return next();
+            }),
+        ], done);
+    });
+});
```
francoisferrand: missing new line at EOF

tests/functional/metadata/mongodb/putObject.spec.js
```diff
@@ -0,0 +1,442 @@
+const async = require('async');
```
francoisferrand, line 1: we have no such convention of having `.spec` extension for test files : so we should stick with the current approach

tests/functional/metadata/mongodb/listObject.spec.js
```diff
@@ -0,0 +1,306 @@
...
+                const params = {
+                    objName: 'pfx1-test-object',
+                    objVal: {
+                        key: 'pfx1-test-object',
+                        versionId: 'null',
+                    },
+                    nbVersions: 5,
+                };
+                putBulkObjectVersions(BUCKET_NAME, params.objName, params.objVal, versionParams,
+                params.nbVersions, next);
```
francoisferrand: followup line should be indented?

tests/functional/metadata/mongodb/listObject.spec.js
```diff
@@ -0,0 +1,306 @@
...
+                const params = {
+                    objName: 'pfx2-test-object',
+                    objVal: {
+                        key: 'pfx2-test-object',
+                        versionId: 'null',
+                    },
+                    nbVersions: 5,
+                };
+                putBulkObjectVersions(BUCKET_NAME, params.objName, params.objVal, versionParams,
+                params.nbVersions, next);
```
francoisferrand: same

tests/functional/metadata/mongodb/listObject.spec.js
```diff
@@ -0,0 +1,306 @@
...
+                const params = {
+                    objName: 'pfx3-test-object',
+                    objVal: {
+                        key: 'pfx3-test-object',
+                        versionId: 'null',
+                    },
+                    nbVersions: 5,
+                };
+                putBulkObjectVersions(BUCKET_NAME, params.objName, params.objVal, versionParams,
+                params.nbVersions, next);
```
francoisferrand: same

tests/functional/metadata/mongodb/listObject.spec.js
francoisferrand: i don't see the point of having this structure: just inline the member to keep the code simpler & cleaner

```suggestion
                const objVal = {
                    key: 'pfx1-test-object',
                    versionId: 'null',
                };
                putBulkObjectVersions(BUCKET_NAME, objVal.key, objVal, versionParams, 5, next);
```

tests/functional/metadata/mongodb/listObject.spec.js
```diff
@@ -0,0 +1,306 @@
...
+    it('Should list master versions of objects', done => {
+        const bucketName = BUCKET_NAME;
+        const params = {
+            listingType: 'DelimiterMaster',
+            maxKeys: 100,
+        };
+
+        return metadata.listObject(bucketName, params, logger, (err, data) => {
+            assert.deepStrictEqual(err, null);
+            assert.strictEqual(data.Contents.length, 3);
```
francoisferrand, line 145: should also check the keys of the returned items?
(for all these listing tests)

tests/functional/metadata/mongodb/listObject.spec.js
```diff
@@ -0,0 +1,306 @@
...
+        }, logger, (err, res) => {
+            if (err) {
+                return cb(err);
+            }
+            res.Versions.forEach(version => {
+                if (versionsPerKey[version.key]) {
+                    versionsPerKey[version.key] += 1;
+                } else {
+                    versionsPerKey[version.key] = 1;
+                }
```
francoisferrand: Could be simplified:

```suggestion
                versionsPerKey[version.key] = (versionsPerKey[version.key] || 0) + 1
```

tests/functional/metadata/mongodb/getObject.spec.js
```diff
@@ -0,0 +1,263 @@
...
+                    },
+                },
+                        ],
+            },
+            {
+                $set: { _id: objName, value: objVal },
+            },
+            { upsert: true },
+            err => {
+                if (err) {
```
francoisferrand, line 71: this is copy-paste of code we don't need (not creating 2 keys at the same time?), and most likely don't want (this is  test, not prod code) : keep things simple.

tests/functional/metadata/mongodb/getObject.spec.js
```diff
@@ -0,0 +1,263 @@
...
+        objName: 'pfx1-test-object',
+        objVal: {
+            key: 'pfx1-test-object',
+            versionId: 'null',
+        },
+        vFormat: 'v0',
+    };
+
+    function updateMasterObject(objName, versionId, objVal, vFormat, cb) {
+        const mKey = formatMasterKey(objName, vFormat);
```
francoisferrand: ideally, this should not be computed but a parameter, hard-coded in each test : to make each test easier to read.

tests/functional/metadata/mongodb/delObject.spec.js
```diff
@@ -0,0 +1,365 @@
...
+                versionParams, logger, err => {
+                    if (err) {
+                        return next(err);
+                    }
+                    return next();
+                });
+            },
+            next => {
+                // we put the master version of a second object
+                params.objName = 'object-to-deleted';
```
francoisferrand: typo, should be `object-to-delete`

```suggestion
                params.objName = 'object-to-delete';
```

tests/functional/metadata/mongodb/delObject.spec.js
```diff
@@ -0,0 +1,365 @@
...
+                params.objVal.key = 'object-to-deleted';
+                metadata.putObjectMD(BUCKET_NAME, params.objName, params.objVal,
+                versionParams, logger, err => {
```
francoisferrand: missing indent

tests/functional/metadata/mongodb/putObject.spec.js
welansari: putObjectMD uses a different internal function depending on the params we give it, so here we are testing a different one from the ones tested in before

tests/functional/metadata/mongodb/delObject.spec.js
```diff
@@ -0,0 +1,365 @@
...
+                    return next();
+                });
+            },
+            next => {
+                // we delete the first version
+                metadata.deleteObjectMD(bucketName, params.objName, { versionId: versionId1 },
+                logger, err => {
+                    assert.deepStrictEqual(err, null);
+                    return next();
+                });
```
francoisferrand: just pass `next` here, and move `err` assert to the end of waterfall (where `done` is passed)... and actually mocha already handles this, so if an error is passed to `done` then the test will fail... So this is very simple actually:

```suggestion
                    logger, next);
```

tests/functional/metadata/mongodb/putObject.spec.js
welansari: this was introduced here https://github.com/scality/Arsenal/pull/1723

francoisferrand: fair enough :-)

tests/functional/metadata/mongodb/putObject.spec.js
francoisferrand: these parameters, at line 342-347 :
```
        const objVal = {
            key: OBJECT_NAME,
            versionId: VERSION_ID,
            updated: false,
        };
        const params = {
            versioning: true,
            versionId: null,
            repairMaster: null,
        };
```
are the same as previous test, with parameters are line 290-299 :
```
        const objVal = {
            key: OBJECT_NAME,
            versionId: VERSION_ID,
            updated: false,
        };
        const params = {
            versioning: true,
            versionId: null,
            repairMaster: null,
        };
```

Though i see now that the previous test also has the same pattern: putObject, check, putObject, checks (with the same initial states as this test). So:
* if this initial check is needed, maybe it should be moved to a dedicated test?
* or maybe this intermediate check is not so relevant, and we can just run the "whole" scenario before verifying (putObject, putObject, checks)

tests/functional/metadata/mongodb/putObject.spec.js
welansari: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/putObject.spec.js
welansari: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/putObject.spec.js
welansari: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/putObject.spec.js
welansari: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/putObject.spec.js
welansari: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/listObject.spec.js
welansari: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/listObject.spec.js
Kerkesni: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/listObject.spec.js
welansari: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/listObject.spec.js
welansari: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/listObject.spec.js
welansari: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/listObject.spec.js
Kerkesni: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/getObject.spec.js
welansari: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/getObject.spec.js
welansari: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/delObject.spec.js
welansari: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/delObject.spec.js
Kerkesni: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/delObject.spec.js
welansari: fixed in https://github.com/scality/Arsenal/pull/1729/commits/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/putObject.spec.js
welansari: fixed in https://github.com/scality/Arsenal/pull/1729/commits/c8a1d93711d6509576c6286052b6c22dfa29f324

tests/functional/metadata/mongodb/listObject.spec.js
welansari: fixed in https://github.com/scality/Arsenal/commit/a6f48b9f342344e6df5ca2e3b7a2d03d8737fbba

tests/functional/metadata/mongodb/getObject.spec.js
```diff
@@ -86,20 +76,6 @@ describe('MongoClientInterface::metadata.getObjectMD', () => {
...
-                    * We check here than than the MongoDB error is related to the
-                    * second operation, the master version update and than the error
-                    * code is the one related to mentionned issue.
-                    */
-                    if (err.code === 11000) {
-                        return cb(null);
-                    }
                     return cb(err);
                 }
                 return cb(null);
```
francoisferrand: this is just `cb` now, right ?

tests/functional/metadata/mongodb/listObject.spec.js
```diff
@@ -105,47 +105,30 @@ describe('MongoClientInterface::metadata.listObject', () => {
...
-                        versionId: 'null',
-                    },
-                    nbVersions: 5,
+                const objVal = {
+                    key: 'pfx1-test-object',
+                    versionId: 'null',
                 };
-                putBulkObjectVersions(BUCKET_NAME, params.objName, params.objVal, versionParams,
-                    params.nbVersions, next);
+                putBulkObjectVersions(BUCKET_NAME, objVal.key, objVal, versionParams,
```
francoisferrand: line break not needed anymore, looks short enough now.

tests/functional/metadata/mongodb/getObject.spec.js
welansari: yes it is

francoisferrand (approved):

PR #1734, opened 2022-02-25
Feature: arsn 76 support new bucket format

welansari:
- Add new `vFormat` attribut to new buckets
- Update object CRUD + List functions to support new bucket format
- Update tests to test both old and new bucket format

### DIFF
--- a/lib/algos/list/skip.js
+++ b/lib/algos/list/skip.js
@@ -59,8 +59,15 @@ class Skip {
         } else if (filteringResult === FILTER_SKIP
                    && skippingRange !== SKIP_NONE) {
             if (++this.streakLength >= MAX_STREAK_LENGTH) {
-                const newRange = this._inc(skippingRange);
-
+                let newRange;
+                if (Array.isArray(skippingRange)) {
+                    newRange = [];
+                    for (let i = 0; i < skippingRange.length; ++i) {
+                        newRange.push(this._inc(skippingRange[i]));
+                    }
+                } else {
+                    newRange = this._inc(skippingRange);
+                }
                 /* Avoid to loop on the same range again and again. */
                 if (newRange === this.gteParams) {
                     this.streakLength = 1;
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -148,4 +148,7 @@ module.exports = {
         'noncurrentVersionExpiration',
         'abortIncompleteMultipartUpload',
     ],
+    // Maximum number of buckets to cache (bucket metadata)
+    maxCachedBuckets: process.env.METADATA_MAX_CACHED_BUCKETS ?
+        Number(process.env.METADATA_MAX_CACHED_BUCKETS) : 1000,
 };
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -856,37 +856,57 @@ class MongoClientInterface {
         });
     }
 
+    /**
+     * gets versioned and non versioned object metadata
+     * @param {String} bucketName bucket name
+     * @param {String} objName object name
+     * @param {object} params params
+     * @param {String} params.versionId version of object (optional)
+     * @param {Object} log logger
+     * @param {Function} cb callback
+     * @return {undefined}
+     */
     getObject(bucketName, objName, params, log, cb) {
         const c = this.getCollection(bucketName);
-        if (params && params.versionId) {
-            // eslint-disable-next-line
-            objName = formatVersionKey(objName, params.versionId);
-        }
-        c.findOne({
-            _id: objName,
-        }, {}, (err, doc) => {
-            if (err) {
-                log.error('findOne: error getting object',
-                    { error: err.message });
-                return cb(errors.InternalError);
-            }
-            if (!doc) {
-                return cb(errors.NoSuchKey);
-            }
-            if (doc.value.isPHD) {
-                this.getLatestVersion(c, objName, log, (err, value) => {
+        let key;
+        async.waterfall([
+            next => this.getBucketVFormat(bucketName, log, next),
+            (vFormat, next) => {
+                if (params && params.versionId) {
+                    key = formatVersionKey(objName, params.versionId, vFormat);
+                } else {
+                    key = formatMasterKey(objName, vFormat);
+                }
+                c.findOne({
+                    _id: key,
+                }, {}, (err, doc) => {
                     if (err) {
-                        log.error('getLatestVersion: getting latest version',
-                            { error: err.message });
-                        return cb(err);
+                        log.error('findOne: error getting object',
+                            { bucket: bucketName, object: objName, error: err.message });
+                        return next(errors.InternalError);
                     }
-                    return cb(null, value);
+                    return next(null, vFormat, doc);
                 });
-                return undefined;
-            }
-            MongoUtils.unserialize(doc.value);
-            return cb(null, doc.value);
-        });
+            },
+            (vFormat, doc, next) => {
+                if (!doc) {
+                    return next(errors.NoSuchKey);
+                }
+                if (doc.value.isPHD) {
+                    this.getLatestVersion(c, objName, vFormat, log, (err, value) => {
+                        if (err) {
+                            log.error('getLatestVersion: getting latest version',
+                                { bucket: bucketName, object: objName, error: err.message });
+                            return next(err);
+                        }
+                        return next(null, value);
+                    });
+                    return undefined;
+                }
+                MongoUtils.unserialize(doc.value);
+                return next(null, doc.value);
+            },
+        ], cb);
     }
 
     /**
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -2015,42 +2015,48 @@ class MongoClientInterface {
     deleteObjectWithCond(bucketName, objName, params, log, cb) {
         const c = this.getCollection(bucketName);
         const method = 'deleteObjectWithCond';
-        const filter = { _id: objName };
-        try {
-            MongoUtils.translateConditions(0, 'value', filter,
-                params.conditions);
-        } catch (err) {
-            log.error('error creating mongodb filter', {
-                error: reshapeExceptionError(err),
-            });
-            return cb(errors.InternalError);
-        }
-        return c.findOneAndDelete(filter, (err, res) => {
+        this.getBucketVFormat(bucketName, log, (err, vFormat) => {
             if (err) {
-                log.error('error occurred when attempting to delete object', {
-                    method,
-                    error: err.message,
+                return cb(err);
+            }
+            const masterKey = formatMasterKey(objName, vFormat);
+            const filter = { _id: masterKey };
+            try {
+                MongoUtils.translateConditions(0, 'value', filter,
+                    params.conditions);
+            } catch (err) {
+                log.error('error creating mongodb filter', {
+                    error: reshapeExceptionError(err),
                 });
                 return cb(errors.InternalError);
             }
-            if (res.ok !== 1) {
-                log.error('failed to delete object', {
-                    method,
-                    error: err.message,
-                });
-                return cb(errors.InternalError);
-            }
-            /*
-             * unable to find an object that matches the conditions
-             */
-            if (!res.value) {
-                log.debug('unable to find target object to delete', {
-                    method,
-                    filter,
-                });
-                return cb(errors.NoSuchKey);
-            }
-            return cb();
+            return c.findOneAndDelete(filter, (err, res) => {
+                if (err) {
+                    log.error('error occurred when attempting to delete object', {
+                        method,
+                        error: err.message,
+                    });
+                    return cb(errors.InternalError);
+                }
+                if (res.ok !== 1) {
+                    log.error('failed to delete object', {
+                        method,
+                        error: err.message,
+                    });
+                    return cb(errors.InternalError);
+                }
+                /*
+                * unable to find an object that matches the conditions
+                */
+                if (!res.value) {
+                    log.debug('unable to find target object to delete', {
+                        method,
+                        filter,
+                    });
+                    return cb(errors.NoSuchKey);
+                }
+                return cb();
+            });
         });
     }
 
@@ -2061,50 +2067,56 @@ class MongoClientInterface {
     putObjectWithCond(bucketName, objName, objVal, params, log, cb) {
         const c = this.getCollection(bucketName);
         const method = 'putObjectWithCond';
-        const filter = { _id: objName };
-        try {
-            MongoUtils.translateConditions(0, 'value', filter,
-                params.conditions);
-        } catch (err) {
-            log.error('error creating mongodb filter', {
-                error: reshapeExceptionError(err),
-            });
-            return cb(errors.InternalError);
-        }
-        return c.findOneAndUpdate(filter, {
-            $set: {
-                _id: objName,
-                value: objVal,
-            },
-        }, {
-            upsert: true,
-        }, (err, res) => {
+        this.getBucketVFormat(bucketName, log, (err, vFormat) => {
             if (err) {
-                log.error('error occurred when attempting to update object', {
-                    method,
-                    error: err,
+                return cb(err);
+            }
+            const masterKey = formatMasterKey(objName, vFormat);
+            const filter = { _id: masterKey };
+            try {
+                MongoUtils.translateConditions(0, 'value', filter,
+                    params.conditions);
+            } catch (err) {
+                log.error('error creating mongodb filter', {
+                    error: reshapeExceptionError(err),
                 });
                 return cb(errors.InternalError);
             }
-            if (res.ok !== 1) {
-                log.error('failed to update object', {
-                    method,
-                    error: err.message,
-                });
-                return cb(errors.InternalError);
-            }
-            if (!res.value) {
-                log.debug('object not found...upserted object', {
+            return c.findOneAndUpdate(filter, {
+                $set: {
+                    _id: masterKey,
+                    value: objVal,
+                },
+            }, {
+                upsert: true,
+            }, (err, res) => {
+                if (err) {
+                    log.error('error occurred when attempting to update object', {
+                        method,
+                        error: err,
+                    });
+                    return cb(errors.InternalError);
+                }
+                if (res.ok !== 1) {
+                    log.error('failed to update object', {
+                        method,
+                        error: err.message,
+                    });
+                    return cb(errors.InternalError);
+                }
+                if (!res.value) {
+                    log.debug('object not found...upserted object', {
+                        method,
+                        filter,
+                    });
+                    return cb();
+                }
+                log.debug('Object found...updated object', {
                     method,
                     filter,
                 });
                 return cb();
-            }
-            log.debug('Object found...updated object', {
-                method,
-                filter,
             });
-            return cb();
         });
     }
 }
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -28,6 +28,7 @@ const LRUCache = require('../../../algos/cache/LRUCache');
 const MongoReadStream = require('./readStream');
 const MongoUtils = require('./utils');
 const Skip = require('../../../algos/list/skip');
+const MergeStream = require('../../../algos/stream/MergeStream');
 
 const { formatMasterKey, formatVersionKey } = require('./utils');
 
@@ -1270,12 +1271,36 @@ class MongoClientInterface {
         });
     }
 
+    /**
+     * internal listing function
+     * @param {String} bucketName bucket name
+     * @param {Object} params internal listing params
+     * @param {Object} params.mainStreamParams internal listing param applied
+     * to the main listing stream (master stream when we have the two streams)
+     * @param {Object} params.secondaryStreamParams internal listing param applied
+     * to the secondary stream (versionStream when having two streams) (is optional)
+     * @param {Object} params.mongifiedSearch search options
+     * @param {Object} extension listing extention
+     * @param {Object} log logger
+     * @param {Function} cb callback
+     * @return {undefined}
+     */
     internalListObject(bucketName, params, extension, log, cb) {
         const c = this.getCollection(bucketName);
-        const stream = new MongoReadStream(c, params, params.mongifiedSearch);
+        let stream;
+        if (!params.secondaryStreamParams) {
+            stream = new MongoReadStream(c, params.mainStreamParams, params.mongifiedSearch);
+        } else {
+            const masterStream = new MongoReadStream(c, params.mainStreamParams, params.mongifiedSearch);
+            const versionStream = new MongoReadStream(c, params.secondaryStreamParams, params.mongifiedSearch);
+            stream = new MergeStream(
+                versionStream, masterStream, extension.compareObjects.bind(extension));
+        }
+        const gteParams = params.secondaryStreamParams ?
+            [params.mainStreamParams.gte, params.secondaryStreamParams.gte] : params.mainStreamParams.gte;
         const skip = new Skip({
             extension,
-            gte: params.gte,
+            gte: gteParams,
         });
         let cbDone = false;
 
@@ -1291,12 +1316,18 @@ class MongoClientInterface {
             const newParams = params;
             newParams.start = undefined; // 'start' is deprecated
             newParams.gt = undefined;
-            newParams.gte = range;
 
+            if (params.secondaryStreamParams) {
+                // eslint-disable-next-line no-param-reassign
+                newParams.mainStreamParams.gte = range[0];
+                newParams.secondaryStreamParams.gte = range[1];
+            } else {
+                // eslint-disable-next-line no-param-reassign
+                newParams.mainStreamParams.gte = range;
+            }
             // then continue listing the next key range
             this.internalListObject(bucketName, newParams, extension, log, cb);
         });
-
         stream
             .on('data', e => {
                 skip.filter(e);
@@ -1324,20 +1355,64 @@ class MongoClientInterface {
         return undefined;
     }
 
+    /**
+     * lists versioned and non versioned objects in a bucket
+     * @param {String} bucketName bucket name
+     * @param {Object} params params
+     * @param {String} params.listingType type of algorithm to use (optional)
+     * @param {Number} params.maxKeys maximum number of keys to list (optional)
+     * @param {String} params.prefix prefix of objects to use (optional)
+     * @param {String} params.delimiter delimiter to use (optional)
+     * @param {Object} params.mongifiedSearch search options (optional)
+     * @param {Object} log logger
+     * @param {Function} cb callback
+     * @return {undefined}
+     */
     listObject(bucketName, params, log, cb) {
-        const extName = params.listingType;
-        const extension = new listAlgos[extName](params, log);
-        const internalParams = extension.genMDParams();
-        internalParams.mongifiedSearch = params.mongifiedSearch;
-        return this.internalListObject(bucketName, internalParams, extension,
-            log, cb);
+        return this.getBucketVFormat(bucketName, log, (err, vFormat) => {
+            if (err) {
+                return cb(err);
+            }
+            const extName = params.listingType;
+            // extention here can either be DelimiterMaster ot DelimiterVersions
+            const extension = new listAlgos[extName](params, log, vFormat);
+            // the params returned depend on the vFormat as well as the algorithm used
+            // DelimiterMaster returns an object of filters to apply when filtering
+            // DelimiterVersions in v0 also returns the same thing
+            // in v1 however it returns two objects containing filters to use for two separate
+            // listing streams (master and version)
+            const extensionParams = extension.genMDParams();
+            const internalParams = {
+                mainStreamParams: Array.isArray(extensionParams) ? extensionParams[0] : extensionParams,
+                secondaryStreamParams: Array.isArray(extensionParams) ? extensionParams[1] : null,
+            };
+            internalParams.mongifiedSearch = params.mongifiedSearch;
+            return this.internalListObject(bucketName, internalParams, extension,
+                log, cb);
+        });
     }
 
+    /**
+     * lists versionned and non versionned objects in a bucket
+     * @param {String} bucketName bucket name
+     * @param {Object} params params
+     * @param {String} params.listingType type of algorithm to use (optional)
+     * @param {Number} params.maxKeys maximum number of keys to list (optional)
+     * @param {String} params.prefix prefix of objects to use (optional)
+     * @param {String} params.delimiter delimiter to use (optional)
+     * @param {Object} params.mongifiedSearch search options (optional)
+     * @param {Object} log logger
+     * @param {Function} cb callback
+     * @return {undefined}
+     */
     listMultipartUploads(bucketName, params, log, cb) {
         const extName = params.listingType;
         const extension = new listAlgos[extName](params, log);
-        const internalParams = extension.genMDParams();
-        internalParams.mongifiedSearch = params.mongifiedSearch;
+        const extensionParams = extension.genMDParams();
+        const internalParams = {
+            mainStreamParams: extensionParams,
+            mongifiedSearch: params.mongifiedSearch,
+        };
         return this.internalListObject(bucketName, internalParams, extension,
             log, cb);
     }
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -23,10 +23,14 @@ const diskusage = require('diskusage');
 
 const genVID = require('../../../versioning/VersionID').generateVersionId;
 const listAlgos = require('../../../algos/list/exportAlgos');
+const LRUCache = require('../../../algos/cache/LRUCache');
 
 const MongoReadStream = require('./readStream');
 const MongoUtils = require('./utils');
 const Skip = require('../../../algos/list/skip');
+
+const { formatMasterKey, formatVersionKey } = require('./utils');
+
 
 const USERSBUCKET = '__usersbucket';
 const METASTORE = '__metastore';
@@ -44,17 +48,13 @@ const initialInstanceID = process.env.INITIAL_INSTANCE_ID;
 
 let uidCounter = 0;
 
-const VID_SEP = require('../../../versioning/constants')
-    .VersioningConstants.VersionId.Separator;
+const BUCKET_VERSIONS = require('../../../versioning/constants')
+    .VersioningConstants.BucketVersioningKeyFormat;
 
 function generateVersionId(replicationGroupId) {
     // generate a unique number for each member of the nodejs cluster
     return genVID(`${process.pid}.${uidCounter++}`,
         replicationGroupId);
-}
-
-function formatVersionKey(key, versionId) {
-    return `${key}${VID_SEP}${versionId}`;
 }
 
 function inc(str) {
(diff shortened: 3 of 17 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -393,7 +419,8 @@ class MongoClientInterface {
         const versionId = generateVersionId(this.replicationGroupId);
         // eslint-disable-next-line
         objVal.versionId = versionId;
-        const vObjName = formatVersionKey(objName, versionId);
+        const vObjName = formatVersionKey(objName, versionId, params.vFormat);
+        const mObjName = formatMasterKey(objName, params.vFormat);
```
francoisferrand: take the chance to use more descriptive name : the `m` or `v` prefix looks like a minor "type" indicator, when it actually is most important...

```suggestion
        const versionKey = formatVersionKey(objName, versionId, params.vFormat);
        const masterKey = formatMasterKey(objName, params.vFormat);
```

francoisferrand: additionally, maybe it is be better to tweak the code a bit more, to compute both the versionKey and masterKey in one call?

since most parameters of both functions are actually the same, and both calls are made almost every time...

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -724,11 +755,14 @@ class MongoClientInterface {
     /*
      * This function return the latest version
      */
-    getLatestVersion(c, objName, log, cb) {
+    getLatestVersion(c, objName, vFormat, log, cb) {
+        const mObjName = formatMasterKey(objName, vFormat);
+        let vObjName = formatVersionKey(objName, '', vFormat);
```
francoisferrand: take the change to remove `VID_SEPPLUS` (not used anymore) ; and introduce an alternaive constant for this empty string (`VID_NONE` ?)

also would be nice to have a comment to explain the "no version than increase" : which breaks the abstraction, and works only because the version is added at the end of the key...

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -743,13 +743,25 @@ class MongoClientInterface {
                             { error: err.message });
                         return cb(err);
                     }
-                    return cb(null, value);
+                    if (!doc) {
+                        return next(errors.NoSuchKey);
+                    }
+                    if (doc.value.isPHD) {
+                        this.getLatestVersion(c, objName, vFormat, log, (err, value) => {
```
francoisferrand: should this be `params.vFormat` ?

francoisferrand: also, strange to do this call twice (and the first time without the `vFormat` parameter : which does not seem like a valid use of this function...)
is this maybe some issue with partial commit?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1053,12 +1063,23 @@ class MongoClientInterface {
...
+                // algo backend skipping() function must
+                // return as many skip keys as listing
+                // param sets.
+                for (let i = 0; i < range.length; ++i) {
+                    // eslint-disable-next-line no-param-reassign
+                    newParams[i].gte = inc(range[i]);
+                }
+            } else {
+                // eslint-disable-next-line no-param-reassign
+                newParams.gte = inc(range);
```
francoisferrand: why change to `inc(range)` ?
if this was a bug, it should be fixed on other brahces as well...

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1034,7 +1035,16 @@ class MongoClientInterface {
...
         const c = this.getCollection(bucketName);
-        const stream = new MongoReadStream(c, params, params.mongifiedSearch);
+        const _params = Array.isArray(params) ? params : [params];
+        let stream;
+        if (_params.length === 1) {
+            stream = new MongoReadStream(c, _params[0], _params.mongifiedSearch);
+        } else {
+            const masterStream = new MongoReadStream(c, _params[0], _params.mongifiedSearch);
+            const versionStream = new MongoReadStream(c, _params[1], _params.mongifiedSearch);
+            stream = new MergeStream(
```
francoisferrand, line 1296: how is this related?
this updates the implementation of an internal function (`internalListObject`), to support passing an array of parameters.... but not the place where it is called : so it actually has no effect...

* if not needed, don't add it ; and avoid splitting commits too much (e.g. code and test should ideally be in same commit ; updating the implementation of a function and making use of it, if not too big...)
* please write a descriptive commit message (https://cbea.ms/git-commit/)
* take the change to add documentation header to the function, even if missing, when you are changing its prototype

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -703,46 +732,53 @@ class MongoClientInterface {
...
-                params, log, cb);
-        } else if (params && params.versionId === '') {
-            return this.putObjectVerCase2(c, bucketName, objName, objVal,
-                params, log, cb);
-        } else if (params && params.versionId && !params.repairMaster) {
-            return this.putObjectVerCase3(c, bucketName, objName, objVal,
-                params, log, cb);
-        } else if (params && params.versionId && params.repairMaster) {
-            return this.putObjectVerCase4(c, bucketName, objName, objVal,
+        return this.getBucketVFormat(bucketName, log, (err, vFormat) => {
```
francoisferrand, line 844: this is much more significant than just "updating tests" !
please try to be more careful when splitting commits...

francoisferrand:
i think we are missing *unit tests* for the code which was added : e.g. each of the `update XXXX function` should typically include unit tests to validate all the new code path (e.g. each error case...). Note i am talking about _unit_ tests only here, e.g. validating the code is behaving as expected, returning the right errors in the right place, ... (and using mocks instead of relying on an actual mongodb deployment, contrary to the functional tests).

lib/storage/metadata/mongoclient/MongoClientInterface.js
francoisferrand, line 844: Actually, I am wondering if this should be done in here.... It seems to me that reading the metastore to find the "type" of bucket is already done in Cloudserver : so IMHO it would be more consistant if it staid there only, and this function would just take the "vFormat" in argument...

Another argument for this approach is that cloudserver has likely alreday queried the metastore to retrieve information on the bucket, so this extra call to the backend here may actually be redundant in many/most case.

How is it implemented in other backeds? Do we have the same call to metastore from within the `putObject` method ?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -703,46 +732,53 @@ class MongoClientInterface {
...
-                    { error: err.message });
-                return cb(errors.InternalError);
-            }
-            if (!doc) {
-                return cb(errors.NoSuchKey);
-            }
-            if (doc.value.isPHD) {
-                this.getLatestVersion(c, objName, log, (err, value) => {
+        async.waterfall([
+            next => this.getBucketVFormat(bucketName, log, next),
```
francoisferrand, line 874: same here (and in the other methods)

tests/functional/metadata/mongodb/getObject.spec.js
```diff
@@ -106,6 +111,25 @@ describe('MongoClientInterface::metadata.getObjectMD', () => {
             });
     }
 
+    function setObjectVFormat(vFormat, cb) {
```
francoisferrand: should not be needed: just create the bucket with the right format in its metadata?

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: > should this be params.vFormat ?

In the waterfall we first get the bucket vFormat then pass it to the next function, we don't have a vFormat in the params

welansari: I'm not sure i see where we call it twice ?

francoisferrand: twice at line 730-740 then line 749-750 :

```javascript
            if (doc.value.isPHD) {
                this.getLatestVersion(c, objName, log, (err, value) => {
                    [...]
                    if (doc.value.isPHD) {
                        this.getLatestVersion(c, objName, vFormat, log, (err, value) => {
```

welansari: That's probably a bug in Github, if you go to this you'll see that there is only one
https://github.com/scality/Arsenal/blob/3194f6941acae5ef7a5ec310ec8adc87c778fc5f/lib/storage/metadata/mongoclient/MongoClientInterface.js#L730-L769

francoisferrand: indeed, very weird...

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: > How is it implemented in other backends?

Cloudserver doesn't have access to the vFormat attribut, the logic that deals with that is offloaded to the metadata backend, this is also the case for other backends. So Cloudserver does query the `metastore` to get bucket attributs but it never gets the vFormat.

> Do we have the same call to `metastore` from within the putObject method ?

In metadata it's a bit different, but as far as i understood yes we always query the db to get the vFormat at each request.

https://github.com/scality/MetaData/blob/613bc56a9f174dd4ba95f2e513dc648260bac3d9/lib/server/VersioningRequestProcessor.js#L407-L411
^ Here `replicateRequest` is used to replicate the putObject (and deleteObject) action among the servers. We get the vFormat each time we call it.

tests/functional/metadata/mongodb/getObject.spec.js
welansari: Bucket metadata is stored inside the `BucketInfo` class, and since it doesn't include vFormat we can't really do that

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: > additionally, maybe it is be better to tweak the code a bit more, to compute both the versionKey and masterKey in one call?

I'm not really a fan of doing that, i think it'll just over complicate things. IMO we should keep them separate just to keep things intuitive

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: In the `listObject` function, we generate the internal params of the listing algorithm here:
https://github.com/scality/Arsenal/blob/ad62d8b2e0fe9b4f1b4a5536fffa075026c83026/lib/storage/metadata/mongoclient/MongoClientInterface.js#L1131-L1132
Those params depend on the version of the algorithm (DelimiterMaster, DelimiterVersions) as well as the vFormat. In `v0` the params are in form of an object and in `v1` they are in form of an array of two objects where each object contain the filters to apply to the master and version stream.

This is the way it is implemented in the `metadata` backend, and to avoid changing everything i reused the code and classes already there (MergeStream)

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -388,27 +421,38 @@ class MongoClientInterface {
      * We also test the existence of the versionId property
      * to manage the case of an object created before the
      * versioning was enabled.
+     * @param {Object} c bucket collection
+     * @param {String} bucketName bucket name
+     * @param {String} objName object name
+     * @param {Object} objVal object metadata
```
francoisferrand, line 481: whenever possible, please specify the expected "type" of the Object, like:

```suggestion
     * @param {ObjectMD} objVal object metadata
```

lib/storage/metadata/mongoclient/MongoClientInterface.js
francoisferrand: ok, let's leave it like this then...

lib/storage/metadata/mongoclient/MongoClientInterface.js
jonathan-gramain: In metadata the bucket metadata is cached once retrieved, which means that we only get them from the db when they're not in the cache.

lib/algos/list/skip.js
```diff
@@ -59,10 +59,17 @@ class Skip {
...
+                    newRange = [];
+                    for (let i = 0; i < skippingRange.length; ++i) {
+                        newRange.push(this._inc(skippingRange[i]));
+                    }
+                } else {
+                    newRange = this._inc(skippingRange);
+                }
                 /* Avoid to loop on the same range again and again. */
-                if (newRange === this.gteParams) {
+                if (JSON.stringify(newRange) === JSON.stringify(this.gteParams)) {
```
jonathan-gramain: This is tricky, but you can actually leave this check alone (in v1, it would actually never be true in either case). For the context, the check was introduced in metadata to fix https://scality.atlassian.net/browse/MD-678.

I cannot tell with full confidence whether a similar issue would never happen with v1 actually, but clearly the fix here wouldn't work in v1 due to how the data is laid out internally, and here we compare the user param `gte` against the internal skip params which include the v1 special prefix. In metadata I did not change the original logic, presumably for that reason (I don't remember since it's been a while!)

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -189,21 +188,36 @@ class MongoClientInterface {
...
         const newBucketMD = JSON.parse(bucketMDStr);
         const m = this.getCollection(METASTORE);
+
+        const payload = {
+            _id: bucketName,
+            value: newBucketMD,
+        };
+        if (bucketName !== constants.usersBucket &&
+            bucketName !== PENSIEVE) {
+            payload.vFormat = BUCKET_VERSIONS.v1;
```
jonathan-gramain: The vFormat for new buckets should be configurable at least (via an env var?), and even default to v0 maybe for the time being until we get confidence that it's reliable (it's still in beta phase for now in metadata, so likely the same should go for Mongo).

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -270,6 +284,25 @@ class MongoClientInterface {
...
     }
 
+    getBucketVFormat(bucketName, log, cb) {
+        const m = this.getCollection(METASTORE);
+        m.findOne({
+            _id: bucketName,
+        }, {}, (err, doc) => {
+            if (err) {
+                log.error(
+                    'getBucketVFormat: error getting bucket vFormat',
```
jonathan-gramain, line 342: We should log the bucket name here

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -668,68 +753,127 @@ class MongoClientInterface {
...
-        } else if (params && params.versionId && !params.repairMaster) {
-            return this.putObjectVerCase3(c, bucketName, objName, objVal,
-                params, log, cb);
-        } else if (params && params.versionId && params.repairMaster) {
-            return this.putObjectVerCase4(c, bucketName, objName, objVal,
+        return this.getBucketVFormat(bucketName, log, (err, vFormat) => {
+            if (err) {
+                return cb(err);
+            }
+            if (!params) {
```
jonathan-gramain: I'd suggest creating a new object `_params = Object.assign({}, params)` to avoid modifying the argument, since it's an API function we cannot assume that it's safe to modify `params`.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -668,68 +753,127 @@ class MongoClientInterface {
...
-                return cb(errors.NoSuchKey);
-            }
-            if (doc.value.isPHD) {
-                this.getLatestVersion(c, objName, log, (err, value) => {
+        let key;
+        async.waterfall([
+            next => this.getBucketVFormat(bucketName, log, next),
+            (vFormat, next) => {
+                if (params && params.versionId) {
+                    // eslint-disable-next-line
```
jonathan-gramain: This eslint exception may not be needed anymore

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -668,68 +753,127 @@ class MongoClientInterface {
...
+                    key = formatVersionKey(objName, params.versionId, vFormat);
+                } else {
+                    // eslint-disable-next-line
```
jonathan-gramain: same

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -668,68 +753,127 @@ class MongoClientInterface {
...
+            return this.putObjectNoVer(c, bucketName, objName, objVal,
                 params, log, cb);
-        }
-        return this.putObjectNoVer(c, bucketName, objName, objVal,
-            params, log, cb);
+        });
     }
 
+    /**
+     * gets versionned and non versionned object metadata
```
```suggestion
     * gets versioned and non-versioned object metadata
```

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -668,68 +753,127 @@ class MongoClientInterface {
...
+                } else {
+                    // eslint-disable-next-line
+                    key = formatMasterKey(objName, vFormat);
+                }
+                c.findOne({
+                    _id: key,
+                }, {}, (err, doc) => {
                     if (err) {
-                        log.error('getLatestVersion: getting latest version',
+                        log.error('findOne: error getting object',
```
jonathan-gramain, line 885: The bucket and object key should be logged as well

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -668,68 +753,127 @@ class MongoClientInterface {
...
+                        return next(errors.InternalError);
                     }
-                    return cb(null, value);
+                    if (!doc) {
+                        return next(errors.NoSuchKey);
+                    }
+                    if (doc.value.isPHD) {
+                        this.getLatestVersion(c, objName, vFormat, log, (err, value) => {
+                            if (err) {
+                                log.error('getLatestVersion: getting latest version',
```
jonathan-gramain: Maybe time to add the bucket/object key here too

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -668,68 +753,127 @@ class MongoClientInterface {
...
+                        log.error('findOne: error getting object',
                             { error: err.message });
-                        return cb(err);
+                        return next(errors.InternalError);
                     }
-                    return cb(null, value);
+                    if (!doc) {
+                        return next(errors.NoSuchKey);
+                    }
+                    if (doc.value.isPHD) {
```
jonathan-gramain: I think we could do `return next(null, doc)` and create a new step in the waterfall for the following logic

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -668,68 +753,127 @@ class MongoClientInterface {
...
+        // version id is added at the end of the key so giving it an empty
+        // string gives us the last key in the range
+        const versionKey = formatVersionKey(objName, VID_NONE, vFormat);
+        const lastVersionKey = inc(versionKey);
+        const filter = {};
+        if (vFormat === BUCKET_VERSIONS.v0) {
+            filter.$gt = masterKey;
+            filter.$lt = lastVersionKey;
+        } else {
+            filter.$gt = versionKey;
```
jonathan-gramain, line 936: It may not be enough to retrieve only the version key in v1, due to the fact that a bucket may be nonversioned for some time and only create master keys without associated version keys. I think we need to check both prefixes for that reason in v1.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -751,20 +895,30 @@ class MongoClientInterface {
...
+     * @param {String} objName object name
+     * @param {Object} objVal new object version metadata
+     * @param {Object} mst master version metadata
+     * @param {String} vFormat key format version
+     * @param {Object} log logger
+     * @param {Function} cb callback
+     * @return {undefined}
      */
-    repair(c, bucketName, objName, objVal, mst, log, cb) {
+    repair(c, bucketName, objName, objVal, mst, vFormat, log, cb) {
```
jonathan-gramain, line 976: In metadata implementation, v1 does not create any PHD key since the repair is done at time of deletion of a specific version, hence the repair code should not have to deal with v1. But if you decided to keep the PHD in v1 in the Mongo implementation, then you can ignore my comment.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -940,20 +1140,29 @@ class MongoClientInterface {
                         if (err) {
                             return cb(err);
                         }
-                        return cb();
+                        return cb(null);
```
jonathan-gramain: Is this change needed

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -971,22 +1180,67 @@ class MongoClientInterface {
...
+     */
     deleteObject(bucketName, objName, params, log, cb) {
         const c = this.getCollection(bucketName);
-        if (params && params.versionId) {
-            return this.deleteObjectVer(c, bucketName, objName,
+        return this.getBucketVFormat(bucketName, log, (err, vFormat) => {
+            if (err) {
+                return cb(err);
+            }
+            if (!params) {
```
jonathan-gramain: Same comment about creating a new `_params` object

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1035,22 +1295,72 @@ class MongoClientInterface {
         return undefined;
     }
 
+    /**
+     * lists versionned and non versionned objects in a bucket
```
```suggestion
     * lists versioned and non-versioned objects in a bucket
```

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1035,22 +1295,72 @@ class MongoClientInterface {
...
+            }
+            const extName = params.listingType;
+            // extention here can either be DelimiterMaster ot DelimiterVersions
+            const extension = new listAlgos[extName](params, log, vFormat);
+            // the params returned depend on the vFormat as well as the algorithm used
+            // DelimiterMaster returns an object of filters to apply when filtering
+            // DelimiterVersions in v0 also returns the same thing
+            // in v1 however it returns two objects containing filters to use for two separate
+            // listing streams (master and version)
+            const extentionParams = extension.genMDParams();
```
```suggestion
            const extensionParams = extension.genMDParams();
```

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1035,22 +1295,72 @@ class MongoClientInterface {
...
+     * @return {undefined}
+     */
     listMultipartUploads(bucketName, params, log, cb) {
-        const extName = params.listingType;
-        const extension = new listAlgos[extName](params, log);
-        const internalParams = extension.genMDParams();
-        internalParams.mongifiedSearch = params.mongifiedSearch;
-        return this.internalListObject(bucketName, internalParams, extension,
-            log, cb);
+        return this.getBucketVFormat(bucketName, log, (err, vFormat) => {
```
jonathan-gramain: I'm not sure how this works with Mongo, but I'm surprised that we need to take into account the vFormat of the bucket to list its multipart-uploads. Normally (in metadata at least), listing multipart uploads goes through a special sidekick bucket that is just a flat list of open multipart uploads, hence always in v0. Can you check this to make sure it's correct?

lib/storage/metadata/mongoclient/utils.js
```diff
@@ -117,11 +120,79 @@ function translateConditions(depth, prefix, object, cond) {
     throw errors.InternalError;
 }
 
+/**
+ * format v0 master key
+ * @param {String} key object key
+ * @return {String} formatted key
+ */
+function formatMasterKeyV0(key) {
+    return `${key}`;
```
jonathan-gramain:
```suggestion
    return key;
```

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: Well i didn't exclude the mpu buckets in the createBucket function so they too have a v0 and v1 key format, but you're right there's no point in having them support the v1 format, i'll fix that

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: The `ObjectMD` example here might bring confusion cause what we get is an object that contains the ObjectMD data not the actual instance of the `ObjectMD` class

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: Do you recommend adding a cache here? and if so how do we decide its sizing ?

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: `getLatestVersion` only returns the versions and never the master, this is the case in previous implementation as well. You can see that it gets keys greater than the master key and not greater or equal.
If i changed the function to also check the master prefix it would mess up the other functions calling it. An example would be the repair function that gets the latest version to decide if it should repair or delete the master.
Having the function in it's current state doesn't create a problem when retrieving objects that don't have a version associated to them as we get them by their master key in `getObject`

jonathan-gramain: OK I didn't notice that this function was for internal use, good point

lib/storage/metadata/mongoclient/MongoClientInterface.js
vrancurel: We could get rid of the PHDs in MongoClientInterface but I suggest we do that in a v2:
E.g.
- Use transactions to atomically update/delete MasterKey + LastVersionKey MD (the rationale is to have the MD up-to-date at all time in the MasterKey for O(1) access of the MD.

So let's make it simple and fix the DeleteMarkers first in v1 (and therefore keep PHD related concepts).

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -114,6 +114,11 @@ class MongoClientInterface {
             !Number.isNaN(process.env.CONCURRENT_CURSORS))
             ? Number.parseInt(process.env.CONCURRENT_CURSORS, 10)
             : CONCURRENT_CURSORS;
+
+        this.bucketMetadataCache = new LRUCache(constants.maxCachedBuckets);
```
vrancurel: A value of 1000 shall be fine.
Can you monitor the cache/hit miss somewhere ?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -148,6 +157,16 @@ class MongoClientInterface {
                 ignoreUndefined: true,
             });
             this.adminDb = client.db('admin');
+            // log cache hit/miss every 5min
+            this.cacheHitMissLoggerInterval = setInterval(() => {
+                let hitRatio = (this.cacheHit / (this.cacheHit + this.cacheMiss)) || 0;
+                hitRatio = hitRatio.toFixed(3);
+                this.logger.info('MongoClientInterface: Bucket vFormat cache hit/miss (5min)',
```
vrancurel: `logger.debug` shall be enough I think

vrancurel (approved):

lib/algos/list/skip.js
```diff
@@ -59,8 +59,15 @@ class Skip {
         } else if (filteringResult === FILTER_SKIP
                    && skippingRange !== SKIP_NONE) {
             if (++this.streakLength >= MAX_STREAK_LENGTH) {
-                const newRange = this._inc(skippingRange);
-
+                let newRange;
```
francoisferrand, line 62: can we add a (unit) test for this?

lib/constants.js
```diff
@@ -148,4 +150,10 @@ module.exports = {
         'noncurrentVersionExpiration',
         'abortIncompleteMultipartUpload',
     ],
+    // Maximum number of buckets to cache (bucket metadata)
+    maxCachedBuckets: process.env.METADATA_MAX_CACHED_BUCKETS ?
+        Number(process.env.METADATA_MAX_CACHED_BUCKETS) : 1000,
+    // default bucket key format version to use
+    defaultBucketVFormat: process.env.DEFAULT_BUCKET_KEY_FORMAT ?
```
francoisferrand: not really a `constant` ; should this be added to the config instead?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -114,6 +114,15 @@ class MongoClientInterface {
             !Number.isNaN(process.env.CONCURRENT_CURSORS))
             ? Number.parseInt(process.env.CONCURRENT_CURSORS, 10)
             : CONCURRENT_CURSORS;
+
+        this.bucketVFormatCache = new LRUCache(constants.maxCachedBuckets);
+        this.defaultBucketKeyFormat = [BUCKET_VERSIONS.v0, BUCKET_VERSIONS.v1]
+            .includes(constants.defaultBucketVFormat) ? constants.defaultBucketVFormat
+            : BUCKET_VERSIONS.v0;
```
francoisferrand: should this not default to v1 ?

btw, you got 2 places where the default value is defined : here and in the "constant" defintion. it would be better to have a single default value.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -148,6 +157,16 @@ class MongoClientInterface {
                 ignoreUndefined: true,
             });
             this.adminDb = client.db('admin');
+            // log cache hit/miss every 5min
+            this.cacheHitMissLoggerInterval = setInterval(() => {
```
francoisferrand, line 161: should we make these (cachehHit/miss) available as prometheus metrics?
this would be easier to consume and monitor...
Should probably be done later, in a separate PR.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -189,21 +209,37 @@ class MongoClientInterface {
...
         const m = this.getCollection(METASTORE);
+
+        const payload = {
+            _id: bucketName,
+            value: newBucketMD,
+        };
+        if (bucketName !== constants.usersBucket &&
+            bucketName !== PENSIEVE &&
+            !bucketName.startsWith(constants.mpuBucketPrefix)) {
+            payload.vFormat = this.defaultBucketKeyFormat;
```
francoisferrand, line 235: generally (and thinking about typing/upcoming typescript migration) : vFormat should be initialized anyway, e.g. to v0 format.
Or if we need an extra different behavior we can introduce yet another value (like "unversioned")

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -212,6 +248,8 @@ class MongoClientInterface {
                     { error: err.message });
                 return cb(errors.InternalError);
             }
+            // caching bucket vFormat
+            this.bucketVFormatCache.add(bucketName, payload.vFormat);
```
francoisferrand, line 254: c.f. previous comment : `paylog.vFormat` may not be defined here.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -268,6 +313,42 @@ class MongoClientInterface {
...
+            if (err) {
+                log.error(
+                    'getBucketVFormat: error getting bucket vFormat',
+                    { bucket: bucketName, error: err.message });
+                return cb(errors.InternalError);
+            }
+            if (!doc) {
+                return cb(null, BUCKET_VERSIONS.v0);
+            }
+            const vFormat = doc.vFormat ? doc.vFormat : BUCKET_VERSIONS.v0;
```
francoisferrand:
```suggestion
            const vFormat = doc.vFormat || BUCKET_VERSIONS.v0;
```

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -309,8 +390,10 @@ class MongoClientInterface {
         m.update({
             _id: bucketName,
         }, {
-            _id: bucketName,
-            value: newBucketMD,
+            $set: {
```
francoisferrand, line 395: what 's the difference (with vs w/o `$set`) ? was it a bug?

welansari: without $set the vFormat gets removed

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: it's out of the scope of this PR, but yes good idea

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: it should yes, fixed that

francoisferrand (approved):

francoisferrand:
/wait

Some open question has been raised about "activation" of this feature: e.g. should we make the feature configurable (at Zenko level), and consequently what should be the default here (e.g. keep v0 for compatibility, or default to v1 as the new normal).

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -753,24 +804,41 @@ class MongoClientInterface {
...
-        } else if (params && params.versionId && !params.repairMaster) {
-            return this.putObjectVerCase3(c, bucketName, objName, objVal,
-                params, log, cb);
-        } else if (params && params.versionId && params.repairMaster) {
-            return this.putObjectVerCase4(c, bucketName, objName, objVal,
-                params, log, cb);
-        }
-        return this.putObjectNoVer(c, bucketName, objName, objVal,
-            params, log, cb);
+        const _params = Object.assign({}, params);
```
jbertran, line 843: Consider `const _params = { ...params };` instead - no side-effect, no doc reading required, same cost in memory.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -753,24 +804,41 @@ class MongoClientInterface {
...
+        return this.getBucketVFormat(bucketName, log, (err, vFormat) => {
+            if (err) {
+                return cb(err);
+            }
+            _params.vFormat = vFormat;
+            if (_params && _params.versioning && !_params.versionId) {
```
jbertran: Interestingly, all the following function calls use the same arguments. You could split this into one function that returns the `putObjectStrategy`, in which you could early-return in a lot of cases and improve readability.

```js
function getPutObjectStrategy(params) {
    if (_params.versionId === '') {
        return this.putObjectVerCase2;
    }
    if (!_params.repairMaster)  {
        return this.putObjectVerCase3;
    }
    [...]
    return this.putObjectNoVer;
}
```

jbertran (commented):
Nice work! That's a lot of moving parts :clap:

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: Linter doesn't seem to like this one

jbertran (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
jonathan-gramain: For the context, in S3C we leave it to v0 for now as a default because we will introduce the feature as NFI, to make sure early adopters do not have issues before making it available to every customer.

For Artesca the context is different since there are far fewer customers and less stored data, and you want customers to benefit from v1 ASAP, so I think it is good to enable v1 by default, but just so you know.

PR #1735, opened 2022-03-02
Feature: ARSN-83 remove master when delete marker in MongoClientInterface 

welansari:
- Update putObject functions to delete master version when last version is a delete marker
- Update getObject function to look fro last object version when master not found
- Update deleteObject functions to delete the placeholder master when last version is a delete marker

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -465,8 +465,43 @@ class MongoClientInterface {
     }
 
     /**
+     * Returns the suitable mongo operation to perform
+     * depending on the version being put
+     * In v1 the master gets deleted instead of being
+     * updated like in v0 when the last version is a delete
+     * marker
+     * @param {Boolean} isDeleteMarker isDeleteMarker tag
+     * @param {String} vFormat vFormat of bucket
+     * @param {Object} filter filter to get master
+     * @param {Object} update value to update master with
+     * @param {Boolean} upsert if upserting is needed
+     * @return {Object} mongo operation
+     */
+    updateDeleteMaster(isDeleteMarker, vFormat, filter, update, upsert) {
+        // delete master when we are in v1 and the version is a delete
+        // marker
+        if (isDeleteMarker && vFormat === BUCKET_VERSIONS.v1) {
+            return {
+                deleteOne: {
+                    filter,
+                },
+            };
+        }
+        // in v0 or if the version is not a delete marker the master
+        // simply gets updated
+        return {
+            updateOne: {
+                filter,
+                update,
+                upsert,
+            },
+        };
+    }
+
+    /**
      * In this case we generate a versionId and
      * sequentially create the object THEN update the master.
+     * Master is deleted when version put is a delete marker
      *
      * It is possible that 2 version creations are inverted
      * in flight so we also check that we update a master only
@@ -492,7 +527,8 @@ class MongoClientInterface {
         objVal.versionId = versionId;
         const versionKey = formatVersionKey(objName, versionId, params.vFormat);
         const masterKey = formatMasterKey(objName, params.vFormat);
-        c.bulkWrite([{
+        // initiating array of operations with version creation
+        const ops = [{
             updateOne: {
                 filter: {
                     _id: versionKey,
@@ -502,29 +538,32 @@ class MongoClientInterface {
                 },
                 upsert: true,
             },
-        }, {
-            updateOne: {
-                // eslint-disable-next-line
-                filter: {
-                    _id: masterKey,
-                    $or: [{
-                        'value.versionId': {
-                            $exists: false,
-                        },
-                    },
-                    {
-                        'value.versionId': {
-                            $gt: objVal.versionId,
-                        },
-                    },
-                    ],
+        }];
+        // filter to get master
+        const filter = {
+            _id: masterKey,
+            $or: [{
+                'value.versionId': {
+                    $exists: false,
                 },
-                update: {
-                    $set: { _id: masterKey, value: objVal },
-                },
-                upsert: true,
             },
-        }], {
+            {
+                'value.versionId': {
+                    $gt: objVal.versionId,
+                },
+            },
+            ],
+        };
+        // values to update master
+        const update = {
+            $set: { _id: masterKey, value: objVal },
+        };
+        // updating or deleting master depending on the last version put
+        // in v0 the master gets updated, in v1 the master gets deleted if version is
+        // a delete marker or updated otherwise.
+        const masterOp = this.updateDeleteMaster(objVal.isDeleteMarker, params.vFormat, filter, update, true);
+        ops.push(masterOp);
+        c.bulkWrite(ops, {
             ordered: true,
         }, err => {
             /*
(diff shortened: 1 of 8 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -470,6 +471,14 @@ class MongoClientInterface {
                 },
                 upsert: true,
             },
+        }, {
+            deleteOne: {
```
jonathan-gramain: I think for performance reasons, we should check from `objVal` whether we are putting a delete marker, and construct the minimal request to achieve the same result, i.e send a delete to the potentially existing master key iff we put a delete marker. I fear that putting + deleting a master key, even in the same batch of operations, could have a significant performance cost compared to just deleting it, but I may be wrong.

I believe we still need to keep the condition on the `versionId` on the delete to cope with race conditions if other writers create a valid non-delete-marker version at the same time.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -825,10 +859,12 @@ class MongoClientInterface {
                             { error: err.message });
                         return next(errors.InternalError);
                     }
-                    if (!doc) {
+                    if (!doc && params && params.versionId) {
```
jonathan-gramain: I don't think we should change the behavior here (I may be wrong):

When we get an object without a version ID (so implicitly the latest version), whether it's a non-existing object or a delete marker, it should always return `NoSuchKey` I believe, per the API. So everything we have to do is check the status of the master key, even in v1.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -983,36 +1019,36 @@ class MongoClientInterface {
...
-                    });
-                    return undefined;
-                }
+        this.getLatestVersion(c, objName, vFormat, log, (err, version) => {
+            if (err && err !== errors.NoSuchKey) {
                 log.error('getLatestVersion: error getting latest version',
                     { error: err.message });
                 return cb(err);
             }
+            if (err === errors.NoSuchKey || version.isDeleteMarker) {
```
jonathan-gramain: Same comment about the use of PHD in v1: the metadata implementation does not create PHDs anymore in v1, so make sure that this change is because Mongo still creates PHDs in v1 and that it's wanted this way.

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: I do agree with the fact that with or without the modification at the end of the day we'll get the same response from Cloudserver (A NoSuchKey error), but i didn't want to change the behaviour of the function cause before, if the master was a delete marker it still got returned and then the logic is handled in Cloudserver.

vrancurel:
If I understand correctly this PR is based on another PR, can you rebase on its branch because this is very hard to see what is new in the code ? Thx

welansari:
Sorry about that forgot to rebase, should be fixed now

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -500,29 +501,51 @@ class MongoClientInterface {
...
-                    },
-                    {
-                        'value.versionId': {
-                            $gt: objVal.versionId,
+                        {
+                            'value.versionId': {
+                                $gt: objVal.versionId,
+                            },
                         },
+                        ],
```
francoisferrand: could this be written the other way round?

```
'value.versionId': {
   $or: [
      { $exists: false },
      { $gt: objVal.versionId }
   ]
}
```

It seems easier to read and less redundant...

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -500,29 +501,51 @@ class MongoClientInterface {
...
-        }, {
-            updateOne: {
-                // eslint-disable-next-line
-                filter: {
-                    _id: masterKey,
-                    $or: [{
-                        'value.versionId': {
-                            $exists: false,
+        }];
+        if (objVal.isDeleteMarker && params.vFormat === BUCKET_VERSIONS.v1) {
```
francoisferrand: seems this "pattern" is quite verbose & used multiple times, would be easier to read with a new function to abstract it a bit : something like

```
buildUpdateMaster(masterKey, versionId /*or filter*/, objVal) {
  if (objVal.isDeleteMarker && params.vFormat === BUCKET_VERSIONS.v1)
     return { deleteOne: .... }
  return { updateOne: .... }
```

francoisferrand:
Seems to be missing some new tests, right?

welansari:
Yes some functional tests are missing, to check if master gets deleted in v1 and if the other functions behave the way they should

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -465,6 +465,7 @@ class MongoClientInterface {
     /**
      * In this case we generate a versionId and
      * sequentially create the object THEN update the master.
+     * Master is deleted when version put is a delete marker
```
francoisferrand, line 504: are we already handling the re-creating of Master when a new verison is added after a delete marker? e.g. 

* Create versionned object
* Delete object, i.e. create delete marker → Master version is removed
* Put another version of the object → Master version should be re-created

welansari: Yes we are, that's what the `upsert: true` param is for, when master is non existent it gets created

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: It is easier to read like that i agree, but unfortunately that is not possible as the `$or` operator doesn't apply to single fields, and so for each item we need to specify the field

lib/storage/metadata/mongoclient/MongoClientInterface.js
francoisferrand: ok nice, just please make sure the tests do check this

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -462,9 +462,37 @@ class MongoClientInterface {
...
+     * Returns the suitable mongo operation to perform
+     * depending on the version being put
+     * @param {Boolean} isDeleteMarker isDeleteMarker tag
+     * @param {String} vFormat vFormat of bucket
+     * @param {Object} filter filter to get master
+     * @param {Object} update value to update master with
+     * @param {Boolean} upsert if upserting is needed
+     * @return {Object} mongo operation
+     */
+    updateDeleteMaster(isDeleteMarker, vFormat, filter, update, upsert) {
```
vrancurel, line 480: The code is very complex to read, the logic is not that common between v0 and v1. Why not having different files from v0 and v1 ? this way we prepare for a v2 with e.g. transactions.

vrancurel (changes requested):

jbertran (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: I don't think having a different file for each version is a good idea in this point in time, because that will introduce a lot of code duplication.
The logic between v0 and v1 is actually pretty similar, all the changes introduced in this PR are very slight, for the most part we only add an extra step where we check if the master needs to be deleted.

vrancurel (approved):

PR #1736, opened 2022-03-04
ARSN-100 Migrate errors to TypeScript

ghivert:
## Arsenal Errors in TypeScript

This PR is a sample to present what TypeScript looks like when integrated in a codebase.
This PR focuses on one aspect: migrating the Arsenal errors into a new TypeScript class, with proper type-checking.
Switching from JS to TS adds a bunch of new cool features.

- Full type-checking of errors, and all affiliated files. The JSON errors is migrated to TS to type-check the different errors and get a list of all the possible errors.
- Some type manipulation: in the `types.ts` and `errors/index.ts` files, we are using [`keyof`](https://www.typescriptlang.org/docs/handbook/2/keyof-types.html), [`typeof`](https://www.typescriptlang.org/docs/handbook/2/typeof-types.html) 
and [Mapped Types](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html). Those are good examples to illustrate the different patterns we can have in TypeScript to create exhaustive types.
- Defining a new field in the class with proper private visibility: `#is`. `is` is the new way to check the type of the error as the previous iteration was untypable due to meta-programming. This allow both to write a nice fluent API and to keep type-checking.
- New features from JS: [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy). This shows an example of how to mix meta-programming and type-checking. The small amount of code to write (thanks to meta-programming) doesn't mean that we're giving up safety. Try to use `errors.is` in the codebase with a TypeScript IDE, and you'll see the auto-complete niceties.

    <details>
    <summary>The screenshot of auto-complete (click to open)</summary>
    
    > Capture d’écran 2022-03-04 à 15 39 06
    </details>
    
- Getters from new classes, with `get` keyword.
- New import/export syntax to follow as closely as possible the ES modules standard.

## Commits

- ARSN-67 Rename errors and arsenalErrors
- ARSN-67 Switch errors to TS
- ARSN-67 Change errors.spec.js to errors.spec.ts
- ARSN-67 Fix all tests


tests/unit/errors.spec.ts
```diff
@@ -0,0 +1,39 @@
+import * as rawErrors from '../../lib/errors/arsenalErrors';
+import { errors } from '../../index';
+import { ArsenalError } from '../../lib/errors';
+
+describe('Errors: ', () => {
+    Object.entries(errors).forEach(([name, error]) => {
+        const raw = rawErrors[name];
+        it(`should return and instance of ${name} Error`, () => {
```
miniscruff:
```suggestion
        it(`should return an instance of ${name} Error`, () => {
```

lib/errors/index.ts
```diff
@@ -0,0 +1,94 @@
...
+// Proxy can be found on MDN.
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy
+const createIs = (type: types.Name) => {
+    const get = (_: {}, value: string | symbol) => type === value;
+    return new Proxy({}, { get }) as Is;
+};
+
+export class ArsenalError extends Error {
+    /** HTTP status code. Example: 401, 403, 500, ... */
+    #code: number;
```
miniscruff, line 25: why use # here?

lib/errors/index.ts
```diff
@@ -0,0 +1,94 @@
...
+    get description() {
+        return this.#description;
+    }
+
+    get type() {
+        return this.#type;
+    }
+
+    /** Generate all possible errors. An instance is created by default. */
+    static errors() {
```
miniscruff, line 86: I would move this out of the class personally, Could also have a return type or does TS auto find it?

lib/errors/types.ts
```diff
@@ -0,0 +1,11 @@
...
+
+const entries = Object.keys(rawErrors).map((v) => [v, v]);
+
+/** All possible errors. */
+export type Name = keyof typeof rawErrors
+/** Object containing all errors. It has the format [Name]: "Name" */
+export type Names = { [Name_ in Name]: Name_ };
+
+/** Use types with error.is(types.InternalError) to have nice autocomplete */
+export const types: Names = Object.fromEntries(entries);
```
miniscruff: types seems like a bad name here, it feels to me like that would conflict with any other generic type value or import. The best I got right now is just to call it "errors" but that can be redundant when reading it. `error is errors internal error`. You would probably glance over the second error word though. We could also add an extra map without the `Error` suffix in the error names so you could do: `error.is(errors.Internal)`. 

Or maybe just rename `is` to `of` so it reads `error of type internal error`. Feels slightly more grammatically correct. 

Kinda bike shedding on this one though so it is probably fine.

lib/errors/index.ts
```diff
@@ -0,0 +1,94 @@
+import * as http from 'http';
```
miniscruff: If we are only importing for the type, doesn't type script have some sort of type import? Feels a bit heavy to import all of http just for the few types.

lib/errors/index.ts
```diff
@@ -0,0 +1,94 @@
...
+export type Is = { [Name in types.Name]: boolean };
+/** Mapping of all possible Errors */
+export type Errors = { [Property in keyof types.Names]: ArsenalError };
+
+const entries = Object.entries(rawErrors);
+
+// This contains some metaprog. Be careful.
+// Proxy can be found on MDN.
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy
+const createIs = (type: types.Name) => {
```
miniscruff: Is this meta thing necessary? Can't we just do?
```js
function is(name: string): boolean {
    return this.#type === name;
}
```

lib/errors/index.ts
```diff
@@ -0,0 +1,94 @@
...
+import * as types from './types';
+
+export * from './types';
+
+/** Mapping used to determine an error type. */
+export type Is = { [Name in types.Name]: boolean };
+/** Mapping of all possible Errors */
+export type Errors = { [Property in keyof types.Names]: ArsenalError };
+
+const entries = Object.entries(rawErrors);
```
miniscruff: Entries is only used one time in the generate function below, I would just use it inside that function. Otherwise we keep this value around indefinitely?

miniscruff (changes requested):

lib/errors/index.ts
miniscruff: Nevermind, read the description... Did not know this, also kinda wish it was just a `readonly` prefix or something.

lib/errors/index.ts
miniscruff: Read the description in more detail...

lib/errors/index.ts
jbertran: The type is inferred from the return statement - the engine sees a map of `types.Name` to `ArsenalError` in the reduce, it just yells at us in the base case because `{}` can be almost anything, so we coerce it to `Errors`.

It needs to be in the class because that's the only place we want to allow calling the `ArsenalError` constructor - all other errors are either specific errors, or created through the `customizeDescription` call, which is in the class.

miniscruff: oh right, the private constructor, that's fine then.

lib/errors/types.ts
ghivert: That's a good point, `types.ts` was not used anymore, so it is now removed. 
The preferred way to check for an error type is to use directly `error.is.Internal` to avoid some parens and namespaces. It's also more similar to what we already have.

lib/errors/index.ts
ghivert: TypeScript usually remove the imports if we are only using types, but there's no guarantee. I'm used to import whatever and let the tree shaking do the work, but you're totally right, import type does not add any overhead at runtime (because types are stripped at compile-time).

lib/errors/index.ts
ghivert: Good point!

miniscruff (approved):

lib/errors/arsenalErrors.ts
```diff
@@ -1,760 +1,1033 @@
...
+export const AccessDenied: ErrorFormat = {
+    code: 403,
+    description: 'Access Denied',
+};
+
+export const AccessForbidden: ErrorFormat = {
+    code: 403,
+    description: 'Access Forbidden',
+};
+export const AccountProblem: ErrorFormat = {
```
francoisferrand, line 17: nit: missing blank line

lib/auth/Vault.js
```diff
@@ -1,4 +1,4 @@
-const errors = require('../errors');
+const errors = require('../errors').default;
```
francoisferrand, line 1: can't we stick with just importing errors simply:

const errors = require('../errors')

lib/errors/index.ts
```diff
@@ -1,89 +1,94 @@
...
-        this.description = desc;
+export class ArsenalError extends Error {
+    /** HTTP status code. Example: 401, 403, 500, ... */
+    #code: number;
+    /** Text description of the error. */
+    #description: string;
+    /** Type of the error. Belongs to errors.types. */
+    #type: types.Name;
+    /** Object used to determine the error type.
+     * Example: error.is.InternalError */
```
francoisferrand, line 31: what is the benefit of this approach, vs a simpler `error.is(InternalError)` ?

Granted, it looks similar to the previous code ; but since we need to change it anyway, we may as well use simpler approach (and avoid the meta-programming)...

lib/auth/Vault.js
ghivert: Unfortunately we can't due to ES Modules. 
We have to write things like this: `export default errors()` because we don't have alternatives to expose errors as an object. In TS, we would like to continue to import errors in writing `import errors from 'errors'`, which is possible by using the `export default`, and as such, using `require('errors').default` in CommonJS.

lib/errors/index.ts
ghivert: This is mainly in comparison with the two ways of writing. Today, we are doing this:

```javascript
function example() {
  myfun(function(err, res) {
    if (err && err.InternalError) {
      // Do some stuff with error.
    } else {
      // Do some stuff with the res.
    }
  });
}
```

The TS equivalent we can do, with proper type-checking is the proposed solution:

```typescript
// Here the export default is there, because we can’t do another way.
function example() {
  myfun(function(err, res) {
    if (err?.is?.InternalError) {
      // Do some stuff with error.
    } else {
      // Do some stuff with the res.
    }
  });
}
```

The advantage here is the low code change, and the fluent API. The object is self-discoverable, you can easily understand what it is even if you are not so much aware of the package. At first we tried a simpler solution like you described, but the best we can do is:

```javascript
import errors from 'errors';

function example() {
  myfun(function(err, res) {
    if (err?.is?.(errors.InternalError)) {
      // Do some stuff with error.
    } else {
      // Do some stuff with the res.
    }
  });
}
```

Besides the syntax for conditional function call (`object?.()`) which feels a little weird, we are forced to import the package errors and to have a function call, which feels less native than just accessing a property of the error. In the end the difference is not so important, and we can still benefit from the type-checking in both case, with the same type casts. It's more a matter of taste and style-guides/guidelines.

On the specific meta-programming part, `Proxy` are a commonly used pattern in such cases, but using meta-prog really depends on teams, so I don't really have an opinion on whether we should keep it or get rid of it. We mainly wanted to show the integration of meta-prog with TypeScript (because in some cases meta-prog is mandatory) and how we can leverage power of modern JS while having TS safety all the way through.

lib/auth/Vault.js
francoisferrand: I missed this was for JS only, and thus "temporary", while we have a cleaner way in typescript

francoisferrand (approved):

ghivert:
Bloop

ghivert:
Bloop

ghivert:
Beep boop

ghivert:
Boop

PR #1738, opened 2022-03-08
Feature/arsn 100 backport build system migration to jest ts

ghivert:
This PR is a cherry-pick backporting the changes of #1723 to development/7.10.

> Switch the test runner to Jest, and upgrade ESLint.
> The test runner is now Jest and the code coverage is Jest + Istanbul.
> All new test files should be written as `[filename].spec.js` or `[filename].spec.ts`.

This PR also introduces TypeScript in the test runner to support further development.
To validate the configuration, it only introduces one ts file: `index.ts` at the root of the folder.
This provides a fully-working Jest + TS configuration, making it a solid foundations for future.

To prove our implementation, besides the tests and functional tests of Arsenal, we also tried an integration into Vault, to check that dependants tests are passing too. The run can be found here: https://github.com/scality/Vault/actions/runs/2022097091

We also made the integration branch, which is here: #1757 and the corresponding run on Vault:
https://github.com/scality/Vault/actions/runs/2028943885

Everything is working! 🎉

**Resolved**
> Unfortunately, the test suite is not working, because of the CI. The CI is not able to build the Vault project with the dependencies because of some `yarn` or `npm` issues. You can still test the branch on your own on a Linux OS. 😊 It is under investigation with RELENG


jbertran:
Boop?

.github/workflows/tests.yaml
```diff
@@ -45,3 +46,31 @@ jobs:
...
+      with:
+        node-version: '16'
+        cache: yarn
+    - name: Install dependencies
+      run: yarn cache clean && yarn install --frozen-lockfile
+      continue-on-error: true # TODO ARSN-97 Remove it when no errors in TS
+    - name: Compile
+      run: yarn build
+      continue-on-error: true # TODO ARSN-97 Remove it when no errors in TS
+    - name: Upload artifacts
```
miniscruff, line 68: what are we uploading here and why? It looks like we are uploading the built arsenal project, are we then going to import it later or something?

.github/workflows/tests.yaml
```diff
@@ -45,3 +46,31 @@ jobs:
...
+      uses: scality/action-artifacts@v2
+      with:
+        url: https://artifacts.scality.net
+        user: ${{ secrets.ARTIFACTS_USER }}
+        password: ${{ secrets.ARTIFACTS_PASSWORD }}
+        source: ./build
+        method: upload
+      if: success()
```
miniscruff, line 76: I can't comment on an empty file, but do we need `.npmignore`?

index.ts
```diff
@@ -0,0 +1,142 @@
+export const auth = require('./lib/auth/auth');
```
miniscruff, line 1: Is there a way around this massive index file? Like some sort of auto export or export * thing? I have forgotten to add new files to this at least 3 times...

tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+    "compilerOptions": {
+        "target": "es6",
+        "module": "commonjs",
+        "rootDir": "./",
```
miniscruff, line 5: Would we be able to avoid a bit of extra importing if we set the rootDir as `./lib`?

.github/workflows/tests.yaml
ghivert: We are uploading the built Arsenal package, because it's not really possible to get the build from GitHub, because we're ignoring the build folder in the git repo. Storing the built is there to be able to depend from a commit hash directly.
I don't know the use case. @jbertran

.github/workflows/tests.yaml
ghivert: Sure, otherwise the resulting `build` folder is ignored by `npm`: `npm` is ignoring everything is listed in the `.gitignore` file. Adding a `.npmignore` file takes precedence over `.gitignore`.

index.ts
ghivert: We can surely do something about that in the future, but for now, the idea is to keep the same structure for this PR. Later, we could do something like `export * as auth from 'auth-folder'` for every sub-folder. I think it's cleaner to do this for every future package rewrite.

tsconfig.json
ghivert: Unfortunately considering the rootDir as `./lib` would ignore `index.ts`. This is not the desired behaviour as of now, but we could move the `index.ts` in the `lib` folder.

ghivert:
Every tests on CI are finally OK!

.github/workflows/tests.yaml
```diff
@@ -31,7 +31,8 @@ jobs:
         node-version: '16'
         cache: 'yarn'
     - name: install dependencies
-      run: yarn install --frozen-lockfile
+      run: yarn cache clean && yarn install --frozen-lockfile
```
francoisferrand: we should not clean the cache: however it makes no sense to use it.

We may also want to add `--prefer-offline` flag to yarn install, to prefer using cached packages when possible.

.github/workflows/tests.yaml
```diff
@@ -45,3 +46,31 @@ jobs:
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Install NodeJS
+      uses: actions/setup-node@v2
+      with:
+        node-version: '16'
+        cache: yarn
+    - name: Install dependencies
+      run: yarn cache clean && yarn install --frozen-lockfile
```
francoisferrand: same, should not remove cache

package.json
```diff
@@ -45,10 +45,11 @@
   },
   "devDependencies": {
     "@sinonjs/fake-timers": "^6.0.1",
-    "eslint": "2.13.1",
+    "eslint": "^8.10.0",
```
francoisferrand: why do we need to update eslint here ?
Seems like it should be a separate commit (or even PR)...

francoisferrand (changes requested):
I don't think we should relint everything: this creates lots of "noise" w/ regard to the real purpose of this PR, and generally will create merge conflicts and "break" git-blame. Thus it would be better not to include `Fix linting with correct indentation and trailing commas` commit.

.github/workflows/tests.yaml
miniscruff: Oh right, cause we don't use NPM

index.ts
miniscruff: Ok, that is what I would expect. Sounds good.

tsconfig.json
miniscruff: Ok so if we didn't use index.ts or placed the index in the lib folder then we could default to ./lib for the rootDir. That is all I wanted to know.

miniscruff (approved):
I agree on not including lint changes but will approve for now

francoisferrand (approved):

PR #1743, opened 2022-03-14
Feature/arsn 114 move in memory metadata backend 7 10

vrancurel:
For now proof-of-concept of moving cloudserver data and md backend back to Arsenal.


jbertran:
Hello, I know Bert-E says `7.10.15`, but it's a bug - see https://scality.atlassian.net/browse/BERTE-564.

I recommend switching to `7.10.14` on the ticket and asking for a `bypass_jira_check` once you're satisfied with the PR, so that we're tracking the right things :)

PR #1744, opened 2022-03-15
bugfix: ARSN-115 Fix listing algo returning phd

welansari:
In the metadata backend, PHD master is not created in v1 hence
it wasn't considered in the skipping method, in the Mongo implementation
however the PHD keys need to be taken into consideration as they are still
created


francoisferrand (approved):

jonathan-gramain (approved):

welansari:
ping

PR #1758, opened 2022-03-22
Bugfix: arsn 116 fix listing master returning phd

welansari:
When the last version of an object is deleted a placeholder master key is created and is kept for 15 seconds before it gets repaired.

To avoid listing the PHD keys we added a transform stream that replaces PHD master keys with the last version of that object

For ease of readability, the internal listing function was divided into two functions one for each bucket version format

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -16,6 +16,7 @@ const constants = require('../../../constants');
 const { reshapeExceptionError } = require('../../../errorUtils');
 const errors = require('../../../errors');
 const BucketInfo = require('../../../models/BucketInfo');
+const jsutil = require('../../../jsutil');
 
 const MongoClient = require('mongodb').MongoClient;
 const Uuid = require('uuid');
@@ -29,6 +30,8 @@ const MongoReadStream = require('./readStream');
 const MongoUtils = require('./utils');
 const Skip = require('../../../algos/list/skip');
 const MergeStream = require('../../../algos/stream/MergeStream');
+const { Transform } = require('stream');
+const { Version } = require('../../../versioning/Version');
 
 const { formatMasterKey, formatVersionKey } = require('./utils');
 
@@ -52,6 +55,8 @@ let uidCounter = 0;
 
 const BUCKET_VERSIONS = require('../../../versioning/constants')
     .VersioningConstants.BucketVersioningKeyFormat;
+const DB_PREFIXES = require('../../../versioning/constants')
+    .VersioningConstants.DbPrefixes;
 
 function generateVersionId(replicationGroupId) {
     // generate a unique number for each member of the nodejs cluster
@@ -1336,7 +1341,7 @@ class MongoClientInterface {
     }
 
     /**
-     * internal listing function
+     * internal listing function for buckets
      * @param {String} bucketName bucket name
      * @param {Object} params internal listing params
      * @param {Object} params.mainStreamParams internal listing param applied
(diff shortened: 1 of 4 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1336,7 +1340,69 @@ class MongoClientInterface {
...
+     * @return {undefined}
+     */
+    internalListObjectV0(bucketName, params, extension, log, cb) {
+        const c = this.getCollection(bucketName);
+        const stream = new MongoReadStream(c, params, params.mongifiedSearch);
+        const skip = new Skip({
+            extension,
+            gte: params.gte,
+        });
+        let cbDone = false;
```
ghost:
```suggestion
        const cbOnce = jsutil.once(callback);
```

Do you think we can use the builtin helper `once` here? It will manage for you the callback.
See https://github.com/scality/Arsenal/blob/development/8.1/lib/jsutil.js

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1336,7 +1340,69 @@ class MongoClientInterface {
...
+            newParams.gt = undefined;
+            newParams.gte = range;
+
+            // then continue listing the next key range
+            this.internalListObject(bucketName, newParams, extension, log, cb);
+        });
+        stream
+            .on('data', e => {
+                skip.filter(e);
+            })
```
ghost:
```suggestion
            .on('data', entry => {
                skip.filter(entry);
            })
```

It is easier to understand with full variable names

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1349,12 +1415,48 @@ class MongoClientInterface {
...
+            const resolvePhdKey = new Transform({
+                objectMode: true,
+                transform(obj, encoding, callback) {
+                    if (Version.isPHD(obj.value)) {
+                        const key = obj.key.slice(DB_PREFIXES.Master.length);
+                        getLatestVersion(c, key, BUCKET_VERSIONS.v1, log, (err, version) => {
+                            if (err) {
+                                log.error(
+                                    'internalListObjectV1: error while getting latest version of PHD key');
+                                callback(errors.InternalError);
```
ghost:
```suggestion
                                return callback(errors.InternalError);
```

I am not sure we want to run the function in case of an error, and return the latestVersion with an undefined value.

ghost (changes requested):

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: yeah it'll make the code a bit cleaner, good idea

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: yep i agree

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: IMO i think it's better to return an error instead of skipping the entry, we don't want to return a false listing result, i.e with missing objects

ghost: yes but without the return, you will run the following code in case of error:

```js
                            MongoUtils.unserialize(version);
                            // we keep the master key and only replace the value
                            const latestVersion = {
                                key: obj.key,
                                value: JSON.stringify(version),
                            };
                            callback(null, latestVersion);
```

Right? In this case, you will call the callback twice.

Edit: edited in the last commit, so all good!

ghost (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: oh i guess i just misunderstood what you said, but yeah i agree with adding the return

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1349,12 +1407,48 @@ class MongoClientInterface {
...
+                            return callback(null, latestVersion);
+                        });
+                    } else {
+                        callback(null, obj);
+                    }
+                },
+            });
+            // listing masters only (DelimiterMaster)
+            stream = new MongoReadStream(c, params.mainStreamParams, params.mongifiedSearch)
+                .pipe(resolvePhdKey);
```
francoisferrand: the change is actually "just" introducing the pipe(resolvePhdKey) here for V1, right ?

so instead of duplicating, could we not just do something like:

```
            stream = new MongoReadStream(c, params.mainStreamParams, params.mongifiedSearch);
            if (use_v1) {
                const resolvePhdKey = ...
                stream = stream.pipe(resolvePhdKey);
            }
```

vrancurel (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: IMO i think to make the code a bit easier to read it's better to separate them otherwise the function would just be too complex

francoisferrand (approved):

XinLi0207 (approved):

Kerkesni:
ping

PR #1769, opened 2022-03-31
improvement: ARSN-129 fix-critical-dependency-package-socketio

XinLi0207:
according to [CVE](https://github.com/advisories/GHSA-72mh-269x-7mh5), the security issue of xmlhttprequest-ssl has been fixed after v1.6.1, so we need to at least upgrade the 

socket.ioand socket.io-client packages to v2.4.x, I will try to bump it to the latest version 4.4.x and then release new arsenal and merge it into vault, vaultclient, uatpi etc, solve potential incompatibilities.


XinLi0207:
duplicated

PR #1770, opened 2022-03-31
bugfix: ARSN-129-upgrade-socketio-fix-critical

XinLi0207:
according to [CVE](https://github.com/advisories/GHSA-72mh-269x-7mh5), the security issue of xmlhttprequest-ssl has been fixed after v1.6.1, so we need to at least upgrade the socket.io and socket.io-client packages to v2.4.x, I will try to bump it to version 2.4.x in order to be safe and then release new arsenal and merge it into vault, vaultclient, uatpi, bucketclient etc, solve potential incompatibilities.

all related PRs:
https://github.com/scality/vaultclient/pull/393
https://github.com/scality/bucketclient/pull/243
https://github.com/scality/utapi/pull/1229
https://github.com/scality/Vault/pull/2021


ghost (approved):

KillianG (approved):

anurag4DSB (changes requested):
This is a major bump, we should analyze the change log to find breaking changes between 2.3.0 and 4.4.1
Do we have any breaking changes?

XinLi0207:
> This is a major bump, we should analyze the change log to find breaking changes between 2.3.0 and 4.4.1 Do we have any breaking changes?

the breaking changes are described in this doc 
https://socket.io/docs/v3/migrating-from-2-x-to-3-0/
https://socket.io/docs/v3/migrating-from-3-x-to-4-0/

breaking changes are these:

- [io.to() is now immutable](https://socket.io/docs/v3/migrating-from-3-x-to-4-0/#io-to-is-now-immutable)
- [wsEngine option](https://socket.io/docs/v3/migrating-from-3-x-to-4-0/#wsEngine-option)


we didn't use these features in arsenal, so that's why I think it should be ok to bump a major version, but maybe it's not a good idea to bump a major version directly, it might introduce some uncertainties that we didn't expect

anurag4DSB (approved):

PR #1801, opened 2022-04-19
ARSN-153 Improve errors API

ghivert:
We found some improvement for errors. Right now, the `error.is` object is empty, and metaprogramming takes care of everything. In debugging it can be weird to not have a filled object like the type indicates.

Currenty we got this.
```js
const error = errors.InternalError;
console.log(error.is); // {}
console.log(Object.keys(error.is)); // []
```

After the PR, we got this.
```js
const error = errors.InternalError;
console.log(error.is); // { InternalError: true, InvalidURI: false, NoSuchBucket: false, ... }
console.log(Object.keys(error.is)); // ['InternalError', 'InvalidURI', 'NoSuchBucket', ...]
```

We ensure the `Is` object cannot be modified through `Object.freeze` and we still take care of everything in `Is` through metaprog to make sure the function correctly returns false everytime.

### DIFF
--- a/lib/errors/index.ts
+++ b/lib/errors/index.ts
@@ -10,14 +10,23 @@ export type Is = { [_ in Name]: boolean };
 /** Mapping of all possible Errors. It has the format { [Name]: Error } */
 export type Errors = { [_ in Name]: ArsenalError };
 
+// This object is reused constantly through createIs, we store it there
+// to avoid recomputation.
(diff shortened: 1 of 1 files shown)


ghivert:
Boop

ghivert:
Boop again

jbertran (approved):

miniscruff (approved):

PR #1812, opened 2022-04-20
Improvement/arsn 157 short i ds

watsaqat:


lib/versioning/VersionID.js
```diff
@@ -116,12 +116,17 @@ function generateVersionId(info, replicationGroupId) {
     lastSeq = (lastTimestamp === ts) ? lastSeq + 1 : 0;
     lastTimestamp = ts;
 
+    // if USE_HEX_VERSION_ID is set, info is used. By default, it is not used.
+    if (process.env.USE_HEX_VERSION_ID === 'no' || !process.env.USE_HEX_VERSION_ID) {
```
watsaqat: better name for this?

lib/versioning/VersionID.js
```diff
@@ -221,13 +226,13 @@ const ENC_TYPE_BASE62 = 1; // new (tiny) encoding
...
  * @return {string} - the encoded versionId
  */
-function encode(str, encType = ENC_TYPE_HEX) {
-    if (encType === ENC_TYPE_BASE62) {
-        return base62Encode(str);
-    }
+function encode(str) {
+    // default format without 'info' field will always be 27 characters
+    if (str.length === 27) {
+        return padRight(base62Encode(str), ' '.repeat(32));
```
watsaqat: any potential issues with padding right with spaces?

tests/unit/versioning/VersionID.spec.js
```diff
@@ -9,83 +10,124 @@ function randkey(length) {
...
+                }
+            }
+        });
 
-    it('error case: when invalid base62 key part 3 has invalid base62 character', () => {
-        const invalidBase62VersionId = 'aJLWKz4Ko9IjBBgXKj5KQT.G9UHv0g7P';
-        const decoded = VID.base62Decode(invalidBase62VersionId);
-        assert(decoded instanceof Error);
-    });
+        it('simple base62 version test', () => {
```
watsaqat, line 73: base62Encode/Decode and hexEncode/Decode don't depend on USE_HEX_VERSION_ID, so they should be moved to their own section as was previously done.

lib/versioning/VersionID.js
jonathan-gramain: as discussed, I suggest checking for `process.env.S3_VERSION_ID_ENCODING_TYPE=hex` specifically

tests/unit/versioning/VersionID.spec.js
```diff
@@ -9,83 +10,124 @@ function randkey(length) {
...
+function generateRandomVIDs(count) {
     const vids = Array(count).fill(null);
     for (let i = 0; i < count; i++) {
         vids[i] = VID.generateVersionId(randkey(15), 'PARIS');
     }
     process.env.VID_CRYPTO_PASSWORD = randkey(64);
+    return vids;
+}
+
+const count = 1000;
```
vrancurel: is it possible to have 1,000,000 ?

lib/versioning/VersionID.js
```diff
@@ -202,8 +215,11 @@ function base62Decode(str) {
         const enc2 = str.substring(start, start + B62V_EPAD.length);
         const orig2 = base62Integer.decode(enc2);
         start += B62V_EPAD.length;
-        const enc3 = str.substring(start);
+        let enc3 = str.substring(start);
+        // take off null bytes padding
+        enc3 = enc3.replaceAll('\u0000', '');
```
vrancurel: is there a better way to trim the null bytes ?

vrancurel (changes requested):

vrancurel:
a few details but otherwise it looks good.

lib/versioning/VersionID.js
```diff
@@ -202,8 +215,12 @@ function base62Decode(str) {
         const enc2 = str.substring(start, start + B62V_EPAD.length);
         const orig2 = base62Integer.decode(enc2);
         start += B62V_EPAD.length;
-        const enc3 = str.substring(start);
+        let enc3 = str.substring(start);
+        // take off null bytes padding only from the end
+        const re = new RegExp('\u0000+$'); // eslint-disable-line
```
vrancurel: isn't it more efficient to do:
`const i = b.indexOf(0x00);
  const nb = b.slice(0, i);`

vrancurel (changes requested):

lib/versioning/VersionID.js
watsaqat: as long as we are guaranteed to never have a null byte besides the padding at the end, we can do this. Benchmark is about 2x speedup in favor of slice

watsaqat: https://measurethat.net/Benchmarks/Show/18405/0/regex-vs-slice
I tried binary search too lol, but the strings are too short to take advantage of over a simple while loop

vrancurel (dismissed):

lib/versioning/VersionID.js
vrancurel: > as long as we are guaranteed to never have a null byte besides the padding at the end, we can do this. Benchmark is about 2x speedup in favor of slice

why would we have a null byte ?

lib/versioning/VersionID.js
```diff
@@ -116,12 +116,19 @@ function generateVersionId(info, replicationGroupId) {
     lastSeq = (lastTimestamp === ts) ? lastSeq + 1 : 0;
     lastTimestamp = ts;
 
+    // if USE_HEX_VERSION_ID is set, info is used. By default, it is not used.
+    if (process.env.S3_VERSION_ID_ENCODING_TYPE === 'hex') {
+        // info field stays as is
+    } else {
+        info = ''; // eslint-disable-line
+    }
```
miniscruff, line 125:
```suggestion
    if (process.env.S3_VERSION_ID_ENCODING_TYPE !== 'hex') {
        info = ''; // eslint-disable-line
    }
```
Having an empty if looks weird to me, I would flip the condition

lib/versioning/VersionID.js
```diff
@@ -202,8 +215,19 @@ function base62Decode(str) {
         const enc2 = str.substring(start, start + B62V_EPAD.length);
         const orig2 = base62Integer.decode(enc2);
         start += B62V_EPAD.length;
-        const enc3 = str.substring(start);
+        let enc3 = str.substring(start);
+        // take off null bytes padding only from the end
```
miniscruff: You can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/lastIndexOf to find the last character you need. TrimEnd might work but doesn't seem like it unless \u0000 counts as white space.

tests/unit/versioning/VersionID.spec.js
```diff
@@ -9,83 +10,124 @@ function randkey(length) {
...
+        const vids = generateRandomVIDs(count);
 
-    it('sorted in reversed chronological and alphabetical order', () => {
-        for (let i = 0; i < count; i++) {
-            if (i !== 0) {
-                assert(vids[i - 1] > vids[i],
-                    'previous VersionID is higher than its next');
+        it('sorted in reversed chronological and alphabetical order', () => {
+            for (let i = 0; i < count; i++) {
+                if (i !== 0) {
```
miniscruff: You can just start at 1 to avoid this extra check

tests/unit/versioning/VersionID.spec.js
```diff
@@ -9,83 +10,124 @@ function randkey(length) {
...
-    });
+        });
 
-    // nodejs 10 no longer returns error for non-hex string versionIds
-    it.skip('should return error decoding non-hex string versionIds', () => {
-        const encoded = vids.map(vid => VID.hexEncode(vid));
-        const decoded = encoded.map(vid => VID.hexDecode(`${vid}foo`));
-        decoded.forEach(result => assert(result instanceof Error));
-    });
+        // nodejs 10 no longer returns error for non-hex string versionIds
```
miniscruff, line 37: Can probably remove this since it was a node 10 thing and we are on node 16 now

tests/unit/versioning/VersionID.spec.js
```diff
@@ -9,83 +10,124 @@ function randkey(length) {
...
-    it('simple base62 version test', () => {
-        const vid = '98376906954349999999RG001  145.20.5';
-        const encoded = VID.base62Encode(vid);
-        assert.strictEqual(encoded, 'aJLWKz4Ko9IjBBgXKj5KQT2G9UHv0g7P');
-        const decoded = VID.base62Decode(encoded);
-        assert.strictEqual(vid, decoded);
-    });
+        it('should encode and decode correctly with legacy format', () => {
+            const encoded = vids.map(vid => VID.encode(vid));
+            const decoded = encoded.map(vid => VID.decode(vid));
```
miniscruff:
```suggestion
            const encoded = vids.map(VID.encode);
            const decoded = encoded.map(VID.decode);
```
Does this work?

tests/unit/versioning/VersionID.spec.js
```diff
@@ -9,83 +10,124 @@ function randkey(length) {
...
 
-    it('base62 version test with smaller part3 - 2', () => {
-        const vid = '98376906950000099999R1x';
-        const encoded = VID.base62Encode(vid);
-        const decoded = VID.base62Decode(encoded);
-        assert.strictEqual(vid, decoded);
-    });
+        it('sorted in reversed chronological and alphabetical order', () => {
+            for (let i = 0; i < count; i++) {
+                if (i !== 0) {
```
miniscruff: again just start at 1

miniscruff (changes requested):

lib/versioning/VersionID.js
watsaqat: we're not trying to find the last index with null but the first index with null. TrimEnd doesn't work unfortunately.

miniscruff: oh well the first index would just be `indexOf` unless you specifically needed to start from the end for some reason.

lib/versioning/VersionID.js
watsaqat: I wrote it this way to be explicit about what's happening and to avoid a negative predicate check. I agree it looks weird, but I feel the behavior is obfuscated the other way.

tests/unit/versioning/VersionID.spec.js
watsaqat: I didn't write this test originally, just moved it around, so I'd rather leave it in case the original author wanted to revisit it/ modify it. I agree we probably don't need it but whoever added the skip intentionally didn't delete it and skipped it instead.

tests/unit/versioning/VersionID.spec.js
watsaqat: with branch prediction this is going to be barely noticeable 
https://measurethat.net/Benchmarks/Show/18409/0/branch-prediction
I don't see the need for micro-optimizing unit tests in general, but that's just my opinion.

tests/unit/versioning/VersionID.spec.js
watsaqat: I'm trying to find the es-lint rule for this. Do you have a link? We don't enforce it right now though and I'm not sure if we should be introducing linting rules in a random pr.

lib/versioning/VersionID.js
watsaqat: oh I was thinking in case we add a 4th field in the future and use it a separator or something. But we would just use the offsets so just disregard that.

lib/versioning/VersionID.js
watsaqat: there's a comment above with the benchmarks.

watsaqat: I originally did it with just regex but Vianney suggested to optimize

lib/versioning/VersionID.js
```diff
@@ -181,9 +188,15 @@ function base62Encode(str) {
...
-        enc3;
+    const padded1 = (B62V_EPAD + enc1).slice(-B62V_EPAD.length);
+    const padded2 = (B62V_EPAD + enc2).slice(-B62V_EPAD.length);
+    const enc3String = enc3.toString();
+    const totalLength = padded1.toString().length +
+        padded2.toString().length +
+        enc3String.length;
+    const enc3PaddingNeeded = 32 - totalLength + enc3String.length;
+
+    return padded1 + padded2 + enc3.toString().padEnd(enc3PaddingNeeded, '\u0000');
```
jonathan-gramain: I don't think we can return null bytes in the encoded version: it should only contain base62 characters. Even if we are allowed other characters than base62 (letters and digits) we have to consider what the application would do with such characters, it's surely calling for issues to include nonprintable characters in the exposed version ID.

jonathan-gramain (changes requested):

tests/unit/versioning/VersionID.spec.js
miniscruff: This is a linting rule? I was just suggesting it to pass the function directly instead of passing an inline function for the same thing. Not a big deal either way.

tests/unit/versioning/VersionID.spec.js
miniscruff: My comment was not about optimization it was about reducing a tab and readability. I rarely try and optimize tests unless its like super obvious or less code in general.

lib/versioning/VersionID.js
watsaqat: I think the only way then is to rely on the fact that the unpadded string should be 22 bytes and then add noise. Then decoding is a matter of stripping the last 10 bytes

lib/versioning/VersionID.js
miniscruff: oh yea missed it, sorry

tests/unit/versioning/VersionID.spec.js
miniscruff: hmm fair, maybe we can propose someone else look at it. It will just stay around indefinitely otherwise is my guess.

lib/versioning/VersionID.js
watsaqat: @vrancurel thoughts on this?

vrancurel: oh right I though we were padding null bytes in the decoded version. Yes the encoded key shall only return base62 characters.

watsaqat:
putting back in draft

vrancurel (approved):

lib/versioning/VersionID.js
```diff
@@ -202,8 +211,19 @@ function base62Decode(str) {
...
+        // take off prefix 0s which represent null bytes
+        let idx = 0;
+        while (idx < enc3.length) {
+            if (enc3[idx] === '0') {
+                idx++;
+            } else {
+                break;
+            }
+        }
+        enc3 = enc3.slice(idx);
```
jonathan-gramain, line 225: In JS this should be relatively efficient and simplifies code a bit:

```suggestion
        enc3 = enc3.replace(/^0*/, '');
```

tests/unit/versioning/VersionID.spec.js
jonathan-gramain:
```suggestion
            for (let i = 1; i < count; i++) {
```

tests/unit/versioning/VersionID.spec.js
```diff
@@ -9,83 +10,124 @@ function randkey(length) {
...
-    it('should encode and decode base62 versionIds', () => {
-        const encoded = vids.map(vid => VID.base62Encode(vid));
-        const decoded = encoded.map(vid => VID.base62Decode(vid));
-        assert.strictEqual(vids.length, count);
-        assert.deepStrictEqual(vids, decoded);
+        it('should encode and decode correctly with new 32 byte format', () => {
+            const encoded = vids.map(vid => VID.encode(vid));
+            const decoded = encoded.map(vid => VID.decode(vid));
+            assert.strictEqual(vids.every(x => x.length === 27), true);
+            assert.strictEqual(encoded.every(x => x.length === 32), true);
```
jonathan-gramain:
```suggestion
            assert(vids.every(x => x.length === 27));
            assert(encoded.every(x => x.length === 32));
```

lib/versioning/VersionID.js
```diff
@@ -240,14 +260,19 @@ function encode(str, encType = ENC_TYPE_HEX) {
  * @return {(string|Error)} - the decoded versionId or an error
  */
 function decode(str) {
-    if (str.length < 40) {
+    // default format is exactly 32 characters when encoded
+    if (str.length === 32) {
         return base62Decode(str);
```
jonathan-gramain: I'd suggest adding an assert to check the decoded size, because for such an important invariant, in the unlikely case it would be broken at some point it seems better to crash than to store invalid version IDs with the risk of corruption:

```suggestion
        const decoded = base62Decode(str);
        assert.strictEqual(decoded.length, 27);
        return decoded;
```

tests/unit/versioning/VersionID.spec.js
```diff
@@ -9,83 +10,124 @@ function randkey(length) {
...
 }
 
-describe('test generating versionIds', () => {
-    const count = 1000;
+function generateRandomVIDs(count) {
     const vids = Array(count).fill(null);
     for (let i = 0; i < count; i++) {
         vids[i] = VID.generateVersionId(randkey(15), 'PARIS');
     }
     process.env.VID_CRYPTO_PASSWORD = randkey(64);
```
jonathan-gramain: I'd put this line out of the function since it does not look related

tests/unit/versioning/VersionID.spec.js
```diff
@@ -9,83 +10,124 @@ function randkey(length) {
...
+        env.S3_VERSION_ID_ENCODING_TYPE = 'base62';
+        const vids = generateRandomVIDs(count);
 
-    it('base62 version test with smaller part3 - 2', () => {
-        const vid = '98376906950000099999R1x';
-        const encoded = VID.base62Encode(vid);
-        const decoded = VID.base62Decode(encoded);
-        assert.strictEqual(vid, decoded);
-    });
+        it('sorted in reversed chronological and alphabetical order', () => {
```
jonathan-gramain, line 65: Woulnd't this test better be done outside of the encoding scheme `describe` blocks, only once? Because I don't think it depends on the encoding scheme, it's just checking that internal version IDs are properly ordered without encoding/decoding anything.

tests/unit/versioning/VersionID.spec.js
```diff
@@ -9,83 +10,124 @@ function randkey(length) {
...
     process.env.VID_CRYPTO_PASSWORD = randkey(64);
+    return vids;
+}
+
+const count = 1000000;
+
+describe('test generating versionIds', () => {
+    describe('legaxy hex encoding', () => {
+        env.S3_VERSION_ID_ENCODING_TYPE = 'hex';
+        const vids = generateRandomVIDs(count);
```
jonathan-gramain, line 28: Linked with my other comment, we may also move the vid generation outside of the encoding-specific test block, after `const count = 1000000`

tests/unit/versioning/VersionID.spec.js
```diff
@@ -9,83 +10,124 @@ function randkey(length) {
...
-    it('should encode and decode versionIds', () => {
-        const encoded = vids.map(vid => VID.hexEncode(vid));
-        const decoded = encoded.map(vid => VID.hexDecode(vid));
-        assert.strictEqual(vids.length, count);
-        assert.deepStrictEqual(vids, decoded);
-    });
+        it('should encode and decode versionIds', () => {
+            const encoded = vids.map(vid => VID.hexEncode(vid));
+            const decoded = encoded.map(vid => VID.hexDecode(vid));
+            assert.strictEqual(vids.length, count);
```
jonathan-gramain: You could get rid of this line as it does not check anything more than the next line and only checks that the original `vids` array indeed has `count` versions in it (so testing the test case in a way)

jonathan-gramain (approved):
LGTM with some suggestions

tests/unit/versioning/VersionID.spec.js
watsaqat: the env var `env.S3_VERSION_ID_ENCODING_TYPE = 'hex';` needs to be set before the run. Otherwise the info field gets chopped off by default.

tests/unit/versioning/VersionID.spec.js
watsaqat: I think its better to explicitly show that the new format won't break ordering in the spec

watsaqat: its only obvious to us that ordering is preserved because we know exactly what its doing (just chopping off the info field if there is one) in the new format.

jonathan-gramain: Yes, it makes sense since it is removing the info, I did not get this at first.

lib/versioning/VersionID.js
```diff
@@ -116,12 +117,19 @@ function generateVersionId(info, replicationGroupId) {
     lastSeq = (lastTimestamp === ts) ? lastSeq + 1 : 0;
     lastTimestamp = ts;
 
+    // if USE_HEX_VERSION_ID is set, info is used. By default, it is not used.
```
jonathan-gramain:
```suggestion
    // if S3_VERSION_ID_ENCODING_TYPE is "hex", info is used. By default, it is not used.
```

jonathan-gramain (approved):

lib/versioning/VersionID.js
watsaqat: leaving as is for a bit of additional perf: https://measurethat.net/Benchmarks/Show/18405/0/regex-vs-slice

miniscruff (approved):
approving just curious on one point

PR #1813, opened 2022-04-21
feature: ARSN-160 add support for invalid arguments in errors

welansari:
Issue: [ARSN-160](https://scality.atlassian.net/browse/ARSN-160)
This will be used [here](https://github.com/scality/cloudserver/blob/ffd1fcd2c173b77d716567fc8b412af0484b081f/lib/api/apiUtils/bucket/getNotificationConfiguration.js#L18-L22), where for each invalid destination in the target we should add an invalid argument. This is done to match the AWS's error

### DIFF
--- a/lib/errors/index.ts
+++ b/lib/errors/index.ts
@@ -2,7 +2,7 @@ import type { ServerResponse } from 'http';
 import * as rawErrors from './arsenalErrors';
 
 /** All possible errors names. */
-export type Name = keyof typeof rawErrors
+export type Name = keyof typeof rawErrors;
 /** Object containing all errors names. It has the format { [Name]: "Name" } */
 export type Names = { [Name_ in Name]: Name_ };
 /** Mapping used to determine an error type. It has the format { [Name]: boolean } */
@@ -13,7 +13,7 @@ export type Errors = { [_ in Name]: ArsenalError };
 // This object is reused constantly through createIs, we store it there
 // to avoid recomputation.
 const isBase = Object.fromEntries(
-    Object.keys(rawErrors).map(key => [key, false])
+    Object.keys(rawErrors).map((key) => [key, false])
 ) as Is;
 
 // This contains some metaprog. Be careful.
@@ -25,7 +25,7 @@ const isBase = Object.fromEntries(
 // the Proxy will return false.
 const createIs = (type: Name): Is => {
     const get = (is: Is, value: string | symbol) => is[value] ?? false;
-    const final = Object.freeze({ ...isBase, [type]: true })
+    const final = Object.freeze({ ...isBase, [type]: true });
     return new Proxy(final, { get });
 };
 
@@ -39,13 +39,18 @@ export class ArsenalError extends Error {
     /** Object used to determine the error type.
      * Example: error.is.InternalError */
     #is: Is;
+    /** A map of error metadata (can be extra fields
+     * that only show in debug mode) */
+    #metadata: Map<string, Object[]>;
 
-    private constructor(type: Name, code: number, description: string) {
+    private constructor(type: Name, code: number, description: string,
+        metadata?: Map<string, Object[]>) {
         super(type);
         this.#code = code;
         this.#description = description;
         this.#type = type;
         this.#is = createIs(type);
+        this.#metadata = metadata ?? new Map<string, Object[]>();
     }
 
     /** Output the error as a JSON string */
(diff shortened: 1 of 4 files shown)


jbertran (commented):
I'm not sure modifying the base type for every error is the way forward here. It's always possible to define a contract between Arsenal and Cloudserver like receiving an (error, error detail list) tuple, but maybe there's a finer way of adding a field just for InvalidArgument.

Maybe @ghivert has seen this pattern before?

KillianG (approved):

lib/errors/index.ts
```diff
@@ -39,13 +39,18 @@ export class ArsenalError extends Error {
     /** Object used to determine the error type.
      * Example: error.is.InternalError */
     #is: Is;
+    /** A map of error metadata (can be extra fields
+     * that only show in debug mode) */
+    #metadata: Map<string, Array<Object>>;
```
ghivert: I usually try to write arrays as `Object[]` instead of `Array<Object>`, in case where type can be inlined. (I don't do it for types spanning accros multiple lines.)
It's mainly a matter of style, and would be good to have an opinion about that to avoid commits changing style later. @jbertran @Kerkesni any opinion?

lib/errors/index.ts
```diff
@@ -90,7 +95,22 @@ export class ArsenalError extends Error {
     customizeDescription(description: string): ArsenalError {
         const type = this.#type;
         const code = this.#code;
-        return new ArsenalError(type, code, description);
+        const metadata = this.#metadata;
+        const err = new ArsenalError(type, code, description, metadata);
```
ghivert, line 99: Because Maps are pointers and are subject to state modifications, it would be a huge improvement to duplicate the `Map` structure in the new error, just like we are duplicating the error. Pure code tends to avoid bugs. In fact, doing something like this:

```javascript
const foo = () => {
  const error = errors.InternalError
  const newError = error.customizeDescription('bla')
  newError.addMetadataEntry('bar', 'baz')
  error.medatata.get('bar') // ==> 'baz'
  // We probably want error to not be touched.
}
```

We could probably solve the problem by just doing `const metadata = new Map(this.#metadata)`.

lib/errors/index.ts
```diff
@@ -90,7 +95,22 @@ export class ArsenalError extends Error {
...
+    /** Clone the error with a new metadata field */
+    addMetadataEntry(key: string, value: Array<Object>): ArsenalError {
+        this.#metadata.set(key, value);
+        const type = this.#type;
+        const code = this.#code;
+        const description = this.#description;
+        const metadata = this.#metadata;
+        const err = new ArsenalError(type, code, description, metadata);
+        err.stack = this.stack;
+        return err;
```
ghivert, line 113: I get your point, but you're not duplicating the `Map` here unfortunately… 😕

tests/unit/errors.spec.ts
```diff
@@ -22,6 +22,14 @@ describe('Errors: ', () => {
...
     });
 
+    it('should allow additional metadata fields', () => {
+        const error = errors.InvalidArgument;
+        const custom = error.addMetadataEntry('fieldName', [{ key: 'value' }]);
+        expect(custom.metadata.get('fieldName')).toBeTruthy();
+        expect(custom.metadata.get('fieldName')).toEqual([{ key: 'value' }]);
+        expect(custom.is).toHaveProperty('InvalidArgument', true);
+    });
+
```
ghivert, line 40: We probably should check that `error` is not modified also.

ghivert (changes requested):

lib/errors/index.ts
jbertran: To me, `Type[]` seems less verbose, but still quite obvious, so I'm in favour of pushing in that direction

welansari: yeah i agree, it's cleaner that way 🙂

tests/unit/errors.spec.ts
ghivert: To be more precise, I would add something like `expect(error.metadata.get('fieldName')).toBeNull()`.

tests/unit/errors.spec.ts
```diff
@@ -22,6 +22,23 @@ describe('Errors: ', () => {
         expect(custom.is).toHaveProperty('NoSuchEntity', true);
     });
 
+    it('should allow additional metadata fields', () => {
+        const error = errors.InvalidArgument;
+        const originDescription = error.description;
+        const origincode = error.code;
+        const origintype = error.type;
+        const custom = error.addMetadataEntry('fieldName', [{ key: 'value' }]);
+        expect(custom.metadata.get('fieldName')).toBeTruthy();
```
ghivert: That line is useless thanks to Jest and the next line. :)

tests/unit/errors.spec.ts
```diff
@@ -22,6 +22,23 @@ describe('Errors: ', () => {
...
+        expect(custom.metadata.get('fieldName')).toEqual([{ key: 'value' }]);
+        // Other properties should not change
+        expect(custom.is).toHaveProperty('InvalidArgument', true);
```
ghivert: `expect(custom.is[origintype]).toBeTruthy()` a little more precise ? Not sure about that one, do as you feel it right.

ghivert (approved):
Main modifications are there, mainly style details remaining, so approving.
Feel free to includes the comments or not and close them. 😁

PR #1815, opened 2022-04-21
Dont merge, just creating a PR to see my changes easier

miniscruff:
Work to fix cloudserver issues that come from arsenal, will remake with a proper jira ticket.

### DIFF
--- a/lib/network/rpc/utils.js
+++ b/lib/network/rpc/utils.js
@@ -40,7 +40,9 @@ module.exports.reconstructError = function reconstructError(err) {
         return err;
     }
     const reconstructedErr = new Error(err.message);
-
+    reconstructedErr.is = {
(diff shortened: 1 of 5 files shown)


lib/network/rpc/utils.js
```diff
@@ -40,7 +40,9 @@ module.exports.reconstructError = function reconstructError(err) {
         return err;
     }
     const reconstructedErr = new Error(err.message);
-
+    reconstructedErr.is = {
```
miniscruff, line 43: @jbertran I added this so we can use `is` when dealing with these errors, it feels kinda awkward as they are not really arsenal errors but it also felt weird to not have `.is` in the bucket file interface.

miniscruff:
Moved work to the proper ticket and PR

PR #1839, opened 2022-04-27
ARSN-161 Type network folder

ghivert:


miniscruff (approved):

miniscruff (approved):

lib/network/RoundRobin.ts
```diff
@@ -17,19 +18,29 @@ function shuffle(array) {
...
+    hostIndex: number;
+    pickCount: number;
+
     /**
      * @constructor
      * @param {object[]|string[]} hostsList - list of hosts to query
      *   in round-robin fashion.
      * @param {string} hostsList[].host - host name or IP address
-     * @param {number} [hostsList[].port] - port number to contact
+     * @param {number} [hostsList[].port] - port number to contacts
```
ghost: Nit:
```suggestion
     * @param {number} [hostsList[].port] - port numbers to contact
```

lib/network/RoundRobin.ts
```diff
@@ -55,8 +69,8 @@ class RoundRobin {
         this.pickCount = 0;
     }
 
-    _validateHostObj(hostItem) {
-        const hostItemObj = {};
+    _validateHostObj(hostItem: string | { host: string; port: string }): { host: string; port?: number } {
```
ghost, line 72: Why not using an interface to avoid repeating this custom type everywhere?

lib/network/http/utils.ts
```diff
@@ -72,8 +79,8 @@ function getByteRangeFromSpec(rangeSpec, objectSize) {
     if (rangeSpec.start < objectSize) {
         // test is false if end is undefined
         return { range: [rangeSpec.start,
-            (rangeSpec.end < objectSize ?
-                rangeSpec.end : objectSize - 1)] };
+            (rangeSpec.end && rangeSpec.end < objectSize ?
+            rangeSpec.end : objectSize - 1)] };
```
ghost: Just wondering if we can make this more clear with parenthesis?

```suggestion
            ((rangeSpec.end && rangeSpec.end < objectSize) ?
            rangeSpec.end : objectSize - 1)] };
```

lib/network/kmip/Client.ts
```diff
@@ -483,19 +493,20 @@ class Client {
...
-        masterKeyId,
-        plainTextDataKey,
-        logger,
-        cb) {
+    cipherDataKey(
+        cryptoScheme: number,
+        masterKeyId: string,
+        plainTextDataKey: Buffer,
+        logger: werelogs.Logger,
+        cb: any,
```
ghost, line 508: Maybe being more specific on the callbacks type is better? (maybe not, so it's just a suggestion)

```suggestion
        cb: Function,
```

lib/network/kmip/codec/ttlv.ts
```diff
@@ -369,7 +369,7 @@ function TTLVCodec() {
 
 
     /* Public Methods Definition */
-
+    // @ts-ignore
```
ghost, line 372: Instead of ts-ignores, if this is a rule we don't want, why not just adding a one liner comment at the top of the file or in the `tsconfig.json` file?

lib/network/rpc/sio-stream.ts
```diff
@@ -1,29 +1,34 @@
...
 
-const uuid = require('uuid');
-const stream = require('stream');
-const debug = require('debug')('sio-stream');
-const assert = require('assert');
-const async = require('async');
-
-const flattenError = require('./utils').flattenError;
-const reconstructError = require('./utils').reconstructError;
+const debug = debug_('sio-stream')
```
ghost: Nit:

```suggestion
const debug = debug_('sio-stream');
```

ghost (commented):
Some comments, feel free to ignore some if they are not relevant (it's the first time I am reviewing the typescript migration PRs :smile: )

Feel free to ping me on slack once you are done with the potential changes, and I will approve.

lib/network/RoundRobin.ts
ghivert: Well see, a declaration was useless

lib/network/http/utils.ts
ghivert: Sure

lib/network/kmip/Client.ts
ghivert: Usually we are writing callbacks as `(error: Error | null, data: YourDataType) => void` and this add type-checking. Some type-checking still needs to be added, so I prefer to let it as `any` to really indicates "This is not properly typed".

lib/network/kmip/codec/ttlv.ts
ghivert: We want someone else to can find that and fix it. We don't want to get rid of the rule, only placing a marker for others to know where they need to rewrite something. 🙂

ghost (approved):

PR #1841, opened 2022-04-28
feature: ARSN-172 fix invalid timestamp in the oplog entries

welansari:
Fixed the timestamp from giving `"Invalid Date"` as a value


miniscruff (approved):
nitpick but I think the branch and ticket should be bugfix and bug respectively and not feature.

miniscruff (changes requested):
Any change in lib/storage needs to be done in 7x and 8x as they are moving to Armory. They are meant to be identical.

PR #1846, opened 2022-04-29
ARSN-175 Restores old behaviors of errors

ghivert:


lib/errors/index.ts
```diff
@@ -16,6 +16,10 @@ const isBase = Object.fromEntries(
     Object.keys(rawErrors).map(key => [key, false])
 ) as Is;
 
+// This allows to conditionally add the old behavior of errors to properly
+//   test migration. Activate CI tests with MIGRATED=true yarn test.
+const notMigrated = (process.env.MIGRATED ?? 'false') === 'false'
```
miniscruff: I feel like migrated is a little too vague. Maybe something like "ErrorIsSupported" or something.

lib/network/rpc/utils.ts
```diff
@@ -55,6 +55,9 @@ export function reconstructError(err: Error) {
     }
 
     const reconstructedErr = new Error(err.message);
+    // This restores the old behavior of errors. This should be removed as soon
```
miniscruff, line 58: We should link a Jira ticket to this so it is easy to spot / search

miniscruff (changes requested):

lib/network/rpc/utils.ts
jbertran: Done below the comment

lib/errors/index.ts
jbertran: Updated. Please check my boolean logic, I'm easily confused on this for some reason :grimacing:

nicolas2bert (approved):

lib/network/rpc/utils.ts
```diff
@@ -55,6 +55,9 @@ export function reconstructError(err: Error) {
...
+    // as all dependent codebases have been migrated to `is` accessors (ARSN-176).
+    reconstructedErr[err.message] = true;
```
francoisferrand, line 60: should this be conditioned on `allowUnsafeErrComp` ?

lib/errors/index.ts
miniscruff: Looks good now, much easier to make flags like this in the positive term like isX or allowY which might be why you were confused.

miniscruff (approved):

lib/network/rpc/utils.ts
jbertran: Indeed. Done.

PR #1849, opened 2022-05-02
test cold storage changes for arsenal

ghost:
Will be closed, branch used for test only. Didn't keep fujifilm authorship either, but will be done in the actual implementation.

### DIFF
--- a/lib/models/ObjectMD.js
+++ b/lib/models/ObjectMD.js
@@ -106,7 +106,9 @@ class ObjectMD {
             'x-amz-restore': {
                 'ongoing-request': false,
             },
-            'archive': {},
+            'archive': {
+                archiveInfo: {},
+            },
             'x-amz-server-side-encryption': '',
             'x-amz-server-side-encryption-aws-kms-key-id': '',
             'x-amz-server-side-encryption-customer-algorithm': '',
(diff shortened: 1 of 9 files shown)


lib/models/ObjectMD.js
```diff
@@ -1194,6 +1200,54 @@ class ObjectMD {
     getValue() {
         return this._data;
     }
+
```
ghost, line 1205: Here we have getters and setters for the new fields in the object MD. Both are using a class because the fields are objects, and we need proper input validation + getter/setters for them.

lib/models/ObjectMDArchive.js
```diff
@@ -0,0 +1,169 @@
...
+            archiveMD.setArchiveInfo(data.archiveInfo);
+            archiveMD.setRestoreRequestedAt(data.restoreRequestedAt);
+            archiveMD.setRestoreRequestedDays(data.restoreRequestedDays);
+            archiveMD.setRestoreCompletedAt(data.restoreCompletedAt);
+            archiveMD.setRestoreWillExpireAt(data.restoreWillExpireAt);
+            return true;
+        } catch (err) {
+            return false;
+        }
+    }
```
ghost, line 47: The reason why we have this function is that the current Arsenal doesn't accept class objects in the MD in the tests, but at the same time, we want to make sure the provided object is correct. So I introduce a "isValid" function in both new models to make sure, when we use the setter, that the provided object is either of the right class or is valid, using the model function. It allows to pass the unit tests.

PR #1853, opened 2022-05-03
ARSN-182 : Return error from callback intead of internal error

KillianG:


ghost (approved):

jbertran:
Can we have a test for this, if only to have a guarantee of that contract in the tests?

ghost:
Also nit: the ticket number in the commit is not the right one :slightly_smiling_face:

KillianG:
> Can we have a test for this, if only to have a guarantee of that contract in the tests?

As mentionned in private on slack, this is already "tested" [here](https://github.com/scality/Arsenal/blob/0e3edb847ec1e8adff9a1eca79bbd9b61231db7d/tests/unit/storage/metadata/mongoclient/listObject.spec.js#L31)
Not totally tested but we cannot test deeper as this is in a pipe and I'm note able to mock it :/

ghost (approved):

XinLi0207 (approved):

jonathan-gramain (approved):

PR #1855, opened 2022-05-03
Feature/arsn 178 introduce x amz restore header

ghost:
Introduce the new `x-amz-restore` header based on Fujifilm contributions. I added a note in the file owned by its creator.

The new header is introduced as a new class. In order to be able to set the object MD with an object or a class object, I introduced the `isValid` static function that will ensure the provided object is valid, i.e., that its fields are of the right type.

### DIFF
--- a/lib/models/ObjectMD.js
+++ b/lib/models/ObjectMD.js
@@ -101,6 +101,9 @@ class ObjectMD {
             // similar to normalizing request but after checkAuth so
             // string to sign is not impacted.  This is GH Issue#89.
             'x-amz-storage-class': 'STANDARD',
+            'x-amz-restore': {
+                'ongoing-request': false,
+            },
             'x-amz-server-side-encryption': '',
             'x-amz-server-side-encryption-aws-kms-key-id': '',
             'x-amz-server-side-encryption-customer-algorithm': '',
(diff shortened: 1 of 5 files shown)


ghost (approved):
lgtm!

XinLi0207 (approved):

lib/models/ObjectMDAmzRestore.js
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     *
+     * @param {Object} data archiveInfo
+     * @returns {boolean} true if the provided object is valid
+     */
+    static isValid(data) {
+        try {
+            const object = new ObjectMDAmzRestore(data['ongoing-request'], data['expiry-date']);
+            return !!object;
```
tmacro: Why not return true since the constructor will throw if it's invalid?
```suggestion
            return true;
```

ghost: I think it was mostly a linter issue with the current pattern: if `object` is never used we will have an error, but yes, you are right!

ghost: Edit: I am now returning `true` and ignoring the `no-new` linter rule in https://github.com/scality/Arsenal/pull/1855/commits/2fecb99cd2ba0508c9bbc6e2e1c6edc2e1e02b13#diff-f7e9b6b502cb91cc82d0dc4928a661ac320ea8b750c025c0cd759dd2142fca2fR31-R33

tmacro (approved):

lib/models/ObjectMD.js
```diff
@@ -1194,6 +1195,31 @@ class ObjectMD {
...
+    }
+
+    /**
+     * Set x-amz-restore
+     *
+     * @param {ObjectMDAmzRestore} value x-amz-restore object
+     * @returns {ObjectMD} itself
+     * @throws {Error} case of invalid parameter
+     */
+    setAmzRestore(value) {
```
rachedbenmustapha, line 1215: Any way to clear it after it's not needed anymore?

ghost: Thanks, I agree this is needed. I added a way to clear it when the setter is used with an undefined (or falsy...) value. See https://github.com/scality/Arsenal/pull/1855/commits/3388de6fb6a8e66d6cdf3ac8f32d86cebeef1e34

XinLi0207 (approved):

ghost:
Hi @RachedBenMustapha , I am approving our Arsenal PRs with @XinLiScality , so we can progress. Please let me know if you see a logical change that is required, we will take care of it as part of our work. Thanks!

PR #1856, opened 2022-05-03
Feature/arsn 177 introduce archive field in metadata

ghost:
This new metadata field is introduced as part of the Cold Storage epic.

The suggested format is a class in the same way as https://github.com/scality/Arsenal/pull/1855.

Here, we enforce a specific order of fields edition:
- The `archiveInfo` is required when creating the object
- Both `restoreRequestedAt` and `restoreRequestedDays` can be only set if `archiveInfo` is set (because the latter is mandatory when creating the object)
- Both `restoreCompletedAt` and `restoreWillExpireAt` can be only set if `restoreRequestedAt` and `restoreRequestedDays` are set, because the restoration completion comes after the restoration request.

Once the object is restored and the restoration expires, the `archive` field can be re-computed by keeping the `archiveInfo` field only.

The field setter will make sure the provided object is of the right type or is valid using the `isValid` static function of the new class.

Unit tests are added to make sure the class behave as expected.

### DIFF
--- a/lib/models/ObjectMD.js
+++ b/lib/models/ObjectMD.js
@@ -5,6 +5,7 @@ const VersionIDUtils = require('../versioning/VersionID');
 
 const ObjectMDLocation = require('./ObjectMDLocation');
 const ObjectMDAmzRestore = require('./ObjectMDAmzRestore');
+const ObjectMDArchive = require('./ObjectMDArchive');
 
 /**
  * Class to manage metadata object for regular s3 objects (instead of
@@ -1222,6 +1223,31 @@ class ObjectMD {
         } else {
             delete this._data['x-amz-restore'];
         }
+        return this;
+    }
+
+    /**
+     * Get archive
+     *
+     * @returns {ObjectMDArchive} archive
+     */
+    getArchive() {
+        return this._data.archive;
+    }
+
+    /**
+     * Set archive
+     *
+     * @param {ObjectMDArchive} value archive object
+     * @returns {ObjectMD} itself
+     * @throws {Error} case of invalid parameter
+     */
+    setArchive(value) {
+        // Accept object instance of ObjectMDArchive and Object
+        if (!(value instanceof ObjectMDArchive) && !ObjectMDArchive.isValid(value)) {
+            throw new Error('archive is must be type of ObjectMDArchive.');
+        }
+        this._data.archive = value;
         return this;
     }
 }
--- a/lib/models/ObjectMD.js
+++ b/lib/models/ObjectMD.js
@@ -1238,16 +1238,20 @@ class ObjectMD {
     /**
      * Set archive
      *
-     * @param {ObjectMDArchive} value archive object
+     * @param {ObjectMDArchive|undefined} value archive object
      * @returns {ObjectMD} itself
      * @throws {Error} case of invalid parameter
      */
     setArchive(value) {
-        // Accept object instance of ObjectMDArchive and Object
-        if (!(value instanceof ObjectMDArchive) && !ObjectMDArchive.isValid(value)) {
-            throw new Error('archive is must be type of ObjectMDArchive.');
+        if (value) {
+            // Accept object instance of ObjectMDArchive and Object
+            if (!(value instanceof ObjectMDArchive) && !ObjectMDArchive.isValid(value)) {
+                throw new Error('archive is must be type of ObjectMDArchive.');
+            }
+            this._data.archive = value;
+        } else {
+            delete this._data.archive;
         }
-        this._data.archive = value;
         return this;
     }
 }
--- a/lib/models/ObjectMDArchive.js
+++ b/lib/models/ObjectMDArchive.js
@@ -19,7 +19,7 @@ class ObjectMDArchive {
         restoreRequestedAt = undefined,
         restoreRequestedDays = undefined,
         restoreCompletedAt = undefined,
-        restoreWillExpireAt = undefined
+        restoreWillExpireAt = undefined,
     ) {
         this.setArchiveInfo(archiveInfo);
         this.setRestoreRequestedAt(restoreRequestedAt);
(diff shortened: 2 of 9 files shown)


lib/models/ObjectMDArchive.js
```diff
@@ -0,0 +1,171 @@
...
+    static isValid(data) {
+        try {
+            const object = new ObjectMDArchive(
+                data.archiveInfo,
+                data.restoreRequestedAt,
+                data.restoreRequestedDays,
+                data.restoreCompletedAt,
+                data.restoreWillExpireAt,
+            );
+            return !!object;
```
ghost: Well, I just learned something today hahaha

lib/models/ObjectMDArchive.js
```diff
@@ -0,0 +1,171 @@
...
+     */
+    getRestoreWillExpireAt() {
+        return this.restoreWillExpireAt;
+    }
+    /**
+     * @param {Date} value restoreWillExpireAt
+     * @returns {void}
+     * @throws {Error} case of invalid parameter
+     */
+    setRestoreWillExpireAt(value) {
```
ghost, line 153: I'm supposing there will be a function like 'handleRestore' in the future that will calculate the value seperately then call this?

ghost (approved):
lgtm

XinLi0207 (approved):

lib/models/ObjectMDArchive.js
ghost: It will be done in the APIs, with internal logic :slightly_smiling_face: 
But this object will mainly be edited by backbeat and the TLP client (the "driver" we will use to communicate with a TAPE cold backend), in an async way, so here we only need to provide a class with a set of getter/setters to let the service properly manage the metadata.

lib/models/ObjectMDArchive.js
tmacro: Why not return true since the constructor will throw if it's invalid?
```suggestion
            return true;
```

ghost: yes, done in https://github.com/scality/Arsenal/pull/1856/commits/de35c1e434dfe5f4e7eea348488e8433058ea8f8, along with some new tests to have a 100% code coverage on my additions: https://github.com/scality/Arsenal/pull/1856/commits/af71f12dbbf1918318809fc9aac2a8a25e8b5a3d

tmacro (approved):

lib/models/ObjectMDArchive.js
```diff
@@ -0,0 +1,172 @@
...
+     * @param {Date} [restoreCompletedAt] set at the time of successful restore
+     * @param {Date} [restoreWillExpireAt] computed and stored at the time of restore
+     * @throws {Error} case of invalid parameter
+     */
+    constructor(
+        archiveInfo,
+        restoreRequestedAt = undefined,
+        restoreRequestedDays = undefined,
+        restoreCompletedAt = undefined,
+        restoreWillExpireAt = undefined,
```
rachedbenmustapha: It's never really useful to explicitly write `undefined` because if you write it explicitly, it's no longer `undefined` as in "missing value". A general rule of thumb to make any inspection/debugging easier later for others is, if you want the value to be missing and the code *knows* that it's missing at that time use `null`, and if it's a value that the code was never aware of, it will automatically be `undefined` and that will be clear.

In this particular case, you can just ommit the `undefined` as it's the implicit value (which again does not make sense as explicit).

tests/unit/models/ObjectMDArchive.spec.js
```diff
@@ -0,0 +1,109 @@
...
+
+describe('ObjectMDArchive setters/getters', () => {
+    let archived = null;
+    beforeEach(() => {
+        archived = new ObjectMDArchive(
+            testArchive.archiveInfo,
+            undefined,
+            undefined,
+            undefined,
+            undefined,
```
rachedbenmustapha: Same here, the function has de facto variable arity so no need to fill in the blanks with `undefined`s, it defeats the purpose.

lib/models/ObjectMD.js
```diff
@@ -105,6 +106,9 @@ class ObjectMD {
             'x-amz-restore': {
                 'ongoing-request': false,
             },
+            'archive': {
```
rachedbenmustapha: Is this necessary to store on every single new object that's created or updated, even if there's no cold storage capability on the platform (probably the majority of cases). This adds 20-ish bytes per object, multiplied by billions, where metadata space is precious.

Just noticed the same for `x-amz-restore`.

lib/models/ObjectMDArchive.js
ghost: Thanks, I udpated the function in https://github.com/scality/Arsenal/pull/1856/commits/11106e9db5c2a08f652d9ebf5564020125678863#diff-31d508ca346b0ada575d33da2a05b00e78a940bde0bc7b52e4c76408553fcc79R19
In the other PR for AmzRestore, https://github.com/scality/Arsenal/pull/1855/commits/da9090d3e20de3f102aa3a7f62274384ec80dc24#diff-f7e9b6b502cb91cc82d0dc4928a661ac320ea8b750c025c0cd759dd2142fca2fR19

I made sure not to edit the original commit from Fujifilm.

tests/unit/models/ObjectMDArchive.spec.js
ghost: Done in https://github.com/scality/Arsenal/pull/1856/commits/11106e9db5c2a08f652d9ebf5564020125678863#diff-4a93125ecd0054d2891dd3ea883b3012792a1a1a05594172ed4383a0ddf80eaaL34

lib/models/ObjectMD.js
ghost: I guess you meant "not necessary", and I agree :smile: I removed the fields from the default object MD: https://github.com/scality/Arsenal/pull/1856/files#diff-073d3a5a6a278c5758cbf757c53c2a5e04843f623799fbfd51cd8f953fad6916R8

ghost: Same in https://github.com/scality/Arsenal/pull/1855/commits/da9090d3e20de3f102aa3a7f62274384ec80dc24#diff-073d3a5a6a278c5758cbf757c53c2a5e04843f623799fbfd51cd8f953fad6916L105 for the AmzRestore.

rachedbenmustapha (approved):

PR #1866, opened 2022-05-04
bugfix: ARSN-191 fix wrong notification type when master version is d…

welansari:
…eleted

Ticket: [ARSN-191](https://scality.atlassian.net/browse/ARSN-191)

Issue:
When the last version of an object is deleted, the master gets updated with the metadata of the new last version of the object. The originOp is kept the same during this update, however it should be updated into a "delete operation"

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -1033,6 +1033,7 @@ class MongoClientInterface {
         const masterKey = formatMasterKey(objName, vFormat);
         MongoUtils.serialize(objVal);
         // eslint-disable-next-line
+        objVal.originOp = 's3:ObjectRemoved:Delete';
         c.findOneAndReplace({
             '_id': masterKey,
             'value.isPHD': true,
--- a/tests/unit/storage/metadata/mongoclient/delObject.spec.js
+++ b/tests/unit/storage/metadata/mongoclient/delObject.spec.js
@@ -202,4 +202,25 @@ describe('MongoClientInterface:delObject', () => {
             return done();
         });
     });
+
+    it('repair:: should set correct originOp', done => {
+        const collection = {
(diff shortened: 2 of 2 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1033,6 +1033,7 @@ class MongoClientInterface {
         const masterKey = formatMasterKey(objName, vFormat);
         MongoUtils.serialize(objVal);
         // eslint-disable-next-line
+        objVal.originOp = 's3:ObjectRemoved:Delete'
```
ghost: Nit:
```suggestion
        objVal.originOp = 's3:ObjectRemoved:Delete';
```

(by the way is the line above really needed? By adding a new line we still don't have any linter error...)

welansari: good catch !
Yes the line is needed, we get the following warning otherwise
`1035:9  warning  Assignment to property of function parameter 'objVal'  no-param-reassign`

francoisferrand (approved):

ghost (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1033,6 +1033,7 @@ class MongoClientInterface {
         const masterKey = formatMasterKey(objName, vFormat);
         MongoUtils.serialize(objVal);
         // eslint-disable-next-line
+        objVal.originOp = 's3:ObjectRemoved:Delete';
```
jbertran, line 1036: It seems the `eslint-disable` isn't necessary any longer since it isn't complaining any longer

jbertran (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: just tried it's still complaining

PR #1877, opened 2022-05-06
ARSN-193: more flexible use of error metadata

ghost:
Today, errors going through the errorXMLResponse may not be always ArsenalErrors.

### DIFF
--- a/lib/s3routes/routesUtils.js
+++ b/lib/s3routes/routesUtils.js
@@ -118,7 +118,7 @@ const XMLResponseBackend = {
             `<Message>${errCode.description}</Message>`,
         );
         let counter = 1;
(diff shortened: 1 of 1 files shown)


welansari (approved):
Should probably also update these, and maybe also give them a better description

https://github.com/scality/Arsenal/blob/579bcdf08f21dba80fb3eaa12df8b5dcf16a4672/lib/s3routes/routesUtils.js#L432

https://github.com/scality/Arsenal/blob/579bcdf08f21dba80fb3eaa12df8b5dcf16a4672/lib/s3routes/routesUtils.js#L446

XinLi0207 (approved):

ghost:
Closing this as we found that the error was coming from another place, errors should indeed always be of ArsenalError type.

PR #1878, opened 2022-05-06
ARSN-184 type check s3routes

ghivert:


jbertran (approved):

lib/s3routes/routes/routePUT.ts
```diff
@@ -200,14 +212,13 @@ function routePUT(request, response, api, log, statsClient) {
...
-            if (Number.isNaN(request.parsedContentLength) ||
-            request.parsedContentLength < 0) {
+            if (Number.isNaN(parsedContentLength) || parsedContentLength < 0) {
                 return routesUtils.responseNoBody(errors.BadRequest,
                     null, response, 400, log);
             }
-            log.end().addDefaultFields({
-                contentLength: request.parsedContentLength,
-            });
+            // TODO What's happening?
```
nicolas2bert: What is the issue here? Do we have a ticket number related to this that we can add to the comments?

lib/s3routes/routesUtils.ts
```diff
@@ -0,0 +1,1149 @@
...
+ */
+export function normalizeRequest(
+    request: http.IncomingMessage,
+    validHosts: string[],
+) {
+    /* eslint-disable no-param-reassign */
+    const parsedUrl = url.parse(request.url!, true);
+    // @ts-expect-error
+    request.query = parsedUrl.query;
+    // TODO: make the namespace come from a config variable.
```
nicolas2bert, line 1020: It is a good practice to have TODO attached with a JIRA ticket number to track todos :)

nicolas2bert (approved):
Looking good :)

lib/s3routes/routesUtils.ts
jbertran: This one comes from existing code :grimacing:

lib/s3routes/routes/routePUT.ts
jbertran: Perhaps we can blanket-add ARSN-174 references to these TODOs. These all to track the second pass required to understand the business logic to type these accurately.

PR #1888, opened 2022-05-10
improvement: ARSN-197 implement object restore request xml parser

XinLi0207:
add objectRestore api required request xml parser, for cloudserver side usage

although we don't support Bulk or Expedited Tier as per[ the design ](https://github.com/scality/citadel/blob/development/1.0/docs/design/cold-storage-s3-api.md#cloudserver-changes), we will have this limitation in cloudserver side(throw NotImplemented error if specified Bulk or Expited), here we keep it in line with aws, we allow Standard, Bulk and Expedited, otherwise throw MalformedXML error.

### DIFF
--- a/lib/s3middleware/objectRestore.ts
+++ b/lib/s3middleware/objectRestore.ts
@@ -1,90 +1,49 @@
-import { parseString } from 'xml2js';
-import errors, { ArsenalError } from '../errors';
+import {parseStringPromise} from 'xml2js';
+import errors, {ArsenalError} from '../errors';
 import * as werelogs from 'werelogs';
+import {validRestoreObjectTiers} from "../constants";
 
 /*
     Format of xml request:
     <RestoreRequest>
       <Days>integer</Days>
-      <Tier>Standard|Bulk|Expedited</RetainUntilDate>
+      <Tier>Standard|Bulk|Expedited</Tier>
     </RestoreRequest>
 */
 
 /**
- * validateDays - validate restore days
- * @param days - parsed days string
- * @return - contains days or error
- */
-function validateDays(days?: string) {
-    if (!days || !days[0]) {
-        const desc = 'request xml does not contain RestoreRequest.Days';
-        const error = errors.MalformedXML.customizeDescription(desc);
-        return { error };
-    }
-    // RestoreRequest.Days must be greater than or equal to 1
-    const daysValue = Number.parseInt(days[0], 10);
-    if (Number.isNaN(daysValue)) {
-        const desc = `RestoreRequest.Days is invalid type. [${days}]`;
-        const error = errors.MalformedXML.customizeDescription(desc);
-        return { error };
-    }
-    if (daysValue < 1) {
-        const desc = `RestoreRequest.Days must be greater than 0. [${days}]`;
-        const error = errors.MalformedXML.customizeDescription(desc);
-        return { error };
-    }
-    if (daysValue > 2147483647) {
-        const desc = `RestoreRequest.Days must be less than 2147483648. [${days}]`;
-        const error = errors.MalformedXML.customizeDescription(desc);
-        return { error };
-    }
-    return { days: daysValue }
-}
-
-/**
- * validateTier - validate retain until date
- * @param tier - parsed tier string
- * @return - contains retain until date or error
- */
-function validateTier(tier?: string) {
-    // If Tier is specified, must be "Expedited" or "Standard" or "Bulk"
-    const tierList = ['Expedited', 'Standard', 'Bulk'];
-    if (tier && tier[0] && !tierList.includes(tier[0])) {
-        const desc = `RestoreRequest.Tier is invalid value. [${tier}]`;
-        const error = errors.MalformedXML.customizeDescription(desc);
-        return { error };
-    }
-    // If do not specify, set "Standard"
-    return { tier: tier && tier[0] ? tier[0] : 'Standard' };
-}
-
-/**
  * validate restore request xml
- * @param parsedXml - parsed restore request xml object
- * @return - contains restore request information on success,
- * error on failure
+ * @param restoreRequest - parsed restore request object
+ * @return{ArsenalError|undefined} - error on failure, undefined on success
  */
-function validateRestoreRequestParameters(parsedXml?: any) {
-    if (!parsedXml) {
-        const desc = 'request xml is undefined or empty';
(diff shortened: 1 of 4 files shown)


lib/s3middleware/objectRestore.ts
```diff
@@ -0,0 +1,141 @@
...
+    <RestoreRequest>
+      <Days>integer</Days>
+      <Tier>Standard|Bulk|Expedited</RetainUntilDate>
+    </RestoreRequest>
+*/
+
+/**
+ * validateDays - validate restore days
+ * @param days - parsed days string
+ * @return - contains days or error
```
miniscruff: I would not return something or error, especially since this is an object with errors defined or days defined and not X|Y.
I would consider either renaming the function to something like `getDays` that throws an exception on a bad value.
Or maybe we find another style that works but I wouldn't return two different things for two different reasons.

lib/s3middleware/objectRestore.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+    return { days: daysValue }
+}
+
+/**
+ * validateTier - validate retain until date
+ * @param tier - parsed tier string
+ * @return - contains retain until date or error
+ */
+function validateTier(tier?: string) {
```
miniscruff: Same as validate days

lib/s3middleware/objectRestore.ts
```diff
@@ -0,0 +1,141 @@
...
+    // If Tier is specified, must be "Expedited" or "Standard" or "Bulk"
+    const tierList = ['Expedited', 'Standard', 'Bulk'];
```
miniscruff: Use a set if we just need to check includes ( and probably move it out of the function scope )

lib/s3middleware/objectRestore.ts
```diff
@@ -0,0 +1,141 @@
...
+    if (tier && tier[0] && !tierList.includes(tier[0])) {
```
miniscruff: I would use a guard clause for `if (!tier) { ... }`

lib/s3middleware/objectRestore.ts
```diff
@@ -0,0 +1,141 @@
...
+    return { tier: tier && tier[0] ? tier[0] : 'Standard' };
+}
+
+/**
+ * validate restore request xml
+ * @param parsedXml - parsed restore request xml object
+ * @return - contains restore request information on success,
+ * error on failure
+ */
+function validateRestoreRequestParameters(parsedXml?: any) {
```
miniscruff: Same as other validates.

lib/s3middleware/objectRestore.ts
```diff
@@ -0,0 +1,141 @@
...
+ * @param log - Werelogs logger
+ * @param cb - callback to server
+ * @return - calls callback with object restore request or error
+ */
+export function parseRestoreRequestXml(
+    xml: string,
+    log: werelogs.Logger,
+    cb: (err: ArsenalError | null, data?: any) => void,
+) {
+    parseString(xml, (err, result) => {
```
miniscruff: I would use async await here instead of a callback

tests/unit/s3middleware/objectRestore.spec.js
```diff
@@ -0,0 +1,146 @@
...
+
+const generateXml = (days, tier) => {
+    const daysElement = days ? `<Days>${days}</Days>` : '';
+    const tierElement = tier ? `<Tier>${tier}</Tier>` : '';
+    return [
+        '<RestoreRequest xmlns="http://s3.amazonaws.com/doc/2006-03-01/">',
+        `${daysElement}`,
+        `${tierElement}`,
+        '</RestoreRequest>',
+    ].join('');
```
miniscruff:
```suggestion
    const ret = ['<RestoreRequest xmlns="http://s3.amazonaws.com/doc/2006-03-01/">'];
    if (days) {
        ret.push(`<Days>${days}</Days>`);
    }
    if (tier) {
        ret.push(`<Tier>${tier}</Tier>`);
    }
    ret.push('</RestoreRequest>');
    return ret.join('');
```

This style is a bit more flexible.

lib/s3middleware/objectRestore.ts
miniscruff: Is there a non-callback form of this? Kinda weird this needs a callback.

miniscruff (changes requested):

lib/s3middleware/objectRestore.ts
XinLi0207: agree with you, I changed the structure to have one function only for validating, returning errors on failure or undefined on success

lib/s3middleware/objectRestore.ts
XinLi0207: yeah, make sense, put in in `constants` file

lib/s3middleware/objectRestore.ts
XinLi0207: yes, use parseStringPromise instead

tests/unit/s3middleware/objectRestore.spec.js
XinLi0207: thank you! this looks clearer

ghost (approved):
lgtm

lib/s3middleware/objectRestore.ts
```diff
@@ -0,0 +1,110 @@
...
+) {
+    let result;
+    try {
+        result = await parseStringPromise(xml);
+    } catch (err) {
+        log.trace('xml parsing failed', {
+            error: err,
+            method: 'parseRestoreXml',
+        });
+        log.debug('invalid xml', { xml });
```
ghost: Can we keep only one log here? The first one looks good, maybe like this:
```suggestion
        log.debug('xml parsing failed', {
            error: err,
            method: 'parseRestoreXml',
            xml,
        });
```

lib/s3middleware/objectRestore.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * convertToXml - Convert restore request info object to xml
+ * @param days - restore days
+ * @param tier - restore tier
+ * @return - returns restore request information xml string
+ */
+export function convertToXml(days: string, tier: string) {
+    if (!(days && tier)) {
+        return ''
```
ghost: Nit:
```suggestion
        return '';
```

miniscruff (approved):

PR #1889, opened 2022-05-10
ARSN-200: fix probe server readiness path

alexanderchan-scality:


miniscruff (approved):
not that it matters much but using /live for both was my intention as we had only used one form of readiness at the time so I just implemented the one handler.

naren-rajendran (approved):

alexanderchan-scality:
@miniscruff I see. I had made this change as it was not backwards compatible with previous versions of backbeat images. I had probably missed the intention for the change since it was made as part of the typescript refactor.

miniscruff:
> @miniscruff I see. I had made this change as it was not backward compatible with previous versions of backbeat images. I had probably missed the intention for the change since it was made as part of the typescript refactor.

We should definitely do this change, just making a point as to why it was the way it was. It looks like a typo bug but was more of a shortcut. Just have to check backbeat uses to make sure the two routes are used properly.

alexanderchan-scality:
ping

PR #1894, opened 2022-05-12
ARSN-201 type check versioning

ghivert:


jbertran (approved):

lib/versioning/Version.ts
```diff
@@ -11,36 +10,48 @@ const VID_SEP = require('./constants').VersioningConstants.VersionId.Separator;
  * with the full stringify/parse cycle, which is too low a number for
  * use with a production setup with Metadata).
  */
-class Version {
+export class Version {
+    version: {
```
miniscruff, line 14: would making this a general type reduce the copy paste between here and the constructor?

miniscruff: On second thought, do we need this extra layer at all?

ghivert: Right now, I would like to avoid breaking changes. We try to keep the same compatibility as long as we're not TypeScript compatible all over the codebases.

The only problem I have with a general type is finding a good name for it. Because we can't name it `Version` obviously, I was deeply in seek of creativity and ended up with just dumping everything in the constructor and the version. If you have a good name, I'll take it!

miniscruff: Could we remove the extra layer and just have the 4 values on the Version class itself?

ghivert: Unfortunately I have no idea on whether the `version` field is used or not throughout the rest of the codebase. Removing the nesting would also mean changing the interface of `Version`, and thus, breaking the promise of not breaking changes during the refacto. I think we had enough breaking changes during this migration at the moment, and we should both document and open a specific issue on the subject instead of doing it silently. 🙂

I'd like to have a fully working TS solution without too much changes to highlight all the problems, and then address them all at once in a complete breaking change of the API. Having TS on the components requiring Arsenal would be a great help, because the refactor of breaking changes would be really easy to tackle.

miniscruff (approved):

PR #1912, opened 2022-05-13
ARSN-202 Overwrite master version metadata

nicolas2bert:
Introduce a new metadata logic to update the master version metadata given a version id: `masterVersionId`.

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -867,12 +867,51 @@ class MongoClientInterface {
     }
 
     /**
+     * Update master with specific version if provided.
+     * @param {Object} c bucket collection
+     * @param {String} bucketName bucket name
+     * @param {String} objName object name
+     * @param {Object} objVal object metadata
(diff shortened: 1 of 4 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -866,13 +866,52 @@ class MongoClientInterface {
...
+     */
+    putObjectMasterVer(c, bucketName, objName, objVal, params, log, cb) {
+        const { masterVersionId, vFormat } = params;
+        if (masterVersionId) {
+            // eslint-disable-next-line no-param-reassign
+            objVal.versionId = masterVersionId;
+        }
+        const masterKey = formatMasterKey(objName, vFormat);
+        c.update({
+            _id: masterKey,
```
vrancurel, line 889: Can you also update the microVersionId ? (look at object model)

`models/ObjectMD.js:    updateMicroVersionId()`

This will be used to know internally that there was an update.

vrancurel (changes requested):

PR #1920, opened 2022-05-20
ARSN-209 type check models

ghivert:


lib/models/WebsiteConfiguration.ts
```diff
@@ -0,0 +1,215 @@
...
+    * @param [params.redirect.hostName] - hostname to use for redirect
+    * @param [params.redirect.replaceKeyPrefixWith] - string to replace
+    *   keyPrefixEquals specified in condition
+    * @param [params.redirect.replaceKeyWith] - string to replace key
+    * @param [params.redirect.httpRedirectCode] - http redirect code
+    * @param [params.condition] - specifies conditions for a redirect
+    * @param [params.condition.keyPrefixEquals] - key prefix that
+    *   triggers a redirect
+    * @param [params.condition.httpErrorCodeReturnedEquals] - http code
+    *   that triggers a redirect
```
miniscruff: There seems to be a lot more params then are used, is this right?

lib/models/WebsiteConfiguration.ts
```diff
@@ -0,0 +1,215 @@
...
+    constructor(params?: RoutingRuleParams) {
+        if (params) {
+            this._redirect = params.redirect;
+            this._condition = params.condition;
+        }
+    }
+
+    /**
+    * Return copy of rule as plain object
+    * @return rule;
```
miniscruff, line 47: Is this rule a valid type?

lib/models/WebsiteConfiguration.ts
```diff
@@ -0,0 +1,215 @@
...
+        const rule = {
+            redirect: this._redirect,
+            condition: this._condition,
+        };
+        return rule;
+    }
+
+    /**
+    * Return the condition object
+    * @return condition;
```
miniscruff, line 59: Is this condition a valid type?

lib/models/WebsiteConfiguration.ts
```diff
@@ -0,0 +1,215 @@
...
+    */
+    getCondition() {
+        return this._condition;
+    }
+
+    /**
+    * Return the redirect object
+    * @return redirect;
```
miniscruff, line 67: Same type question

lib/models/WebsiteConfiguration.ts
```diff
@@ -0,0 +1,215 @@
...
+    * @param [params.redirect.replaceKeyWith] - string to replace key
+    * @param [params.redirect.httpRedirectCode] - http redirect code
+    * @param [params.condition] - specifies conditions for a redirect
+    * @param [params.condition.keyPrefixEquals] - key prefix that
+    *   triggers a redirect
+    * @param [params.condition.httpErrorCodeReturnedEquals] - http code
+    *   that triggers a redirect
+    */
+    constructor(params?: RoutingRuleParams) {
+        if (params) {
```
miniscruff, line 39: Is params optional? The JSDoc has it as required but the JS code has this check to

lib/models/WebsiteConfiguration.ts
```diff
@@ -0,0 +1,215 @@
...
+            this._redirectAllRequestsTo = params.redirectAllRequestsTo;
+            this.setRoutingRules(params.routingRules);
+        }
+    }
+
+    /**
+    * Return plain object with configuration info
+    * @return - Object copy of class instance
+    */
+    getConfig() {
```
miniscruff, line 118: Isn't there some typescript way to define return types in the code, like this? And if so should it be used?

```suggestion
    * @return - Object copy of class instance
    */
    getConfig(): Object {
```

lib/models/WebsiteConfiguration.ts
ghivert: Sure, but it would be better to handle it in a specific ticket. The scope is the migration, we try to keep the code as similar as possible for now. Any opinion @jbertran ?

lib/models/WebsiteConfiguration.ts
ghivert: The type of `rule` is infered by TypeScript.

lib/models/WebsiteConfiguration.ts
ghivert: The type of `condition` is infered by TypeScript.

lib/models/WebsiteConfiguration.ts
ghivert: The type of `redirect` is infered by TypeScript.

lib/models/WebsiteConfiguration.ts
ghivert: I prefered to stick with the implementation because the JSDoc was sometimes outdated. The type is mainly to reflect the construct.

lib/models/WebsiteConfiguration.ts
ghivert: Sure we could, but as written, the type is directly infered by TypeScript. Try to build the project with TSC, you'll see the correct signature for the function, which is something like 
```typescript
type Return = { 
    indexDocument?: string;
    errorDocument?: string;
    redirectAllRequestsTo?: RedirectAllRequestsTo;
    routingRules?: {
        redirect: Redirect;
        condition: Condition;
    };
}
```

lib/models/WebsiteConfiguration.ts
jbertran: This specific bit of JSDoc looks like an attempt at type hints more than anything, so I'm not surprised it contains more than what we actually use.

What we could do is move the JSDoc to the types (`Redirect` and `Condition`), because it no longer has a lot of meaning directly on the constructor

ghivert:
Boop

miniscruff (approved):
thought I approved this one already

welansari (approved):

PR #1933, opened 2022-06-14
Improvement/arsn 227 support bucket tagging

ghost:
Added @KillianG as an author because the changes for the 8.1 branch are the same as what was done for 7.10 branch.

This PR adds support for tags in the bucket MD. Unit tests were updated to validate the changes.

### DIFF
--- a/documentation/BucketInfoModelVersion.md
+++ b/documentation/BucketInfoModelVersion.md
@@ -187,6 +187,17 @@ Used to store the users configured KMS key id
 this._tags = tags || null;
 ```
 
+The Tag Set of a bucket is an array of objects with Key and Value:
+
+```javascript
+[
+    {
+        Key: 'something',
+        Value: 'some_data'
+    }
+]
+```
+
 ### Usage
 
 Used to store bucket tagging
--- a/documentation/BucketInfoModelVersion.md
+++ b/documentation/BucketInfoModelVersion.md
@@ -178,3 +178,15 @@ this._serverSideEncryption.configuredMasterKeyId = configuredMasterKeyId || unde
 ### Usage
 
 Used to store the users configured KMS key id
+
+## Model version 15
+
+### Properties Added
+
+```javascript
+this._tags = tags || null;
+```
+
+### Usage
+
+Used to store bucket tagging
--- a/lib/models/BucketInfo.ts
+++ b/lib/models/BucketInfo.ts
@@ -60,7 +60,7 @@ export default class BucketInfo {
     _objectLockEnabled?: boolean;
     _objectLockConfiguration?: any;
     _notificationConfiguration?: any;
-    _tags?: Array<BucketTag> | [];
+    _tags?: Array<BucketTag>;
     _readLocationConstraint: string | null;
     _isNFS: boolean | null;
     _azureInfo: any | null;
@@ -244,7 +244,7 @@ export default class BucketInfo {
         };
 
         if (tags === undefined) {
-            tags = [];
+            tags = [] as BucketTag[];
         }
         validateTags(tags);
 
@@ -864,7 +864,7 @@ export default class BucketInfo {
      * Set bucket tags
      * @return - bucket info instance
      */
-    setTags(tags: Array<BucketTag> | []) {
+    setTags(tags: Array<BucketTag>) {
         this._tags = tags;
         return this;
     }
--- a/lib/models/BucketInfo.ts
+++ b/lib/models/BucketInfo.ts
@@ -8,7 +8,7 @@ import ObjectLockConfiguration from './ObjectLockConfiguration';
 import BucketPolicy from './BucketPolicy';
 import NotificationConfiguration from './NotificationConfiguration';
 import { ACL as OACL } from './ObjectMD';
-import { validateTags, BucketTag } from '../s3middleware/tagging';
+import { areTagsValid, BucketTag } from '../s3middleware/tagging';
 
 // WHEN UPDATING THIS NUMBER, UPDATE BucketInfoModelVersion.md CHANGELOG
 // BucketInfoModelVersion.md can be found in documentation/ at the root
@@ -246,7 +246,7 @@ export default class BucketInfo {
         if (tags === undefined) {
             tags = [] as BucketTag[];
         }
-        validateTags(tags);
+        assert.strictEqual(areTagsValid(tags), true);
 
         // IF UPDATING PROPERTIES, INCREMENT MODELVERSION NUMBER ABOVE
         this._acl = aclInstance;
--- a/lib/models/BucketInfo.ts
+++ b/lib/models/BucketInfo.ts
@@ -144,6 +144,7 @@ export default class BucketInfo {
         objectLockEnabled?: boolean,
         objectLockConfiguration?: any,
         notificationConfiguration?: any,
+        tags?: { key: string; value: string }[] | null,
     ) {
         assert.strictEqual(typeof name, 'string');
         assert.strictEqual(typeof owner, 'string');
@@ -240,6 +241,10 @@ export default class BucketInfo {
             READ_ACP: [],
         };
 
+        if (tags) {
+            assert(Array.isArray(tags));
+        }
+
         // IF UPDATING PROPERTIES, INCREMENT MODELVERSION NUMBER ABOVE
         this._acl = aclInstance;
         this._name = name;
@@ -265,6 +270,7 @@ export default class BucketInfo {
         this._objectLockEnabled = objectLockEnabled || false;
         this._objectLockConfiguration = objectLockConfiguration || null;
         this._notificationConfiguration = notificationConfiguration || null;
+        this._tags = tags || null;
         return this;
     }
 
@@ -298,6 +304,7 @@ export default class BucketInfo {
             objectLockEnabled: this._objectLockEnabled,
             objectLockConfiguration: this._objectLockConfiguration,
             notificationConfiguration: this._notificationConfiguration,
+            tags: this._tags,
         };
         const final = this._websiteConfiguration
             ? {
@@ -323,7 +330,7 @@ export default class BucketInfo {
             obj.cors, obj.replicationConfiguration, obj.lifecycleConfiguration,
             obj.bucketPolicy, obj.uid, obj.readLocationConstraint, obj.isNFS,
             obj.ingestion, obj.azureInfo, obj.objectLockEnabled,
-            obj.objectLockConfiguration, obj.notificationConfiguration);
+            obj.objectLockConfiguration, obj.notificationConfiguration, obj.tags);
     }
 
     /**
@@ -350,7 +357,7 @@ export default class BucketInfo {
             data._bucketPolicy, data._uid, data._readLocationConstraint,
             data._isNFS, data._ingestion, data._azureInfo,
             data._objectLockEnabled, data._objectLockConfiguration,
-            data._notificationConfiguration);
+            data._notificationConfiguration, data._tags);
     }
 
     /**
@@ -839,6 +846,23 @@ export default class BucketInfo {
     */
     setObjectLockEnabled(enabled: boolean) {
         this._objectLockEnabled = enabled;
+        return this;
+    }
+
+    /**
+     * Get the value of bucket tags
+     * @return - Array of bucket tags
+     */
+    getTags() {
+        return this._tags;
+    }
+    
+    /**
+     * Set bucket tags
+     * @return - bucket info instance
+     */
+    setTags(tags: { key: string; value: string }[] | null) {
+        this._tags = tags;
         return this;
     }
 }
--- a/lib/models/BucketInfo.ts
+++ b/lib/models/BucketInfo.ts
@@ -8,6 +8,7 @@ import ObjectLockConfiguration from './ObjectLockConfiguration';
 import BucketPolicy from './BucketPolicy';
 import NotificationConfiguration from './NotificationConfiguration';
 import { ACL as OACL } from './ObjectMD';
+import { validateTags, BucketTag } from '../s3middleware/tagging';
 
 // WHEN UPDATING THIS NUMBER, UPDATE BucketInfoModelVersion.md CHANGELOG
 // BucketInfoModelVersion.md can be found in documentation/ at the root
@@ -59,7 +60,7 @@ export default class BucketInfo {
     _objectLockEnabled?: boolean;
     _objectLockConfiguration?: any;
     _notificationConfiguration?: any;
-    _tags?: { key: string; value: string }[] | null;
+    _tags?: Array<BucketTag> | [];
     _readLocationConstraint: string | null;
     _isNFS: boolean | null;
     _azureInfo: any | null;
@@ -118,6 +119,7 @@ export default class BucketInfo {
     * @param [objectLockEnabled] - true when object lock enabled
     * @param [objectLockConfiguration] - object lock configuration
     * @param [notificationConfiguration] - bucket notification configuration
+    * @param [tags] - bucket tag set
     */
     constructor(
         name: string,
@@ -144,7 +146,7 @@ export default class BucketInfo {
         objectLockEnabled?: boolean,
         objectLockConfiguration?: any,
         notificationConfiguration?: any,
-        tags?: { key: string; value: string }[] | null,
+        tags?: Array<BucketTag> | [],
     ) {
         assert.strictEqual(typeof name, 'string');
         assert.strictEqual(typeof owner, 'string');
@@ -241,9 +243,10 @@ export default class BucketInfo {
             READ_ACP: [],
         };
 
-        if (tags) {
-            assert(Array.isArray(tags));
+        if (tags === undefined) {
+            tags = [];
         }
+        validateTags(tags);
 
         // IF UPDATING PROPERTIES, INCREMENT MODELVERSION NUMBER ABOVE
         this._acl = aclInstance;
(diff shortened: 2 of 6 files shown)


tests/unit/models/BucketInfo.spec.js
```diff
@@ -205,7 +220,8 @@ Object.keys(acl).forEach(
             true, undefined, testAzureInfo,
             testobjectLockEnabled,
             testObjectLockConfiguration,
-            testNotificationConfiguration);
+            testNotificationConfiguration,
+            testBucketTagging);
```
miniscruff:
```suggestion
            testBucketTagging,
        );
```
This helps a little with diffs and it makes the start and end on the same column

lib/models/BucketInfo.ts
```diff
@@ -144,6 +144,7 @@ export default class BucketInfo {
         objectLockEnabled?: boolean,
         objectLockConfiguration?: any,
         notificationConfiguration?: any,
+        tags?: { key: string; value: string }[] | null,
```
miniscruff: Instead of defaulting to null why not default to an empty array? That is, tags are required but can be empty, rather than optionally null. `if (tags)` will still as well. For compatibility you can do: 
```
if (tags === undefined) { tags = [] }
```

miniscruff: On another note, you missed the JSDoc for the new param.

documentation/BucketInfoModelVersion.md
```diff
@@ -178,3 +178,15 @@ this._serverSideEncryption.configuredMasterKeyId = configuredMasterKeyId || unde
 ### Usage
 
 Used to store the users configured KMS key id
+
+## Model version 15
+
+### Properties Added
+
+```javascript
+this._tags = tags || null;
```
miniscruff, line 187: I think we should include what is in a tag

lib/models/BucketInfo.ts
```diff
@@ -240,6 +241,10 @@ export default class BucketInfo {
             READ_ACP: [],
         };
 
+        if (tags) {
+            assert(Array.isArray(tags));
```
miniscruff: We should probably also check what is in each tag

lib/models/BucketInfo.ts
miniscruff: Ok one more point on this line, I would use a TS type or whatever its called instead of building it each time.

miniscruff (changes requested):

tests/unit/models/BucketInfo.spec.js
ghost: First, thanks for your review, very helpful!
Done: https://github.com/scality/Arsenal/pull/1933/commits/b964834d3135e98184538a8176b0b5630c8c147b#diff-31604c11f0c75e4caa3bd813de17aa4ff66eee7a586dd1a583acd1e1feaf48e6R224

lib/models/BucketInfo.ts
ghost: Agree and done in https://github.com/scality/Arsenal/pull/1933/commits/b964834d3135e98184538a8176b0b5630c8c147b#diff-4ee3cd5de98073dadb72b27f034f4979480c04388cc4f4538f4895373527f174R63

ghost: :smile:  https://github.com/scality/Arsenal/pull/1933/commits/b964834d3135e98184538a8176b0b5630c8c147b#diff-4ee3cd5de98073dadb72b27f034f4979480c04388cc4f4538f4895373527f174R122

lib/models/BucketInfo.ts
ghost: I found that the validation function internally used was not using the exact same format so I created another one that is exported and that will throw an error instead of returning one. It's the same function but using a different format, i.e., the format that we are supposed to use when we create a `BucketInfo` object. The other function deals with XML.

https://github.com/scality/Arsenal/pull/1933/commits/b964834d3135e98184538a8176b0b5630c8c147b#diff-7aa7eeaed88643ea97d57a5f15ddc2e64c380190641c53bb0c3f64d67807a475R120

So now I am calling this validation function in the constructor, after assigning the default value (`[]`) if the parameter is undefined

documentation/BucketInfoModelVersion.md
ghost: I added some doc here: https://github.com/scality/Arsenal/pull/1933/commits/b964834d3135e98184538a8176b0b5630c8c147b#diff-ffb3da4096d0bcb75ae13838bd7090485735b50eebf5e1ea1535641969bc9e07R190

lib/models/BucketInfo.ts
ghost: I created the `BucketTag` type here: https://github.com/scality/Arsenal/commit/b964834d3135e98184538a8176b0b5630c8c147b#diff-7aa7eeaed88643ea97d57a5f15ddc2e64c380190641c53bb0c3f64d67807a475R6

And I updated the associated functions.

XinLi0207 (approved):

lib/models/BucketInfo.ts
miniscruff: Yea that is what I was thinking, cool.

lib/s3middleware/tagging.ts
```diff
@@ -32,6 +37,12 @@ export const _validator = {
         && tag.Key[0] !== undefined && tag.Value[0] !== undefined
         && typeof tag.Key[0] === 'string' && typeof tag.Value[0] === 'string',
 
+    validateTagObjectStructure: (tag: BucketTag) => tag
+        && Object.keys(tag).length === 2
+        && tag.Key && tag.Value
+        && tag.Key.length >= 1 && tag.Value.length >= 1
+        && typeof tag.Key === 'string' && typeof tag.Value === 'string',
```
miniscruff: I would switch these two comparisons, if the tag key or value is not a string it is not always going to have a `.length`, eg if either were numbers.

lib/s3middleware/tagging.ts
```diff
@@ -32,6 +37,12 @@ export const _validator = {
         && tag.Key[0] !== undefined && tag.Value[0] !== undefined
         && typeof tag.Key[0] === 'string' && typeof tag.Value[0] === 'string',
 
+    validateTagObjectStructure: (tag: BucketTag) => tag
+        && Object.keys(tag).length === 2
+        && tag.Key && tag.Value
```
miniscruff: you can remove this check if you swap the other two like my other comment, undefined and null are both not a typeof string.

lib/models/BucketInfo.ts
```diff
@@ -59,7 +60,7 @@ export default class BucketInfo {
     _objectLockEnabled?: boolean;
     _objectLockConfiguration?: any;
     _notificationConfiguration?: any;
-    _tags?: { key: string; value: string }[] | null;
+    _tags?: Array<BucketTag> | [];
```
miniscruff: Do we need the `| []`? I believe an empty array should still have the bucket tag type so we wouldn't need this or empty array thing. Could be wrong though.

lib/models/BucketInfo.ts
```diff
@@ -240,6 +243,11 @@ export default class BucketInfo {
             READ_ACP: [],
         };
 
+        if (tags === undefined) {
+            tags = [];
```
miniscruff:
```suggestion
            tags = [] as BucketTag[];
```

I think you need something like this to signify it is of a specific type, even if empty

lib/models/BucketInfo.ts
```diff
@@ -841,4 +851,21 @@ export default class BucketInfo {
...
+     */
+    getTags() {
+        return this._tags;
+    }
+    
+    /**
+     * Set bucket tags
+     * @return - bucket info instance
+     */
+    setTags(tags: Array<BucketTag> | []) {
```
miniscruff: Same comment as above, I don't think we need or want the `| []`

lib/s3middleware/tagging.ts
```diff
@@ -106,6 +117,54 @@ function _validateTags(tags: Array<{ Key: string[], Value: string[] }>) {
...
+    for (const tag of tags) {
+        if (!_validator.validateTagObjectStructure(tag)) {
+            throw errors.MalformedXML;
+        }
+        const key = tag.Key;
+        const value = tag.Value;
+
+        if (!key) {
+            throw errors.InvalidTag.customizeDescription('The TagKey you ' +
+            'have provided is invalid');
```
miniscruff:
```suggestion
            throw errors.InvalidTag.customizeDescription(
                'The TagKey you have provided is invalid'
            );
```
I think this method of splitting lines is easier to read. It has the same column open and close. There is not a string starting on a normal code line, don't need to read two lines for a string. That sort of stuff.

lib/s3middleware/tagging.ts
```diff
@@ -106,6 +117,54 @@ function _validateTags(tags: Array<{ Key: string[], Value: string[] }>) {
...
+        return errorBadRequestLimit50();
+    }
+
+    const tagsResult = {};
+    for (const tag of tags) {
+        if (!_validator.validateTagObjectStructure(tag)) {
+            throw errors.MalformedXML;
+        }
+        const key = tag.Key;
+        const value = tag.Value;
```
miniscruff:
```suggestion
        const {Key: key, Value: value } = tag;
```

lib/s3middleware/tagging.ts
```diff
@@ -106,6 +117,54 @@ function _validateTags(tags: Array<{ Key: string[], Value: string[] }>) {
...
+
+        if (!key) {
+            throw errors.InvalidTag.customizeDescription('The TagKey you ' +
+            'have provided is invalid');
+        }
+
+        // Allowed characters are letters, whitespace, and numbers, plus
+        // the following special characters: + - = . _ : /
+        // Maximum key length: 128 Unicode characters
+        // Maximum value length: 256 Unicode characters
```
miniscruff: This description more belongs above the validate key value function then it does here. That is where we would look if we wanted to know what is being validated, not in one of the uses.

lib/s3middleware/tagging.ts
```diff
@@ -106,6 +117,54 @@ function _validateTags(tags: Array<{ Key: string[], Value: string[] }>) {
...
+        if (result instanceof Error) {
+            throw result;
+        }
+
+        tagsResult[key] = value;
+    }
+    // not repeating keys
+    if (tags.length > Object.keys(tagsResult).length) {
+        throw errors.InvalidTag.customizeDescription('Cannot provide ' +
+        'multiple Tags with the same key');
```
miniscruff:
```suggestion
        throw errors.InvalidTag.customizeDescription(
            'Cannot provide multiple Tags with the same key'
        );
```

lib/s3middleware/tagging.ts
```diff
@@ -106,6 +117,54 @@ function _validateTags(tags: Array<{ Key: string[], Value: string[] }>) {
...
+        return {};
+    }
+    // Maximum number of tags per resource: 50
+    if (tags.length > 50) {
+        return errorBadRequestLimit50();
+    }
+
+    const tagsResult = {};
+    for (const tag of tags) {
+        if (!_validator.validateTagObjectStructure(tag)) {
```
miniscruff, line 139: This will return false if key is falsey, negating the if below I believe. Unless there is some other condition I am missing.

miniscruff (changes requested):
I don't see as many tests as I would expect, like nothing for the validation which could be small unit tests.

ghost (approved):

lib/s3middleware/tagging.ts
ghost: Right, that's an artifact from the original function where we were using an array.

lib/s3middleware/tagging.ts
ghost: Done: https://github.com/scality/Arsenal/pull/1933/commits/4b19cc2101f4a15666f8746ad8fb12533e9b45d8#diff-7aa7eeaed88643ea97d57a5f15ddc2e64c380190641c53bb0c3f64d67807a475R46

lib/s3middleware/tagging.ts
ghost: Done as well: https://github.com/scality/Arsenal/pull/1933/commits/4b19cc2101f4a15666f8746ad8fb12533e9b45d8#diff-7aa7eeaed88643ea97d57a5f15ddc2e64c380190641c53bb0c3f64d67807a475R46

lib/models/BucketInfo.ts
ghost: No, we don't need it, it was mostly to help the user know that he can pass an empty array, but yeah, you are right! https://github.com/scality/Arsenal/pull/1933/commits/4b19cc2101f4a15666f8746ad8fb12533e9b45d8#diff-4ee3cd5de98073dadb72b27f034f4979480c04388cc4f4538f4895373527f174R63

lib/models/BucketInfo.ts
ghost: It was not needed by the transpiler but it's good to be explicit about the type https://github.com/scality/Arsenal/pull/1933/commits/4b19cc2101f4a15666f8746ad8fb12533e9b45d8#diff-4ee3cd5de98073dadb72b27f034f4979480c04388cc4f4538f4895373527f174R247t

lib/models/BucketInfo.ts
ghost: https://github.com/scality/Arsenal/pull/1933/commits/4b19cc2101f4a15666f8746ad8fb12533e9b45d8#diff-4ee3cd5de98073dadb72b27f034f4979480c04388cc4f4538f4895373527f174R867

lib/s3middleware/tagging.ts
ghost: Removed as the condiiton was not needed anymore as you said.

lib/s3middleware/tagging.ts
ghost: https://github.com/scality/Arsenal/pull/1933/commits/4b19cc2101f4a15666f8746ad8fb12533e9b45d8#diff-7aa7eeaed88643ea97d57a5f15ddc2e64c380190641c53bb0c3f64d67807a475R145

lib/s3middleware/tagging.ts
ghost: Yes! https://github.com/scality/Arsenal/pull/1933/commits/4b19cc2101f4a15666f8746ad8fb12533e9b45d8#diff-7aa7eeaed88643ea97d57a5f15ddc2e64c380190641c53bb0c3f64d67807a475R40

lib/s3middleware/tagging.ts
ghost: Done: https://github.com/scality/Arsenal/pull/1933/commits/4b19cc2101f4a15666f8746ad8fb12533e9b45d8#diff-7aa7eeaed88643ea97d57a5f15ddc2e64c380190641c53bb0c3f64d67807a475R117

lib/s3middleware/tagging.ts
```diff
@@ -100,7 +114,46 @@ function _validateTags(tags: Array<{ Key: string[], Value: string[] }>) {
...
+            'Cannot provide multiple Tags with the same key'
+        );
+    }
+    return tagsResult;
+}
+
+/** validateTags - Validate tags, throwing an error if tags are invalid
+* @param tags - tags parsed from xml to be validated
+* @param tags[].Key - Name of the tag
+* @param tags[].Value - Value of the tag
```
miniscruff: I don't think either of these are required as the tags type already describes them

lib/s3middleware/tagging.ts
```diff
@@ -100,7 +114,46 @@ function _validateTags(tags: Array<{ Key: string[], Value: string[] }>) {
...
+* @return tagsResult - return object tags on success
+* { key: value }; error on failure
+*/
+export function validateTags(tags: Array<BucketTag>) {
```
miniscruff: Ok after looking again I have two comments on this function.
1. Why do we return the tags back? If the tags are valid then the input is the same as the output, if they are not valid then its an error response. Why not just rename the function to `areTagsValid` and return a bool or false? 
2. Or if you want an explicit error keep it validate tags but return `error | undefined` instead.

lib/models/BucketInfo.ts
```diff
@@ -240,6 +243,11 @@ export default class BucketInfo {
             READ_ACP: [],
         };
 
+        if (tags === undefined) {
+            tags = [] as BucketTag[];
+        }
+        validateTags(tags);
```
miniscruff: Should this handle the return value? Calling this without handling the return seems weird.

miniscruff (changes requested):

lib/s3middleware/tagging.ts
ghost: I removed them and updated the function jsdoc: https://github.com/scality/Arsenal/pull/1933/commits/f7d360fe0b9fd632254350844ac8e1c4e50de1c3#diff-7aa7eeaed88643ea97d57a5f15ddc2e64c380190641c53bb0c3f64d67807a475R124

lib/s3middleware/tagging.ts
ghost: I agree with approach 1, the tags are also validated in the APIs from the provided XML, so throwing an error here is not required. Done in https://github.com/scality/Arsenal/pull/1933/commits/f7d360fe0b9fd632254350844ac8e1c4e50de1c3#diff-7aa7eeaed88643ea97d57a5f15ddc2e64c380190641c53bb0c3f64d67807a475R128

lib/models/BucketInfo.ts
ghost: Thanks to your suggestion above, the new function can be passed to an assert to make sure it returns `true`: https://github.com/scality/Arsenal/pull/1933/commits/f7d360fe0b9fd632254350844ac8e1c4e50de1c3#diff-4ee3cd5de98073dadb72b27f034f4979480c04388cc4f4538f4895373527f174R249

miniscruff (approved):

miniscruff (approved):
William didn't like my first approval

PR #1934, opened 2022-06-15
Bugfix/arsn 232 add missing actions

ghost:
We are missing two permissions for the Versioned variant of these APIs.


XinLi0207 (approved):

ghost (approved):

ghost:
Approving as I confirmed in the product that the fix is working.

PR #1935, opened 2022-06-17
improvement: ARSN-225-add User Tag actionMaps

XinLi0207:
add new iam actions for User Tag feature
these actions are iam policy controlled.


ghost (approved):
LGTM, maybe you also want to release Arsenal? In this case, you can bump the project version, and also deal with integration branches as these APIs will be also added to 8.x.

miniscruff (approved):

rachedbenmustapha (changes requested):
According to [docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_tags_users.html) the case is wrong. Not that matching is case-sensitive (not sure about that), but the golden source should be properly written.

XinLi0207:
@RachedBenMustapha thanks a lot for catching this!

rachedbenmustapha (approved):

PR #1944, opened 2022-06-30
Feature: ARSN-235 update object before deleting it

welansari:
Issue: [ARSN-235](https://scality.atlassian.net/browse/ARSN-235)

Note: _The full changes are in the integration PR_

Due to Armory work, any changes affecting the MongoClientInterface needs to be pushed both in 7.x and 8.x.
Some tests are incompatible with 7.x due to the ObjectMD model being different. These tests were skipped in 7.x.

Two new properties were added to the ObjectMD (deleted, isPHD). The commits are only applied to the 8.x branch for obvious reasons.

Bucket Notification as well as Cold Storage need object metadata when this one gets deleted. The MongoDB oplog's delete events do not contain any of the metadata of the deleted object. They only contain the "_id" of the object.

This issue was addressed by updating the metadata of the object while adding a new tag that shows that the object is being deleted, and then immediately deleting the object. This will add an update event to the oplog containing all of the needed metadata.

Cloudserver Build : https://eve.devsca.com/github/scality/cloudserver/#/builders/15/builds/9413

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -16,6 +16,7 @@ const constants = require('../../../constants');
 const { reshapeExceptionError } = require('../../../errorUtils');
 const errors = require('../../../errors').default;
 const BucketInfo = require('../../../models/BucketInfo').default;
+const ObjectMD = require('../../../models/ObjectMD').default;
 const jsutil = require('../../../jsutil');
 
 const MongoClient = require('mongodb').MongoClient;
@@ -67,13 +68,6 @@ function generateVersionId(replicationGroupId) {
 function inc(str) {
     return str ? (str.slice(0, str.length - 1) +
             String.fromCharCode(str.charCodeAt(str.length - 1) + 1)) : str;
-}
-
-function generatePHDVersion(versionId) {
-    return {
-        isPHD: true,
-        versionId,
-    };
 }
 
 /**
@@ -1107,7 +1101,7 @@ class MongoClientInterface {
         this.getLatestVersion(c, objName, vFormat, log, (err, version) => {
             if (err && !err.is.NoSuchKey) {
                 log.error('getLatestVersion: error getting latest version',
-                    { error: err.message });
+                    { error: err.message, bucket: bucketName, key: objName });
                 return cb(err);
             }
             if ((err && err.is.NoSuchKey) || (version.isDeleteMarker && vFormat === BUCKET_VERSIONS.v1)) {
@@ -1117,15 +1111,21 @@ class MongoClientInterface {
                 // atomic condition on the PHD flag and the mst
                 // version:
                 // eslint-disable-next-line
-                c.findOneAndDelete({
-                    '_id': masterKey,
+                const filter = {
                     'value.isPHD': true,
                     'value.versionId': mst.versionId,
-                }, {}, err => {
+                };
+                this.internalDeleteObject(c, bucketName, masterKey, filter, log, err => {
                     if (err) {
+                        // the PHD master might get updated when a PUT is performed
+                        // before the repair is done, we don't want to return an error
+                        // in this case
+                        if (err.is.NoSuchKey) {
+                            return cb(null);
+                        }
                         log.error(
-                            'findOneAndDelete: error finding and deleting',
-                            { error: err.message });
+                            'deleteOrRepairPHD: error deleting object',
+                            { error: err.message, bucket: bucketName, key: objName });
                         return cb(errors.InternalError);
                     }
                     // do not test result.ok === 1 because
@@ -1162,33 +1162,47 @@ class MongoClientInterface {
         const masterKey = formatMasterKey(objName, params.vFormat);
         const versionKey = formatVersionKey(objName, params.versionId, params.vFormat);
         const _vid = generateVersionId(this.replicationGroupId);
-        const mst = generatePHDVersion(_vid);
-        c.bulkWrite([{
-            updateOne: {
-                filter: {
+        async.series([
+            next => c.updateOne(
+                {
+                    // Can't filter out objects with deletiong flag
+                    // as it will try and recreate an object with the same _id
+                    // instead we reset the flag to false, the data might be
+                    // inconsistent with the current state of the object but
+                    // this is not an issue as the object is in a temporary
+                    // placeholder (PHD) state
                     _id: masterKey,
                 },
-                update: {
-                    $set: { _id: masterKey, value: mst },
+                {
+                    $set: {
+                        '_id': masterKey,
+                        'value.isPHD': true,
+                        'value.versionId': _vid,
+                        'value.deleted': false,
+                    },
                 },
-                upsert: true,
-            },
-        }, {
-            deleteOne: {
-                filter: {
-                    _id: versionKey,
-                },
-            },
-        }], {
-            ordered: true,
-        }, err => {
+                { upsert: true },
+                next,
+            ),
+            // delete version
+            next => this.internalDeleteObject(c, bucketName, versionKey, {}, log,
+                err => {
+                    // we don't return an error in case we don't find
+                    // a version as we expect this case when dealing with
+                    // a versioning suspended object.
+                    if (err && err.is.NoSuchKey) {
+                        return next(null);
+                    }
+                    return next(err);
+                }),
+        ], err => {
             if (err) {
                 log.error(
-                    'deleteObjectVerMaster: error deleting object',
-                    { error: err.message });
+                    'deleteObjectVerMaster: error deleting the object',
+                    { error: err.message, bucket: bucketName, key: objName });
                 return cb(errors.InternalError);
             }
-            return this.deleteOrRepairPHD(c, bucketName, objName, mst, params.vFormat, log, cb);
+            return this.deleteOrRepairPHD(c, bucketName, objName, { versionId: _vid }, params.vFormat, log, cb);
         });
     }
 
@@ -1207,19 +1221,17 @@ class MongoClientInterface {
      */
     deleteObjectVerNotMaster(c, bucketName, objName, params, log, cb) {
         const versionKey = formatVersionKey(objName, params.versionId, params.vFormat);
-        c.findOneAndDelete({
-            _id: versionKey,
-        }, {}, (err, result) => {
+        this.internalDeleteObject(c, bucketName, versionKey, {}, log, err => {
             if (err) {
+                if (err.is.NoSuchKey) {
+                    log.error(
+                        'deleteObjectVerNotMaster: unable to find target object to delete',
+                        { error: err.message, bucket: bucketName, key: objName });
+                    return cb(errors.NoSuchKey);
+                }
                 log.error(
-                    'findOneAndDelete: error when version is not master',
-                    { error: err.message });
-                return cb(errors.InternalError);
-            }
-            if (result.ok !== 1) {
-                log.error(
-                    'findOneAndDelete: failed when version is not master',
-                    { error: err.message });
+                    'deleteObjectVerNotMaster: error deleting object with no version',
+                    { error: err.message, bucket: bucketName, key: objName });
                 return cb(errors.InternalError);
             }
             return cb(null);
@@ -1248,10 +1260,14 @@ class MongoClientInterface {
                 // find the master version
                 c.findOne({
                     _id: masterKey,
+                    $or: [
+                        { 'value.deleted': { $exists: false } },
+                        { 'value.deleted': { $eq: false } },
+                    ],
                 }, {}, (err, mst) => {
                     if (err) {
                         log.error('deleteObjectVer: error deleting versioned object',
-                            { error: err.message });
+                            { error: err.message, bucket: bucketName, key: objName });
                         return cb(errors.InternalError);
                     }
                     return next(null, mst);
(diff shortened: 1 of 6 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1295,25 +1341,88 @@ class MongoClientInterface {
...
+                const objMetadata = new ObjectMD(doc.value);
+                // adding delete flag to object metadata
+                objMetadata.setDeleted(true);
+                // updating originOp
+                objMetadata.setOriginOp('s3:ObjectRemoved:Delete');
+                return next(null, objMetadata.getValue());
+            }),
+            (objMetadata, next) => collection.bulkWrite([
+                {
+                    updateOne: {
```
francoisferrand, line 1397: there may be some race condition here: e.g. if the document is updated after it has been read (line 1389) but before we re-set its metadata... and in this case we would reset some older (outdated) metadata...

this should not happen with version documents, which are immutable, but may be the case when updating the master (in non-versioned bucket).

→ i think we should add an extra condition, to ensure we only update/delete the right document.

welansari: IMO Adding that extra condition will lead to adding more logic for handling what to do in case we don't find that version of the object (retry logic). I think that'll hinder the performance of deletion operations for little benefit.

I think deletion events should not be used to get volatile information such as tags and such, we always get update events for those types of operations, and those events should be the ones used instead.

Kerkesni: Also i think having deletion operations that occur at the same time as update ones should not be a common occurence so we should be good for the most part

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1162,30 +1172,62 @@ class MongoClientInterface {
...
+                objMetadata.setIsPHD(true);
+                // updating versionId
+                objMetadata.setVersionId(_vid);
+                return next(null, objMetadata.getValue());
+            }),
+            // Updating/creating master object
+            (mst, next) => c.updateOne(
+                { _id: masterKey },
+                {
+                    $set: { _id: masterKey, value: mst },
```
francoisferrand: Somehow, i am a bit inconfortable with this : we get the (mutable) metadata and store in the master, which may create a race condition (e.g. we put an older copy of the metadata)...

For many fields this may be acceptable, as they are actually immutable; and for some other fields we don't really care... But putting everything in there means we may later be tempted to use the whole information, even if parts of it may not be correct.

Maybe that is a trade-off we must make, but I would like other opinions on this...

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1295,25 +1341,88 @@ class MongoClientInterface {
...
+     * @param {object} filter additional query filters
+     * @param {Logger}log logger instance
+     * @param {Function} cb callback containing error
+     * and BulkWriteResult
+     * @return {undefined}
+     */
+    internalDeleteObject(collection, bucketName, key, filter, log, cb) {
+        // filter used when finding and updating object
+        const findUpdateFilter = Object.assign({
+            _id: key,
```
francoisferrand, line 1355: should check 'value.deleted' as well, as a safety measure?

lib/storage/metadata/mongoclient/MongoClientInterface.js
francoisferrand, line 1397: same here.... though the race condition may be easy to fix:

1. Do a simple update to simply set the delete flag
2. Once it has been marked as deleted, get the full metadata (actually, it may be returned by the first call, if we do a `findAndUpdate(_id: key, delete: false)`)
3. Finalize the delete with a bulkwrite to replace the value and delete the object

francoisferrand: > I think deletion events should not be used to get volatile information such as tags and such, we always get update events for those types of operations, and those events should be the ones used instead.

Yes, i agree. But if the information is available, it will eventually be used (see [Hyrum's law](https://www.hyrumslaw.com/)) : so i think either we need to go the extra mile and protect against this, or we need to remove the mutable metadata...

lib/storage/metadata/mongoclient/MongoClientInterface.js
nicolas2bert: This master value seems to be either deleted or overwritten by deleteOrRepairPHD or external "PUT/DELETE" object calls.
It seems that, here, we use it exclusively for the MongoDB oplog's delete events https://github.com/scality/Arsenal/pull/1944/files#diff-9c207befbd8f4d79a3b6e3d7d143444e804ca1684236b8fd9fa0f9e03c53477aR1124 and since any "outdated master" won't be deleted, oplog state should be fine.
However, even if it does not seem to be mandatory in our case, we should find a way to avoid potential race conditions (with findAndModify or condition expressions...).

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1295,25 +1341,88 @@ class MongoClientInterface {
...
+        }, filter);
+        // filter used when deleting object
+        const deleteFilter = Object.assign({
+            '_id': key,
+            'value.deleted': true,
+        }, filter);
+        async.waterfall([
```
rachedbenmustapha, line 1366: Just to make sure I understand correctly, we're only having a race condition here because:
- we first do a get+unserialize+modify using application-level setters
- we then do a bulk sequence of write+delete
- these 2 operations, done non-atomically, can be interrupted in the middle or interleaved with a change from another request

Can't the waterfall be replaced by a single `bulk-write(ordered: true, upsert: false)` with direct `$set`s of `originOp` and `deleted` (doesn't really matter that encapsulation's broken, since we're already direct-querying on `deleted`) and conditional flags - as long as we have `value.deleted: false` for the conditional update. Upon checking the bulk-write result, if the first step failed, just return an error to the caller; if the second step failed, return success (some mopping mechanism may have to run outside though as although rare, it's quite impossible to guarantee that this won't happen without transactions, as mongodb's quorum could be broken at any time).

Once we move to mongodb 4.2+, transactions can be used to do all of this atomically on sharded clusters.

Am I missing something?

(cc @francoisferrand)

lib/storage/metadata/mongoclient/MongoClientInterface.js
rachedbenmustapha: +1

rachedbenmustapha (commented):
I'm curious, why is this implemented on the 7.10 branch while skipping incompatible bits? don't all users of this API use arsenal 8.x?

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: Oplog update events only show the updated fields, so if we directly update only `originOp` and `deleted`, that's all we are going to see in the oplog, and the event will look like this :
```
  "ts" : Timestamp(1666765379, 3),
    "t" : NumberLong(14),
    "h" : NumberLong(-8924670320665905898),
    "v" : 2,
    "op" : "u",
    "ns" : "metadata.arsenal-bucket",
    "ui" : UUID("09301a87-0bb6-4eb3-9ca4-11c890492d1f"),
    "o2" : {
        "_id" : "Vobject-1\u000098333234628661999999RG001  54937.0"
    },
    "wall" : ISODate("2022-10-26T06:22:59.019Z"),
    "lsid" : {
        "id" : UUID("901cb895-39af-42ce-8082-6f78f68bca8f"),
        "uid" : { "$binary" : "47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=", "$type" : "00" }
    },
    "txnNumber" : NumberLong(5),
    "stmtId" : 0,
    "prevOpTime" : {
        "ts" : Timestamp(0, 0),
        "t" : NumberLong(-1)
    },
    "postImageOpTime" : {
        "ts" : Timestamp(1666765379, 2),
        "t" : NumberLong(14)
    },
    "o" : {
        "$v" : 1,
        "$set" : {
            "value.originOp" : "s3:ObjectRemoved:Delete"
            "value.deleted" : true
        }
    }
```
This is why we first `Find` and then `Update` the whole metadata

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: To remove any race condition, i decided to stick with the original implementation of the function, where we directly update the master, this should hopefully remove any inconsistency in the metadata.

The original reason of why we first `Find` and then `Update` the metadata is because of one particular case where we don't have a master object. This happens when the last version is a delete marker, and only happens when the bucket is versioned. In this case we don't really have a choice i think but to get the version and then create the master with its data.

So now in this case we'll have an object with empty metadata kind of like it was before, only containing the `isPHD` flag and the versionId, this is not really an issue as the solution to this would just be to use version object's events when dealing with a versioned bucket as they contain everything.

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: The reason this was implemented in 7.x is as Ronnie said [here](https://scality.slack.com/archives/CFD81UFEK/p1657214530430809?thread_ts=1657208435.407199&cid=CFD81UFEK), cause Armory is meant to be used in both 7.x and 8.x branch, so we want to keep the lib/storage folder identical in both branches to avoid issues until the code is moved to Armory

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1162,33 +1172,40 @@ class MongoClientInterface {
...
-                    $set: { _id: masterKey, value: mst },
-                },
-                upsert: true,
-            },
-        }, {
-            deleteOne: {
-                filter: {
-                    _id: versionKey,
+                {
+                    $set: {
```
francoisferrand, line 1187: wondering if we should have a condition here, e.g. update the master conditionally when value is the indeed the versionId we are removing ? If the condition fails, we could then just fallback to deleting the object without PHD, as another version has been pushed already...

(not really needed for this specific bug fix, but may avoid setting a PHD when the version pointing to master is removed at the same time as a new version is added)

Not sure we need to rush to fix that, we should also consider what is the "best" imprecision.... Maybe to be discussed with experts first and/or create a separate improvement ticket?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1295,25 +1316,112 @@ class MongoClientInterface {
...
+                }
+                if (!doc.value) {
+                    log.error('internalDeleteObject: unable to find target object to delete',
+                        { bucket: bucketName, object: key });
+                    return next(errors.NoSuchKey);
+                }
+                const obj = doc.value;
+                const objMetadata = new ObjectMD(obj.value);
+                objMetadata.setOriginOp('s3:ObjectRemoved:Delete');
+                objMetadata.setDeleted(true);
```
francoisferrand, line 1390: should already be set?
or we need to add it, since it is not "known" by ObjectMD (and it should be added there, possibly as an "optional" field,  to make sure the model is actually representative)

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: Yes i think we should do this in a separate improvement ticket

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: The mongo memory server we use for our functional tests don't support the `returnDocument: 'after'` argument for some reason even when explicitly setting the mongo version to 4.0.27. so it always returns the version of the object before the update

lib/storage/metadata/mongoclient/MongoClientInterface.js
welansari: ticket: ARSN-275

lib/storage/metadata/mongoclient/MongoClientInterface.js
francoisferrand: it is weird that it does not work... we may accept it, but then the code still seamingly depends on `returnDocumet:'after'` (requested line 1377) : so either we need to fix this, or do not even try to request the updated document (or at least at a comment and create a TechDebt ticket)

francoisferrand (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1295,25 +1316,111 @@ class MongoClientInterface {
...
+            '_id': key,
+            'value.deleted': true,
+        }, filter);
+        async.waterfall([
+            // Adding delete flag when getting the object
+            // to avoid having race conditions.
+            next => collection.findOneAndUpdate(findFilter, {
+                $set: {
+                    '_id': key,
+                    'value.deleted': true,
```
nicolas2bert, line 1372: About `Bucket Notification as well as Cold Storage need object metadata when this one gets deleted`, what metadata do we need exactly? (and why?)

If the full object metadata is needed,

"updating `'value.deleted': true` here" would make the function non-atomic.
`collection.findOneAndUpdate()` could succeed then `collection.bulkWrite(updateOne + deleteOne)` could fail, leaving an object with "deleted:true" state. Then, if we try the `internalDeleteObject` call again `collection.findOneAndUpdate()` would fail with a "NoSuchKey" error. 

I don't think it is an issue since we can use a cleaning logic to get rid of those `deleted:true` object metadata. (If it doest not exist already)

We would ignore the `collection.bulkWrite(updateOne + deleteOne)` potential failures and return no error on failure.
(The issue I see with that is the oplog's delete events with metadata won't be fired whereas the object will be considered deleted. "Clean up job" should be able to take care of it?)

We need to make sure "objects with `deleted:true` state" are considered as "deleted" internally when we CRUD them.

francoisferrand: > About Bucket Notification as well as Cold Storage need object metadata when this one gets deleted, what metadata do we need exactly? (and why?)

AFAIK:
- For bucket notification, we need to find out if this is a MASTER without versioning (→ notify) or a MASTER with versioning (→ notify only when removing the 'version' entry)
- For cold storage, we need to remove (or mark as removed) the object from cold storage, so we need the archive id i guess

welansari: @francoisferrand yes, but during the meeting we had with the architects (Vianney & Rached), we were asked to keep all of the metadata.

welansari: > We need to make sure "objects with deleted:true state" are considered as "deleted" internally when we CRUD them.

This is the case, i fixed the last missing part for the delete operations.

What's left to do is making a job that gets rid of the objects tagged for deletion. This job will also need to update the metadata before deletion to keep them in the oplog. This would need to be done in another ticket

welansari: Ticket number : [S3UTILS-113](https://scality.atlassian.net/browse/S3UTILS-113)

francoisferrand: > @nicolas2bert we can use a cleaning logic to get rid of those deleted:true object metadata. (If it doest not exist already)
> @Kerkesni  What's left to do is making a job that gets rid of the objects tagged for deletion

Seems to me we may use lifecycle for this purpose (possibly with a hidden/hard-coded workflow?), esp. once we have the "indexing" optimizations ? Or we can just add a cronjob, should be fairly simple to do.

Anyway, I think this should happen only infrequently (e.g. not in a properly working system): since this would be the only place where we try to "use" the `delete=true` entry. So I guess we can go ahead like this, do you agree @nicolas2bert ?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1295,25 +1327,111 @@ class MongoClientInterface {
...
+                            $set: { _id: key, value: objMetadata },
+                        },
+                        upsert: false,
+                    },
+                }, {
+                    deleteOne: {
+                        filter: updateDeleteFilter,
+                    },
+                },
+            ], { ordered: true }, next),
```
nicolas2bert: We might not want to return an error if the `collection.bulkWrite()` fails.
At this stage, the operation is already considered "completed" with `deleted:true` set.
Also, the user/client won't be able to retry the operation in case of failure at this stage and it would assume the operation has failed.

welansari: should be good now

nicolas2bert (approved):

PR #1946, opened 2022-07-05
ARSN-236 Put bucket replication to dmf is not supported

nicolas2bert:


nicolas2bert:
ping

lib/models/ReplicationConfiguration.ts
```diff
@@ -359,6 +359,11 @@ export default class ReplicationConfiguration {
                 (endpoint: any) => endpoint.site === storageClass
             );
             if (endpoint) {
+                // We do not support replication to dmf location. 
+                // Only transition to dmf location is supported.
+                if (endpoint.type === 'dmf') {
```
francoisferrand: to make future updates (e.g. adding new kinds of cold-storage locations) easier & safer, should we not introduce some kind of abstraction here already?
e.g. some method/function to check if an endpoint is cold-storage only (or if it supports replication)

francoisferrand (approved):

lib/models/ReplicationConfiguration.ts
nicolas2bert: I want to introduce as little disruption as possible since it should be a temporary limitation. (We should be able to replicate to cold storage location later).
Since I just remembered that the locations' list was part of the config I will leverage it, thank you :).

rachedbenmustapha (approved):

nicolas2bert:
@bert-e approve

PR #1958, opened 2022-07-08
ARSN-242 Fix non-current version transition

nicolas2bert:


lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
```diff
@@ -239,6 +316,16 @@ export default class LifecycleUtils {
                     currentDate,
                 });
             }
+
+            const hasNoncurrentVersionTransitions = Array.isArray(rule.NoncurrentVersionTransitions) && rule.NoncurrentVersionTransitions.length > 0;
+            if (hasNoncurrentVersionTransitions && this._supportedRules.includes('noncurrentVersionExpiration')) {
```
rachedbenmustapha: should the supported rule be `noncurrentVersionTransition` ?

lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
```diff
@@ -98,6 +139,42 @@ export default class LifecycleUtils {
         });
     }
 
+    /**
+    * Find the most relevant non-current version trantition rule for the given transitions array
```
rachedbenmustapha:
```suggestion
    * Find the most relevant non-current version transition rule for the given transitions array
```

lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
```diff
@@ -98,6 +139,42 @@ export default class LifecycleUtils {
...
+    * and any previously stored non-current version transition from another rule.
+    * @param params - The function parameters
+    * @param params.transitions - Array of lifecycle non-current version transitions
+    * @param params.lastModified - The object's last modified
+    * date
+    * @return The most applicable non-current version transition rule
+    */
+    getApplicableNCVTransition(params: {
```
rachedbenmustapha, line 151: More of a question: given the new constraint that we can't transition out of cold storage (yet?) is it something that should be handled here, or at the call site? Call site seems to make more sense as the backbeat bucket processor has access to the location's `isCold` attribute.

lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
nicolas2bert: Yes, thanks!

lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
nicolas2bert: Yes, I think "backbeat bucket processor" is a good place to handle it. 

Also, we can't replication an object located in a Cold storage location. We will need to handle this limitation as well. 

Note: We already handle replicating "to" a Cold storage location but not "from" there.

nicolas2bert: I created a ticket for it: https://scality.atlassian.net/browse/BB-241

rachedbenmustapha (approved):

miniscruff (approved):

nicolas2bert:
@bert-e approve

PR #1970, opened 2022-07-21
bf(ARSN-250): Fix getByteRangeFromSpec when range is 0-0

tmacro:


jonathan-gramain (approved):
good catch :monocle_face:

watsaqat (approved):

PR #1975, opened 2022-08-01
bugfix/ARSN-253: issue with special unicode chars

KillianG:
fixing the problem with unicode special chars by encoding them with URI
Problem was that our encoreURI function was not working properly for special chars

### DIFF
--- a/lib/auth/v4/awsURIencode.ts
+++ b/lib/auth/v4/awsURIencode.ts
@@ -44,8 +44,9 @@ export default function awsURIencode(
     if (typeof input !== 'string') {
         return encoded;
     }
-    for (let i = 0; i < input.length; i++) {
-        const ch = input.charAt(i);
+    let charArray = Array.from(input);
+    for (let i = 0; i < charArray.length; i++) {
+        const ch = charArray[i];
         if ((ch >= 'A' && ch <= 'Z') ||
             (ch >= 'a' && ch <= 'z') ||
             (ch >= '0' && ch <= '9') ||
@@ -59,7 +60,7 @@ export default function awsURIencode(
         } else if (ch === '*') {
             encoded = encoded.concat(noEncodeStar ? '*' : '%2A');
         } else {
-            encoded = encoded.concat(ch);
+            encoded = encoded.concat(_toHexUTF8(ch));
         }
     }
     return encoded;
--- a/lib/auth/v4/awsURIencode.ts
+++ b/lib/auth/v4/awsURIencode.ts
@@ -59,7 +59,7 @@ export default function awsURIencode(
         } else if (ch === '*') {
             encoded = encoded.concat(noEncodeStar ? '*' : '%2A');
         } else {
-            encoded = encoded.concat(_toHexUTF8(ch));
+            encoded = encoded.concat(ch);
         }
     }
     return encoded;
--- a/lib/auth/v4/awsURIencode.ts
+++ b/lib/auth/v4/awsURIencode.ts
@@ -44,7 +44,7 @@ export default function awsURIencode(
     if (typeof input !== 'string') {
         return encoded;
     }
-    let charArray = Array.from(input);
+    const charArray = Array.from(input);
     for (let i = 0; i < charArray.length; i++) {
         const ch = charArray[i];
         if ((ch >= 'A' && ch <= 'Z') ||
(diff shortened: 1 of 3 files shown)


lib/auth/v4/awsURIencode.ts
```diff
@@ -44,8 +44,9 @@ export default function awsURIencode(
     if (typeof input !== 'string') {
         return encoded;
     }
-    for (let i = 0; i < input.length; i++) {
-        const ch = input.charAt(i);
+    let charArray = Array.from(input);
```
ghost: Nit: can be a const?

KillianG: Surely yeah: https://github.com/scality/Arsenal/pull/1975/commits/46c24c5cc39675cc921e6a22872c60dc23fdcf41

lib/auth/v4/awsURIencode.ts
```diff
@@ -44,8 +44,9 @@ export default function awsURIencode(
     if (typeof input !== 'string') {
         return encoded;
     }
-    for (let i = 0; i < input.length; i++) {
-        const ch = input.charAt(i);
+    const charArray = Array.from(input);
```
miniscruff, line 51: ( unrelated to this line, just using it for inline comments )
Idk how big the inputs can be but this func is actually quite inefficient using string concats and even worse the concat method itself ( https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat#performance ). I could rewrite it more efficiently if this is something we want to tackle. If the inputs are only ever like 15 characters long its whatever but I believe we use this quite often and the inputs could be quite long?

ghost: I just double checked and yes, the function is used to encode policy documents, canonical requests (they can be very long) and other but smaller resource names.

ghost (approved):
From my current knowledge, lgtm

lib/auth/v4/awsURIencode.ts
KillianG: If you know how to make it more efficiently we surely can make it, this is always a benefits to be more efficient..

miniscruff: Ok let me give it a shot ( closing conversation to open another one. )

miniscruff, line 51: Here is what I came up with:
```js
    /**
     * Duplicate query params are not suppported by AWS S3 APIs. These params
     * are parsed as Arrays by Node.js HTTP parser which breaks this method
    */
    if (typeof input !== 'string') {
        return '';
    }

    // precalc slash and star based on configs
    const slash = encodeSlash !== undefined && encodeSlash ? '%2F' : '/';
    const star = noEncodeStar === undefined || noEncodeStar ? '*' : '%2A';
    const encoded: string[] = [];

    const charArray = Array.from(input);
    for (const ch of charArray) {
        switch (true) {
            case ch >= 'A' && ch <= 'Z':
            case ch >= 'a' && ch <= 'z':
            case ch >= '0' && ch <= '9':
            case ch === '-':
            case ch === '_':
            case ch === '~':
            case ch === '.':
                encoded.push(ch);
                break;
            case ch === '/':
                encoded.push(slash);
                break;
            case ch === '*':
                encoded.push(star);
                break;
            case ch === ' ':
                encoded.push('%20');
                break;
            default:
                encoded.push(_toHexUTF8(ch));
                break;
        }
    }
    return encoded.join('');
```

KillianG: Thanks a lot Ronnie, I added it with this commit: https://github.com/scality/Arsenal/pull/1975/commits/e9db7da105a898a198590a9fe0b450f52085b504

ghost (approved):

lib/auth/v4/awsURIencode.ts
```diff
@@ -35,32 +35,44 @@ export default function awsURIencode(
...
-    const encSlash = encodeSlash === undefined ? true : encodeSlash;
-    let encoded = '';
     /**
      * Duplicate query params are not suppported by AWS S3 APIs. These params
      * are parsed as Arrays by Node.js HTTP parser which breaks this method
-    */
-    if (typeof input !== 'string') {
-        return encoded;
+     */
+     if (typeof input !== 'string') {
```
miniscruff, line 42: is there an extra space here?

miniscruff (approved):
Remove the space if so, otherwise its good

PR #1982, opened 2022-08-10
Bugfix/arsn 252 listing bug versioned bucket

watsaqat:
Addresses https://scality.atlassian.net/browse/S3C-4682.

The `skippingBase` fn is used in Metadata:RepdServer to determine the token from which we can skip to the next key range. Because of [Zenko-1048](https://scality.atlassian.net/browse/ZENKO-1048), this token is not updated correctly and when a key ends with a delimiter, an optimization in RepdServer ends up being allowed to skip all sub-prefixes in the range. This PR adds a version id separator char after tokens ending with the delimiter. Now, when the range is incremented the sub-prefixes  are listed.

### DIFF
--- a/lib/algos/list/delimiterMaster.js
+++ b/lib/algos/list/delimiterMaster.js
@@ -62,6 +62,10 @@ class DelimiterMaster extends Delimiter {
         let key = obj.key;
         const value = obj.value;
 
+        if (key === this.prefix) {
+            return this.addContents(key, value);
+        }
+
         if (key.startsWith(DbPrefixes.Replay)) {
             this.inReplayPrefix = true;
             return FILTER_SKIP;
@@ -96,8 +100,7 @@ class DelimiterMaster extends Delimiter {
              *   value. (TODO: remove this test once ZENKO-1048 is fixed)
              *   */
             if (key === this.prvKey || key === this[this.nextContinueMarker] ||
-                (this.delimiter &&
-                key.startsWith(this[this.nextContinueMarker]))) {
+                (this.delimiter && key.startsWith(this[this.nextContinueMarker]))) {
                 /* master version already filtered */
                 return FILTER_SKIP;
             }
--- a/tests/unit/algos/list/delimiterMaster.spec.js
+++ b/tests/unit/algos/list/delimiterMaster.spec.js
@@ -1,7 +1,9 @@
 'use strict'; // eslint-disable-line strict
 
 const assert = require('assert');
+const chance = require('chance').Chance(); // eslint-disable-line
 
+const { getCommonPrefix } = require('../../../../lib/algos/list/delimiter');
 const DelimiterMaster =
     require('../../../../lib/algos/list/delimiterMaster').DelimiterMaster;
 const {
@@ -16,8 +18,6 @@ const Version = require('../../../../lib/versioning/Version').Version;
 const { generateVersionId } = require('../../../../lib/versioning/VersionID');
 const { DbPrefixes } = VSConst;
 const zpad = require('../../helpers').zpad;
-
-
 const VID_SEP = VSConst.VersionId.Separator;
 const EmptyResult = {
     CommonPrefixes: [],
@@ -46,6 +46,111 @@ function getListingKey(key, vFormat) {
     return assert.fail(`bad vFormat ${vFormat}`);
 }
 
+const MAX_STREAK_LENGTH = 100;
+
+function createStreakState(prefix, vFormat) {
+    return {
+        vFormat,
+        prefix,
+        prefixes: new Set(),
+        params: {},
+        streakLength: 0,
+        gteparams: null,
+    };
+}
+
+/**
+ * Simplified version of logic found in Metadata's RepdServer. When MAX_STREAK_LENGTH (typically 100)
+ * is reached, we attempt to skip to the next listing range as a performance optimization.
+ * @param {Integer} filteringResult - 0, 1, -1 indicating if we can skip to the next character range.
+ * @param {String} skippingRange - lower bound that the listing can begin from.
+ * @returns {undefined}
+ */
+/* eslint-disable no-param-reassign */
+function handleStreak(filteringResult, skippingRange, state) {
+    if (filteringResult < 0) {
+        // readStream would be destroyed. In this mock, we just continue.
+    } else if (filteringResult === 0 && skippingRange) {
+        // check if MAX_STREAK_LENGTH consecutive keys have been
+        // skipped
+        if (++state.streakLength === MAX_STREAK_LENGTH) {
+            if (Array.isArray(skippingRange)) {
+                // With synchronized listing, the listing
+                // algo backend skipping() function must
+                // return as many skip keys as listing
+                // param sets.
+                for (let i = 0; i < skippingRange.length; ++i) {
+                    state.params[i].gte = inc(skippingRange[i]);
+                }
+            } else {
+                state.params.gte = inc(skippingRange);
+            }
+            if (state.gteparams && state.gteparams === state.params.gte) {
+                state.streakLength = 1;
+            } else {
+                // stop listing this key range
+                state.gteparams = state.params.gte;
+            }
+        }
+    } else {
+        state.streakLength = 1;
+    }
+}/* eslint-enable */
+
+/**
+ * Generate a random number of versioned keys.
+ * @param {String} masterKey - base key used to derive version keys
+ * @param {Integer} numKeys - how many versioned keys to generate
+ * @param {Object} state - test case state
+ * @yields {Object} - { key, isDeleteMarker }
+ * @returns {undefined}
+ */
+function *generateVersionedKeys(masterKey, numKeys, state) {
+    for (let i = 0; i < numKeys; i++) {
+        yield {
+            key: `${masterKey}${VID_SEP}${zpad(i)}`,
+            isDeleteMarker: state.vFormat === 'v0' && chance.bool(),
+            canSkip: true,
+        };
+    }
+}
+
+/**
+ * Generate raw listing keys in alphabetical order.
+ * @param {Integer} count - how many keys to generate.
+ * @param {Object} state - state for test run.
+ * @yields {Object} - { key, isDeleteMarker }
+ */
+
+function *generateKeys(count, state) {
+    let idx = 0;
+    let masterKey = '_Common-Prefix/';
+    yield { key: masterKey, isDeleteMarker: false, canSkip: false };
+
+    idx++;
+    while (idx < count) {
+        masterKey = `_Common-Prefix/${zpad(idx)}`;
+        const masterKeyWithSubprefix = `_Common-Prefix/${zpad(idx)}/${zpad(idx)}`;
+
+        const hasSubprefix = chance.bool();
+        const isDeleteMarker = state.vFormat === 'v0' && chance.bool();
+        const baseKey = hasSubprefix ? masterKeyWithSubprefix : masterKey;
+
+        let canSkip = isDeleteMarker;
+        if (hasSubprefix || state.prefix[state.prefix.length - 1] !== '/') {
+            canSkip = canSkip || state.handleSubprefixKey(baseKey);
+        }
+
+        yield { key: baseKey, isDeleteMarker, canSkip };
+        idx++;
+
+        const versioned = chance.integer({ min: 0, max: 200 });
+        const allowedVersionedKeys = Math.min(count - idx, versioned);
+        yield* generateVersionedKeys(baseKey, allowedVersionedKeys, state);
+        idx += allowedVersionedKeys;
+    }
+}
+
 ['v0', 'v1'].forEach(vFormat => {
     describe(`Delimiter All masters listing algorithm vFormat=${vFormat}`, () => {
         it('should return SKIP_NONE for DelimiterMaster when both NextMarker ' +
@@ -57,6 +162,70 @@ function getListingKey(key, vFormat) {
             // When there is no NextMarker or NextContinuationToken, it should
             // return SKIP_NONE
             assert.strictEqual(delimiter.skipping(), SKIP_NONE);
+        });
+
+        it('should list all subprefixes when handleStreak is applied as in RepdServer', () => {
+            ['_Common-Prefix', '_Common-Prefix/'].forEach(prefix => {
+                const state = createStreakState(prefix, vFormat);
+                state.handleSubprefixKey = baseKey => {
+                    const isLastDelim = state.prefix[state.prefix.length - 1] === '/';
+                    const lastIdx = isLastDelim ? baseKey.lastIndexOf('/') : state.prefix.length;
+                    const prefix = isLastDelim ? getCommonPrefix(baseKey, '/', lastIdx) : baseKey.slice(0, lastIdx);
+                    if (state.prefixes.has(prefix) || (!isLastDelim && baseKey.length > lastIdx)) {
+                        return true;
+                    }
+
+                    state.prefixes.add(prefix);
+                    return false;
+                };
+
+                const maxKeys = 1000000;
+                const delimiter = new DelimiterMaster({
+                    maxKeys,
+                    prefix,
+                    delimiter: '/',
+                    startAfter: '',
+                    continuationToken: '',
+                    v2: true,
+                    fetchOwner: false }, fakeLogger, vFormat);
+
+                [...generateKeys(maxKeys, state)].forEach(ob => {
+                    const { key, isDeleteMarker, canSkip } = ob;
+                    // simulate not doing a raw listing when key < params.gte
+                    // this represents a key not reaching the read stream from leveldb
+                    if (state.params.gte && key < state.params.gte) {
+                        return;
+                    }
+
+                    if (!key.includes(VID_SEP)) { // master key
+                        const version = new Version({ isDeleteMarker });
+                        const obj = {
+                            key: getListingKey(key, vFormat),
+                            value: version.toString(),
+                        };
+
+                        const res = delimiter.filter(obj);
+                        const skippingRange = delimiter.skipping();
+                        handleStreak(res, skippingRange, state);
+
+                        const expected = canSkip ? FILTER_SKIP : FILTER_ACCEPT;
+                        assert.strictEqual(res, expected);
+                    } else { // versioned key
+                        if (vFormat === 'v0') {
+                            const vid = key.split(VID_SEP).slice(-1)[0];
+                            const version = new Version({ versionId: vid, isDeleteMarker });
+                            const obj2 = {
+                                key: getListingKey(key, vFormat),
+                                value: version.toString(),
+                            };
+                            const res = delimiter.filter(obj2);
+                            const skippingRange = delimiter.skipping();
+                            handleStreak(res, skippingRange, state);
+                            assert.strictEqual(res, FILTER_SKIP);
+                        }
+                    }
+                });
+            });
         });
 
         it('should return <key><VersionIdSeparator> for DelimiterMaster when ' +
--- a/lib/algos/list/delimiter.js
+++ b/lib/algos/list/delimiter.js
@@ -271,4 +271,4 @@ class Delimiter extends Extension {
     }
 }
 
-module.exports = { Delimiter };
+module.exports = { Delimiter, getCommonPrefix };
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "@types/jest": "^27.4.1",
     "@types/node": "^17.0.21",
     "@types/xml2js": "^0.4.11",
+    "chance": "^1.1.8",
     "eslint": "^8.12.0",
     "eslint-config-airbnb": "6.2.0",
     "eslint-config-scality": "scality/Guidelines#7.10.2",
--- a/tests/unit/network/http/server.spec.js
+++ b/tests/unit/network/http/server.spec.js
@@ -166,7 +166,7 @@ describe('network.Server: ', () => {
                 if (err) {
                     return ws.onStop(() => {
                         clearTimeout(requestTimeout);
-                        if (err.code === 'EPROTO') {
+                        if (err.code === 'EPROTO' || err.code === 'ECONNRESET') {
                             return done();
                         }
                         return done(err);
--- a/yarn.lock
+++ b/yarn.lock
@@ -2144,6 +2144,11 @@ chalk@^4.0.0:
     ansi-styles "^4.1.0"
     supports-color "^7.1.0"
 
+chance@^1.1.8:
+  version "1.1.8"
+  resolved "https://registry.yarnpkg.com/chance/-/chance-1.1.8.tgz#5d6c2b78c9170bf6eb9df7acdda04363085be909"
+  integrity sha512-v7fi5Hj2VbR6dJEGRWLmJBA83LJMS47pkAbmROFxHWd9qmE1esHRZW8Clf1Fhzr3rjxnNZVCjOEv/ivFxeIMtg==
+
 char-regex@^1.0.2:
   version "1.0.2"
   resolved "https://registry.yarnpkg.com/char-regex/-/char-regex-1.0.2.tgz#d744358226217f981ed58f479b1d6bcc29545dcf"


lib/algos/list/delimiterMaster.js
```diff
@@ -95,12 +95,29 @@ class DelimiterMaster extends Delimiter {
...
+                    return FILTER_SKIP;
+                }
+                if (len > 1) {
+                    // key=foo/ or key=foo/bar
+                    if (key[len - 1] === '') {
+                        // key=foo/ : version of a previously seen master key.
+                        return FILTER_SKIP;
+                    }
+                    // Else key=foo/bar (or further nesting as foo/bar/baz/..)
+                    // This case should not be skipped and be handled in this.addContents below.
```
jonathan-gramain: Why so? If the key is `foo/bar` and the value of `this[this.nextContinueMarker]` is `foo/`, we must skip that key as it is part of a prefix already seen, I think?

jonathan-gramain (commented):
Can you confirm if this issue only happening with versioned buckets?

Also, make sure you check V1 format, or create a ticket to keep track that we need to do this check.

lib/algos/list/delimiterMaster.js
watsaqat: this is what gets handled on line 98

jonathan-gramain: Can you be more specific

watsaqat: I worded this poorly. Should be "should not be skipped *yet*.. ". It does indeed get skipped but this case is handled later in the function. Updated now.

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -59,6 +59,93 @@ function getListingKey(key, vFormat) {
             assert.strictEqual(delimiter.skipping(), SKIP_NONE);
         });
 
+        it('should not hide keys when a common prefix master is filtered first', () => {
+            // See S3C-4682 for details.
+            // Delimiter will call .filter multiple times with different keys.
+            // It should list them all masters keys except those with delete markers.
+            const delimiter = new DelimiterMaster({
+                prefix: '_Common-Prefix/',
```
jonathan-gramain:
```suggestion
                prefix: 'Common-Prefix/',
```

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -59,6 +59,93 @@ function getListingKey(key, vFormat) {
...
+                value: version.toString(),
+            };
+            // Do not skip master with lexicographically smallest key.
+            assert.strictEqual(delimiter.filter(obj), FILTER_ACCEPT);
+
+            // Skip these with delete markers.
+            // In S3C-4682 there are 514 ids with delete markers and a common prefix.
+            for (let idx = 0; idx < 514; idx++) {
+                // Keys have delete markers and are versioned
+                const keyVersion = `${masterKey}${VID_SEP}${idx}`;
```
jonathan-gramain: Correct me if I'm wrong but I don't think the test is maintaining the key ordering for two reasons, on one hand:

- in https://github.com/scality/Arsenal/pull/1982/files#diff-52fb3163eb0083606af1b564f1535ed6ae456e8b9ee0fbc42be4b22c749ed17dR79 we introduce the `obj.key` ending with the delimiter `/`, and we call `filter` with it
- here we create a version from the `masterKey` which is the same key without the final delimiter, then appending `VID_SEP`, but `VID_SEP` is the null character, lower than `/` in the key ordering. Did you mean to add versions to the same key ending with the delimiter character?

The second reason I see is more practical, we need to pad `idx`, e.g.:

```
const paddedIdx = `0000${idx}`.slice(-4);
```

If I'm not mistaken, the test could have to be reworked to filter keys in order, which maybe could trigger some new testing conditions.

lib/algos/list/delimiterMaster.js
```diff
@@ -95,12 +95,29 @@ class DelimiterMaster extends Delimiter {
...
-                (this.delimiter &&
-                key.startsWith(this[this.nextContinueMarker]))) {
+            if (key === this.prvKey || key === this[this.nextContinueMarker]) {
                 /* master version already filtered */
                 return FILTER_SKIP;
             }
+
+            if (this.delimiter && key.startsWith(this[this.nextContinueMarker])) {
+                // If this[continueMarker] ends with the delimiter then we have to treat it differently.
+                const candidate = key.split(this.delimiter);
```
jonathan-gramain: I don't think splitting the key is really needed for what we want to check, essentially:
- if the key ends with a delimiter character
- if the key contains no delimiter at all in its whole length (not sure why this check)

I'd rather avoid the split call and just do the same kind of checks on the whole key to avoid unnecessary processing and simplify the code.

lib/algos/list/delimiterMaster.js
```diff
@@ -95,12 +95,29 @@ class DelimiterMaster extends Delimiter {
...
+                const len = candidate.length;
+                if (len === 1) {
+                    // case key=foo: this is a version of a previously seen master key.
```
jonathan-gramain: I don't understand the logic, can you try to clarify? Basically I don't see the relationship between not having any delimiter character in the key (hence `len === 1` when we split) and this key being a version of a previously seen master key. I see that a similar case is skipped below when the key ends with a delimiter, but I think if the logic is correct it could deserve some clarification.

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -59,6 +59,93 @@ function getListingKey(key, vFormat) {
...
+            for (const ob of versionedSuffixes) {
+                const { name, vid } = ob;
+                // Does not contain delete markers but contains versions to simulate turning on versioning.
+                const keyVersion = `${commonPrefix}${name}${VID_SEP}${vid}`;
+                const version = new Version({ versionId: vid });
+                const obj = {
+                    key: getListingKey(keyVersion, vFormat),
+                    value: version.toString(),
+                };
+                const recieved = delimiter.filter(obj);
```
jonathan-gramain: To simplify we could simply check that all results are `FILTER_ACCEPT` since we expect 0 `FILTER_SKIP`

Also a minor spelling typo: ~recieved~ received

jonathan-gramain: In v0 format, all version keys of the same object are preceded by a single master key, so since all those keys are different objects they should all be preceded by a master key (i.e. the same key without `${VID_SEP}${vid}`, and the value of the latest version)

jonathan-gramain (changes requested):

tests/unit/algos/list/delimiterMaster.spec.js
watsaqat: It is harder to debug if you assert FILTER_ACCEPT at each iteration.

tests/unit/algos/list/delimiterMaster.spec.js
watsaqat: if you're just referring to only the keys with isDeleteMarker, we can get rid of those and the behavior wouldn't change. I left them in to more accurately reproduce the bug, but they don't actually matter.

tests/unit/algos/list/delimiterMaster.spec.js
watsaqat: why change this? Its a closer representation to the key that caused the bug than without the underscore. And in general testing with more complicated inputs should lead to us catching more bugs

jonathan-gramain: > And in general testing with more complicated inputs should lead to us catching more bugs

I have to disagree, but let's move on it does not really matter just a readability suggestion

tests/unit/algos/list/delimiterMaster.spec.js
jonathan-gramain: It's fine to pass extra keys to more accurately represent the issue, but they should be passed in order, otherwise we're testing an impossible input.

tests/unit/algos/list/delimiterMaster.spec.js
jonathan-gramain: Fair enough, I personally think the readability and simplicity of the test matters more, and if an issue is reproducible it's easy to get down to the cause by adding some debug when the time comes to fix an issue (and it's possible to add context to the `assert.strictEqual` if needed for debugging).

lib/algos/list/delimiterMaster.js
```diff
@@ -122,11 +120,13 @@ class DelimiterMaster extends Delimiter {
              *   when an entry is a PHD avoids the skip on version for the
              *   next entry). In that case we expect the master version to
              *   follow. */
+
             if (key === this.prvPHDKey) {
                 this.prvKey = undefined;
                 return FILTER_ACCEPT;
             }
             this.prvKey = key;
+            this[this.nextContinueMarker] = key;
```
watsaqat, line 129: the key changes are setting the nextContinuationToken to key instead of commonPrefix here and on line 179.

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -296,7 +514,7 @@ function getListingKey(key, vFormat) {
                 /* When filtered, it should return FILTER_SKIP and set the prvKey. It
                  * should not be added to the result content or common prefixes. */
                 assert.strictEqual(delimiter.filter(obj), FILTER_SKIP);
-                assert.strictEqual(delimiter.NextMarker, undefined);
+                // assert.strictEqual(delimiter.NextMarker, undefined);
```
watsaqat, line 517: I am unsure why having the NextContinuationToken be set to undefined is desirable or needed here and on line 533. Both cases occur after filtering a deleteMarker key.

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -486,7 +704,7 @@ function getListingKey(key, vFormat) {
                     }),
                     FILTER_ACCEPT);
                 // ...it should return to skipping by prefix as usual
-                assert.strictEqual(delimiter.skipping(), `${inc(DbPrefixes.Replay)}foo/`);
+                assert.strictEqual(delimiter.skipping(), `${inc(DbPrefixes.Replay)}foo/bar${VID_SEP}`);
```
watsaqat, line 707: since we change to returning the key instead of the commonPrefix, this test is changed to reflect this.

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -46,7 +48,91 @@ function getListingKey(key, vFormat) {
...
+
+function *generateKeys(number) {
+    let idx = 0;
+    let masterKey = '_Common-Prefix/';
+    yield { key: masterKey, isDeleteMarker: false }
+
+    idx++;
+    while (idx < number) {
+        masterKey = `_Common-Prefix/${zpad(idx)}`;
+        yield { key: masterKey, isDeleteMarker: chance.bool() }
```
miniscruff: is chance.bool() just a 50/50 random bool? I feel like a simple util function would work instead of a new import but idk.

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -46,7 +48,91 @@ function getListingKey(key, vFormat) {
...
+    }
+}
+
+/**
+ * Generate raw listing keys in alphabetical order.
+ * @param {Integer} number - how many keys to generate.
+ * @yields {Object} - { key, isDeleteMarker }
+ */
+
+function *generateKeys(number) {
```
miniscruff:
```suggestion
function *generateKeys(count) {
```
I would call this var something like count or amount? Just number is kinda vague.

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -46,7 +48,91 @@ function getListingKey(key, vFormat) {
...
+ * @param {Integer} maxKeys - how many versioned keys to generate
+ * @yields {Object} - { key, isDeleteMarker }
+ */
+function *generateVersionedKeys(masterKey, numKeys) {
+    let [idx, key] = [0, masterKey];
+    while (idx < numKeys) {
+        key = `${masterKey}${VID_SEP}${zpad(idx)}`
+        yield {key, isDeleteMarker: chance.bool()}
+        idx++
+    }
```
miniscruff: I would just use a for loop here
```suggestion
    for (let i = 0; i < numKeys; i++) {
        yield {
            key: `${masterKey}${VID_SEP}${zpad(i)}`, 
            isDeleteMarker: chance.bool(),
        }
    }
```
For key you could also do `[masterKey, VID_SEP, zpad(i)].join('')` or just `+`, I never know what is better :smirk:

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -46,7 +48,91 @@ function getListingKey(key, vFormat) {
...
+            } else {
+                // stop listing this key range
+                // eslint-disable-next-line no-param-reassign
+                gteparams = params.gte;
+            }
+        }
+    } else {
+        streakLength = 1;
+    }
+    return undefined;
```
miniscruff:
```suggestion
```
Unless ESLint complains you don't need this

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -59,6 +145,61 @@ function getListingKey(key, vFormat) {
...
+            [...generateKeys(1000)].forEach((ob, idx) => {
+                const {key, isDeleteMarker} = ob;
+
+                // simulate not doing a raw listing when key < params.gte
+                // this represents a key not reaching the read stream from leveldb
+                if (params.gte && key < params.gte) {
+                    throw new Error('Invariant violated. Keys may be missing from listing.');
+                }
+
+                if (key.indexOf(VID_SEP) === -1) { // master key
```
miniscruff: Instead of `indexOf` you can do includes: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -59,6 +145,61 @@ function getListingKey(key, vFormat) {
...
+                    handleStreak(res, skippingRange, delimiter, idx);
+
+                    let expected = null;
+                    if (vFormat === 'v0') {
+                        expected = isDeleteMarker ? FILTER_SKIP : FILTER_ACCEPT
+                    } else { // v1
+                        expected = FILTER_ACCEPT
+                    }
+                    
+                    assert.equal(res, expected);
```
miniscruff: should always use strict assert checks

tests/unit/algos/list/delimiterMaster.spec.js
miniscruff: I see you use chance.integer as well but that is also a one liner :smirk:

miniscruff (changes requested):

tests/unit/algos/list/delimiterMaster.spec.js
watsaqat: fair enough. I'd still rather use a library for these types of tests in general so the homespun versions of chance.bool, chance.integer, etc don't accidentally get written in a way where the distribution is non-uniform. Easy mistake to make, though here it is likely easy enough.

miniscruff (approved):

naren-rajendran (approved):

lib/algos/list/delimiterMaster.js
```diff
@@ -131,6 +128,22 @@ class DelimiterMaster extends Delimiter {
         }
 
         this.prvKey = key;
+        if (key === this.prefix) {
+            // key is of the form foo/
+            this.delimiterEndingKeySeen = true;
```
jonathan-gramain: The variable name does not seem to match what it's supposed to mean, as it could refer to any key with an ending delimiter, where it's set only if we've seen a key that `is` the prefix to list. If this is needed, why not calling this variable `keyMatchingPrefixSeen` for example instead.

lib/algos/list/delimiterMaster.js
```diff
@@ -131,6 +128,22 @@ class DelimiterMaster extends Delimiter {
...
+        }
+
+        if (this.delimiterEndingKeySeen === true && // previously seen a key of the form 'prefix/'
```
jonathan-gramain:
```suggestion
        if (this.delimiterEndingKeySeen && // previously seen a key of the form 'prefix/'
```

lib/algos/list/delimiterMaster.js
```diff
@@ -131,6 +128,22 @@ class DelimiterMaster extends Delimiter {
...
+            !Version.isDeleteMarker(value) && // not delete marker
```
jonathan-gramain: We can remove this check since delete markers are treated earlier in this function and always return.

lib/algos/list/delimiterMaster.js
```diff
@@ -131,6 +128,22 @@ class DelimiterMaster extends Delimiter {
...
+            key.indexOf(VID_SEP) === -1 && // master key
+            key.startsWith(this.prefix) &&
```
jonathan-gramain: same, can be removed as we have an earlier check for keys not starting with the prefix

lib/algos/list/delimiterMaster.js
```diff
@@ -91,13 +92,8 @@ class DelimiterMaster extends Delimiter {
              *   way to know  we should not accept this version. This test is
              *   not redundant with the one at the beginning of this function,
              *   we are comparing here the key without the version suffix,
-             * - key startsWith the previous NextMarker happens because we set
-             *   NextMarker to the common prefix instead of the whole key
-             *   value. (TODO: remove this test once ZENKO-1048 is fixed)
              *   */
-            if (key === this.prvKey || key === this[this.nextContinueMarker] ||
-                (this.delimiter &&
```
jonathan-gramain, line 99: Aren't we introducing a regression by removing this check since ZENKO-1048 is not fixed? If so, can we consider the regression acceptable, or was it causing a more annoying bug in place of what it was supposed to fix? I'd rather try to clarify and keep a comment around related to ZENKO-1048 since there seems to be some corner case here.

lib/algos/list/delimiterMaster.js
```diff
@@ -131,6 +128,22 @@ class DelimiterMaster extends Delimiter {
...
 
         this.prvKey = key;
+        if (key === this.prefix) {
+            // key is of the form foo/
+            this.delimiterEndingKeySeen = true;
+        }
+
+        if (this.delimiterEndingKeySeen === true && // previously seen a key of the form 'prefix/'
+            !Version.isDeleteMarker(value) && // not delete marker
+            key.indexOf(VID_SEP) === -1 && // master key
```
jonathan-gramain:
```suggestion
            versionIdIndex === -1 && // master key
```

lib/algos/list/delimiterMaster.js
```diff
@@ -131,6 +128,22 @@ class DelimiterMaster extends Delimiter {
...
+            key.startsWith(this.prefix) &&
+            key.indexOf(this.delimiter, this.prefix.length) === -1 // key is of the form 'prefix/foo'
```
jonathan-gramain: I am thinking this check is redundant with the one below (https://github.com/scality/Arsenal/pull/1982/files#diff-9e520dd5b6e739ad1fba611bf9fdffc713fbaf71ed390ad15565a5263578a1f6R151).

If it works I'd rather limit this case specifically to the key equalling the prefix (`key === this.prefix`) which is the case not properly handled and wrongly added as a common prefix, likely the other subkeys would be properly handled below, without needing a specific check. If that's not the case you can keep this but I think it's not needed and potentially could introduce more corner cases, so better try to keep things as simple as possible.

Otherwise if we keep this here I'd recommend re-using the result set in `delimiterIndex` below to avoid a double computation.

lib/algos/list/delimiterMaster.js
```diff
@@ -91,13 +97,9 @@ class DelimiterMaster extends Delimiter {
...
              *   not redundant with the one at the beginning of this function,
              *   we are comparing here the key without the version suffix,
-             * - key startsWith the previous NextMarker happens because we set
-             *   NextMarker to the common prefix instead of the whole key
-             *   value. (TODO: remove this test once ZENKO-1048 is fixed)
              *   */
             if (key === this.prvKey || key === this[this.nextContinueMarker] ||
-                (this.delimiter &&
-                key.startsWith(this[this.nextContinueMarker]))) {
+                this.delimiter && key.startsWith(this[this.nextContinueMarker])) {
```
jonathan-gramain: Why removing the parenthesis?

lib/algos/list/delimiterMaster.js
```diff
@@ -62,6 +63,11 @@ class DelimiterMaster extends Delimiter {
         let key = obj.key;
         const value = obj.value;
 
+        if (key === this.prefix) {
+            // key is of the form prefix/
+            return this.addContents(key, value);
```
jonathan-gramain, line 66: Please also test with a prefix that does not end with a delimiter character, since it should be allowed by the spec. The comment may have to be rephrased or removed then.

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -59,6 +153,63 @@ function getListingKey(key, vFormat) {
...
+            [...generateKeys(maxKeys)].forEach(ob => {
+                const {key, isDeleteMarker, canSkip} = ob;
+                // simulate not doing a raw listing when key < params.gte
+                // this represents a key not reaching the read stream from leveldb
+                if (state.params.gte && key < state.params.gte) {
+                    return;
+                }
+
+                if (!key.includes(VID_SEP)) { // master key
+                    const version = new Version({ isDeleteMarker });
```
jonathan-gramain: in v1 there is no delete marker key in the master prefix, so we would have to not include them in the listing.

By the way, we should either fix v1 as well or open a ticket for it, if it is not fixed in v1. We can have the test ready for v1 but skip it in the meantime.

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -46,6 +47,99 @@ function getListingKey(key, vFormat) {
...
+    let masterKey = '_Common-Prefix/';
+    yield { key: masterKey, isDeleteMarker: false, canSkip: false }
+
+    idx++;
+    while (idx < count) {
+        masterKey = `_Common-Prefix/${zpad(idx)}`;
+        const masterKeyWithSubprefix = `_Common-Prefix/${zpad(idx)}/${zpad(idx)}`
+        const baseKey = chance.bool() ? masterKeyWithSubprefix : masterKey;
+        const isDeleteMarker = chance.bool();
+        yield { key: baseKey, isDeleteMarker, canSkip: isDeleteMarker}
```
jonathan-gramain: I am not sure the `canSkip` value is right here: we should be able to skip keys in subprefixes if we have already seen the same subprefix. I think the current code is doing this (in delimiter.js `addCommonPrefix`) but the test does not seem to reflect this.

tests/unit/algos/list/delimiterMaster.spec.js
jonathan-gramain: If there is no such issue with v1, it's fine then with just running the same test on v1.

lib/algos/list/delimiterMaster.js
watsaqat: copy/paste error. Will update

tests/unit/algos/list/delimiterMaster.spec.js
watsaqat: good catch

tests/unit/algos/list/delimiterMaster.spec.js
watsaqat: I moved the test to only run under v0; v1 doesn't have this bug and the code change only affects v0 (v1 has separate logic for filtering)

lib/algos/list/delimiterMaster.js
```diff
@@ -122,6 +124,7 @@ class DelimiterMaster extends Delimiter {
              *   when an entry is a PHD avoids the skip on version for the
              *   next entry). In that case we expect the master version to
              *   follow. */
+
```
jonathan-gramain: this extra newline does not seem to be helping, should be removed

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -46,6 +48,116 @@ function getListingKey(key, vFormat) {
...
+
+        if (hasSubprefix) {
+            // 'prefix -> liof = -1, pin = 5'
+            // 'prefix/ -> liof = 6, pin = 6'
+            canSkip = state.handleSubprefixKey(baseKey, canSkip)
+        }
+
+        if (state.prefix[state.prefix.length - 1] !== '/') {
+            canSkip = state.handleSubprefixKey(baseKey, canSkip)
+        }
```
jonathan-gramain:
```suggestion
        if (hasSubprefix || state.prefix[state.prefix.length - 1] !== '/') {
            canSkip = state.handleSubprefixKey(baseKey, canSkip);
        }
```

Also from a style perspective, I'd rather avoid passing `canSkip` to the helper function which should only deal with what it's named for, i.e. subprefix handling, otherwise it becomes hard to follow the logic, we can always use boolean operators to cope with this, something like `canSkip = isDeleteMarker || state.handleSubprefixKey(baseKey)` would likely work.

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -488,6 +600,92 @@ function getListingKey(key, vFormat) {
...
             });
+
+            it('should list all subprefixes when handleStreak is applied as in RepdServer', () => {
+                ['_Common-Prefix', '_Common-Prefix/'].forEach(prefix => {
+                    const state = createStreakState(prefix);
+
+                    if (prefix[prefix.length - 1] === '/') {
+                        state.handleSubprefixKey = (baseKey, canSkip) => {
+                            let lastIdx = baseKey.lastIndexOf('/');
+                            // const prefix = baseKey.slice(0, lastIdx);
```
jonathan-gramain: to be removed

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -488,6 +600,92 @@ function getListingKey(key, vFormat) {
                 // ...it should return to skipping by prefix as usual
                 assert.strictEqual(delimiter.skipping(), `${inc(DbPrefixes.Replay)}foo/`);
             });
+
+            it('should list all subprefixes when handleStreak is applied as in RepdServer', () => {
+                ['_Common-Prefix', '_Common-Prefix/'].forEach(prefix => {
+                    const state = createStreakState(prefix);
+
+                    if (prefix[prefix.length - 1] === '/') {
+                        state.handleSubprefixKey = (baseKey, canSkip) => {
```
jonathan-gramain: Do we really need two different implementations of this function, why not merging all logic into a single helper function?

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -488,6 +600,92 @@ function getListingKey(key, vFormat) {
...
+        
+                            const res = delimiter.filter(obj);
+                            const skippingRange = delimiter.skipping();
+                            handleStreak(res, skippingRange, state);
+        
+                            let expected = null;
+                            if (vFormat === 'v0') {
+                                expected = canSkip ? FILTER_SKIP : FILTER_ACCEPT;
+                            } else { // v1
+                                expected = FILTER_ACCEPT
```
jonathan-gramain: I don't understand why in v1 all master keys are accepted. I would think that the behavior should be similar for v0 and v1 when it comes to master keys handling.

lib/algos/list/delimiterMaster.js
```diff
@@ -33,6 +33,7 @@ class DelimiterMaster extends Delimiter {
         this.prvKey = undefined;
         this.prvPHDKey = undefined;
         this.inReplayPrefix = false;
+        this.delimiterEndingKeySeen = false;
```
jonathan-gramain: looks like this should be removed

lib/algos/list/delimiterMaster.js
```diff
@@ -91,13 +97,9 @@ class DelimiterMaster extends Delimiter {
              *   way to know  we should not accept this version. This test is
              *   not redundant with the one at the beginning of this function,
              *   we are comparing here the key without the version suffix,
-             * - key startsWith the previous NextMarker happens because we set
```
jonathan-gramain: Why removing this comment

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -46,6 +47,111 @@ function getListingKey(key, vFormat) {
     return assert.fail(`bad vFormat ${vFormat}`);
 }
 
+const MAX_STREAK_LENGTH = 100
```
jonathan-gramain: I though that the linter would complain on this, maybe the rules have been updated? Anyway I think it's a good practice to end all instructions with a semicolon.

```suggestion
const MAX_STREAK_LENGTH = 100;
```

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -46,6 +47,111 @@ function getListingKey(key, vFormat) {
...
+const createStreakState = (prefix, vFormat) => {
+    return {
+        vFormat,
+        prefix,
+        prefixes: new Set(),
+        params: {},
+        streakLength: 0,
+        gteparams: null,
+    }
+}
```
jonathan-gramain, line 60:
```suggestion
};
```

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -46,6 +47,111 @@ function getListingKey(key, vFormat) {
...
+        const isDeleteMarker = state.vFormat === 'v0' && chance.bool() ;
+        const baseKey = hasSubprefix ? masterKeyWithSubprefix : masterKey;
+
+        let canSkip = isDeleteMarker;
+        if (hasSubprefix || state.prefix[state.prefix.length - 1] !== '/') {
+            canSkip =  canSkip || state.handleSubprefixKey(baseKey);
+        }
+
+        yield { key: baseKey, isDeleteMarker, canSkip }
+        idx++
```
jonathan-gramain:
```suggestion
        yield { key: baseKey, isDeleteMarker, canSkip };
        idx++;
```

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -46,6 +47,111 @@ function getListingKey(key, vFormat) {
...
+
+        const versioned = chance.integer({ min: 0, max: 200});
+        const allowedVersionedKeys = Math.min(count - idx, versioned)
```
jonathan-gramain:
```suggestion
        const allowedVersionedKeys = Math.min(count - idx, versioned);
```

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -59,6 +165,71 @@ function getListingKey(key, vFormat) {
...
+            ['_Common-Prefix', '_Common-Prefix/'].forEach(prefix => {
+                const state = createStreakState(prefix, vFormat);
+                state.handleSubprefixKey = baseKey => {
+                    const isLastDelim = state.prefix[state.prefix.length - 1] === '/';
+                    let lastIdx = isLastDelim ? baseKey.lastIndexOf('/') : state.prefix.length;
+                    const prefix = isLastDelim ? getCommonPrefix(baseKey, '/', lastIdx): baseKey.slice(0, lastIdx);
+                    if (state.prefixes.has(prefix) || (!isLastDelim && baseKey.length > lastIdx)) {
+                        return true;
+                    } else {
+                        state.prefixes.add(prefix)
```
jonathan-gramain:
```suggestion
                        state.prefixes.add(prefix);
```

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -59,6 +165,71 @@ function getListingKey(key, vFormat) {
...
+        it('should list all subprefixes when handleStreak is applied as in RepdServer', () => {
+            ['_Common-Prefix', '_Common-Prefix/'].forEach(prefix => {
+                const state = createStreakState(prefix, vFormat);
+                state.handleSubprefixKey = baseKey => {
+                    const isLastDelim = state.prefix[state.prefix.length - 1] === '/';
+                    let lastIdx = isLastDelim ? baseKey.lastIndexOf('/') : state.prefix.length;
+                    const prefix = isLastDelim ? getCommonPrefix(baseKey, '/', lastIdx): baseKey.slice(0, lastIdx);
+                    if (state.prefixes.has(prefix) || (!isLastDelim && baseKey.length > lastIdx)) {
+                        return true;
+                    } else {
```
jonathan-gramain: The `else` is not needed after the `return`

jonathan-gramain (commented):
There's something strange with the linter rules, anyway let's fix this and get this merged

jonathan-gramain (commented):
ok with the code, please squash commits

jonathan-gramain (commented):
Can you please add a description of the bugfix in the commit message (you can use the PR's description for example)

jonathan-gramain (approved):

PR #1983, opened 2022-08-10
improvement: ARSN-260 improve efficiency of findConditionKey

jonathan-gramain:
Instead of pre-creating a Map with all supported condition keys before
returning the wanted one, use a switch/case construct to directly
return the attribute from the request context.


jonathan-gramain:
ping

miniscruff (approved):

naren-rajendran (approved):

PR #1989, opened 2022-08-10
bugfix: ARSN-255 revamp evaluatePolicy logic for tag conditions

jonathan-gramain:
Rethink the logic of tag condition evaluation, so that the"evaluateAllPolicies" function appropriately returns the verdict:
Allow or Deny or NeedTagConditionEval, the latter being when tag values (request and/or object tags) are needed to settle the verdict to Allow or Deny, in which case, Cloudserver knows it has to resend the request to Vault along with tag info.


jonathan-gramain:
ping

tmacro (approved):

miniscruff (approved):

PR #2011, opened 2022-09-20
Bugfix/arsn 269 listing bug versioned bucket edge case

watsaqat:
address https://scality.atlassian.net/browse/S3C-4682 and corrects missed faulty logic in  [ARSN-262](https://scality.atlassian.net/browse/ARSN-262)

Simplifies testing that was used in [ARSN-262](https://scality.atlassian.net/browse/ARSN-262). 
Adds a function allowDelimiterRangeSkip to determine when a nextContinueMarker range can be skipped when .skipping is called. This function uses a new state variable prefixKeySeen and the nextContinueMarker to determine if a range of the form prefix/ can be skipped. An additional check is added when processing delete markers of the form prefix/foo/(bar) so that the prefix/foo/ range can still be skipped as an optimization.

### DIFF
--- a/lib/algos/list/delimiterMaster.js
+++ b/lib/algos/list/delimiterMaster.js
@@ -33,6 +33,8 @@ class DelimiterMaster extends Delimiter {
         this.prvKey = undefined;
         this.prvPHDKey = undefined;
         this.inReplayPrefix = false;
+        this.prefixKeySeen = false;
+        this.prefixEndsWithDelim = this.prefix && this.prefix.endsWith(this.delimiter);
 
         Object.assign(this, {
             [BucketVersioningKeyFormat.v0]: {
@@ -63,7 +65,7 @@ class DelimiterMaster extends Delimiter {
         const value = obj.value;
 
         if (key === this.prefix) {
-            return this.addContents(key, value);
+            this.prefixKeySeen = true;
         }
 
         if (key.startsWith(DbPrefixes.Replay)) {
@@ -99,8 +101,8 @@ class DelimiterMaster extends Delimiter {
              *   NextMarker to the common prefix instead of the whole key
              *   value. (TODO: remove this test once ZENKO-1048 is fixed)
              *   */
-            if (key === this.prvKey || key === this[this.nextContinueMarker] ||
-                (this.delimiter && key.startsWith(this[this.nextContinueMarker]))) {
+            if (key === this.prvKey || key === this[this.nextContinueMarker]
+                || (this.delimiter && key.startsWith(this[this.nextContinueMarker]))) {
                 /* master version already filtered */
                 return FILTER_SKIP;
             }
@@ -130,6 +132,14 @@ class DelimiterMaster extends Delimiter {
                 return FILTER_ACCEPT;
             }
             this.prvKey = key;
+            if (this.prefixEndsWithDelim) {
+                /* When the prefix ends with a delimiter, update nextContinueMarker
+                 * to be able to skip ranges of the form prefix/subprefix/ as an optimization.
+                 * The marker may also end up being prefix/, in which case .skipping will determine
+                 * if a skip over the full range is allowed or a smaller skipping range of prefix/{VID_SEP}
+                 * must be used. */
+                this[this.nextContinueMarker] = key.slice(0, key.lastIndexOf(this.delimiter) + 1);
+            }
             return FILTER_SKIP;
         }
 
@@ -165,6 +175,27 @@ class DelimiterMaster extends Delimiter {
         return super.filter(obj);
     }
 
+    /**
+     * Determine if a nextContinueMarker ending with a delimiter
+     * can be skipped.
+     * @returns {bool}
+     */
+
+    allowDelimiterRangeSkip() {
+        if (!this.prefixKeySeen) {
+            // A prefix key is a master key equal to the prefix. If it has
+            // not been encountered, can skip.
+            return true;
+        }
+        const marker = this[this.nextContinueMarker];
+        // prefix = prefix, key = prefix/. Can skip since key will be part of commonPrefixes.
+        if (marker.length > this.prefix.length && marker.startsWith(this.prefix)) {
+            return true;
+        }
+        const lastIdx = marker.lastIndexOf(this.delimiter); // prefix/foo is a masterKey following a prefix key.
+        return marker.slice(0, lastIdx + 1) !== this.prefix; // cannot skip the full range prefix/ range.
+    }
+
     skippingBase() {
         if (this[this.nextContinueMarker]) {
             // next marker or next continuation token:
@@ -172,7 +203,7 @@ class DelimiterMaster extends Delimiter {
             // - foo  : skipping foo.
             const index = this[this.nextContinueMarker].
                 lastIndexOf(this.delimiter);
-            if (index === this[this.nextContinueMarker].length - 1) {
+            if (index === this[this.nextContinueMarker].length - 1 && this.allowDelimiterRangeSkip()) {
                 return this[this.nextContinueMarker];
             }
             return this[this.nextContinueMarker] + VID_SEP;
--- a/tests/unit/algos/list/delimiterMaster.spec.js
+++ b/tests/unit/algos/list/delimiterMaster.spec.js
@@ -2,8 +2,6 @@
 
 const assert = require('assert');
 const chance = require('chance').Chance(); // eslint-disable-line
-
-const { getCommonPrefix } = require('../../../../lib/algos/list/delimiter');
 const DelimiterMaster =
     require('../../../../lib/algos/list/delimiterMaster').DelimiterMaster;
 const {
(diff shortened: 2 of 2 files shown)


lib/algos/list/delimiterMaster.js
```diff
@@ -62,7 +63,8 @@ class DelimiterMaster extends Delimiter {
         let key = obj.key;
         const value = obj.value;
 
-        if (key === this.prefix) {
+        if (key === this.prefix || key[key.length - 1] === this.delimiter) {
```
jonathan-gramain: Why do we have `key[key.length - 1] === this.delimiter` as an extra condition? It does not look right to me to say that we've seen the prefix key when we actually have seen a key that ends with a delimiter.

lib/algos/list/delimiterMaster.js
```diff
@@ -130,6 +132,9 @@ class DelimiterMaster extends Delimiter {
                 return FILTER_ACCEPT;
             }
             this.prvKey = key;
+            if (this.prefixKeySeen) {
+                this[this.nextContinueMarker] = key;
```
jonathan-gramain: I'm not sure I get why we do this

lib/algos/list/delimiterMaster.js
```diff
@@ -130,6 +131,9 @@ class DelimiterMaster extends Delimiter {
                 return FILTER_ACCEPT;
             }
             this.prvKey = key;
+            if (!this.allowDelimiterRangeSkip(key)) {
+                this[this.nextContinueMarker] = key;
```
jonathan-gramain: The problem I see with this piece of code is that it will prevent skipping over subprefixes when some/all keys are delete markers in those subprefixes, leading to a potentially very long listing or timing out.

Note that today we have a similar issue in v0 but only when delete marker keys are not part of a subprefix, we still have the ability to skip a whole subprefix.

Here's a test that catches this issue:

```
            it.only('foobar', () => {
                for (const prefix of ['prefix/']) {
                    const delimiter = new DelimiterMaster({
                        prefix,
                        delimiter: '/',
                    }, fakeLogger, vFormat);
                    for (const testEntry of [
                        {
                            key: prefix,
                            expectedRes: FILTER_ACCEPT,
                            expectedSkipping: `${prefix}${VID_SEP}`,
                        },
                        {
                            key: `${prefix}${VID_SEP}v1`,
                            //isDeleteMarker: true,
                            expectedRes: FILTER_SKIP, // versions get skipped after master
                            expectedSkipping: `${prefix}${VID_SEP}`,
                        },
                        {
                            key: `${prefix}foo/`,
                            isDeleteMarker: true,
                            expectedRes: FILTER_SKIP,
                            expectedSkipping: `${prefix}foo/`,
                        },
                        {
                            key: `${prefix}foo/1`,
                            isDeleteMarker: true,
                            expectedRes: FILTER_SKIP,
                            expectedSkipping: `${prefix}foo/`,
                        },
                        {
                            key: `${prefix}foo/2`,
                            isDeleteMarker: true,
                            expectedRes: FILTER_SKIP,
                            expectedSkipping: `${prefix}foo/`,
                        },
                    ]) {
                        const entry = {
                            key: testEntry.key,
                        };
                        if (testEntry.isDeleteMarker) {
                            entry.value = '{"isDeleteMarker":true}';
                        } else {
                            entry.value = '{}';
                        }
                        const res = delimiter.filter(entry);
                        const skipping = delimiter.skipping();
                        assert.strictEqual(res, testEntry.expectedRes);
                        assert.strictEqual(skipping, testEntry.expectedSkipping);
                    }
                }
            });
```

tests/unit/algos/list/delimiterMaster.spec.js
```diff
@@ -657,6 +677,326 @@ function *generateKeys(count, state) {
...
+                            isDeleteMarker: true,
+                            expectedRes: FILTER_SKIP,
+                            expectedSkipping: `${prefix}${VID_SEP}`,
+                        },
+                        {
+                            key: `${prefix}deleted${VID_SEP}v2`,
+                            expectedRes: FILTER_SKIP,
+                            expectedSkipping: `${prefix}${VID_SEP}`,
+                        },
+                        {
```
jonathan-gramain, line 525: Could be useful to add a `notdeleted` key here similarly to the test below, to make sure it is not skipped

lib/algos/list/delimiterMaster.js
```diff
@@ -165,14 +170,54 @@ class DelimiterMaster extends Delimiter {
...
+        //         return true;
+        //     }
+        // }
+        // return false;
+
+        const delimEnd = this.prefix && this.prefix.endsWith(this.delimiter); // prefix/
+        if (!delimEnd) {
+            return null;
+        }
+        const delimIndex = this.prefix.lastIndexOf(this.delimiter);
```
jonathan-gramain:
```suggestion
        const delimIndex = this.prefix.length - 1;
```

lib/algos/list/delimiterMaster.js
```diff
@@ -165,14 +170,54 @@ class DelimiterMaster extends Delimiter {
...
+        const keySuffix = key.slice(delimIndex + 1);
+
+
+        if (keySuffix.endsWith(this.delimiter)) { // prefix/foo/
+            return key;
+        }
+        return key.slice(0, key.lastIndexOf(this.delimiter) + 1); // prefix/foo/1
```
jonathan-gramain: Test should not be needed (unless there is a real efficiency reason to do the first check but I don't think so)

```suggestion
        return key.slice(0, key.lastIndexOf(this.delimiter) + 1); // prefix/foo/bar -> prefix/foo/
```

lib/algos/list/delimiterMaster.js
```diff
@@ -165,14 +170,54 @@ class DelimiterMaster extends Delimiter {
...
+        //     const commonPrefix = getCommonPrefix(key, this.delimiter, baseIndex);
+        //     console.log('inside can add common', {key, commonPrefix, prefix: this.prefix});
+        //     if (this.CommonPrefixes.indexOf(commonPrefix) === -1
+        //             && this[this.nextContinueMarker] !== commonPrefix) {
+        //         return true;
+        //     }
+        // }
+        // return false;
+
+        const delimEnd = this.prefix && this.prefix.endsWith(this.delimiter); // prefix/
```
jonathan-gramain: Ideally we would precompute this and store it in a class member to save some CPU cycles

lib/algos/list/delimiterMaster.js
```diff
@@ -165,14 +170,54 @@ class DelimiterMaster extends Delimiter {
         return super.filter(obj);
     }
 
+    allowDelimiterRangeSkip(key) {
+        if (!this.prefixKeySeen) {
+            return true;
+        }
+        // prefix = prefix, key = prefix/. Can skip since key will be part of commonPrefixes.
+        if (key[key.length - 1] === this.delimiter && key.length > this.prefix.length && key.startsWith(this.prefix)) {
```
jonathan-gramain: To be checked but I believe we can remove the two last tests as they are already checked earlier in the function (that the key starts with the prefix), and if it equals the prefix it's already handled as well.

```suggestion
        if (key[key.length - 1] === this.delimiter) {
```

watsaqat: removing this will actually allow prefix/ to be skipped. This is why we need to check that the key is of the form prefix/X

lib/algos/list/delimiterMaster.js
```diff
@@ -33,6 +33,8 @@ class DelimiterMaster extends Delimiter {
         this.prvKey = undefined;
         this.prvPHDKey = undefined;
         this.inReplayPrefix = false;
+        this.prefixKeySeen = false;
+        this.delimEnd = this.prefix && this.prefix.endsWith(this.delimiter);
```
jonathan-gramain: Let's be explicit in the variable name

```suggestion
        this.prefixEndsWithDelim = this.prefix && this.prefix.endsWith(this.delimiter);
```

lib/algos/list/delimiterMaster.js
jonathan-gramain: Actually, we can remove the first test then because it's already checked before calling this function at https://github.com/scality/Arsenal/pull/2011/files#diff-9e520dd5b6e739ad1fba611bf9fdffc713fbaf71ed390ad15565a5263578a1f6R206

```suggestion
        if (key.length > this.prefix.length && key.startsWith(this.prefix)) {
```

lib/algos/list/delimiterMaster.js
```diff
@@ -165,14 +175,36 @@ class DelimiterMaster extends Delimiter {
...
     }
 
+    /**
+     * Determine if a nextContinueMarker ending with a delimiter
+     * can be skipped.
+     * @param {String} key - a nextContinueMarker
+     * @returns {bool}
+     */
+
+    allowDelimiterRangeSkip(key) {
```
jonathan-gramain: I believe we don't really need to pass a parameter here, let's just use the member `this[this.nextContinueMarker]` in place (it's good to use a local variable like `currentMarker = this[this.nextContinueMarker]` though)

jonathan-gramain (approved):
OK for me, a few suggestions for improvement

tmacro (approved):

watsaqat:
ping

PR #2017, opened 2022-10-11
ARSN-252 - listing bug in DelimisterMaster

watsaqat:
DelimiterMaster.filter is used to determine when a key range can be skipped in Metadata:RepdServer to optimize listing performance. When a bucket is created with vFormat=v0, and subsequently a listing is done with a prefix, DelimiterMaster.filter was incorrectly determining that a range could be skipped if a key was listed such that key == prefix. This case is now correctly handled in filterV0.

(cherry picked from commit f62c3d22edf84a47c85fa7753f80bedeeec19636)

### DIFF
--- a/lib/algos/list/delimiter.js
+++ b/lib/algos/list/delimiter.js
@@ -271,4 +271,4 @@ class Delimiter extends Extension {
     }
 }
 
-module.exports = { Delimiter };
+module.exports = { Delimiter, getCommonPrefix };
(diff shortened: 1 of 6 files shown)


jonathan-gramain (approved):

rahulreddy (approved):

tmacro (approved):

watsaqat:
ping

PR #2030, opened 2022-11-04
ARSN-276: putObjectVerCase3 -add case for v1 and versioned entry update

alexanderchan-scality:
add condition check in putObjectVerCase3 to avoid creating erroneous master entry for v1 format and versioned updates.
sequence:
* check if master entry exists
* check if version entry exists
* new condition: if version exists and master entry doesn't, update only the versioned entry


naren-rajendran (approved):

alexanderchan-scality:
ping

alexanderchan-scality:
ping

nicolas2bert (approved):

PR #2033, opened 2022-11-08
bugfix: ARSN-278 handle getting versionId when object is versioning suspended

welansari:
…uspended

When replicating a versioning suspended object, we need to specify 'null' as the encoded versionId as the versionId contained within the object's metadata is strictly internal

In the replication processor we use getVersionId() when putting/deleting a tag. It's used by the mongoClient to fetch the object from MongoDB. Here again we need to specify 'null' to get the versioning suspended object (Cloudserver already knows how to handle 'null' versionId and transforms it to undefined before giving it to the mongoClient)

https://github.com/scality/cloudserver/blob/2c8968ef4a0d6f44f10222c90d2ae3066ce04068/lib/metadata/metadataUtils.js#L94-L96

Linked to: https://github.com/scality/backbeat/pull/2306

### DIFF
--- a/lib/models/ObjectMD.ts
+++ b/lib/models/ObjectMD.ts
@@ -1,6 +1,7 @@
 import * as crypto from 'crypto';
 import * as constants from '../constants';
 import * as VersionIDUtils from '../versioning/VersionID';
+import { VersioningConstants } from '../versioning/constants';
 import ObjectMDLocation, {
     ObjectMDLocationData,
     Location,
@@ -901,6 +902,9 @@ export default class ObjectMD {
      * @return The object versionId
      */
     getVersionId() {
+        if (this.getIsNull()) {
+            return VersioningConstants.ExternalNullVersionId;
+        }
         return this._data.versionId;
     }
 
@@ -908,13 +912,16 @@ export default class ObjectMD {
      * Get metadata versionId value in encoded form (the one visible
      * to the S3 API user)
      *
-     * @return The encoded object versionId
+     * @return {undefined|string} The encoded object versionId
      */
     getEncodedVersionId() {
         const versionId = this.getVersionId();
-        if (versionId) {
+        if (versionId === VersioningConstants.ExternalNullVersionId) {
+            return versionId;
+        } else if (versionId) {
             return VersionIDUtils.encode(versionId);
         }
+        return undefined;
     }
 
     /**
--- a/lib/versioning/constants.ts
+++ b/lib/versioning/constants.ts
@@ -15,4 +15,5 @@ export const VersioningConstants = {
         v1mig: 'v1mig',
         v1: 'v1',
     },
+    ExternalNullVersionId: 'null',
 };
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=16"
   },
-  "version": "8.1.75",
+  "version": "8.1.76",
   "description": "Common utilities for the S3 project components",
   "main": "build/index.js",
   "repository": {
--- a/tests/unit/models/ObjectMD.spec.js
+++ b/tests/unit/models/ObjectMD.spec.js
@@ -1,6 +1,8 @@
 const assert = require('assert');
 const ObjectMD = require('../../../lib/models/ObjectMD').default;
 const constants = require('../../../lib/constants');
+const ExternalNullVersionId = require('../../../lib/versioning/constants')
+    .VersioningConstants.ExternalNullVersionId;
 
 const retainDate = new Date();
 retainDate.setDate(retainDate.getDate() + 1);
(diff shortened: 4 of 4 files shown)


lib/models/ObjectMD.ts
```diff
@@ -892,17 +892,36 @@ export default class ObjectMD {
         return this._data.versionId;
     }
 
+    /**
+     * Get metadata raw versionId value, when object
+     * is in a versioning suspended state it returns 'null'
+     * 
+     * @returns {undefined|string} The object's versionId
+     */
+    getUnencodedVersionId() {
```
francoisferrand: should this not be in `getVersionId` ?

naively, it seems that having 3 methods to get the versionId is a bit much : so I am thinking we may be missing something here. What are the "use case" for each of the 3 methods?

welansari: I believe `getVersionId` is a getter for `versionId` with no extra logic, that's why i decided to add a function for our specific case.

`getEncodedVersionId` is used to get the encoded versionId, so the versionId seen by the clients, a null version should never return it's internal versionId, and so here we return `'null'`

`getUnencodedVersionId` has the same logic as the previous but returns the unencoded versionId

francoisferrand: that's my point : I don't think we need to keep the "accessor with no extra logic" if we don't use it, and it seems we should always either get the encoded or unencoded versionid...

On top of that: there are pieces of code which can be used kind of interchangeably with ObjectMD or other types (KafkaEntry for exemple) : there it expects to have the same method in both types, and I am wondering if that code should benefit from the change "for free"...

francoisferrand:
BTW : we should also check if this change impacts (or can simplify) the code in s3utils.
There is some code there which deals with versionId, may be easier/safer/... to use the code form arsenal (since it already depends on arsenal anyway)

lib/models/ObjectMD.ts
welansari: Looking at where and how the accessor is used, it seems we can indeed add logic to it.

The other types in backbeat all inherit from ObjectMD, so there's nothing to do there. We use ObjectMD function for version logic.

welansari:
@francoisferrand I looked into S3Utils `CrrExistingObjects`. It seems versioning suspended objects are not handled in a proper manner i created a ticket for fixing that : [S3UTILS-114](https://scality.atlassian.net/browse/S3UTILS-114).

The issue is that we are using the  internal version id of the versioning suspended objects to get their metadata, which will not work as they only have a master version and so we need to specify `undefined` in this case.

lib/models/ObjectMD.ts
```diff
@@ -901,20 +901,27 @@ export default class ObjectMD {
      * @return The object versionId
      */
     getVersionId() {
+        const isNull = this.getIsNull();
+        if (isNull) {
+            return 'null';
```
ghost: Let's have it as a const somewhere to avoid duplication, WDYT?
Plus `'null'` is not a usual value in js, it will simplify the code understanding

lib/models/ObjectMD.ts
```diff
@@ -901,20 +901,27 @@ export default class ObjectMD {
      * @return The object versionId
      */
     getVersionId() {
+        const isNull = this.getIsNull();
+        if (isNull) {
```
ghost: It's purely an opinion here, but that can be simplified as follow, as the variable name does not add any more information:

```suggestion
        if (this.getIsNull()) {
```

lib/models/ObjectMD.ts
```diff
@@ -901,20 +901,27 @@ export default class ObjectMD {
      * @return The object versionId
      */
     getVersionId() {
+        const isNull = this.getIsNull();
```
ghost: (picking random line)
Feel free to bump project version is you want to use it after int he BB-264 PR.

ghost (commented):
Some quick comments, the main reason I'm not approving is because of the missing tests.
Otherwise, I reviewed the code structure/context (BB-264).

lib/models/ObjectMD.ts
```diff
@@ -901,20 +901,27 @@ export default class ObjectMD {
...
+        }
         return this._data.versionId;
     }
 
     /**
      * Get metadata versionId value in encoded form (the one visible
      * to the S3 API user)
      *
-     * @return The encoded object versionId
+     * @return {undefined|string} The encoded object versionId
```
ghost, line 915: Sorry forgot to include the main comment:
(picking a random line here)

As it is a bugfix ticket, I would be expecting to see at least one new or updated test to make sure we don't face this issue again, and that we do not add more non-covered lines of code.

ghost (approved):

jonathan-gramain (approved):

PR #2044, opened 2022-11-16
Feature: ARSN-280 abstract update

welansari:
Issue: [ARSN-280](https://scality.atlassian.net/browse/ARSN-280)

**Context**
When deleting an object, and to keep the latest object's metadata in the oplog, we perform an update on the object, setting a deletion flag before actually deleting it.

Backbeat extensions will now use the update operation we do just before deleting an object as the new deletion event, the other "actual" delete will be ignored as it doesn't contain any metadata.

To do that we need to overwrite the event type at the logConsumer level.

### DIFF
--- a/lib/storage/metadata/mongoclient/ListRecordStream.js
+++ b/lib/storage/metadata/mongoclient/ListRecordStream.js
@@ -108,9 +108,26 @@ class ListRecordStream extends stream.Readable {
             if (value && value.tags) {
                 value.tags = unescape(value.tags);
             }
+            // updates overwrite the whole metadata,
+            // so they are considered as puts
+            let type = 'put';
+            // When the object metadata contain the "deleted"
+            // flag, it means that the operation is the update
+            // we perform before the deletion of an object. We
+            // perform the update to keep all the metadata in the
+            // oplog. This update is what will be used by backbeat
+            // as the delete operation so we put the type of operation
+            // for this event to a delete.
+            // Backbeat still receives the actual delete operations
+            // but they are ignored as they don't contain any metadata.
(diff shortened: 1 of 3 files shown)


lib/storage/metadata/mongoclient/ListRecordStream.js
```diff
@@ -108,9 +108,17 @@ class ListRecordStream extends stream.Readable {
...
+            // we perform an update operation before each
+            // delete to keep the metadata in the oplog
+            // this operation should be considered a delete
+            if (value && value.deleted) {
+                type = 'delete';
+            }
             entry = {
-                type: 'put', // updates overwrite the whole metadata,
-                // so they are considered as puts
+                type,
```
francoisferrand, line 130: should we not also filter the "delete" event accordingly (e.g. line 131) ?

welansari: I'm afraid ignoring all deletion events would mean loosing delete events for all other types of collections (__metastore, ...) in case we ever decide to use them. The update->delete logic is only done on bucket "type" collections.

francoisferrand (approved):

lib/storage/metadata/mongoclient/ListRecordStream.js
francoisferrand: would be nice to add a comment for posterity, to explain why we cannot filter the "real" delete event and why this is not a problem (e.g. it will be filtered by LogReader in Backbeat, as the event will not have any metadata)

ghost (approved):

PR #2048, opened 2022-11-23
Migrate to newer azure SDK

francoisferrand:
- Upgrade azure sdk
- Fix yarn warning
- Use new url parser in mongoclient
- Cleanup errorWrapper to better fit with promise-based API used by the new SDK

Note:
- [x] Tested in cloudserver → https://eve.devsca.com/github/scality/cloudserver/#/builders/3/builds/19396

Issue: ARSN-281

### DIFF
--- a/lib/s3middleware/azureHelpers/mpuUtils.ts
+++ b/lib/s3middleware/azureHelpers/mpuUtils.ts
@@ -62,7 +62,7 @@ export const getBlockId = (
     const paddedSubPart = padString(subPartIndex, 'subPart');
     const blockId = `${uploadId}${splitter}partNumber${paddedPartNumber}` +
         `${splitter}subPart${paddedSubPart}${splitter}`;
-    return padString(blockId, 'part');
+    return Buffer.from(padString(blockId, 'part')).toString('base64');
 };
 
 export const getSummaryPartId = (partNumber: number, eTag: string, size: number) => {
--- a/lib/s3middleware/azureHelpers/mpuUtils.ts
+++ b/lib/s3middleware/azureHelpers/mpuUtils.ts
@@ -105,7 +105,6 @@ type ErrorWrapperFn = (
     s3Method: string,
     azureMethod: string,
     command: (client: azure.ContainerClient) => Promise<any>,
-    resultCb: (err: azure.RestError | null | undefined, result?: any) => void,
     log: RequestLogger,
     cb: (err: ArsenalError | null | undefined) => void,
 ) => void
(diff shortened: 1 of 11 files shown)


lib/storage/data/external/AzureClient.js
```diff
@@ -1,62 +1,77 @@
...
+            }
+            // NOTE: config.proxy.certs is not supported
+            const parsedUrl = new URL(config.proxy.url);
+            return {
+                host: parsedUrl.host,
+                port: parsedUrl.port || 80,
+                username: parsedUrl.username || undefined,
+                password: parsedUrl.password || undefined,
+            };
+        })();
```
tmacro, line 35: Why is a lambda needed for this?
```suggestion
        let proxyOptions;
        if (config.proxy && config.proxy.url) {
            // NOTE: config.proxy.certs is not supported
            const parsedUrl = new URL(config.proxy.url);
            proxyOptions = {
                host: parsedUrl.host,
                port: parsedUrl.port || 80,
                username: parsedUrl.username || undefined,
                password: parsedUrl.password || undefined,
            };
        }
```

francoisferrand: It is not _needed_, there are obviously other ways to write the code. But it allows keeping the variable `const` with a single initialization, to better convey intention and avoid mis-assignments, all the while keeping the all logic in the same place.

tmacro (approved):

lib/s3middleware/azureHelpers/mpuUtils.ts
```diff
@@ -172,19 +171,22 @@ const putNextSubPart = (
...
-        log, cb);
+    errorWrapperFn('uploadPart', 'createBlockFromStream', async client => {
+        try {
+            const result = await client.getBlockBlobClient(objectKey)
+                .stageBlock(subPartId, () => subPartStream, subPartSize, {});
+            resultsCollector.pushResult(null, subPartIndex);
+        } catch (err: any) {
+            resultsCollector.pushResult(err, subPartIndex);
+        }
+    }, log, () => {});
```
welansari: I'm not sure i fully understand why we no longer wait for the blocks to be PUT.

it seems to me that we still do asynchronous operations  as we still await for the client to be retrieved ?

I know there's an async alternative to `stageBlock` (`stageBlockAsync`), so i'm assuming this part is synchronous. but not sure about `getBlockBlobClient`.

francoisferrand: getBlockBlobClient() is synchronous, it just creates the block-blob client object (from the container client) : i.e. fundamentally it just adds the blob name in the url...

That said, it is as synchronous as before : just the error handling bit changed, so the error "erasure" needs to be done in the callback, which is now an async function (instead of just returning a promise, as we did in the earlier commit of this PR)

welansari (approved):

PR #2054, opened 2022-12-09
bugfix: ARSN-284 fix and refactor Delimiter and DelimiterMaster

jonathan-gramain:
**Note to reviewers**: I apoligize for the size of this PR, as it was hard to split it into multiple pieces, most of it is in tests though. To ease the review I suggest to look at the new files `delimiter.ts` and `delimiterMaster.ts` directly instead of comparing line-by-line with the old code.

Large refactor of Delimiter and DelimiterMaster classes to typescript,that fixes most known issues with the previous implementation.

The new implementation uses explicit states to manage various conditions, instead of relying on a bunch of internal variable values and maintaining their state. It allows a more robust code flow and
fixes issues related to prefix skipping that were hard to fix by keeping the overall logic of the previous implementation.

This refactor brings the following bug fixes and enhancements:

- prefixes with delete markers and non-deleted objects are now always included in CommonPrefixes (S3C-7248)

- no more duplication of internal range listings when doing skip-scan over prefixes (discovered when analyzing regressions for S3C-4682)

- the skip-scan mecanism for prefixes and versions is no more disturbed by delete markers and PHD keys (S3C-2930)

- NextMarker is now always set to a valid, listed or listable key (that may still be hidden under a CommonPrefix), no more manipulation of next marker to avoid corner-cases with keys ending with a prefix (S3C-4682 and S3C-7274)

- deleting a delete marker immediately allows the new current version to be visible in the listing (S3C-7272)

- Expecting lower CPU usage overall, as the number of checks to do in each state is reduced (may help to reduce the load and reduce impact of cases such as S3C-3946)

- Uses typescript to allow more sanity checks


jonathan-gramain:
ping

lib/algos/list/delimiterMaster.ts
```diff
@@ -0,0 +1,171 @@
...
+
+const VID_SEP = VSConst.VersionId.Separator;
+const { DbPrefixes } = VSConst;
+
+const enum DelimiterMasterFilterStateId {
+    SkippingVersionsV0 = 101,
+    WaitVersionAfterPHDV0 = 102,
+};
+
+interface DelimiterMasterFilterState_SkippingVersionsV0 extends FilterState {
```
watsaqat, line 22: instead of creating a new interface for the different states, you could do:

```
interface DelimiterMasterFilterStateId {
     SkippingVersionsV0: 101,
     WaitVersionAfterPHDV0: 102,
 };

 type DelimiterMasterFilterStateIdValue = DelimiterMasterFilterStateId[keyof DelimiterMasterFilterStateId]

 export interface DelimiterMasterFilterState extends FilterState {
    id: DelimiterMasterFilterStateIdValue,
    masterKey: string,
 }
 ```
 

 The advantage of `DelimiterMasterFilterStateId[keyof DelimiterMasterFilterStateId]` is that it extracts the values of an interface into a union type. So now, the type is restricted to only `101 | 102`. I'd also recommend we make use of this in a few functions that right now are returning `number` when there's only a limited amount of states they can return.

File: lib/algos/list/delimiterMaster.ts [hunk 2]
```diff
@@ -0,0 +1,171 @@
...
+                this.state = <DelimiterFilterState_NotSkipping> {
+                    id: DelimiterFilterStateId.NotSkipping,
+                };
+            }
+        }
+        // in v1, we can directly use Delimiter's implementation,
+        // which is already set to the proper state
+    }
+
+    filter_onNewMasterKeyV0(key, value) {
```
watsaqat: Missing return type. One option is creating an interface such as 
```
interface DelimiterReturnStates {
   FILTER_ACCEPT: 1,
   FILTER_SKIP: 2,
   ...
}
```
And using `Pick` to narrow down the allowed returns for functions

`filter_onNewMasterKeyV0(key: string, value: string): Pick<DelimiterReturnStates, 'FILTER_ACCEPT | FILTER_SKIP'>`
or simply define types individually and have the return type be `ACCEPT | SKIP`

lib/algos/list/delimiterMaster.ts
```diff
@@ -0,0 +1,171 @@
...
+            this.setState(<DelimiterFilterState_SkippingPrefix> {
+                id: DelimiterFilterStateId.SkippingPrefix,
+                prefix: DbPrefixes.Replay,
+            });
+            return FILTER_SKIP;
+        }
+        return this.addCommonPrefixOrContents(key, value);
+    }
+
+    keyHandler_NotSkippingPrefixNorVersionsV0(key, value) {
```
watsaqat: types

lib/algos/list/delimiterMaster.ts
```diff
@@ -0,0 +1,171 @@
...
+        return this.filter_onNewMasterKeyV0(key, value);
+    }
+
+    keyHandler_SkippingVersionsV0(key, value) {
```
watsaqat: types

lib/algos/list/delimiterMaster.ts
```diff
@@ -0,0 +1,171 @@
...
+        /* In the SkippingVersionsV0 state, skip all version keys
+         * (<key><versionIdSeparator><version>) */
+        const versionIdIndex = key.indexOf(VID_SEP);
+        if (versionIdIndex !== -1) {
+            return FILTER_SKIP;
+        }
+        return this.filter_onNewMasterKeyV0(key, value);
+    }
+
+    keyHandler_WaitVersionAfterPHDV0(key, value) {
```
watsaqat: types. And, should the comments be a block comment?

lib/algos/list/delimiterVersions.js
```diff
@@ -151,6 +151,27 @@ class DelimiterVersions extends Delimiter {
         return FILTER_ACCEPT;
     }
 
+    /**
+     * Add a Common Prefix in the list
+     * @param {String} key   - object name
+     * @param {Number} index - after prefix starting point
+     * @return {Boolean}     - indicates if iteration should continue
+     */
+    addCommonPrefix(key, index) {
```
watsaqat, line 160: types

lib/algos/list/delimiter.ts
```diff
@@ -180,67 +253,65 @@ class Delimiter extends Extension {
...
+
+    setState(state: FilterState): void {
+        this.state = state;
+    }
+
+    setKeyHandler(stateId: number, keyHandler: KeyHandler): void {
+        this.keyHandlers[stateId] = keyHandler;
+    }
+
+    handleKey(key: string, value: string): number {
```
watsaqat: if we're returning a state, try to make use of union types to restrict the returned values to certain allowed states.

lib/algos/list/delimiter.ts
```diff
@@ -2,21 +2,41 @@
...
+export interface FilterState {
+    id: number,
+};
+
+export const enum DelimiterFilterStateId {
+    NotSkipping = 1,
+    SkippingPrefix = 2,
+};
+
+export interface DelimiterFilterState_NotSkipping extends FilterState {
```
watsaqat, line 24: see this [comment in delimiterMaster.ts](https://github.com/scality/Arsenal/pull/2054/files?w=0#r1050257767)

lib/algos/list/delimiterMaster.ts
jonathan-gramain: The `masterKey` field here, although named the same, is specific to each state (`SkippingVersionsV0` and `WaitVersionsAfterPHDV0` each have their own `masterKey` in their own state), hence I'd rather not set it in a generic way.

I'd like to improve type checking on the `id` field but the issue is that the `Delimiter` class is generic and must be able to handle states outside of its own list of states (e.g. those from `DelimiterMaster` which inherits from `Delimiter`). I'm not familiar enough with typescript to know if that's doable in a clean way, but I did not find a way, but if you know of a way and checked that it works in this case, feel free to suggest it.

Regarding the `keyof` operator, it does not look to give the expected results, as we cannot assign a value to `id` from the enum afterwards, typescript complains. But for the above reason I'm not even sure we can make it work.

lib/algos/list/delimiterMaster.ts
jonathan-gramain: We seem to get the same result by specifying the return type as the interface name directly, so I'm going to do this instead. By the way, the `Pick` [documentation](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys) says that it's used to pick specific fields from a struct, but maybe it also works in this case.

lib/algos/list/delimiterMaster.ts
jonathan-gramain: Do you mean `/* */` instead of multiple `//`? I don't know to be honest, what is the best practice in such case (or otherwise, what is the issue with single-line comments)

lib/algos/list/delimiter.ts
jonathan-gramain: Should be the case now

tmacro (approved):

watsaqat (approved):

PR #2059, opened 2022-12-22
Improvement/arsn 291 create new bucketmd version

ghost:
Goal: support a new bucketMd scheme to store any capability we might want to support. Here the focus is on Veeam12, where the Veeam12 XML files are stored in the bucket MD directly, under the capabilities.

### DIFF
--- a/lib/models/BucketInfo.ts
+++ b/lib/models/BucketInfo.ts
@@ -912,14 +912,23 @@ export default class BucketInfo {
 
     /**
      * Get the value of bucket capabilities
-     * @param capability? - if provided, will return a specific capacity
-     * @return - capabilities of the bucket, or null
+     * @return - capabilities of the bucket
      */
-    getCapabilities(capability?: string) {
+    getCapabilities() {
+        return this._capabilities;
+    }
+
+    /**
+     * Get a specific bucket capability
+     * 
+     * @param capability? - if provided, will return a specific capacity
+     * @return - capability of the bucket
+     */
+    getCapability(capability: string) : VeeamSOSApi | undefined {
         if (capability && this._capabilities && this._capabilities[capability]) {
             return this._capabilities[capability];
         }
-        return this._capabilities;
+        return undefined;
     }
     
     /**
--- a/lib/models/BucketInfo.ts
+++ b/lib/models/BucketInfo.ts
@@ -13,7 +13,7 @@ import { areTagsValid, BucketTag } from '../s3middleware/tagging';
 // WHEN UPDATING THIS NUMBER, UPDATE BucketInfoModelVersion.md CHANGELOG
 // BucketInfoModelVersion.md can be found in documentation/ at the root
 // of this repository
-const modelVersion = 14;
+const modelVersion = 16;
 
 export type CORS = {
     id: string;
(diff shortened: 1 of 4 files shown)


XinLi0207 (approved):

lib/models/BucketInfo.ts
```diff
@@ -37,6 +37,41 @@ export type VersioningConfiguration = {
...
+        Capacity: number,
+        Available: number,
+        Used: number,
+        LastModified?: string,
+    },
+};
+
+// Capacity contains all specifics from external products supported by
+// our S3 implementation, at bucket level
+export type Capacity = {
```
francoisferrand: this type should be `Capabilies` : it may contain 'capacity' for veam, but other info if we do other integration...

lib/models/BucketInfo.ts
```diff
@@ -37,6 +37,41 @@ export type VersioningConfiguration = {
     MfaDelete: any;
 };
 
+export type VeeamCapacity = {
```
francoisferrand: type should be either `VeamCapability` (to make it generic and support other veam capabilities in the feature) or `VeamSosApi` (and we would possibly have other VeamXXXX capabilities in the future)

francoisferrand: type should be either `VeamCapability` (to make it generic and support other veam capabilities in the feature) or `VeamSosApi` (and we would possibly have other VeamXXXX capabilities in the future)

lib/models/BucketInfo.ts
ghost: Done: https://github.com/scality/Arsenal/pull/2059/files#diff-4ee3cd5de98073dadb72b27f034f4979480c04388cc4f4538f4895373527f174R71

lib/models/BucketInfo.ts
ghost: I chose `VeeamCapabilities` as it may contain more than SOSAPI, if needed: https://github.com/scality/Arsenal/pull/2059/files#diff-4ee3cd5de98073dadb72b27f034f4979480c04388cc4f4538f4895373527f174R40

francoisferrand: would be fine by me, except the field in capabilities is named `VeeamSOSApi` : so the field should be renamed accordingly!

ghost: Ok, i'll use `VeeamSOSApi` for simplicity here, to avoid logic change in child PRs :smile: I'm fine with both anyway

Done & squashed in https://github.com/scality/Arsenal/pull/2059/commits/a6b234b7a8604d2c2d9dd49e2e35b1a7d9f8a50b#diff-4ee3cd5de98073dadb72b27f034f4979480c04388cc4f4538f4895373527f174R72

lib/models/BucketInfo.ts
```diff
@@ -13,7 +13,7 @@ import { areTagsValid, BucketTag } from '../s3middleware/tagging';
 // WHEN UPDATING THIS NUMBER, UPDATE BucketInfoModelVersion.md CHANGELOG
 // BucketInfoModelVersion.md can be found in documentation/ at the root
 // of this repository
-const modelVersion = 14;
+const modelVersion = 16;
```
francoisferrand, line 16: do we need to skip one version?

lib/models/BucketInfo.ts
```diff
@@ -868,4 +909,25 @@ export default class BucketInfo {
         this._tags = tags;
         return this;
     }
+
+    /**
+     * Get the value of bucket capabilities
+     * @param capability? - if provided, will return a specific capacity
+     * @return - capabilities of the bucket, or null
+     */
+    getCapabilities(capability?: string) {
```
francoisferrand: nit : this actually gets a single capability, when specified...

lib/models/BucketInfo.ts
```diff
@@ -868,4 +909,25 @@ export default class BucketInfo {
...
+        if (capability && this._capabilities && this._capabilities[capability]) {
+            return this._capabilities[capability];
+        }
+        return this._capabilities;
```
francoisferrand: would  it be better to either have 2 functions, to either get a single capability or get the whole list?

i fear typescript may not properly infer typing which depend on the parameter, and using this function may essentially be the same as having `any` type (or a choice of any capability) : where doing `this._capabilites?.<capability>` (or dedicated helper function like `getVeeamSosApi()` has a specific type....

francoisferrand (approved):

lib/models/BucketInfo.ts
anurag4DSB: 15 seems already defined in documentation/BucketInfoModelVersion.md

anurag4DSB (approved):

lib/models/BucketInfo.ts
ghost: yes, the bump in the ts file was missed, I wanted to be consistent with what is documented.

lib/models/BucketInfo.ts
ghost: Sure, I will create two functions, it will also let me have the right naming

ghost: I created this commit: https://github.com/scality/Arsenal/pull/2059/commits/f17d52b60281b8f33600a3b60816ffc5d482a918

Keeping this PR open for now, please feel free if you were thinking of something else!

PR #2073, opened 2023-01-31
ARSN-299 add location pause status model

welansari:
Issue: [ARSN-299](https://scality.atlassian.net/browse/ARSN-299)

Add model to handle pause/resume status of a location

[ARSN-299]: https://scality.atlassian.net/browse/ARSN-299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/models/LocationStatus.ts
+++ b/lib/models/LocationStatus.ts
@@ -1,0 +1,166 @@
+type status = {
+    paused: boolean,
+    scheduledResume: string|null,
+};
+
+type serviceStatus = {
+    crr: status|null,
+    ingestion: status|null,
+    lifecycle: status|null,
+};
+
+/**
+ * Class to manage a location's pause/resume state on its
+ * different services.
+ * Current supported services are : crr, ingestion, lifecycle
+ */
+export default class LocationStatus {
+    _data: serviceStatus;
+
+    /**
+     * @constructor
+     * @param services services to init
+     * @param locationStatus initial location status values
+     */
+    constructor(services: string[], locationStatus?: {
+        crr: serviceStatus|null,
+        ingestion: serviceStatus|null,
+        lifecycle: serviceStatus|null,
+    } | serviceStatus) {
+        this._data = this._initStatus(services);
+        if (locationStatus) {
+            const data = locationStatus instanceof LocationStatus ?
+                locationStatus._data : locationStatus;
+            Object.keys(this._data).forEach(svc => {
+                this._data[svc] = {
+                    paused: data[svc]?.paused || false,
+                    scheduledResume: data[svc]?.scheduledResume || null
+                }
+            });
+        }
+    }
+
+    /**
+     * Initializes the status of all services
+     * The default status of a service is unpaused
+     * @returns {LocationServiceStatus}
+     */
+    _initStatus(servicesToInit: string[]): serviceStatus {
+        const initStatus = {
+            paused: false,
+            scheduledResume: null,
+        };
+        return {
+            crr: servicesToInit.includes('crr') ? initStatus : null,
+            ingestion: servicesToInit.includes('ingestion') ? initStatus : null,
+            lifecycle: servicesToInit.includes('lifecycle') ? initStatus : null,
+        };
+    }
+
+    /**
+     * initializes a service status
+     * @param service service name
+     */
+    _initService(service: string) {
+        this._data[service] = {
+            paused: false,
+            scheduledResume: null,
+        };
+    }
+
+    /**
+     * @param service service name
+     * @param paused true if paused
+     */
+    setServicePauseStatus(service: string, paused: boolean) {
+        if (Object.keys(this._data).includes(service)) {
+            if (!this._data[service]) {
+                this._initService(service);
+            }
+            this._data[service].paused = paused;
+        }
+    }
+
+    /**
+     * @param service service name
+     * @returns true if paused
+     */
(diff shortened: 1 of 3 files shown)


lib/models/LocationStatus.ts
```diff
@@ -0,0 +1,143 @@
...
+export type LocationServiceStatus = {
+    replication: serviceStatus,
+    ingestion: serviceStatus,
+    lifecycle: serviceStatus,
+};
+
+/**
+ * Class to manage a location's pause/resume state on its
+ * different services.
+ * Current supported services are : repication, ingestion, lifecycle
```
francoisferrand: typo:

```suggestion
 * Current supported services are : replication, ingestion, lifecycle
```

francoisferrand (commented):
do we need to put this class in arsenal?
it really seems to just be part of backbeat-api implementations?

(and actually does not really represent the state, since ingestion & replication are not actually stored there [yet]...)

lib/models/LocationStatus.ts
```diff
@@ -0,0 +1,143 @@
...
+            this.setServicePauseStatus(svc, true);
+        });
+    }
+
+    /**
+     * @param service service(s) name
+     * @param schedule date to resume service(s)
+     */
+    resumeLocation(service: string | string[], schedule?: Date) {
+        const servicesList = Array.isArray(service) ?
```
XinLi0207, line 138: can we resume a location if the location is not paused? I am thinking if we need to first check if a location is paused, then resume it

XinLi0207: I see now, you checked [here](https://github.com/scality/backbeat/pull/2361/commits/d54b244609ff97e235483c7c9035c3edf0ba8e37#diff-feb400da3fd9b9f1e83c06416f27265ba93371e0f311b75512ec575ebd590b51R1267)

lib/models/LocationStatus.ts
```diff
@@ -0,0 +1,150 @@
+export type serviceStatus = {
+    paused: boolean,
+    scheduledResume: string|null,
+};
+
+export type LocationServiceStatus = {
```
francoisferrand: this is not the *Service* status : more like the overall location status (just like the class...)

→ maybe just `locationStatus`, or declare the type inside the LocationStatus class ?
→ do we need to export these 2 types ?

lib/models/LocationStatus.ts
```diff
@@ -0,0 +1,150 @@
...
+    crr: serviceStatus,
```
francoisferrand: can the fields be 'null' here as well (instead of in the constructor) ?

"functionally", it may be nice to store null instead of the details of the field : so that we don't make it look like the state is management here (for crr/ingestion)

lib/models/LocationStatus.ts
```diff
@@ -0,0 +1,166 @@
...
+    getServicePauseStatus(service: string) : boolean | null {
+        if (!this._data[service]) {
+            return null;
+        }
+        return this._data[service].paused;
+    }
+
+    /**
+     * @param service service name
+     */
```
ghost, line 97: Missing date param description

lib/models/LocationStatus.ts
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * @return location status object
+     */
+    getValue() : serviceStatus{
+        return this._data;
+    }
+
+    /**
+     * @returns serialized location status data
+     */
```
ghost, line 162: not consistent `@return` / `@returns`

lib/models/LocationStatus.ts
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * @param service service name
+     * @returns scheduled resume date
+     */
+    getServiceResumeSchedule(service: string) : Date | null {
+        const schedule = this._data[service].scheduledResume;
+        if (!schedule) {
+            return null;
+        }
+        return new Date(schedule);
```
ghost, line 117: Just in case the service is not yet defined, should we make sure it's defined before accessing `scheduledResume` ?

```suggestion
        const schedule = this._data[service]?.scheduledResume;
        if (!schedule) {
            return null;
        }
        return new Date(schedule);
```

lib/models/LocationStatus.ts
```diff
@@ -0,0 +1,166 @@
...
+            service : [service];
+        servicesList.forEach(svc => {
+            if (!this.getServicePauseStatus(svc)) {
+                return;
+            }
+            let shouldPause = false;
+            if (schedule) {
+                shouldPause = true
+            }
+            this.setServicePauseStatus(svc, shouldPause);
```
ghost, line 148: Just a suggestion, but only my personal opinion here :smile: 
```suggestion
            this.setServicePauseStatus(svc, !!schedule);
```

lib/models/LocationStatus.ts
```diff
@@ -0,0 +1,166 @@
...
+            this.setServiceResumeSchedule(svc, schedule||null);
```
ghost, line 149:
```suggestion
            this.setServiceResumeSchedule(svc, schedule || null);
```

ghost (approved):

welansari:
Closing this PR as it will be implemented directly in backbeat (the model is highly correlated with backbeat).

PR #2075, opened 2023-02-06
ARSN-298: support `NewerNoncurrentVersions`

alexanderchan-scality:
add support for `NewerNoncurrentVersions` params for
`NoncurrentVersionsExpiration` actions

### DIFF
--- a/lib/models/LifecycleConfiguration.ts
+++ b/lib/models/LifecycleConfiguration.ts
@@ -508,6 +508,7 @@ export default class LifecycleConfiguration {
      *              days: <value>,
      *              date: <value>,
      *              deleteMarker: <value>
+     *              newerNoncurrentVersions: <value>,
      *          },
      *      ],
      * }
@@ -520,7 +521,8 @@ export default class LifecycleConfiguration {
                 actionName: string;
                 days?: number;
                 date?: number;
-                deleteMarker?: boolean
+                deleteMarker?: boolean;
+                newerNoncurrentVersions?: number
             }[];
         } = {
             propName: 'actions',
@@ -544,7 +546,7 @@ export default class LifecycleConfiguration {
             if (action.error) {
                 actionsObj.error = action.error;
             } else {
-                const actionTimes = ['days', 'date', 'deleteMarker'];
+                const actionTimes = ['days', 'date', 'deleteMarker', 'newerNoncurrentVersions'];
                 actionTimes.forEach(t => {
                     if (action[t]) {
                         // eslint-disable-next-line no-param-reassign
@@ -693,6 +695,7 @@ export default class LifecycleConfiguration {
      * nvExpObj = {
      *      error: <error>,
      *      days: <value>,
+     *      newerNoncurrentVersions: <value>,
      * }
      */
     _parseNoncurrentVersionExpiration(rule: any) {
@@ -703,14 +706,41 @@ export default class LifecycleConfiguration {
                 'NoncurrentDays');
             return { error };
         }
+
+        const actionParams: {
+            error?: ArsenalError;
+            days: number;
+            newerNoncurrentVersions: number;
+        } = {
+            days: 0,
+            newerNoncurrentVersions: 0,
+        };
+
         const daysInt = parseInt(subNVExp.NoncurrentDays[0], 10);
         if (daysInt < 1) {
             const msg = 'NoncurrentDays is not a positive integer';
             const error = errors.InvalidArgument.customizeDescription(msg);
             return { error };
         } else {
-            return { days: daysInt };
+            actionParams.days = daysInt;
         }
+
+        if (subNVExp.NewerNoncurrentVersions) {
+            const newerVersionsInt = parseInt(subNVExp.NewerNoncurrentVersions[0], 10);
+
+            if (Number.isNaN(newerVersionsInt) || newerVersionsInt < 1) {
+                const msg = 'NewerNoncurrentVersions is not a positive integer';
+                const error = errors.InvalidArgument.customizeDescription(msg);
+                return { error };
+            }
+
+            actionParams.newerNoncurrentVersions = newerVersionsInt;
+
+        } else {
+            actionParams.newerNoncurrentVersions = 0;
+        }
+
+        return actionParams;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


lib/models/LifecycleConfiguration.ts
```diff
@@ -703,14 +706,38 @@ export default class LifecycleConfiguration {
...
+            if (newerVersionsInt < 1) {
+                const msg = 'NewerNoncurrentVersions is not a positive integer';
+                const error = errors.InvalidArgument.customizeDescription(msg);
+                return { error };
+            }
+
+            actionParams.newerNoncurrentVersions = newerVersionsInt;
+
+        } else {
+            actionParams.newerNoncurrentVersions = 0;
```
ghost, line 740: If we set `newerNoncurrentVersions` everytime, it may be better not to have it as an optional field, what about

```ts
        const actionParams: {
            error?: ArsenalError;
            days: number;
            newerNoncurrentVersions: number;
        } = { days: 0 };
```
?

lib/models/LifecycleConfiguration.ts
```diff
@@ -852,13 +885,18 @@ export default class LifecycleConfiguration {
...
                     entry.addAbortMPU(days!);
                     return;
                 }
                 if (actionName === 'NoncurrentVersionExpiration') {
-                    entry.addNCVExpiration(days!);
+                    entry.addNCVExpiration('NoncurrentDays', days!);
+
+                    if (newerNoncurrentVersions) {
+                        entry.addNCVExpiration('NewerNoncurrentVersions', newerNoncurrentVersions);
+                    }
```
ghost: If the above comment makes sense:

```suggestion
                    entry.addNCVExpiration('NewerNoncurrentVersions', newerNoncurrentVersions!);
```

lib/models/LifecycleRule.ts
```diff
@@ -136,10 +140,18 @@ export default class LifecycleRule {
...
     /**
      * NoncurrentVersionExpiration
-     * @param days - NoncurrentDays
+     * @param prop - Property must be defined in `validProps`
+     * @param value - integer for `NoncurrentDays` and `NewerNoncurrentVersions` 
      */
-    addNCVExpiration(days: number) {
-        this.ncvExpiration = { NoncurrentDays: days };
+    addNCVExpiration(prop: 'NoncurrentDays' | 'NewerNoncurrentVersions', value: number): this;
+    addNCVExpiration(prop: string, value: number ) {
```
ghost: Nit:
```suggestion
    addNCVExpiration(prop: string, value: number) {
```

ghost (approved):
LGTM, just some quick comments (I also spent time reading about how this is used by AWS)

nicolas2bert (approved):

lib/models/LifecycleRule.ts
```diff
@@ -136,10 +140,16 @@ export default class LifecycleRule {
...
+     * @param prop - Property must be defined in `validProps`
+     * @param value - integer for `NoncurrentDays` and `NewerNoncurrentVersions` 
      */
-    addNCVExpiration(days: number) {
-        this.ncvExpiration = { NoncurrentDays: days };
+    addNCVExpiration(prop: 'NoncurrentDays' | 'NewerNoncurrentVersions', value: number): this;
+    addNCVExpiration(prop: string, value: number) {
+        const validProps = ['NoncurrentDays', 'NewerNoncurrentVersions'];
+        if (validProps.includes(prop)) {
+            this.ncvExpiration = this.ncvExpiration || { NoncurrentDays: 0 };
```
jonathan-gramain: Are we setting 0 days as a default because the two NCV rules combine with an *and* operator, so an object must match both of them in order to be expired?

package.json
```diff
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=16"
   },
-  "version": "7.10.42",
+  "version": "7.10.43",
```
jonathan-gramain, line 6: Could you bump the version in a separate commit? Makes it easier to cherry-pick things later

lib/models/LifecycleConfiguration.ts
```diff
@@ -703,14 +706,41 @@ export default class LifecycleConfiguration {
...
         if (daysInt < 1) {
             const msg = 'NoncurrentDays is not a positive integer';
             const error = errors.InvalidArgument.customizeDescription(msg);
             return { error };
         } else {
-            return { days: daysInt };
+            actionParams.days = daysInt;
+        }
+
+        if (subNVExp.NewerNoncurrentVersions) {
```
jonathan-gramain, line 728: shouldn't we remove the check at line 703 ` if (!subNVExp.NoncurrentDays)`, now that we introduced a new type of check, or make sure at least one of the two checks is present instead? According to https://docs.aws.amazon.com/AmazonS3/latest/API/API_NoncurrentVersionExpiration.html, it is not required to have `NoncurrentDays` attribute.

lib/models/LifecycleRule.ts
```diff
@@ -49,7 +53,7 @@ export default class LifecycleRule {
             rule.Expiration = this.expiration;
         }
         if (this.ncvExpiration) {
-            rule.NoncurrentVersionExpiration = this.ncvExpiration;
+            rule.NoncurrentVersionExpiration = this.ncvExpiration
```
jonathan-gramain, line 57: probably an unwanted change

lib/models/LifecycleConfiguration.ts
```diff
@@ -703,14 +706,41 @@ export default class LifecycleConfiguration {
...
+            const newerVersionsInt = parseInt(subNVExp.NewerNoncurrentVersions[0], 10);
```
jonathan-gramain, line 729: Is there a validation earlier that we have a parsable integer value at this point, to avoid getting `NaN` back?

lib/models/LifecycleConfiguration.ts
alexanderchan-scality: Tested against  AWS S3 and the `NoncurrentDays` field looks to be required. Schema validation would fail otherwise

lib/models/LifecycleRule.ts
alexanderchan-scality: the initial `0` is due to typescript enforcing the field; will  fix this will  this to be an explicit null and a value

alexanderchan-scality:
ping

PR #2078, opened 2023-02-09
feature/ARSN 298/addHeapDataStructure

alexanderchan-scality:
- ARSN-298: support `NewerNoncurrentVersions`
- ARSN-298: add Min/Max heap data structure

### DIFF
--- a/lib/algos/heap/Heap.ts
+++ b/lib/algos/heap/Heap.ts
@@ -1,0 +1,124 @@
+export enum HeapOrder {
+    Min = -1,
+    Max = 1,
+}
+
+export enum CompareResult {
+    LT = -1,
+    EQ = 0,
+    GT = 1,
+}
+
+export type CompareFunction = (x: any, y: any) => CompareResult;
+
+export class Heap {
+    size: number;
+    _maxSize: number;
+    _order: HeapOrder;
+    _heap: any[];
+    _cmpFn: CompareFunction;
+    
+    constructor(size: number, order: HeapOrder, cmpFn: CompareFunction) {
+        this.size = 0;
+        this._maxSize = size;
+        this._order = order;
+        this._cmpFn = cmpFn;
+        this._heap = new Array<any>(this._maxSize);
+    }
+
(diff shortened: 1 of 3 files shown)


lib/algos/heap/Heap.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export enum CompareResult {
+    LT = -1,
+    EQ = 0,
+    GT = 1,
+}
+
+export type CompareFunction = (x: any, y: any) => CompareResult;
+
+export class Heap {
```
ghost, line 14: I am curious, what was the motivation to implement a heap in Arsenal? Don't we want to reuse existing heaps from proven open source projects?
Also, I am not sure (yet) to understand how it links to the addition of `NewerNoncurrentVersions`, is it to keep track of the objects?

alexanderchan-scality: @williamlardier
not sure about introducing a new dependency for this;  the heap structure should be a relatively simple structure to not need to  do so.
yes. the heap will be used to keep track of the `latest` objects. In the case of `NewNoncurrentVersion`, a heap is used over a list/array due to the asynchronous methods needed to determine if a version needs to be expired which may not be executed in order.

lib/algos/heap/Heap.ts
```diff
@@ -0,0 +1,124 @@
...
+    size: number;
+    _maxSize: number;
+    _order: HeapOrder;
+    _heap: any[];
+    _cmpFn: CompareFunction;
```
ghost, line 19: Do we want to enforce private members?

```suggestion
    #maxSize: number;
    #order: HeapOrder;
    #heap: any[];
    #cmpFn: CompareFunction;
```

ghost (approved):

lib/algos/heap/Heap.ts
ghost: Thanks for the explanation!

nicolas2bert (approved):

PR #2080, opened 2023-02-17
ARSN-300 implement mongoDB logic for lifecycle listings

nicolas2bert:
Introduce a new metadata method: `listLifecycleObject()`

Implement 3 new "delimiters":

- **DelimiterCurrent** used for listing current versions
The Metadata call returns the masters (M prefix) younger than a defined date: `beforeDate`.
No extra filtering action is needed on the Metadata call response.

- **DelimiterNonCurrent** used for listing non-current version: 
The MD response is then processed to only return the non-current versions that became non-current before a defined date: `beforeDate`.

- **DelimiterOrphan** used for listing orphan delete marker:
The MD response is then processed to only return the delete markers with zero noncurrent versions before a defined date: `beforeDate`.

### DIFF
--- a/lib/algos/list/delimiterCurrent.ts
+++ b/lib/algos/list/delimiterCurrent.ts
@@ -1,0 +1,114 @@
+const { Delimiter } = require('./delimiter');
+const Version = require('../../versioning/Version').Version;
+const VSConst = require('../../versioning/constants').VersioningConstants;
+const { BucketVersioningKeyFormat } = VSConst;
+const { inc, FILTER_ACCEPT, FILTER_END, FILTER_SKIP, SKIP_NONE } = require('./tools');
+const VID_SEP = VSConst.VersionId.Separator;
+const { DbPrefixes } = VSConst;
+const TRIM_METADATA_MIN_BLOB_SIZE = 10000;
+import { MDParams, ResultObject } from './types';
+
+/**
+ * Handle object listing with parameters. This extends the base class Delimiter
+ * to return the master/current versions.
+ */
+class DelimiterCurrent extends Delimiter {
+    /**
+     * Delimiter listing of current versions.
+     * @param {Object}  parameters            - listing parameters
+     * @param {String}  parameters.beforeDate - limit the response to keys older than beforeDate
+     * @param {String}  parameters.keyMarker  - key marker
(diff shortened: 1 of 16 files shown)


lib/algos/list/delimiterCurrent.ts
```diff
@@ -0,0 +1,114 @@
...
+            params.gt = `${DbPrefixes.Master}${this.keyMarker}`;
+        }
+
+        if (this.beforeDate) {
+            params.lastModified = {
+                lt: this.beforeDate,
+            };
+        }
+
+        this.start = Date.now();
```
jonathan-gramain, line 76: This does not look to be used in this class

lib/algos/list/delimiterCurrent.ts
```diff
@@ -0,0 +1,114 @@
...
+            params.gte = `${DbPrefixes.Master}${this.prefix}`;
+            params.lt = `${DbPrefixes.Master}${inc(this.prefix)}`;
+        }
+
+        if (this.keyMarker && params.gte && `${DbPrefixes.Master}${this.keyMarker}` >= params.gte) {
+            delete params.gte;
+            params.gt = `${DbPrefixes.Master}${this.keyMarker}`;
+        }
+
+        if (this.beforeDate) {
```
jonathan-gramain, line 70: Is there anything different in `DelimiterCurrent` than in `DelimiterMaster`, except the support for `beforeDate`? In which case I strongly recommend to rely on `DelimiterMaster` and only add the support for this new optional parameter. Listing code is tricky and touchy, the less new code we have to add, the better it is :)

jonathan-gramain (commented):
This PR should be split, it's too large as is and hard to review. For example, only introduce one listing class and its set of tests at a time.

jonathan-gramain (commented):
Also, it's nice to have functional tests, I'd suggest to add unit tests as well, in the vein of what exists for Delimiter, DelimiterMaster and DelimiterVersions, those are more apt to catch specific corner cases that could be hard to catch via functional tests, as we can precisely check inputs and outputs, including for example what the `filter()` function returns.

nicolas2bert:
I am going to split the PR into 3 PRs.

PR #2085, opened 2023-03-01
bugfix/ARSN 308/addLifecycleUtilsNoncurrentVersionSupport

alexanderchan-scality:
- version bump
- ARSN-308: update lifecycle utils to support noncurrent version

### DIFF
--- a/lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
+++ b/lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
@@ -207,6 +207,7 @@ export default class LifecycleUtils {
                 // Names are long, so obscuring a bit
                 const ncve = 'NoncurrentVersionExpiration';
                 const ncd = 'NoncurrentDays';
+                const nncv = 'NewerNoncurrentVersions';
 
                 if (!store[ncve]) {
                     store[ncve] = {};
@@ -214,6 +215,7 @@ export default class LifecycleUtils {
                 if (!store[ncve][ncd] || rule[ncve][ncd] < store[ncve][ncd]) {
                     store[ncve].ID = rule.ID;
                     store[ncve][ncd] = rule[ncve][ncd];
+                    store[ncve][nncv] = rule[ncve][nncv];
                 }
             }
             if (rule.AbortIncompleteMultipartUpload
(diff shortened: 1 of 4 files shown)


nicolas2bert (approved):

lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
```diff
@@ -207,13 +207,15 @@ export default class LifecycleUtils {
...
                 const ncd = 'NoncurrentDays';
+                const nncv = 'NewerNoncurrentVersions';
 
                 if (!store[ncve]) {
                     store[ncve] = {};
                 }
                 if (!store[ncve][ncd] || rule[ncve][ncd] < store[ncve][ncd]) {
                     store[ncve].ID = rule.ID;
                     store[ncve][ncd] = rule[ncve][ncd];
+                    store[ncve][nncv] = rule[ncve][nncv];
```
jonathan-gramain, line 218: Do we support `NewerNoncurrentVersions` without `NoncurrentDays`, like AWS seems to support? If so we should separate the setting of `nncv` specifically (otherwise, it looks fine as is to me, we will set the value of `NewerNoncurrentVersions` from the rule with the most constrained `NoncurrentDays` rule).

tests/unit/s3middleware/LifecycleUtils.spec.js
```diff
@@ -90,6 +90,37 @@ describe('LifecycleUtils::getApplicableRules', () => {
             res.NoncurrentVersionExpiration.NoncurrentDays, 3);
     });
 
+    it('should return earliest applicable rules', () => {
```
jonathan-gramain, line 93: Is the above test still bringing value, considering this new test? Maybe we should remove it, otherwise setting their name differently could help distinguishing them.

lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
alexanderchan-scality: @jonathan-gramain 

testing AWS S3 behavior, `NoncurrentDays` is a required value; 

there looks to be some inconsistencies in AWS documentation:
* https://docs.aws.amazon.com/AmazonS3/latest/API/API_NoncurrentVersionExpiration.html
* other service integration https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-lifecycleconfig-rule-noncurrentversionexpiration.html

in the S3 console, `NoncurrentDays` is a required value

jonathan-gramain (approved):

naren-rajendran (approved):

PR #2088, opened 2023-03-08
ARSN-310 Add logic to list current/master versions for Lifecycle

nicolas2bert:
DelimiterCurrent is used for listing current versions. 
The Metadata call returns the masters (M prefix) younger than a defined date: `beforeDate`. 
No extra filtering action is needed on the Metadata call response.

### DIFF
--- a/lib/algos/list/delimiterCurrent.ts
+++ b/lib/algos/list/delimiterCurrent.ts
@@ -1,0 +1,54 @@
+const { Delimiter } = require('./delimiter');
+
+type ResultObject = {
+    Contents: {
+        key: string;
+        value: string;
+    }[];
+    IsTruncated: boolean;
+    NextMarker ?: string;
+};
+
+/**
+ * Handle object listing with parameters. This extends the base class Delimiter
+ * to return the master/current versions.
+ */
+class DelimiterCurrent extends Delimiter {
+    /**
+     * Delimiter listing of current versions.
+     * @param {Object}  parameters            - listing parameters
+     * @param {String}  parameters.beforeDate - limit the response to keys older than beforeDate
+     * @param {RequestLogger} logger          - The logger of the request
+     * @param {String} [vFormat]              - versioning key format
+     */
+    constructor(parameters, logger, vFormat) {
+        super(parameters, logger, vFormat);
+
+        this.beforeDate = parameters.beforeDate;
+    }
+
+    genMDParamsV1() {
+        const params = super.genMDParamsV1();
+ 
+        if (this.beforeDate) {
+            params.lastModified = {
+                lt: this.beforeDate,
+            };
+        }
+        return params;
+    }
+
+    result(): ResultObject {
+        const result: ResultObject = {
+            Contents: this.Contents,
+            IsTruncated: this.IsTruncated,
+        };
+
+        if (this.IsTruncated) {
+            result.NextMarker = this.NextMarker;
+        }
+
+        return result;
+    }
+}
+module.exports = { DelimiterCurrent };
--- a/lib/storage/metadata/MetadataWrapper.js
+++ b/lib/storage/metadata/MetadataWrapper.js
@@ -51,6 +51,35 @@ function _parseListEntries(entries) {
     });
 }
 
+/** _parseLifecycleListEntries - parse the values returned in a lifeycle listing by metadata
+ * @param {object[]} entries - Version or Content entries in a metadata listing
+ * @param {string} entries[].key - metadata key
+ * @param {string} entries[].value - stringified object metadata
+ * @return {object} - mapped array with parsed value or JSON parsing err
+ */
+function _parseLifecycleListEntries(entries) {
+    return entries.map(entry => {
+        const tmp = JSON.parse(entry.value);
+        return {
+            key: entry.key,
+            value: {
+                Size: tmp['content-length'],
+                ETag: tmp['content-md5'],
+                VersionId: tmp.versionId,
+                LastModified: tmp['last-modified'],
+                Owner: {
+                    DisplayName: tmp['owner-display-name'],
+                    ID: tmp['owner-id'],
+                },
+                StorageClass: tmp['x-amz-storage-class'],
+                tags: tmp.tags,
+                staleDate: tmp.staleDate,
+                dataStoreName: tmp.dataStoreName,
+            },
+        };
+    });
+}
+
 /** parseListEntries - parse the values returned in a listing by metadata
  * @param {object[]} entries - Version or Content entries in a metadata listing
  * @param {string} entries[].key - metadata key
@@ -272,6 +301,29 @@ class MetadataWrapper {
                     error: data.Contents,
                     listingType: listingParams.listingType,
                     method: 'listObject',
+                });
+                return cb(errors.InternalError);
+            }
+            return cb(null, data);
+        });
+    }
+
+    listLifecycleObject(bucketName, listingParams, log, cb) {
+        log.debug('getting object listing for lifecycle from metadata');
+        this.client.listLifecycleObject(bucketName, listingParams, log, (err, data) => {
+            if (err) {
+                log.error('error from metadata', { implName: this.implName,
+                    err });
+                return cb(err);
+            }
+            log.debug('object listing for lifecycle retrieved from metadata');
+            // eslint-disable-next-line no-param-reassign
+            data.Contents = parseListEntries(data.Contents, _parseLifecycleListEntries);
+            if (data.Contents instanceof Error) {
+                log.error('error parsing metadata listing for lifecycle', {
+                    error: data.Contents,
+                    listingType: listingParams.listingType,
+                    method: 'listLifecycleObject',
                 });
                 return cb(errors.InternalError);
             }
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -1651,6 +1651,42 @@ class MongoClientInterface {
     }
 
     /**
+     * lists current version, non-current version and orphan delete markers in a bucket
+     * @param {String} bucketName bucket name
+     * @param {Object} params params
+     * @param {String} params.listingType type of algorithm to use
+     * @param {Number} [params.maxKeys] maximum number of keys to list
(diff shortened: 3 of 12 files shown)


nicolas2bert:
ping

lib/algos/list/delimiterCurrent.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    result(): ResultObject {
+        const result: ResultObject = {
+            Contents: this.Contents,
+            IsTruncated: this.IsTruncated,
+        };
+
+        if (this.IsTruncated) {
+            result.NextKeyMarker = this.NextMarker;
```
jonathan-gramain: We should rather call this `result.NextMarker`, as `NextKeyMarker` normally means there is also a `NextVersionIdMarker` (by AWS API), to avoid any confusion.

lib/algos/list/types.ts
```diff
@@ -0,0 +1,8 @@
+export type ResultObject = {
```
jonathan-gramain: The name sounds a bit too generic since this file is agnostic to any particular listing type. We could call it `DelimiterCurrentResultObject` for example.

Also is this type going to be used anywhere else than in `delimiterCurrent.ts`? If not, it could be best to declare it inside the `delimiterCurrent.ts` file (in which case `ResultObject` is a fine name as it is local to a particular listing algo)

lib/storage/metadata/MetadataWrapper.js
```diff
@@ -279,6 +308,29 @@ class MetadataWrapper {
         });
     }
 
+    listLifecycleObject(bucketName, listingParams, log, cb) {
+        this.client.listLifecycleObject(bucketName, listingParams, log, (err, data) => {
+            log.debug('getting object listing for lifecycle from metadata');
```
jonathan-gramain: Isn't this line supposed to go just before the call to `this.client.listLifecycleObject`?

lib/storage/metadata/MetadataWrapper.js
```diff
@@ -279,6 +308,29 @@ class MetadataWrapper {
...
+            if (err) {
+                log.error('error from metadata', { implName: this.implName,
+                    err });
+                return cb(err);
+            }
+            log.debug('object listing for lifecycle retrieved from metadata');
+            // eslint-disable-next-line
```
jonathan-gramain: If you need to disable a linter rule for good reason, can you add what rule is explicitly disabled? E.g. `// eslint-disable-next-line no-param-reassign` or whatever rule is broken

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1650,6 +1650,42 @@ class MongoClientInterface {
...
+    listLifecycleObject(bucketName, params, log, cb) {
+        return this.getBucketVFormat(bucketName, log, (err, vFormat) => {
+            if (err) {
+                return cb(err);
+            }
+
+            if (vFormat !== BUCKET_VERSIONS.v1) {
+                log.error('not supported bucket format version',
+                    { method: 'listLifecycleObject', bucket: bucketName, vFormat });
+                return cb(errors.InternalError.customizeDescription('Not supported bucket format version'));
```
```suggestion
                return cb(errors.NotImplemented.customizeDescription('Not supported bucket format version'));
```

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1650,6 +1650,42 @@ class MongoClientInterface {
         });
     }
 
+    /**
+     * lists current version, non-current version and orphan delete markers in a bucket
+     * @param {String} bucketName bucket name
+     * @param {Object} params params
+     * @param {String} params.listingType type of algorithm to use
+     * @param {Number} params.maxKeys maximum number of keys to list (optional)
+     * @param {String} params.prefix prefix of objects to use (optional)
```
```suggestion
     * @param {Number} [params.maxKeys] maximum number of keys to list
     * @param {String} [params.prefix] prefix of objects to use
```

tests/functional/metadata/mongodb/listLifecycleObject/current.spec.js
```diff
@@ -0,0 +1,414 @@
...
+        async.series([
+            next => metadata.createBucket(BUCKET_NAME, bucketMD, logger, err => {
+                if (err) {
+                    return next(err);
+                }
+
+                collection = metadata.client.getCollection(BUCKET_NAME);
+                return next();
+            }),
+            next => {
```
jonathan-gramain, line 75: To improve readability a bit, maybe a single call to `async.eachSeries` could wrap each of the 3 blocks below with a single of them parameterized, just providing the changing params in the array passed to `async.eachSeries` (I can see the object name and timestamp change but that seems to be it).

tests/functional/metadata/mongodb/listLifecycleObject/current.spec.js
```diff
@@ -0,0 +1,414 @@
...
+            assert.strictEqual(data.Contents[0].key, 'pfx1-test-object');
+            assert.strictEqual(data.NextKeyMarker, 'pfx1-test-object');
+
+            params.marker = 'pfx1-test-object';
+
+            return metadata.listLifecycleObject(BUCKET_NAME, params, logger, (err, data) => {
+                assert.deepStrictEqual(err, null);
+                assert.strictEqual(data.IsTruncated, false);
+                assert.strictEqual(data.Contents[0].key, 'pfx3-test-object');
+                assert.strictEqual(data.Contents.length, 1);
```
jonathan-gramain:
```suggestion
                assert.strictEqual(data.Contents.length, 1);
                assert.strictEqual(data.Contents[0].key, 'pfx3-test-object');
```

tests/functional/metadata/mongodb/listLifecycleObject/current.spec.js
```diff
@@ -0,0 +1,414 @@
...
+        });
+    });
+
+    it('Should return empty list when beforeDate is before the objects creation date', done => {
+        const params = {
+            listingType: 'DelimiterCurrent',
+            beforeDate: '1970-01-01T00:00:00.000Z',
+        };
+        return metadata.listLifecycleObject(BUCKET_NAME, params, logger, (err, data) => {
+            assert.deepStrictEqual(err, null);
```
jonathan-gramain:
```suggestion
            assert.ifError(err);
```

tests/functional/metadata/mongodb/listLifecycleObject/current.spec.js
```diff
@@ -0,0 +1,414 @@
...
+        });
+    });
+
+    it('Should return an Internal error when bucket non existent (current)', done => {
+        const bucketName = 'non-existent-bucket';
+        const params = {
+            listingType: 'DelimiterCurrent',
+        };
+        return metadata.listLifecycleObject(bucketName, params, logger, (err, data) => {
+            expect(err.InternalError).toBe(true);
```
jonathan-gramain: We should probably stick to using `assert` everywhere, unless we use `expect` consistenly in the whole file.

tests/unit/algos/list/delimiterCurrent.spec.js
```diff
@@ -0,0 +1,128 @@
...
+                    value: value1,
+                },
+            ],
+            IsTruncated: false,
+        };
+
+        assert.deepStrictEqual(delimiter.result(), expectedResult);
+    });
+
+    it('should accept the first master and return the trucated content', () => {
```
jonathan-gramain:
```suggestion
    it('should accept the first master and return the truncated content', () => {
```

tests/functional/metadata/mongodb/listLifecycleObject/global.spec.js
```diff
@@ -0,0 +1,78 @@
...
+            listingType: 'DelimiterCurrent',
+        };
+        return metadata.listLifecycleObject(BUCKET_NAME, params, logger, (err, data) => {
+            expect(err.InternalError).toBe(true);
+            assert(!data);
+
+            return done();
+        });
+    });
+});
```
KazToozs, line 78: Do we not need more positive test cases?

tests/functional/metadata/mongodb/listLifecycleObject/current.spec.js
nicolas2bert: I flattened it for readability purposes, in order to mention (as comments) what the "keys" the method is creating.

tests/functional/metadata/mongodb/listLifecycleObject/global.spec.js
nicolas2bert: `global.spec.js` is used to test all three lifecycle calls with the same config (v0 key format).

tests/functional/metadata/mongodb/listLifecycleObject/current.spec.js
```diff
@@ -306,18 +313,6 @@ describe('MongoClientInterface::metadata.listLifecycleObject::current', () => {
         });
     });
 
-    it('Should return an Internal error when bucket non existent (current)', done => {
```
nicolas2bert: I removed this test because testing for non-existent bucket is not relevant at the Arsenal stage since the check/test is already done in CloudServer.

tests/functional/metadata/mongodb/listLifecycleObject/current.spec.js
jonathan-gramain: Do you want to address this? I personally don't think it's a big deal but you may have forgotten ;)

jonathan-gramain (approved):

tests/functional/metadata/mongodb/listLifecycleObject/current.spec.js
```diff
@@ -0,0 +1,409 @@
+const async = require('async');
+const assert = require('assert');
+const werelogs = require('werelogs');
+const { MongoMemoryReplSet } = require('mongodb-memory-server');
+const logger = new werelogs.Logger('MongoClientInterface', 'debug', 'debug');
+const MetadataWrapper =
+require('../../../../../lib/storage/metadata/MetadataWrapper');
```
rachedbenmustapha, line 7: Not sure why the linter isn't picking up, this line should be indented for readablity.

rachedbenmustapha (approved):

nicolas2bert:
@bert-e approve

PR #2089, opened 2023-03-08
ARSN-311 Add logic to list non-current versions for Lifecycle

nicolas2bert:
DelimiterNonCurrent is used for listing non-current versions.
The MD response is then processed to only return the non-current versions that became non-current before a defined date: `beforeDate`.

### DIFF
--- a/lib/algos/list/delimiterNonCurrent.js
+++ b/lib/algos/list/delimiterNonCurrent.js
@@ -1,0 +1,213 @@
+'use strict'; // eslint-disable-line strict
+const Delimiter = require('./delimiter').Delimiter;
+const VSConst = require('../../versioning/constants').VersioningConstants;
+const { inc, FILTER_ACCEPT, FILTER_END, SKIP_NONE } = require('./tools');
+const VID_SEP = VSConst.VersionId.Separator;
+const { DbPrefixes } = VSConst;
+
+// TODO: find an acceptable timeout value.
+const DELIMITER_TIMEOUT_MS = 10 * 1000; // 10s
+const TRIM_METADATA_MIN_BLOB_SIZE = 10000;
+
+/**
+ * Handle object listing with parameters. This extends the base class Delimiter
+ * to return the raw non-current versions objects.
+ */
+class DelimiterNonCurrent extends Delimiter {
+    /**
+     * Delimiter listing of non-current versions.
+     * @param {Object}  parameters                  - listing parameters
+     * @param {String}  parameters.versionIdMarker  - version id marker
+     * @param {String}  parameters.beforeDate       - limit the response to keys with stale date older than beforeDate
+     * “stale date” is the date on when a version becomes non-current.
+     * @param {String}  parameters.keyMarker        - key marker
+     * @param {RequestLogger} logger                - The logger of the request
+     * @param {String} [vFormat]                    - versioning key format
+     */
+    constructor(parameters, logger, vFormat) {
+        super(parameters, logger, vFormat);
+
+        this.versionIdMarker = parameters.versionIdMarker;
+        this.beforeDate = parameters.beforeDate;
+        this.keyMarker = parameters.keyMarker;
+        this.NextKeyMarker = null;
+
+        this.skipping = this.skippingV1;
+        this.genMDParams = this.genMDParamsV1;
+
+        this.keyName = null;
+        this.staleDate = null;
+
+        // used for monitoring
+        this.evaluatedKeys = 0;
+    }
+
+    skippingV1() {
+        return SKIP_NONE;
+    }
+
+    genMDParamsV1() {
+        const params = {
+            gte: DbPrefixes.Version,
+            lt: inc(DbPrefixes.Version),
+        };
+
+        if (this.prefix) {
+            params.gte = `${DbPrefixes.Version}${this.prefix}`;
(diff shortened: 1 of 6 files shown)


lib/algos/list/delimiterNonCurrent.js
```diff
@@ -0,0 +1,204 @@
...
+
+        return FILTER_ACCEPT;
+    }
+
+    trimMetadataAddStaleDate(value, staleDate) {
+        let ret = undefined;
+        try {
+            ret = JSON.parse(value);
+            ret.staleDate = staleDate;
+            if (value.length >= TRIM_METADATA_MIN_BLOB_SIZE) {
```
rachedbenmustapha, line 180: There are `TRIM_METADATA_MIN_BLOB_SIZE` and a `trimMetadata` function in `Extension` which delimiters inherit from IIRC, that can be reused before adding the stale date.

I am wondering why the trim is needed here but not in the "current" delimiter?

In the error case, the stale date is not added, does it mean that the client will have to do its own filtering with extra `HEAD` requests?

tests/functional/metadata/mongodb/listLifecycleObject/noncurrent.spec.js
```diff
@@ -0,0 +1,766 @@
...
+    it('Should list non-current versions', done => {
+        const params = {
+            listingType: 'DelimiterNonCurrent',
+        };
+        return metadata.listLifecycleObject(BUCKET_NAME, params, logger, (err, data) => {
+            assert.deepStrictEqual(err, null);
+            assert.strictEqual(data.IsTruncated, false);
+            assert.strictEqual(data.Contents.length, 12);
+            const expected = [
+                {
```
rachedbenmustapha, line 165: Assuming that the version id is returned by the listing, maybe it can be checked as well?

tests/functional/metadata/mongodb/listLifecycleObject/noncurrent.spec.js
```diff
@@ -0,0 +1,766 @@
...
+                        return next(err);
+                    }
+                    lastVersionId = JSON.parse(res).versionId;
+                    return next(null);
+                }),
+            next => metadata.deleteObjectMD(BUCKET_NAME, 'pfx4-test-object', { versionId: lastVersionId },
+                logger, next),
+            next => metadata.listLifecycleObject(BUCKET_NAME, params, logger, (err, data) => {
+                assert.ifError(err);
+                assert.strictEqual(data.Contents[0].value.VersionId, earlyVersionId);
```
rachedbenmustapha, line 792: An assert on the number of entries like in the other tests would make it clearer (I'm thinking there should be only 1 but not sure).

tests/functional/metadata/mongodb/listLifecycleObject/noncurrent.spec.js
```diff
@@ -0,0 +1,766 @@
...
+                    staleDate: '1970-01-01T00:00:00.003Z',
+                },
+            ];
+            assertContents(data.Contents, expected);
+
+            return done();
+        });
+    });
+
+    it('Should not take phd master into account key when listing non-current versions', done => {
```
rachedbenmustapha:
```suggestion
    it('Should not take phd master key into account when listing non-current versions', done => {
```

tests/functional/metadata/mongodb/listLifecycleObject/noncurrent.spec.js
```diff
@@ -0,0 +1,766 @@
...
+                logger, next),
+            next => metadata.listLifecycleObject(BUCKET_NAME, params, logger, (err, data) => {
+                assert.ifError(err);
+                assert.strictEqual(data.Contents[0].value.VersionId, earlyVersionId);
+                return next();
+            }),
+        ], done);
+    });
+
+    it('Should not take phd master into account key when listing non-current versions', done => {
```
rachedbenmustapha: Description is the same as the previous one.

lib/algos/list/delimiterNonCurrent.js
```diff
@@ -0,0 +1,204 @@
...
+        // is available (JSON.parse has not failed) and, if the "beforeDate" argument is specified,
+        // the "stale date" is older than the "beforeDate".
+        if (this.staleDate && (!this.beforeDate || this.staleDate < this.beforeDate)) {
+            this.Contents.push({ key, value: this.trimMetadataAddStaleDate(value, this.staleDate) });
+            ++this.keys;
+        }
+
+        // The in-memory "stale date" is then updated with the version's last-modified date to be used for
+        // the following version.
+        this.staleDate = this.getLastModified(value);
```
rachedbenmustapha: This seems to be double-parsing the MD after `trimMetadataAddStaleDate ` (which could be big).

lib/algos/list/delimiterNonCurrent.js
nicolas2bert: Today client does not do anything if the `staleDate` is missing.

I am adding a check so that the listing only returns keys with `staleDate`.

The logic I had in mind is that we make our best effort to retrieve the `staleDate`, if parsing fails, the key (with missing staleDate) is not returned and the next listing will likely return it.

rachedbenmustapha (approved):

alexanderchan-scality (approved):

nicolas2bert:
@bert-e approve

PR #2090, opened 2023-03-08
ARSN-312 Add logic to list orphan delete markers for Lifecycle

nicolas2bert:
DelimiterOrphan is used for listing orphan delete markers.
The MD response is then processed to only return the delete markers with zero noncurrent versions before a defined date: `beforeDate`.

### DIFF
--- a/lib/algos/list/delimiterOrphanDeleteMarker.js
+++ b/lib/algos/list/delimiterOrphanDeleteMarker.js
@@ -1,0 +1,202 @@
+'use strict'; // eslint-disable-line strict
+const Delimiter = require('./delimiter').Delimiter;
+const VSConst = require('../../versioning/constants').VersioningConstants;
+const { inc, FILTER_ACCEPT, FILTER_END, SKIP_NONE } = require('./tools');
+const VID_SEP = VSConst.VersionId.Separator;
+const { DbPrefixes } = VSConst;
+
+const DELIMITER_TIMEOUT_MS = 10 * 1000; // 10s
+const TRIM_METADATA_MIN_BLOB_SIZE = 10000;
+
+/**
+ * Handle object listing with parameters. This extends the base class Delimiter
+ * to return the orphan delete markers. Orphan delete markers are also
+ * refered as expired object delete marker.
+ * They are delete marker with zero noncurrent versions.
+ */
+class DelimiterOrphanDeleteMarker extends Delimiter {
+    /**
+     * Delimiter listing of non-current versions.
+     * @param {Object}  parameters            - listing parameters
+     * @param {String}  parameters.beforeDate - limit the response to keys older than beforeDate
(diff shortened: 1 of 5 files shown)


lib/algos/list/delimiterOrphan.js
```diff
@@ -0,0 +1,175 @@
...
+            } catch (e) {
+                this.logger.warn('could not parse Object Metadata while listing',
+                    {
+                        method: 'getLastModified',
+                        err: e.toString(),
+                    });
+            }
+            // We then check if the orphan version is a delete marker and if it is older than the "beforeDate"
+            if ((!this.beforeDate || (lastModified && lastModified < this.beforeDate)) && isDeleteMarker) {
+                this.Contents.push({ key: this.keyName, value: this.trimMetadata(this.value) });
```
rachedbenmustapha: This might be double-parsing the metadata as well.

lib/algos/list/delimiterOrphan.js
```diff
@@ -0,0 +1,175 @@
...
+
+        const versionIdIndex = keyVersionSuffix.indexOf(VID_SEP);
+        // key without version suffix
+        const key = keyVersionSuffix.slice(0, versionIdIndex);
+
+        // For a given key, the youngest version is kept in memory since it represents the current version.
+        if (key !== this.keyName) {
+            // If the following version reference a new key,
+            // it means that the previous one (this.value) was an orphan version.
+            this._addOrphan();
```
rachedbenmustapha: I had a hard time understanding why this would not always "swallow" the first old enough delete marker that passes through, as the `if (this.value)` test in `_addOrphan` looks more like a regular null-check than an orphan-state test.

For clarity I'd remove the `if (this.value)` test from `_addOrphan` (to clarify its purpose), and instead add it here, with a comment saying that if `this.value != null` it means that the key's not the first to be evaluated and this is "allowed" to be an orphan.

rachedbenmustapha (approved):

lib/algos/list/delimiterOrphan.js
```diff
@@ -0,0 +1,202 @@
...
+const DELIMITER_TIMEOUT_MS = 10 * 1000; // 10s
+const TRIM_METADATA_MIN_BLOB_SIZE = 10000;
+
+/**
+ * Handle object listing with parameters. This extends the base class Delimiter
+ * to return the orphan delete markers. Orphan delete markers are also
+ * refered as expired object delete marker.
+ * They are delete marker with zero noncurrent versions.
+ */
+class DelimiterOrphan extends Delimiter {
```
alexanderchan-scality: minor: name should be more descriptive of what is being listed
`DelimiterDeleteMarkerOrphan`

alexanderchan-scality (approved):

nicolas2bert:
@bert-e approve

PR #2097, opened 2023-03-15
ft: ARSN-317 null key support in BucketFile backend

jonathan-gramain:
Support null keys in BucketFile backend - null keys are the new way to store null versions, where a single database key with a specific empty version ID is used instead of referencing the null version via "nullVersionId" in object metadata.

Add relevant unit tests to check the new behavior (those were copied and mechanically adapted from the Metadata repository).

**note for reviews**: the first commit enhances logging in general, while the second commit adds the null key support proper (and third is a version bump).

### DIFF
--- a/lib/versioning/VersioningRequestProcessor.ts
+++ b/lib/versioning/VersioningRequestProcessor.ts
@@ -91,7 +91,8 @@ export default class VersioningRequestProcessor {
         const { db, key, options } = request;
         logger.addDefaultFields({ bucket: db, key, options });
         if (options && options.versionId) {
-            const versionKey = formatVersionKey(key, options.versionId);
+            const keyVersionId = options.versionId === 'null' ? '' : options.versionId;
+            const versionKey = formatVersionKey(key, keyVersionId);
             return this.wgm.get({ db, key: versionKey }, logger, callback);
         }
         return this.wgm.get(request, logger, (err, data) => {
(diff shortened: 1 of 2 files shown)


tmacro (approved):

lib/versioning/VersioningRequestProcessor.ts
```diff
@@ -310,14 +310,14 @@ export default class VersioningRequestProcessor {
         nextValue?: string;
     }) {
         const { db, key } = request;
-        logger.info('start repair process', { request });
+        logger.info('start repair process');
         this.writeCache.get({ db, key }, logger, (err, value) => {
             // error or the new version is not a place holder for deletion
             if (err) {
-                return logger.info('error repairing', { request, error: err });
+                return logger.info('error repairing', { error: err });
```
dvasilas: Is it better to use `logger.error` here ?

dvasilas (approved):

lib/versioning/VersioningRequestProcessor.ts
jonathan-gramain: It seems so, I will turn to error, though I'll add a specific case for `ObjNotFound` and log in debug since it may happen if the PHD was deleted (by the user deleting the object) before it got repaired.

jonathan-gramain:
ping

PR #2098, opened 2023-03-15
Bugfix/arsn 316 concurrent bucket md updates

ghost:
Fixes concurrent updates of bucket metadata leading to overrides in the context of Veeam XML uploads, that takes place at the exact same time (so, with concurrency), where one would override the other.

Cloudserver PR: https://github.com/scality/cloudserver/pull/5056

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -412,6 +412,72 @@ class MongoClientInterface {
         });
     }
 
+    /**
+     *
+     * @param {String} bucketName - name of bucket
+     * @param {String} capabilityName - name of capability
+     * @param {String} [capabilityField] - name of capability field
+     * @param {Object} capability - capability object
+     * @param {Sbject} log - logger
+     * @param {Function} cb - callback
+     * @return {undefined}
+     */
+    putBucketAttributesCapabilities(bucketName, capabilityName, capabilityField, capability, log, cb) {
+        const m = this.getCollection(METASTORE);
+        const updateString = capabilityField ?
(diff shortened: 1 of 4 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -412,6 +412,72 @@ class MongoClientInterface {
...
+     *
+     * @param {String} bucketName - name of bucket
+     * @param {String} capabilityName - name of capability
+     * @param {String} [capabilityField] - name of capability field
+     * @param {Object} capability - capability object
+     * @param {Sbject} log - logger
+     * @param {Function} cb - callback
+     * @return {undefined}
+     */
+    putBucketAttributesCapabilities(bucketName, capabilityName, capabilityField, capability, log, cb) {
```
KazToozs, line 425: For my understanding, why are capabilityName and capabilityField separate fields from the capability object? Could they not be recovered from that object to avoid these extra parameters?

KazToozs (approved):
lgtm otherwise

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: It comes from the fact that wa have both `VeeamSOSAPI` and two sub fields, one for each XML file. If we update the whole `VeeamSOSAPI`, it won't fix the concurrency issue: we want to be able to set a unique "file" for any identity, so we need two layers. We could do it a bit more dynamically, but to me it'll reduce the readability...

ghost: I added some tests to cover the mongodb client changes.

welansari (approved):

PR #2100, opened 2023-03-17
impr: ARSN-320 add "isNull2" attribute to ObjectMD

jonathan-gramain:
This new attribute will be set whenever a Cloudserver supporting null keys sets the "isNull" attribute to a master key, along with it.

The purpose of this attribute is to allow Cloudserver to optimize by not having to check and delete a null versioned key when the null master has "isNull2" set, as it is guaranteed not to exist.

We need to introduce a new attribute to keep backward compatibility, the naming is a bit unfortunate but it has the benefit of being short and not too specific to a particular optimization, just stating that it is a "new" null master.


naren-rajendran (approved):

tmacro (approved):

jonathan-gramain:
ping

PR #2106, opened 2023-03-28
ARSN-309: support indexing for mongo

alexanderchan-scality:


lib/storage/metadata/MetadataWrapper.js
```diff
@@ -515,6 +517,75 @@ class MetadataWrapper {
...
+     * @param {Array<Object>} indexSpecs index specification
+     * @param {Object} log logger
+     * @param {Function} cb callback
+     * @return {undefined}
+     */
+    putBucketIndexes(bucketName, indexSpecs, log, cb) {
+        log.debug('put bucket indexes');
+
+        if (!supportsIndexingMDBackends.has(this.implName)) {
+            log.debug('error from metadata', { implName: this.implName });
```
jonathan-gramain: Wouldn't this better be an error log? (unless it's meant to be called with incompatible backends in normal circumstances). Also logging the error type can be useful:

```suggestion
            log.error('error from metadata', { error: 'NotImplemented', implName: this.implName });
```

lib/storage/metadata/MetadataWrapper.js
```diff
@@ -515,6 +517,75 @@ class MetadataWrapper {
...
+     * @param {String} bucketName bucket name
+     * @param {Array<Object>} indexSpecs index specification
+     * @param {Object} log logger
+     * @param {Function} cb callback
+     * @return {undefined}
+     */
+    putBucketIndexes(bucketName, indexSpecs, log, cb) {
+        log.debug('put bucket indexes');
+
+        if (!supportsIndexingMDBackends.has(this.implName)) {
```
jonathan-gramain: Maybe a more straightforward way to check if the backend supports it is checking if the client object has the `putBucketIndexes` function

lib/storage/metadata/mongoclient/utils.js
```diff
@@ -185,6 +185,15 @@ function formatVersionKey(key, versionId, vFormat) {
     return formatVersionKeyV0(key, versionId);
 }
 
+function indexFormatMongoArrayToObject(mongoIndexArray) {
+    const indexObj = {};
+
+    mongoIndexArray.forEach(idx => {
```
jonathan-gramain:
```suggestion
    for (const idx of mongoIndexArray) {
```

lib/constants.ts
```diff
@@ -171,3 +171,17 @@ export const maxCachedBuckets = process.env.METADATA_MAX_CACHED_BUCKETS ?
...
+    lifecycleDataStoreNamePrefixed: new Map()
+        .set('value.dataStoreName', 1 )
+        .set('value.last-modified', 1 )
+        .set('_id', 1 ),
+    lifecycleLastModifiedPrefixed: new Map()
+        .set('value.last-modified', 1 )
+        .set('_id', 1 ),
+};
+
+export const supportsIndexingMDBackends = new Set([
```
jonathan-gramain: do we still need this?

jonathan-gramain (approved):

lib/constants.ts
```diff
@@ -171,3 +171,17 @@ export const maxCachedBuckets = process.env.METADATA_MAX_CACHED_BUCKETS ?
     Number(process.env.METADATA_MAX_CACHED_BUCKETS) : 1000;
 
 export const validRestoreObjectTiers = new Set(['Expedited', 'Standard', 'Bulk']);
+
+export const bucketIndexes = {
```
francoisferrand: do we need to keep these constants here?
they seems more closely related to backbeat, can they not stay there, along with the rest of lifecycle code?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -2471,6 +2471,70 @@ class MongoClientInterface {
...
+
+    /**
+     * Puts bucket indexes
+     * @param {String} bucketName bucket name
+     * @param {Array<Object>} indexSpecs index specification
+     * @param {Object} log logger
+     * @param {Function} cb callback
+     * @return {undefined}
+     */
+    putBucketIndexes(bucketName, indexSpecs, log, cb) {
```
nicolas2bert, line 2483: Can we test `putBucketIndexes`, `deleteBucketIndexes`, `getBucketIndexes` ?

lib/constants.ts
nicolas2bert: Now that we decided to not create indexes on bucket creation (lifecycle creation instead), I think the index references could be moved to Backbeat, since BB will handle the index creation and the "Verification of existence".

lib/storage/metadata/mongoclient/MongoClientInterface.js
alexanderchan-scality: @nicolas2bert imo, there isn't much to test here as the methods are  just calling the mongo client method and doesn't look to have any room for possible regressions

nicolas2bert (approved):

PR #2111, opened 2023-04-10
ARSN-328 Exclude keys based on their dataStoreName

nicolas2bert:
To list the current versions using the **listLifecycleCurrents** function, the filtering is performed at the MongoDB query level.

In contrast, when listing the non-current versions with the **listLifecycleNoncurrents** function, filtering is done at the application level. Here, each key is evaluated and assessed for eligibility based on its datastore name. 

Finally, when calling the **listLifecycleOrphans** function to list the orphan delete markers, no filtering is done based on the datastore name. This is because orphan delete markers do not go through a transition workflow.

### DIFF
--- a/lib/algos/list/delimiterCurrent.ts
+++ b/lib/algos/list/delimiterCurrent.ts
@@ -18,6 +18,7 @@ class DelimiterCurrent extends Delimiter {
      * Delimiter listing of current versions.
      * @param {Object}  parameters            - listing parameters
      * @param {String}  parameters.beforeDate - limit the response to keys older than beforeDate
+     * @param {String}  parameters.excludedDataStoreName - excluded datatore name
      * @param {RequestLogger} logger          - The logger of the request
      * @param {String} [vFormat]              - versioning key format
      */
@@ -25,15 +26,22 @@ class DelimiterCurrent extends Delimiter {
         super(parameters, logger, vFormat);
 
         this.beforeDate = parameters.beforeDate;
+        this.excludedDataStoreName = parameters.excludedDataStoreName;
     }
 
     genMDParamsV1() {
         const params = super.genMDParamsV1();
- 
+
         if (this.beforeDate) {
             params.lastModified = {
                 lt: this.beforeDate,
             };
+        }
+
+        if (this.excludedDataStoreName) {
+            params.dataStoreName = {
+                ne: this.excludedDataStoreName,
+            }
         }
         return params;
     }
--- a/lib/algos/list/delimiterNonCurrent.js
+++ b/lib/algos/list/delimiterNonCurrent.js
@@ -30,6 +30,7 @@ class DelimiterNonCurrent extends Delimiter {
         this.versionIdMarker = parameters.versionIdMarker;
         this.beforeDate = parameters.beforeDate;
         this.keyMarker = parameters.keyMarker;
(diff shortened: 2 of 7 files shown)


nicolas2bert:
ping

lib/algos/list/delimiterCurrent.ts
```diff
@@ -25,6 +25,7 @@ class DelimiterCurrent extends Delimiter {
         super(parameters, logger, vFormat);
 
         this.beforeDate = parameters.beforeDate;
+        this.excludedDataStoreName = parameters.excludedDataStoreName;
```
jonathan-gramain, line 29: Should also update JSDoc with this param

lib/algos/list/delimiterNonCurrent.js
```diff
@@ -170,30 +179,35 @@ class DelimiterNonCurrent extends Delimiter {
...
+        } catch (e) {
+            this.logger.warn('Could not parse Object Metadata while listing', {
+                method: 'DelimiterNonCurrent._parse',
+                err: e.toString(),
+            });
+        }
+        return p;
+    }
+
+    _trimAndStringify(value, staleDate) {
```
jonathan-gramain:
```suggestion
    _trimAndStringify(parsedMD, staleDate) {
```

lib/algos/list/delimiterNonCurrent.js
```diff
@@ -170,30 +179,35 @@ class DelimiterNonCurrent extends Delimiter {
...
+        const p = value;
+        let s = undefined;
+        p.staleDate = staleDate;
+        try {
+            if (p.length >= TRIM_METADATA_MIN_BLOB_SIZE) {
```
jonathan-gramain: This check wouldn't work anymore because p is now parsed metadata

lib/algos/list/delimiterNonCurrent.js
```diff
@@ -148,18 +150,25 @@ class DelimiterNonCurrent extends Delimiter {
             return FILTER_ACCEPT;
         }
 
-        // The following version is pushed only if the "stale date" (picked up from the previous version)
-        // is available (JSON.parse has not failed) and, if the "beforeDate" argument is specified,
-        // the "stale date" is older than the "beforeDate".
+        // The following version is pushed only:
```
jonathan-gramain, line 153: If I understand the code flow, it seems that we always have to parse the metadata in any case. I'd suggest to parse it at the beginning then, and use the fields directly thereafter, as it is now the tentative optimizations only seem to clutter the code.

jonathan-gramain (changes requested):

rachedbenmustapha (approved):

lib/algos/list/delimiterOrphanDeleteMarker.js
```diff
@@ -92,26 +92,27 @@ class DelimiterOrphanDeleteMarker extends Delimiter {
         let p;
         try {
             p = JSON.parse(s);
+            if (this.skipping.length >= TRIM_METADATA_MIN_BLOB_SIZE) {
```
jonathan-gramain:
```suggestion
            if (s.length >= TRIM_METADATA_MIN_BLOB_SIZE) {
```

jonathan-gramain (changes requested):

jonathan-gramain (approved):

PR #2121, opened 2023-05-22
ARSN-336 Log KMS Result Message

scality-gelbart:
fixes S3C-5516

### DIFF
--- a/lib/network/kmip/index.ts
+++ b/lib/network/kmip/index.ts
@@ -312,6 +312,7 @@ export default class KMIP {
                     return cb(err);
                 }
                 const response = this._decodeMessage(logger, rawResponse);
+                logger.error(response);
                 const performedOperation =
                       response.lookup('Response Message/' +
                                       'Batch Item/Operation')[0];


jonathan-gramain (changes requested):
The intent to help troubleshoot issues is valid, but I think we need to change how and what we dump info, keeping in mind the following:

- what is the rate of messages, is there a risk of overflowing logs with too much lines on busy systems
- what is the content of the logged messages, do we risk leaking in logs sensitive info like customer keys etc.
- `logger.error`: this will appear as an error log, should it be a debug log instead?
- we should format the logs, for example: `log.info('something happened', { reason: "bleh", client: "some-client" })`, instead of `log.info(lots_of_things)` which might log unwanted things and/or confuse the reader of the logs
- target should be `development/7.x` (`7.10` or `7.70` for respectively RING 8 or RING 9), `development/8.x` will only appear in Artesca releases.

scality-gelbart:
> The intent to help troubleshoot issues is valid, but I think we need to change how and what we dump info, keeping in mind the following:
> 
>     * what is the rate of messages, is there a risk of overflowing logs with too much lines on busy systems
> 
>     * what is the content of the logged messages, do we risk leaking in logs sensitive info like customer keys etc.
> 
>     * `logger.error`: this will appear as an error log, should it be a debug log instead?
> 
>     * we should format the logs, for example: `log.info('something happened', { reason: "bleh", client: "some-client" })`, instead of `log.info(lots_of_things)` which might log unwanted things and/or confuse the reader of the logs
> 
>     * target should be `development/7.x` (`7.10` or `7.70` for respectively RING 8 or RING 9), `development/8.x` will only appear in Artesca releases.

thx for your reply @jonathan-gramain, I realize this ticket is way beyond my very basic dev skills so I'll stop trying... for now

PR #2122, opened 2023-05-22
Filter delete marker from version suspended buckets

KillianG:
Issue: ARSN-337

### DIFF
--- a/lib/algos/list/delimiterMaster.js
+++ b/lib/algos/list/delimiterMaster.js
@@ -157,16 +157,7 @@ class DelimiterMaster extends Delimiter {
      */
     filterV1(obj) {
         const value = obj.value;
-        if (Version.isDeleteMarker(value)) {
-            /* This entry is a deleteMarker which has not been filtered by the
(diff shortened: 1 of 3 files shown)


lib/algos/list/delimiterMaster.js
```diff
@@ -156,9 +156,10 @@ class DelimiterMaster extends Delimiter {
      *  @return {number}          - indicates if iteration should continue
      */
     filterV1(obj) {
-        // Filtering master keys in v1 is simply listing the master
-        // keys, as the state of version keys do not change the
-        // result, so we can use Delimiter method directly.
+        const value = obj.value;
```
francoisferrand, line 159: we should keep the comment & "improve it" to document what you change : because indeed (as was indicated), filtering in v1 is not really needed..........expect we need it to handle the case of objects deleted while a bucket was version-suspended

jonathan-gramain (approved):

francoisferrand (approved):

PR #2127, opened 2023-06-05
ARSN-339 Introduce the time-progression-factor flag

nicolas2bert:
The "time-progression-factor" variable serves as a testing-specific feature that accelerates the progression of time within a system. 
By reducing the significance of each day, it enables the swift execution of specific actions, such as expiration, transition, and object locking, which are typically associated with longer timeframes. 

This capability allows for efficient testing and evaluation of outcomes, optimizing the observation of processes that would normally take days or even years.
It's important to note that this variable is intended exclusively for testing purposes and is not employed in live production environments, where real-time progression is crucial for accurate results.

### DIFF
--- a/lib/s3middleware/lifecycleHelpers/LifecycleDateTime.ts
+++ b/lib/s3middleware/lifecycleHelpers/LifecycleDateTime.ts
@@ -1,19 +1,25 @@
-const oneDay = 24 * 60 * 60 * 1000; // Milliseconds in a day.
+import { scaleMsPerDay } from '../objectUtils';
+const msInOneDay = 24 * 60 * 60 * 1000; // Milliseconds in a day.
 
 export default class LifecycleDateTime {
     _transitionOneDayEarlier?: boolean;
     _expireOneDayEarlier?: boolean;
+    _timeProgressionFactor?: number;
+    _scaledMsPerDay: number;
 
     constructor(params?: {
         transitionOneDayEarlier: boolean;
         expireOneDayEarlier: boolean;
+        timeProgressionFactor: number;
     }) {
         this._transitionOneDayEarlier = params?.transitionOneDayEarlier;
         this._expireOneDayEarlier = params?.expireOneDayEarlier;
+        this._timeProgressionFactor = params?.timeProgressionFactor || 1;
+        this._scaledMsPerDay = scaleMsPerDay(this._timeProgressionFactor);
     }
 
     getCurrentDate() {
-        const timeTravel = this._expireOneDayEarlier ? oneDay : 0;
+        const timeTravel = this._expireOneDayEarlier ? msInOneDay : 0;
         return Date.now() + timeTravel;
     }
 
@@ -25,7 +31,7 @@ export default class LifecycleDateTime {
     findDaysSince(date: Date) {
         const now = this.getCurrentDate();
         const diff = now - date.getTime();
-        return Math.floor(diff / (1000 * 60 * 60 * 24));
+        return Math.floor(diff / this._scaledMsPerDay);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


lib/s3middleware/lifecycleHelpers/LifecycleDateTime.ts
```diff
@@ -1,19 +1,25 @@
...
+    _scaledMsPerDay: number;
 
     constructor(params?: {
         transitionOneDayEarlier: boolean;
         expireOneDayEarlier: boolean;
+        timeProgressionFactor: number;
     }) {
         this._transitionOneDayEarlier = params?.transitionOneDayEarlier;
         this._expireOneDayEarlier = params?.expireOneDayEarlier;
+        this._timeProgressionFactor = params?.timeProgressionFactor || 1;
```
francoisferrand, line 17: Seems we need to "store" this, may be better to keep only the resulting `scaledMsPerDay` to avoid future inconsistent changes...

lib/s3middleware/lifecycleHelpers/LifecycleDateTime.ts
```diff
@@ -52,8 +58,8 @@ export default class LifecycleDateTime {
         }
         if (transition.Days !== undefined) {
             const lastModifiedTime = this.getTimestamp(lastModified);
-            const timeTravel = this._transitionOneDayEarlier ? -oneDay : 0;
-            return lastModifiedTime + (transition.Days * oneDay) + timeTravel;
+            const timeTravel = this._transitionOneDayEarlier ? -msInOneDay : 0;
+            return lastModifiedTime + (transition.Days * this._scaledMsPerDay) + timeTravel;
```
francoisferrand, line 62: I am wondering about the semantics of the "one day earlier" when time factor scaling is used : should it be a "scaled" day, or a real time day?

I fear we may get unexpected behavior if both real time day and time factor are used, though I don't have a precise case.

To avoid this, maybe we should make these exclusive: eg ignore time factor if oneDayEarlier is used (or the opposite)?

lib/s3middleware/lifecycleHelpers/LifecycleDateTime.ts
```diff
@@ -1,19 +1,25 @@
-const oneDay = 24 * 60 * 60 * 1000; // Milliseconds in a day.
+import { scaleMsPerDay } from '../objectUtils';
+const msInOneDay = 24 * 60 * 60 * 1000; // Milliseconds in a day.
```
francoisferrand, line 2: Should use the one from obejctutils

francoisferrand (approved):

lib/s3middleware/lifecycleHelpers/LifecycleDateTime.ts
nicolas2bert: I intended to be even stricter than your proposal by accepting only one option, ensuring there would be no room for confusion.

francoisferrand: i was about to suggest that, but I am not sure this is practical : we have one branch of arsenal, used in many branches of each component... so we need to keep compatibility for a time, or update the deps/CR/... in every version of Zenko and S3C.

So I would be happy to compromise, just keep the 2 options for now in arsenal/cloudserver/backbeat, just not at the same time. in Zenko/ZKOP we may however indeed remove the "one day" and replace it with the new flag.

nicolas2bert: That is what I meant by "accepting only one option": either `xxxOneDayEarlier` or `timeProgressionFactor`, not both.

nicolas2bert: I'll prioritize pushing the CloudServer PR as soon as possible, which will likely assist you in understanding the logic.

alexanderchan-scality (approved):

nicolas2bert:
@bert-e approve

PR #2134, opened 2023-06-13
Improvement/arsn 345 optimize multiobjectdelete api and batching

ghost:
Pull request for 8.1 : https://github.com/scality/Arsenal/pull/2141

Around MultiDeleteObject optimization, this PR does two things:
- Directly deletes an object if a specific flag is passed (in a separate commit)
- Implements a new metadata method (only supported in MongoDB for now) to batch the reads in order to get multiple objects metadata at once.
- Finally, unit and functional tests are added, where the functional ones are only in the 8.x branch, as we do not have any of those in 7.x (unused).

### DIFF
--- a/lib/storage/metadata/MetadataWrapper.js
+++ b/lib/storage/metadata/MetadataWrapper.js
@@ -220,14 +220,14 @@ class MetadataWrapper {
             });
             return cb(errors.NotImplemented);
         }
-        log.debug('getting object from metadata');
+        log.debug('getting objects from metadata', { objects: objNamesWithParams });
         return this.client.getObjects(bucketName, objNamesWithParams, log, (err, data) => {
             if (err) {
-                log.debug('error from metadata', { implName: this.implName,
+                log.debug('error getting objects from metadata', { implName: this.implName, objects: objNamesWithParams,
                     err });
                 return cb(err);
             }
-            log.debug('object retrieved from metadata');
+            log.debug('objects retrieved from metadata', { objects: objNamesWithParams });
             return cb(err, data);
         });
     }
--- a/lib/storage/metadata/MetadataWrapper.js
+++ b/lib/storage/metadata/MetadataWrapper.js
@@ -213,6 +213,25 @@ class MetadataWrapper {
             });
     }
 
+    getObjectsMD(bucketName, objNamesWithParams, log, cb) {
+        if (typeof this.client.getObjects !== 'function') {
+            log.debug('backend does not support get object metadata with batching', {
+                implName: this.implName,
+            });
+            return cb(errors.NotImplemented);
+        }
+        log.debug('getting object from metadata');
+        return this.client.getObjects(bucketName, objNamesWithParams, log, (err, data) => {
+            if (err) {
+                log.debug('error from metadata', { implName: this.implName,
+                    err });
+                return cb(err);
+            }
+            log.debug('object retrieved from metadata');
+            return cb(err, data);
+        });
+    }
+
     getObjectMD(bucketName, objName, params, log, cb) {
         log.debug('getting object from metadata');
         this.client.getObject(bucketName, objName, params, log, (err, data) => {
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -979,32 +979,35 @@ class MongoClientInterface {
         if (!Array.isArray(objects)) {
             return callback(errors.InternalError.customizeDescription('objects must be an array'));
         }
+        // We do not accept more than 1000 keys in a single request
+        if (objects.length > 1000) {
+            return callback(errors.InternalError.customizeDescription('cannot get more than 1000 objects'));
+        }
         // Function to process each document
         const processDoc = (doc, objName, params, key, cb) => {
             const versionIdValue = params && params.versionId ? params.versionId : undefined;
             if (!doc && versionIdValue) {
+                // If no document and a version ID is provided, return an error.
                 return cb(null, {
+                    err: errors.NoSuchKey,
                     doc: null,
                     versionId: versionIdValue,
                     key,
                 });
             }
-            // If no master found then ob ject is either non existent
-            // or last version is delete marker
+            // If no master found then object is either non existent or last
+            // version is delete marker
             if (!doc || doc.value.isPHD) {
-                return this.getLatestVersion(c, objName, vFormat, log, (err, _doc) => {
-                    if (err && !err.is.NoSuchKey) {
-                        return cb(err);
-                    }
-                    return cb(null, {
-                        doc: _doc || null,
-                        versionId: versionIdValue,
-                        key,
-                    });
-                });
+                return this.getLatestVersion(c, objName, vFormat, log, (err, _doc) => cb(null, {
+                    err,
+                    doc: _doc || null,
+                    versionId: versionIdValue,
+                    key,
+                }));
             }
             MongoUtils.unserialize(doc.value);
             return cb(null, {
+                err: null,
                 doc: doc.value,
                 versionId: versionIdValue,
                 key,
@@ -1015,15 +1018,9 @@ class MongoClientInterface {
                 return callback(err);
             }
             vFormat = _vFormat;
-            // Create keys, maintaining the context with each key
-            const keysAndObjects = objects.map(({ key: objName, params }) => {
-                const _key = params && params.versionId
-                    ? formatVersionKey(objName, params.versionId, vFormat)
-                    : formatMasterKey(objName, vFormat);
-                return { key: _key, objName, params };
-            });
-            // Extract keys and find documents
-            const keys = keysAndObjects.map(o => o.key);
+            const keys = objects.map(({ key: objName, params }) => (params && params.versionId
+                ? formatVersionKey(objName, params.versionId, vFormat)
+                : formatMasterKey(objName, vFormat)));
             return c.find({
                 _id: { $in: keys },
                 $or: [
@@ -1037,14 +1034,17 @@ class MongoClientInterface {
                 // Create a Map to quickly find docs by their keys
                 const docByKey = new Map(docs.map(doc => [doc._id, doc]));
                 // Process each document using associated context (objName, params)
-                return async.mapLimit(keysAndObjects, 5, ({ key, objName, params }, cb) => {
-                    const doc = docByKey.get(key);
-                    processDoc(doc, objName, params, key, cb);
-                }, callback);
+                return async.mapLimit(objects, constants.maxBatchingConcurrentOperations,
+                    ({ key: objName, params }, cb) => {
+                        const key = params && params.versionId
+                            ? formatVersionKey(objName, params.versionId, vFormat)
+                            : formatMasterKey(objName, vFormat);
+                        const doc = docByKey.get(key);
+                        processDoc(doc, objName, params, key, cb);
+                    }, callback);
             });
         });
     }
-
     /**
      * This function return the latest version of an object
      * by getting all keys related to an object's versions, ordering them
@@ -1205,7 +1205,7 @@ class MongoClientInterface {
                     'value.isPHD': true,
                     'value.versionId': mst.versionId,
                 };
-                this.internalDeleteObject(c, bucketName, masterKey, filter, log, err => {
+                this.internalDeleteObject(c, bucketName, masterKey, filter, null, log, err => {
                     if (err) {
                         // the PHD master might get updated when a PUT is performed
                         // before the repair is done, we don't want to return an error
@@ -1275,7 +1275,7 @@ class MongoClientInterface {
                 next,
             ),
             // delete version
-            next => this.internalDeleteObject(c, bucketName, versionKey, {}, log,
+            next => this.internalDeleteObject(c, bucketName, versionKey, {}, params, log,
                 err => {
                     // we don't return an error in case we don't find
                     // a version as we expect this case when dealing with
@@ -1284,7 +1284,7 @@ class MongoClientInterface {
                         return next(null);
                     }
                     return next(err);
-                }, params),
+                }),
         ], err => {
             if (err) {
                 log.error(
@@ -1311,7 +1311,7 @@ class MongoClientInterface {
      */
     deleteObjectVerNotMaster(c, bucketName, objName, params, log, cb) {
         const versionKey = formatVersionKey(objName, params.versionId, params.vFormat);
-        this.internalDeleteObject(c, bucketName, versionKey, {}, log, err => {
+        this.internalDeleteObject(c, bucketName, versionKey, {}, params, log, err => {
             if (err) {
                 if (err.is.NoSuchKey) {
                     log.error(
@@ -1325,7 +1325,7 @@ class MongoClientInterface {
                 return cb(errors.InternalError);
             }
             return cb(null);
-        }, params);
+        });
     }
 
     /**
@@ -1401,7 +1401,7 @@ class MongoClientInterface {
      */
     deleteObjectNoVer(c, bucketName, objName, params, log, cb) {
         const masterKey = formatMasterKey(objName, params.vFormat);
-        this.internalDeleteObject(c, bucketName, masterKey, {}, log, err => {
+        this.internalDeleteObject(c, bucketName, masterKey, {}, params, log, err => {
             if (err) {
                 // Should not return an error when no object is found
                 if (err.is.NoSuchKey) {
@@ -1413,7 +1413,7 @@ class MongoClientInterface {
                 return cb(errors.InternalError);
             }
             return cb(null);
-        }, params);
+        });
     }
 
     /**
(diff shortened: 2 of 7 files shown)


lib/storage/metadata/MetadataWrapper.js
```diff
@@ -213,6 +213,25 @@ class MetadataWrapper {
...
     }
 
+    getObjectsMD(bucketName, objNamesWithParams, log, cb) {
+        if (typeof this.client.getObjects !== 'function') {
+            log.debug('backend does not support get object metadata with batching', {
+                implName: this.implName,
+            });
+            return cb(errors.NotImplemented);
+        }
+        log.debug('getting object from metadata');
```
jonathan-gramain:
```suggestion
        log.debug('getting objects from metadata', { objects: objNamesWithParams });
```

lib/storage/metadata/MetadataWrapper.js
```diff
@@ -213,6 +213,25 @@ class MetadataWrapper {
...
+        return this.client.getObjects(bucketName, objNamesWithParams, log, (err, data) => {
+            if (err) {
+                log.debug('error from metadata', { implName: this.implName,
```
jonathan-gramain:
```suggestion
                log.debug('error getting objects from metadata', { implName: this.implName, objects: objNamesWithParams,
```

lib/storage/metadata/MetadataWrapper.js
```diff
@@ -213,6 +213,25 @@ class MetadataWrapper {
...
+                    err });
+                return cb(err);
+            }
+            log.debug('object retrieved from metadata');
```
jonathan-gramain:
```suggestion
            log.debug('objects retrieved from metadata', { objects: objNamesWithParams });
```

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -965,6 +965,86 @@ class MongoClientInterface {
...
+        const processDoc = (doc, objName, params, key, cb) => {
+            const versionIdValue = params && params.versionId ? params.versionId : undefined;
+            if (!doc && versionIdValue) {
+                return cb(null, {
+                    doc: null,
+                    versionId: versionIdValue,
+                    key,
+                });
+            }
+            // If no master found then ob ject is either non existent
```
```suggestion
            // If no master found then object is either non existent
```

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -965,6 +965,86 @@ class MongoClientInterface {
...
+            // or last version is delete marker
+            if (!doc || doc.value.isPHD) {
+                return this.getLatestVersion(c, objName, vFormat, log, (err, _doc) => {
+                    if (err && !err.is.NoSuchKey) {
+                        return cb(err);
```
jonathan-gramain: It might be worth it to return individual error values for each object, in an `error` field, including "not found" cases:
- following the same API than `getObject` to avoid any surprise (e.g. set the error `NoSuchKey` to the corresponding response item when the equivalent `getObject` API would have returned this error)
- when calling `getLatestVersion`, we might have errors that do not prevent getting the other objects (the code using this API would likely just treat the response as individual items anyway, not changing its core logic compared to using `getObject`)
- maybe in the future we would have things like conditional GETs, which specific error codes could also be embedded as individual errors

ghost: Thanks for the comment, I have two questions around this:

* When getting all objects at once with batching, we can "only" handle the NoSuchKeys right? As any other error preventing the DB call to complete will result in the `catch` being called.
* If I understand correctly, you would like to have, instead of a `null` document (or more precisely, in addition to the `null` document), an `err`-like field for each of the returned object with the appropriate error, right? If yes, then we can return one error per call to `getLatestVersion` and let the upper code handle these.

ghost: Here is my commit with the change: https://github.com/scality/Arsenal/pull/2134/commits/34b8ddbbe8b8515e31ea7eb7705c8988c9a786fa

jonathan-gramain (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -965,6 +965,85 @@ class MongoClientInterface {
...
+            const keysAndObjects = objects.map(({ key: objName, params }) => {
+                const _key = params && params.versionId
+                    ? formatVersionKey(objName, params.versionId, vFormat)
+                    : formatMasterKey(objName, vFormat);
+                return { key: _key, objName, params };
+            });
+            // Extract keys and find documents
+            const keys = keysAndObjects.map(o => o.key);
+            return c.find({
+                _id: { $in: keys },
```
francoisferrand, line 1025: mongo query size is limited by the max BSON object length (16MB) : so we need to make sure we don't hit that limit...

Now multi-delete is limited to 1000 objects at a time, and object name is limited to 1024 characters : so we should be fine, worse-case at around 1.1MB.

We may want to check the number of objects passed to `getObjects`, though, to be on the safe side.

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -965,6 +965,85 @@ class MongoClientInterface {
...
+                    { 'value.deleted': { $eq: false } },
+                ],
+            }).toArray((err, docs) => {
+                if (err) {
+                    return callback(err);
+                }
+                // Create a Map to quickly find docs by their keys
+                const docByKey = new Map(docs.map(doc => [doc._id, doc]));
+                // Process each document using associated context (objName, params)
+                return async.mapLimit(keysAndObjects, 5, ({ key, objName, params }, cb) => {
```
francoisferrand: this `5` should moved to properly identified a constant

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -965,6 +965,85 @@ class MongoClientInterface {
...
+            vFormat = _vFormat;
+            // Create keys, maintaining the context with each key
+            const keysAndObjects = objects.map(({ key: objName, params }) => {
+                const _key = params && params.versionId
+                    ? formatVersionKey(objName, params.versionId, vFormat)
+                    : formatMasterKey(objName, vFormat);
+                return { key: _key, objName, params };
+            });
+            // Extract keys and find documents
+            const keys = keysAndObjects.map(o => o.key);
```
francoisferrand: is this "map" really useful?
it seems we really just need the list of keys (for mongo), and later we can iterate over `objects` directly...

seems lots of data structure are created in here (`keysAndObjects`, `keys`, `docByKey`) which may end up costing memory/gc : and maybe 'cost' more than re-formatting the key twice...

not sure if this is an issue or if we need to/can improve this (but i would prefer to avoid having 2 maps at leas...)

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1347,9 +1426,26 @@ class MongoClientInterface {
      * @param {Logger}log logger instance
      * @param {Function} cb callback containing error
      * and BulkWriteResult
+     * @param {object} [params] request params
      * @return {undefined}
      */
-    internalDeleteObject(collection, bucketName, key, filter, log, cb) {
+    internalDeleteObject(collection, bucketName, key, filter, log, cb, params = null) {
```
francoisferrand: should keep `cb` as last parameter (anyway you modified all call sites to pass `params`)

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1387,26 +1483,31 @@ class MongoClientInterface {
                 const obj = doc.value;
                 const objMetadata = new ObjectMD(obj.value);
                 objMetadata.setOriginOp('s3:ObjectRemoved:Delete');
-                objMetadata.setDeleted(true);
+                // Not supported in 7.x
+                // objMetadata.setDeleted(true);
```
francoisferrand, line 1488: do you mean S3C 7.x ?

this code is used exclusively for mongo, which is used only in 8.x ; and this is javascript, so no type checking : i.e. no reason to remove this line (and possibly have an issue on 8)

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1387,26 +1483,31 @@ class MongoClientInterface {
...
+                // Add the delete operation if the shouldDelete flag is true
+                operations.push({
                     deleteOne: {
                         filter: updateDeleteFilter,
                     },
-                },
-            ], { ordered: true }, () => next(null)),
+                });
+                collection.bulkWrite(operations, { ordered: true }, () => next(null));
+            },
```
francoisferrand: no change here, esp. the introduced comment does not seem valid?
→ keep as the code was before...

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1347,9 +1426,26 @@ class MongoClientInterface {
...
      * @return {undefined}
      */
-    internalDeleteObject(collection, bucketName, key, filter, log, cb) {
+    internalDeleteObject(collection, bucketName, key, filter, log, cb, params = null) {
+        // filter used when deleting object
+        const deleteFilter = Object.assign({
+            _id: key,
+        }, filter);
+
+        if (params && params.shouldOnlyDelete) {
```
francoisferrand: i fear `shouldOnlyDelete` is not very clear about what happens: I would prefer expliciting the specific behavior we (sometimes) request, e.g. sometimes like `needObjectDeletedOplog`

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -965,6 +965,85 @@ class MongoClientInterface {
...
+                    err: errors.NoSuchKey,
+                    doc: null,
+                    versionId: versionIdValue,
+                    key,
+                });
+            }
+            // If no master found then object is either non existent or last
+            // version is delete marker
+            if (!doc || doc.value.isPHD) {
+                return this.getLatestVersion(c, objName, vFormat, log, (err, _doc) => cb(null, {
```
francoisferrand, line 1001: (probably not for this PR, though)
could we leverage the `find()` to retrieve as well the masters all at once?

or do we think this case will happen seldom enough that it is best to perform the extra find for these cases and keep the first find() faster (with less results) ?

alternatively, maybe we could rework this to 'batch' the getLatestVersion: e.g.
1. find() all keys
2. go over the returned documents and 'process' them
3. find() the master for all remaining objects [e.g. those not processed in 2.)
4. go over the returned documents and 'process' them
5. process the remaining documents (e.g. NoSuchKey)

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: I kept this line in Arsenal 8.x, but because we have this code in Arsenal 7.x as well, and this line relies on a non implemented function, it's better to comment it... Otherwise the unit test fails

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: So you want to invert the default behavior and only do the extra calls to MongoDB when we request it? How to detect when we "need" it?

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: For the first point, please create a ticket for it if you think this should be included :smile:

For the alternative, we do that in parallel already, I don't understand how, from your list, we can "batch" in 3)? Do you mean doing a single `find` with all the master we need? I.e., do 2 batched calls to MongoDB? 
I don't think we should introduce more risks in this PR, as batching would mean creating a new method for the `getLatestVersion` function, hence more things to maintain...

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: Any reason why we cannot put optional params after the callback as we do in other functions? Like the `originOp`, `isRetry` and all?

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: So far no issues with the memory with testing... We already greatly improve it thanks to the batching, instead of doing 1000s of calls to MongoDB...

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: Done in https://github.com/scality/Arsenal/pull/2134/commits/cc75b89115a7a8d4f26417fde5103f80421bedef#diff-9c207befbd8f4d79a3b6e3d7d143444e804ca1684236b8fd9fa0f9e03c53477aR979

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: https://github.com/scality/Arsenal/pull/2134/commits/cc75b89115a7a8d4f26417fde5103f80421bedef#diff-9c207befbd8f4d79a3b6e3d7d143444e804ca1684236b8fd9fa0f9e03c53477aR1037 & https://github.com/scality/Arsenal/pull/2134/commits/cc75b89115a7a8d4f26417fde5103f80421bedef#diff-561ab956a0b904914573a9544f92493c5493b6e14d30372365bcdd9c5c7f6e59R129

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: Here is my tentative: https://github.com/scality/Arsenal/pull/2134/commits/cc75b89115a7a8d4f26417fde5103f80421bedef#diff-9c207befbd8f4d79a3b6e3d7d143444e804ca1684236b8fd9fa0f9e03c53477aR976
It removes one of the maps

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: Reverted in https://github.com/scality/Arsenal/pull/2134/commits/cc75b89115a7a8d4f26417fde5103f80421bedef#diff-9c207befbd8f4d79a3b6e3d7d143444e804ca1684236b8fd9fa0f9e03c53477aR1494

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: I updated the flag with a more meaningful name without changing the logic: https://github.com/scality/Arsenal/pull/2134/commits/cc75b89115a7a8d4f26417fde5103f80421bedef#diff-9c207befbd8f4d79a3b6e3d7d143444e804ca1684236b8fd9fa0f9e03c53477aR1439

Let me know if you want a better name or something else...

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: Done in https://github.com/scality/Arsenal/pull/2134/files#diff-9c207befbd8f4d79a3b6e3d7d143444e804ca1684236b8fd9fa0f9e03c53477aR1433

lib/storage/metadata/mongoclient/MongoClientInterface.js
francoisferrand: > Do you mean doing a single find with all the master we need? I.e., do 2 batched calls to MongoDB

Yes, this was the idea. Not sure if it is more efficient overall (depend on how much we reach that case), the point is that here we want to do batching, but may fall back to making one call for each master anyway...

> For the first point, please create a ticket for it if you think this should be included

I don't know if it _should_ be included, it is a compromise depending on the workload... hence the question:  do you think this case will happen seldom enough that it is best to perform the extra find for these cases and keep the first find() faster (with less results) ?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -965,6 +965,86 @@ class MongoClientInterface {
...
+     * @param {String} bucketName bucket name
+     * @param {Array} objects array of objects
+     * @param {Object} log logger
+     * @param {Function} callback callback
+     * @return {undefined}
+     */
+    getObjects(bucketName, objects, log, callback) {
+        const c = this.getCollection(bucketName);
+        // We do not accept more than 1000 keys in a single request
+        if (objects.length > 1000) {
```
francoisferrand: logically, should this check not be done after the next one (`Array.isArray(objects)`) ?

KazToozs (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: To me, this is definitely an optimization we can do. For our use case here, this is not needed, so let's not add more risks in this PR.

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: yes, definitely: https://github.com/scality/Arsenal/pull/2134/commits/74218ad3ef820998d9970bfdb367e50f38608f25

lib/storage/metadata/mongoclient/MongoClientInterface.js
francoisferrand: no need to add in this PR indeed, but please file a ticket if you think we should do it (eventually).

francoisferrand (approved):

ghost:
ping

PR #2150, opened 2023-07-12
ARSN-350 Missing Null Version in Lifecycle List of Non-Current Versions

nicolas2bert:
_Note: We only support the v1 bucket format for "list lifecycle" in Artesca._

We made the assumption that the first version key stored the current/latest version, which is true in most cases except for "null" versions. In the case of a "null" version, the current version is stored in the master key alone, rather than being stored in both the master key and a new version key. Here's an example of the key structure:

`Mkey0`: Represents the null version ID.
`VKey0<versionID>`: Represents a non-current version.

Additionally, we assumed that the versions for a given key were ordered by creation date, from newest to oldest. However, in Ring S3C, for non-current null versions, the metadata version ID is not part of the metadata key id. Therefore, the non-current null version is listed before the current version that has a version ID. Here's an example of the key ordering:

`Mkey0`: Master version
`Vkey0`: "Null" Non-current version
`VKey0<versionID>`: Current version

**The listing was using only versions, but because those assumptions are incorrect, we now use both the master and the versions for each given key to ensure that we return the correct non-current versions.**

### DIFF
--- a/lib/algos/list/delimiterNonCurrent.js
+++ b/lib/algos/list/delimiterNonCurrent.js
@@ -3,6 +3,7 @@ const Delimiter = require('./delimiter').Delimiter;
 const VSConst = require('../../versioning/constants').VersioningConstants;
 const { inc, FILTER_ACCEPT, FILTER_END, SKIP_NONE } = require('./tools');
 const VID_SEP = VSConst.VersionId.Separator;
+const Version = require('../../versioning/Version').Version;
 const { DbPrefixes } = VSConst;
 
 // TODO: find an acceptable timeout value.
@@ -36,8 +37,10 @@ class DelimiterNonCurrent extends Delimiter {
         this.skipping = this.skippingV1;
         this.genMDParams = this.genMDParamsV1;
 
-        this.keyName = null;
+        // internal state
         this.staleDate = null;
+        this.masterKey = undefined;
+        this.masterVersionId = undefined;
 
         // used for monitoring
         this.evaluatedKeys = 0;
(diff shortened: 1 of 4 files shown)


lib/algos/list/delimiterNonCurrent.js
```diff
@@ -131,25 +178,29 @@ class DelimiterNonCurrent extends Delimiter {
...
+            this.masterVersionId = Version.from(value).getVersionId() || 'null';
 
-        if (isLatestVersion) {
-            this.keyName = key;
-            // The current last-modified date is kept in memory and used as a "stale date" for the following version.
             this.staleDate = this.getLastModified(value);
             return FILTER_ACCEPT;
         }
 
+        const isCurrentVersion = this.masterKey === key && this.masterVersionId === versionId;
```
jonathan-gramain, line 198: If delete markers are also candidates for expiration, they don't have an associated master key even though they can be current. The way to know whether they are current is if the previous key in the merged M+V listing is a different key, or if it's the very first key of the listing.

jonathan-gramain (commented):
We should also keep in mind that at some point, Artesca may enable null keys on Cloudserver (the change to allow V1 metadata to be more performant), and this listing algo doesn't seem to support them properly. Not needed urgently as currently Cloudserver on Artesca is hard-coded not to use null keys although the code is there, but we should probably track this in some way to avoid forgetting about it the day we decide to switch to null keys in Artesca.

This is where the compat mode (not to use null keys) is forcefully enabled today:

https://github.com/scality/cloudserver/blob/652bf925365e8ec1bd94dec9b54ac61b675da7f2/lib/Config.js#L1611

lib/algos/list/delimiterNonCurrent.js
nicolas2bert: - Delete markers have a version ID that is correctly ordered.
- It doesn't matter if the current delete marker doesn't have a master when listing non-current versions.
The first version key in the listing should store the "delete marker" version and will be skipped.

nicolas2bert: Also, just to make sure, I added tests for delete marker here: https://github.com/scality/cloudserver/pull/5231

nicolas2bert:
@jonathan-gramain Is there a ticket for Artesca to support "null keys"?
Based on this comment: `"null version compatibility mode is enforced because null keys are not supported by the  MongoDB backend."`, it seems like no listing algo supports "null keys".

jonathan-gramain:
@nicolas2bert it would be part of the work to use 8.x branches for S3C, I don't think there is an EPIC for this but it should go in there. I'll check this.

The current Arsenal code for DelimiterVersions supports null keys, as it was forward-ported from 7.70 (for example: https://github.com/scality/Arsenal/blob/development/8.1/lib/algos/list/delimiterVersions.ts#L302)

jonathan-gramain (approved):

naren-rajendran (approved):

nicolas2bert:
@bert-e approve

PR #2152, opened 2023-07-24
Adapt delimiterCurrent for S3C Metadata

nicolas2bert:
In Artesca, for listing the current versions, filtering on “last-modified“ was done at the MongoDB level (query filter).

Metadata does not support built-in filters today.

Listing algorism for the current versions has to be modified to filter out non-eligible masters based on their “last-modified” date and “dataStoreName“.

### DIFF
--- a/lib/algos/list/delimiterCurrent.ts
+++ b/lib/algos/list/delimiterCurrent.ts
@@ -1,4 +1,5 @@
 const { Delimiter } = require('./delimiter');
+const { FILTER_ACCEPT, FILTER_END } = require('./tools');
 
 type ResultObject = {
     Contents: {
@@ -8,6 +9,8 @@ type ResultObject = {
     IsTruncated: boolean;
     NextMarker ?: string;
 };
+
+const DELIMITER_TIMEOUT_MS = 10 * 1000; // 10s
 
 /**
  * Handle object listing with parameters. This extends the base class Delimiter
@@ -27,6 +30,9 @@ class DelimiterCurrent extends Delimiter {
 
         this.beforeDate = parameters.beforeDate;
         this.excludedDataStoreName = parameters.excludedDataStoreName;
+        // used for monitoring
+        this.start = null;
+        this.evaluatedKeys = 0;
     }
 
     genMDParamsV1() {
@@ -43,7 +49,60 @@ class DelimiterCurrent extends Delimiter {
                 ne: this.excludedDataStoreName,
             }
         }
+
+        this.start = Date.now();
+
         return params;
+    }
+
+    /**
+     * Parses the stringified entry's value.
+     * @param s - sringified value
+     * @return - undefined if parsing fails, otherwise it contains the parsed value.
+     */
+    _parse(s) {
+        let p;
+        try {
+            p = JSON.parse(s);
+        } catch (e: any) {
+            this.logger.warn(
+                'Could not parse Object Metadata while listing',
+                { err: e.toString() });
+        }
+        return p;
+    }
+
+    filter(obj) {
+        if (this.start && Date.now() - this.start > DELIMITER_TIMEOUT_MS) {
+            this.IsTruncated = true;
+            this.logger.info('listing stopped after expected internal timeout',
+                {
+                    timeoutMs: DELIMITER_TIMEOUT_MS,
+                    evaluatedKeys: this.evaluatedKeys,
+                });
+            return FILTER_END;
+        }
+        ++this.evaluatedKeys;
+
+        const parsedValue = this._parse(obj.value);
+        // if parsing fails, skip the key.
+        if (parsedValue) {
+            const lastModified = parsedValue['last-modified'];
+            const dataStoreName = parsedValue.dataStoreName;
+            // We then check if the current version is older than the "beforeDate" and
+            // "excludedDataStoreName" is not specified or if specified and the data store name is different.
+            if ((!this.beforeDate || (lastModified && lastModified < this.beforeDate)) && 
+            (!this.excludedDataStoreName || dataStoreName !== this.excludedDataStoreName)) {
+                return super.filter(obj);
+            }
+            // In the event of a timeout occurring before any content is added,
+            // NextMarker is updated even if the object is not eligible.
+            // It minimizes the amount of data that the client needs to re-process if the request times out.
+            const key = this.getObjectKey(obj);
+            this.NextMarker = key;
+        }
+
+        return FILTER_ACCEPT;
     }
 
     result(): ResultObject {
--- a/tests/unit/algos/list/delimiterCurrent.spec.js
+++ b/tests/unit/algos/list/delimiterCurrent.spec.js
@@ -13,6 +13,8 @@ const VSConst =
     require('../../../../lib/versioning/constants').VersioningConstants;
 const { DbPrefixes } = VSConst;
 
+const DELIMITER_TIMEOUT_MS = 10 * 1000; // 10s
+
 const VID_SEP = VSConst.VersionId.Separator;
 const EmptyResult = {
     Contents: [],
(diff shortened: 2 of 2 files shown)


lib/algos/list/delimiterCurrent.ts
```diff
@@ -43,9 +49,57 @@ class DelimiterCurrent extends Delimiter {
...
+                    timeoutMs: DELIMITER_TIMEOUT_MS,
+                    evaluatedKeys: this.evaluatedKeys,
+                });
+            return FILTER_END;
+        }
+        ++this.evaluatedKeys;
+
+        const parsedValue = this._parse(obj.value);
+        // if parsing fails, skip the key.
+        if (parsedValue) {
```
ghost, line 89: I see that after this condition (when parsing succeed), we have `return FILTER_ACCEPT;`, does that mean we skip the key? It looks like we accept it instead...

lib/algos/list/delimiterCurrent.ts
```diff
@@ -43,9 +49,57 @@ class DelimiterCurrent extends Delimiter {
...
+            const lastModified = parsedValue['last-modified'];
+            const dataStoreName = parsedValue.dataStoreName;
+            // We then check if the current version is older than the "beforeDate" and
+            // "excludedDataStoreName" is not specified or if specified and the data store name is different.
+            if ((!this.beforeDate || (lastModified && lastModified < this.beforeDate)) &&
```
ghost, line 94: Do we have cases where `lastModified` can be undefined? If so, does such case requires to skip the super.filter call? Here, It looks like we are accepting the key, in this case.

tests/unit/algos/list/delimiterCurrent.spec.js
```diff
@@ -125,4 +129,213 @@ describe('DelimiterCurrent', () => {
...
+            value: value3,
+        }), FILTER_END);
+
+        const expectedResult = {
+            Contents: [],
+            NextMarker: masterKey2,
+            IsTruncated: true,
+        };
+
+        assert.deepStrictEqual(delimiter.result(), expectedResult);
```
ghost, line 379: Should we have a unit test in case we have no returned lastModified?

lib/algos/list/delimiterCurrent.ts
```diff
@@ -9,6 +10,8 @@ type ResultObject = {
     NextMarker ?: string;
 };
 
+const DELIMITER_TIMEOUT_MS = 10 * 1000; // 10s
```
ghost, line 13: I'm curious to understand why 10s, is it related to some kind of limit we already have when making requests?

lib/algos/list/delimiterCurrent.ts
nicolas2bert: - If beforeDate is undefined, we "add" the key without checking for the last-modified property.

- If beforeDate is defined, we perform the following check before adding the key:

1. Ensure that the last-modified property exists as a safeguard, considering that it might never happen.
2. Verify that the last-modified property's value is a valid date and older than the beforeDate.

The reason for adding the step to check for the existence of the `last-modified` property is to prevent unexpected behavior when `last-modified` is not present (especially empty). This check acts as a safeguard since it might never happen, but it ensures that the comparison is performed only when the property is available and avoids potential errors or incorrect results in such scenarios. 
By confirming the existence of the property, we can safely proceed with the comparison and accurately determine if the entry should be added.

lib/algos/list/delimiterCurrent.ts
nicolas2bert: `return FILTER_ACCEPT;` basically means that we keep reading from the stream.
To actually add an entry to the response's content, we use the `addContents()` function from the `super.filter(obj)`.
If parsing fails, we do not add the entry to the response's content but keep reading from the stream.

lib/algos/list/delimiterCurrent.ts
nicolas2bert: We are working on defining an acceptable approach/value: https://scality.atlassian.net/browse/S3C-7928

lib/algos/list/delimiterCurrent.ts
ghost: Thanks for the explanation, then it makes sense

lib/algos/list/delimiterCurrent.ts
ghost: It's more a safety check I got it, thank you for the very detailed explanation!

lib/algos/list/delimiterCurrent.ts
ghost: I see, my 2 cents on this would be to add a scality header to let the client decide, and default to the solution that would ensure expected responses (2nd approach), as the first one may requires implementation changes for our customers. But that's only my opinion :smile:

ghost (approved):
Thank you for adding the unit test and the explanation, LGTM

lib/algos/list/delimiterCurrent.ts
jonathan-gramain: I am in favor of a limit on the listing, easier to test and levelup module supports the `limit` argument to stop listing after a certain number of entries is returned (https://github.com/Level/levelup#dbcreatereadstreamoptions)

jonathan-gramain (commented):
As discussed in Slack, we can/should reuse and filter results from `DelimiterMaster` class to abstract the metadata format handling.

I have a general question though: `Delimiter` would return both master and version keys (in a raw form without the V1 prefix), and more importantly it returns all version keys, although I think we only want to process current versions here. Is this wanted?

lib/algos/list/delimiterCurrent.ts
```diff
@@ -43,9 +49,57 @@ class DelimiterCurrent extends Delimiter {
...
+        const parsedValue = this._parse(obj.value);
+        // if parsing fails, skip the key.
+        if (parsedValue) {
+            const lastModified = parsedValue['last-modified'];
+            const dataStoreName = parsedValue.dataStoreName;
+            // We then check if the current version is older than the "beforeDate" and
+            // "excludedDataStoreName" is not specified or if specified and the data store name is different.
+            if ((!this.beforeDate || (lastModified && lastModified < this.beforeDate)) && 
+            (!this.excludedDataStoreName || dataStoreName !== this.excludedDataStoreName)) {
+                return super.filter(obj);
```
jonathan-gramain, line 96: cf my comment for the review, I think this would be called for all keys (master and versions), hence the Delimiter class would output all versions of an object in the result.

nicolas2bert:
As we discussed in Slack, leveraging DelimiterMaster and DelimiterVersion classes is still the plan to support all the different bucket version formats. 
However, this PR is a simple follow up of the backporting of lifecycle listing done in Artesca (that only support V1 bucket format). 
Today, Ring S3C logic only supports v1 bucket format.
I will soon push the changes to support both v0 and v1 format leveraging DelimiterMaster and DelimiterVersion.

lib/algos/list/delimiterCurrent.ts
nicolas2bert: With v1 bucket format only masters are returned. I already wrote tests for it in Cloudserver development/8.x branches.

To support v0, we are planing on leveraging DelimiterMaster.

lib/algos/list/delimiterCurrent.ts
```diff
@@ -43,9 +49,57 @@ class DelimiterCurrent extends Delimiter {
                 ne: this.excludedDataStoreName,
             }
         }
+
+        this.start = Date.now();
+
         return params;
     }
 
+    _parse(s) {
```
jonathan-gramain, line 63: A little JSDoc could help, if only to mention that it may return an object or `undefined` if parsing fails

tests/unit/algos/list/delimiterCurrent.spec.js
```diff
@@ -125,4 +129,253 @@ describe('DelimiterCurrent', () => {
...
+    });
+
+    it('should return an empty list if last-modified is undefined', () => {
+        const beforeDate = '1970-01-01T00:00:00.003Z';
+        const delimiter = new DelimiterCurrent({ beforeDate }, fakeLogger, 'v1');
+
+        const masterKey0 = 'key0';
+        const value0 = '';
```
jonathan-gramain: Let's make the value a valid JSON object to avoid triggering the "invalid JSON" condition:

```suggestion
        const value0 = '{}';
```

tests/unit/algos/list/delimiterCurrent.spec.js
```diff
@@ -125,4 +129,253 @@ describe('DelimiterCurrent', () => {
...
+
+        const expectedResult = {
+            Contents: [],
+            IsTruncated: false,
+        };
+
+        assert.deepStrictEqual(delimiter.result(), expectedResult);
+    });
+
+    it('should return the object with dataStore name that does not match', () => {
```
jonathan-gramain, line 208: This test doesn't seem to test anything more than the next test, maybe could be removed to simplify?

jonathan-gramain (approved):
LGTM with a few comments

nicolas2bert:
@bert-e reset

nicolas2bert:
@bert-e approve

PR #2155, opened 2023-07-28
ARSN-354 List lifecycle current versions supports V0 bucket format

nicolas2bert:

- The implementation, now, leverages `DelimiterMaster` to provide an accurate list of objects, which then gets filtered based on `last-modified` date and `dataStoreName` location.

### DIFF
--- a/lib/algos/list/delimiterCurrent.ts
+++ b/lib/algos/list/delimiterCurrent.ts
@@ -1,4 +1,4 @@
-const { Delimiter } = require('./delimiter');
+const { DelimiterMaster } = require('./delimiterMaster');
 const { FILTER_ACCEPT, FILTER_END } = require('./tools');
 
 type ResultObject = {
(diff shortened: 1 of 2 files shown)


lib/algos/list/delimiterCurrent.ts
```diff
@@ -32,12 +32,13 @@ class DelimiterCurrent extends Delimiter {
...
-        this.evaluatedKeys = 0;
+        this.scannedKeys = 0;
     }
 
-    genMDParamsV1() {
-        const params = super.genMDParamsV1();
-
+    genMDParamsV0() {
+        const params = super.genMDParamsV0();
+        // lastModified and dataStoreName parameters are used by metadata that enables built-in filtering,
```
jonathan-gramain, line 40: It feels odd to see those parameters now in the `genMDParamsV0` section, since MongoDB uses V1 metadata. But it's more theoretical as Metadata doesn't use them anyway (we could maybe remove them, but we'd have to add them back in the 8.x branches, so good for me, they should not have side-effects in Metadata).

jonathan-gramain (approved):

rachedbenmustapha (approved):

lib/algos/list/delimiterCurrent.ts
nicolas2bert: Could "query/built-in filter" be supported one day in Metadata?

nicolas2bert:
@bert-e approve

PR #2156, opened 2023-07-31
ARSN-355 List lifecycle non-current versions supports V0

nicolas2bert:

- The implementation, now, leverages `DelimiterVersions` to provide an accurate list of versions.

### DIFF
--- a/lib/algos/list/delimiterNonCurrent.js
+++ b/lib/algos/list/delimiterNonCurrent.js
@@ -1,103 +1,43 @@
-'use strict'; // eslint-disable-line strict
-const Delimiter = require('./delimiter').Delimiter;
-const VSConst = require('../../versioning/constants').VersioningConstants;
-const { inc, FILTER_ACCEPT, FILTER_END, SKIP_NONE } = require('./tools');
-const VID_SEP = VSConst.VersionId.Separator;
-const Version = require('../../versioning/Version').Version;
-const { DbPrefixes } = VSConst;
+const { DelimiterVersions } = require('./delimiterVersions');
+const { FILTER_END, FILTER_SKIP } = require('./tools');
 
 // TODO: find an acceptable timeout value.
 const DELIMITER_TIMEOUT_MS = 10 * 1000; // 10s
 const TRIM_METADATA_MIN_BLOB_SIZE = 10000;
 
 /**
- * Handle object listing with parameters. This extends the base class Delimiter
+ * Handle object listing with parameters. This extends the base class DelimiterVersions
  * to return the raw non-current versions objects.
  */
-class DelimiterNonCurrent extends Delimiter {
+class DelimiterNonCurrent extends DelimiterVersions {
     /**
      * Delimiter listing of non-current versions.
      * @param {Object}  parameters                  - listing parameters
-     * @param {String}  parameters.versionIdMarker  - version id marker
-     * @param {String}  parameters.beforeDate       - limit the response to keys with stale date older than beforeDate
-     * “stale date” is the date on when a version becomes non-current.
      * @param {String}  parameters.keyMarker        - key marker
+     * @param {String}  parameters.versionIdMarker  - version id marker
+     * @param {String}  parameters.beforeDate       - limit the response to keys with stale date older than beforeDate.
+     * “stale date” is the date on when a version becomes non-current.
+     *  @param {String} parameters.excludedDataStoreName - exclude dataStoreName matches from the versions
      * @param {RequestLogger} logger                - The logger of the request
      * @param {String} [vFormat]                    - versioning key format
      */
     constructor(parameters, logger, vFormat) {
         super(parameters, logger, vFormat);
 
-        this.versionIdMarker = parameters.versionIdMarker;
         this.beforeDate = parameters.beforeDate;
-        this.keyMarker = parameters.keyMarker;
         this.excludedDataStoreName = parameters.excludedDataStoreName;
-        this.NextKeyMarker = null;
-
-        this.skipping = this.skippingV1;
-        this.genMDParams = this.genMDParamsV1;
 
         // internal state
+        this.prevKey = null;
         this.staleDate = null;
-        this.masterKey = undefined;
-        this.masterVersionId = undefined;
 
         // used for monitoring
-        this.evaluatedKeys = 0;
+        this.scannedKeys = 0;
     }
 
-    skippingV1() {
-        return SKIP_NONE;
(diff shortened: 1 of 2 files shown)


nicolas2bert:
ping

lib/algos/list/delimiterNonCurrent.js
```diff
@@ -1,103 +1,40 @@
...
- * Handle object listing with parameters. This extends the base class Delimiter
+ * Handle object listing with parameters. This extends the base class DelimiterVersions
  * to return the raw non-current versions objects.
  */
-class DelimiterNonCurrent extends Delimiter {
+class DelimiterNonCurrent extends DelimiterVersions {
     /**
      * Delimiter listing of non-current versions.
      * @param {Object}  parameters                  - listing parameters
-     * @param {String}  parameters.versionIdMarker  - version id marker
```
jonathan-gramain, line 21: I think we should leave the jsdoc for `keyMarker` and `versionIdMarker` since those are still accepted by the parent class constructor. (there might be others as well).

lib/algos/list/delimiterNonCurrent.js
```diff
@@ -1,103 +1,40 @@
...
-            } else {
-                delete vParams.gte;
-                delete mParams.gte;
-                vParams.gt = DbPrefixes.Version + inc(this.keyMarker + VID_SEP);
-                mParams.gt = DbPrefixes.Master + inc(this.keyMarker + VID_SEP);
-            }
-        }
-
+    genMDParamsV0() {
         this.start = Date.now();
```
jonathan-gramain, line 39: Don't we need to also set `start` in `genMDParamsV1`?

lib/algos/list/delimiterNonCurrent.js
```diff
@@ -115,39 +52,45 @@ class DelimiterNonCurrent extends Delimiter {
...
+            if (_versionId < this.versionIdMarker) {
+                // skip all versions until marker
+                return FILTER_SKIP;
+            }
         }
-        const nonversionedKey = fullKey.slice(0, versionIdIndex);
-        const versionId = fullKey.slice(versionIdIndex + 1);
-        return { key: nonversionedKey, versionId };
+        this.setState({
+            id: 1 /* NotSkipping */,
```
jonathan-gramain, line 73: Since the `DelimiterVersions` class exports the enum `DelimiterVersionsFilterStateId`, it should be possible to import it in this file and use its named values.

lib/algos/list/delimiterNonCurrent.js
```diff
@@ -158,47 +101,21 @@ class DelimiterNonCurrent extends Delimiter {
...
-        this.NextKeyMarker = key;
-        this.NextVersionIdMarker = versionId;
-
-        // The master key serves two purposes:
-        // - It retrieves the expiration date for the previous version that is no longer current.
-        // - It excludes the current version from the list.
-        const isMasterKey = versionId === undefined;
+        // Skip the version if it represents the master key, but keep its last-modified date,
+        // which will be the stale date of the following version.
+        const isMasterKey = this.masterKey === key && this.masterVersionId === versionId;
```
jonathan-gramain: Could this variable be clearer if named `isCurrentVersion`, like in the previous code? At this stage we don't know if the key being processed is a master or a version key, because this is left to the internal implementation of `DelimiterVersions` which contract is to return actual distinct versions to `addContents`, of which one of them is the current version, hence the suggestion.

For the same reason, it could be better to check if the version is the latest by checking if the `key` is different than the previous one (by saving the previous one in a class member) so to avoid using "internal" `this.masterKey` and `this.masterVersionId` fields from `DelimiterVersions`. I'm fine with keeping it as is if it helps, your call.

rachedbenmustapha (approved):

tests/unit/algos/list/delimiterNonCurrent.spec.js
```diff
@@ -31,342 +30,369 @@ const fakeLogger = {
...
+                        value: `{"versionId":"${versionId3}","last-modified":"${date3}","staleDate":"${date2}"}`,
+                    },
+                ],
+                IsTruncated: false,
+            };
+
+            assert.deepStrictEqual(delimiter.result(), expectedResult);
+        });
+
+        it('should accept one delete marker and one versions and return the non current version', () => {
```
jonathan-gramain:
```suggestion
        it('should accept one delete marker and one version and return the noncurrent version', () => {
```

tests/unit/algos/list/delimiterNonCurrent.spec.js
```diff
@@ -31,342 +30,369 @@ const fakeLogger = {
...
+
+            assert.strictEqual(delimiter.filter({
+                key: getListingKey(versionKey1, v),
+                value: value1,
+            }), FILTER_ACCEPT);
+
+            assert.deepStrictEqual(delimiter.result(), EmptyResult);
+        });
+
+        it('should accept two versions and return the non current version', () => {
```
jonathan-gramain:
```suggestion
        it('should accept two versions and return the noncurrent version', () => {
```

tests/unit/algos/list/delimiterNonCurrent.spec.js
```diff
@@ -31,342 +30,369 @@ const fakeLogger = {
...
+                        value: `{"versionId":"${versionId2}","last-modified":"${date2}","staleDate":"${date1}"}`,
+                    },
+                ],
+                IsTruncated: false,
+            };
+
+            assert.deepStrictEqual(delimiter.result(), expectedResult);
+        });
+
+        it('should accept three versions and return the non current version which stale date before beforeDate', () => {
```
jonathan-gramain:
```suggestion
        it('should accept three versions and return the noncurrent version which stale date before beforeDate', () => {
```

jonathan-gramain (approved):
approving with some minor comments

lib/algos/list/delimiterNonCurrent.js
nicolas2bert: `genMDParamsV1` calls `genMDParamsV0`. 
There is a unit test that ensures `this.start` is always assigned for v0 and v1.

nicolas2bert:
@bert-e reset

nicolas2bert:
@bert-e approve

nicolas2bert:
@bert-e reset

nicolas2bert:
@bert-e reset

nicolas2bert:
@bert-e reset

PR #2157, opened 2023-07-31
ARSN-356 List lifecycle orphan delete markers supports V0

nicolas2bert:

- The implementation, now, leverages `DelimiterVersions` to provide an accurate list of objects.

### DIFF
--- a/lib/algos/list/delimiterOrphanDeleteMarker.js
+++ b/lib/algos/list/delimiterOrphanDeleteMarker.js
@@ -1,45 +1,50 @@
-'use strict'; // eslint-disable-line strict
-const Delimiter = require('./delimiter').Delimiter;
-const VSConst = require('../../versioning/constants').VersioningConstants;
-const { inc, FILTER_ACCEPT, FILTER_END, SKIP_NONE } = require('./tools');
-const VID_SEP = VSConst.VersionId.Separator;
-const { DbPrefixes } = VSConst;
-
+const DelimiterVersions = require('./delimiterVersions').DelimiterVersions;
+const { FILTER_END } = require('./tools');
 const DELIMITER_TIMEOUT_MS = 10 * 1000; // 10s
 const TRIM_METADATA_MIN_BLOB_SIZE = 10000;
-
 /**
- * Handle object listing with parameters. This extends the base class Delimiter
+ * Handle object listing with parameters. This extends the base class DelimiterVersions
  * to return the orphan delete markers. Orphan delete markers are also
  * refered as expired object delete marker.
  * They are delete marker with zero noncurrent versions.
  */
-class DelimiterOrphanDeleteMarker extends Delimiter {
+class DelimiterOrphanDeleteMarker extends DelimiterVersions {
     /**
-     * Delimiter listing of non-current versions.
+     * Delimiter listing of orphan delete markers.
      * @param {Object}  parameters            - listing parameters
      * @param {String}  parameters.beforeDate - limit the response to keys older than beforeDate
      * @param {RequestLogger} logger          - The logger of the request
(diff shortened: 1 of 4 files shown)


lib/algos/list/delimiterOrphanDeleteMarker.js
```diff
@@ -1,45 +1,35 @@
...
-
         // used for monitoring
-        this.evaluatedKeys = 0;
+        this.scannedKeys = 0;
     }
 
-    skippingV1() {
-        return SKIP_NONE;
+    genMDParamsV0() {
+        this.start = Date.now();
```
jonathan-gramain, line 46: Same than for the other PR, shouldn't this also be done in a `genMDParamsV1()` method?

lib/algos/list/delimiterOrphanDeleteMarker.js
```diff
@@ -1,45 +1,35 @@
...
+     * Delimiter listing of orphan delete markers.
      * @param {Object}  parameters            - listing parameters
      * @param {String}  parameters.beforeDate - limit the response to keys older than beforeDate
      * @param {RequestLogger} logger          - The logger of the request
      * @param {String} [vFormat]              - versioning key format
      */
     constructor(parameters, logger, vFormat) {
         super(parameters, logger, vFormat);
-
+        this.keyMarker = parameters.marker;
```
jonathan-gramain: Is `this.keyMarker` used anywhere?

lib/algos/list/delimiterOrphanDeleteMarker.js
nicolas2bert: `genMDParamsV1()` calls `genMDParamsV0()`.
There is a unit test that ensures `this.start` is always assigned for v0 and v1.

lib/algos/list/delimiterOrphanDeleteMarker.js
nicolas2bert: `this.keyMarker` is used by `DelimiterVersions`.

lib/algos/list/delimiterOrphanDeleteMarker.js
jonathan-gramain: OK, I recommend to add a comment then, because the implementation of `genMDParamsV1` is in the base class which makes it not obvious.

lib/algos/list/delimiterOrphanDeleteMarker.js
jonathan-gramain: I see, it's because we use `this.marker` here. I would suggest to instead, pass the `keyMarker` and `versionIdMarker` parameters to the `super` constructor, instead of re-initializing it later, to avoid potential interference with the inner workings of `DelimiterVersions` and for general separation of responsibility.

nicolas2bert: The orphan delete marker logic uses the term 'marker' instead of 'keyMarker', as the latter could suggest the presence of a 'versionIdMarker'.

nicolas2bert: I can still update the constructor like so: 
```
constructor(parameters, logger, vFormat) {
        parameters.keyMarker = parameters.marker;
        super(parameters, logger, vFormat);
```

jonathan-gramain: Or just pass a new object instead of `parameters` constructed with the params you need for `DelimiterVersions`, as if you were instanciating it directly.

jonathan-gramain (approved):

nicolas2bert:
@bert-e reset

rachedbenmustapha (approved):

nicolas2bert:
@bert-e approve

PR #2159, opened 2023-08-08
ARSN-359 Fix NextMarker calculation in listLifecycleCurrent 

nicolas2bert:
Please note that there are no missing entries in the listing and no extra resource used since the next listing will do the fetching anyway. The issue lies in how we determine the NextMarker. It has to be compatible with the current logic merged in Artesca.

When using the listLifecycleCurrent function, we need to calculate the NextMarker correctly. Currently, if the maximum number of keys (max-keys) is reached, the function continues fetching more entries, which is unnecessary and should be done by the next listing. This is because the `_reachedMaxKeys` logic is set after the "filter" logic.

For instance, if max-keys is set to 1 and the first entry (key0) is eligible, while the following two entries (key1 and key2) are not eligible, but the fourth entry (key3) is eligible, the listing should stop at key0 and the NextMarker should be key0 instead the listing keep fetching until key3 and return the NextMarker key2.

### DIFF
--- a/lib/algos/list/delimiterCurrent.ts
+++ b/lib/algos/list/delimiterCurrent.ts
@@ -76,6 +76,10 @@ class DelimiterCurrent extends DelimiterMaster {
     }
 
     addContents(key, value) {
+        if (this._reachedMaxKeys()) {
+            return FILTER_END;
+        }
+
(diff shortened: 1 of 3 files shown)


jonathan-gramain (approved):

francoisferrand (approved):

nicolas2bert:
@bert-e approve

PR #2161, opened 2023-08-11
bf(ARSN-361): Handle infinite versionID correctly on version specific PUT for file MD backend

tmacro:
When passing the infinite versionId (`0099999999999999999999RG001`) the file backend fails to update the master key if the the infinite version is also the latest version.

### DIFF
--- a/lib/versioning/VersioningRequestProcessor.ts
+++ b/lib/versioning/VersioningRequestProcessor.ts
@@ -1,6 +1,6 @@
 import errors, { ArsenalError } from '../errors';
 import { Version } from './Version';
-import { generateVersionId as genVID } from './VersionID';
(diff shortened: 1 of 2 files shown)


jonathan-gramain (commented):
Can you add a test for this new behavior? Even though it's a test backend, we should ensure we don't introduce regressions. For example a simple test inspired by this one would do: https://github.com/scality/Arsenal/blob/development/7.10/tests/unit/versioning/VersioningRequestProcessor.spec.js#L153

jonathan-gramain (approved):

rachedbenmustapha (approved):

tmacro:
closing this PR since it is no longer needed

PR #2162, opened 2023-08-22
Improvement/arsn 362 implicit deny

KazToozs:
Adds ImplicitDeny logic to policy checks, where an ImplicitDeny will be sent back in case no policy validates an action, but does not explicitly Deny it either, allowing for bucket policies and other authorization mechanisms to grant permission.

Part of the bucket policy redo epic: https://scality.atlassian.net/jira/software/c/projects/OS/boards/214?modal=detail&selectedIssue=S3C-7756

### DIFF
--- a/lib/policyEvaluator/evaluator.ts
+++ b/lib/policyEvaluator/evaluator.ts
@@ -310,6 +310,7 @@ export function evaluatePolicy(
 }
 
 /**
+ * @deprecated Upgrade to evaluateAllPoliciesV2
  * Evaluate whether a request is permitted under a policy.
  * @param requestContext - Info necessary to
  * evaluate permission
@@ -325,6 +326,16 @@ export function evaluateAllPolicies(
     allPolicies: any[],
     log: Logger,
 ): string {
+    return evaluateAllPoliciesV2(requestContext, allPolicies, log).verdict;
+}
+export function evaluateAllPoliciesV2(
+    requestContext: RequestContext,
+    allPolicies: any[],
+    log: Logger,
+): {
+    verdict: string;
+    isImplicit: boolean;
+} {
     log.trace('evaluating all policies');
     let allow = false;
     let allowWithTagCondition = false;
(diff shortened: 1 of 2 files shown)


lib/policyEvaluator/evaluator.ts
```diff
@@ -324,7 +324,10 @@ export function evaluateAllPolicies(
     requestContext: RequestContext,
```
francoisferrand: to avoid breaking compatibility, may be better not to touch the prototype of this function : i.e. make the change to the body in a function with a new name, and transform the older function in a simpler wrapper
→ this way existing code using it is not affected, it can decide when they "upgrade" to the new API, and we don't introduce coupling or limit upgrade paths
→ ideally we could even mark the 'old' method as deprecated, though I am not sure what is the impact currently: depending on build flags it could probably fail the build...

Something like this:

```ts
/* @deprecated Upgrade to evaluateAllPoliciesV2
 */
export function evaluateAllPolicies(
    requestContext: RequestContext,
    allPolicies: any[],
    log: Logger,
): string {
    return evaluateAllPoliciesV2(requestContext, allPolicies, log).verdict;
}

export function evaluateAllPoliciesV2(
    requestContext: RequestContext,
    allPolicies: any[],
    log: Logger,
): {
    verdict: string;
    isImplicit: boolean;
} {
   ...
}
```

anurag4DSB (commented):
Code LGTM

The PR needs to target development/7.70
We need a test to ensure the v1 method only returns verdict for future change prevention

KazToozs:
@anurag4DSB  the tests you're speaking of already exist just above
https://github.com/scality/Arsenal/pull/2162/files#diff-f13d18d50beb84716ae3ca3972ea5222c49ac740e32e4dbac083fba8cb51fda7R1432

Is this what you had in mind?

anurag4DSB (approved):
Its worth keeping the v1/v2 consistent across cloudserver, arsenal and vault.

anurag4DSB (changes requested):
We need to aim development/7.10 as per the fix versions specified in the epic: https://scality.atlassian.net/browse/S3C-7756

KazToozs:
closing and opening new PR to facilitate rebase to 7.10

https://github.com/scality/Arsenal/pull/2164

PR #2177, opened 2023-10-05
ARSN-370: fix memory leak in MongoDBReadStreams

ghost:
The MongoDBReadStreams are not properly destroyed in both the Bucket V1 and V0 cases. In the V1 case, only the pipe-ed stream, the Transform one, is cleaned. In the V0 case, we directly call the callback without properly cleaning the stream, leaving open, in both cases, the mongodb cursors, that in turn affect the mongos memory consumption.

The fix was tested end2end:

Before (with k8s memory limits, otherwise it uses up to 100% of the cluster RAM):
image
Causing heap overflows in cloudserver pods and high MongoS memory usage due to open cursors.

After:
image

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -1641,9 +1641,11 @@ class MongoClientInterface {
         const c = this.getCollection(bucketName);
         const getLatestVersion = this.getLatestVersion;
         let stream;
+        let baseStream;
         if (!params.secondaryStreamParams) {
             // listing masters only (DelimiterMaster)
             stream = new MongoReadStream(c, params.mainStreamParams, params.mongifiedSearch);
+            baseStream = stream;
             if (vFormat === BUCKET_VERSIONS.v1) {
                 /**
                  * When listing masters only in v1 we can't just skip PHD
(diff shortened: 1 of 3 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1736,10 +1743,14 @@ class MongoClientInterface {
                 };
                 log.error(
                     'internalListObjectV1: error listing objects', logObj);
+                // call explicitly the destroy method to clean the mongodb cursor
+                stream.destroy();
```
francoisferrand, line 1758: is this call not duplicated: e.g. will `end` not be called on `error` as well?

tests/functional/metadata/mongodb/listObject.spec.js
```diff
@@ -477,6 +479,25 @@ describe('MongoClientInterface::metadata.listObject', () => {
                     }),
                 ], done);
             });
+
```
francoisferrand, line 482: can we add a test for the "error" case as well?

francoisferrand (approved):

nicolas2bert (approved):

benzekrimaha (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: No, and actually it's not even called in the v1 case due to the piping. See https://github.com/scality/Arsenal/pull/2177/commits/4aa8b5cc6e82d4f69773e5d3cedcc60b9a22f39c

if I remove the line, the test would fail because the stream was not properly cleaned.

tests/functional/metadata/mongodb/listObject.spec.js
ghost: Done in https://github.com/scality/Arsenal/pull/2177/commits/4aa8b5cc6e82d4f69773e5d3cedcc60b9a22f39c

PR #2179, opened 2023-10-12
Bugfix/arsn 371 useunifiedtopology

ghost:
Add an option to support and enable this feature, as it is not clear, depending on the driver version, if this flag defaults to true or not.

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -91,7 +91,7 @@ class MongoClientInterface {
     constructor(params) {
         const { replicaSetHosts, writeConcern, replicaSet, readPreference, path,
             database, logger, replicationGroupId, authCredentials,
-            isLocationTransient, shardCollections } = params;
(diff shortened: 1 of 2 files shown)


francoisferrand:
not sure what this brings: we want to use this mode, it should not be disabled, and is the default in the version of arsenal where you add it...

ghost:
@francoisferrand it is not clearly documented if the option is set to `true` by default, if the option is still useful, or not. The only hint is here: https://www.mongodb.com/docs/drivers/node/current/whats-new/#mongoclientoptions-interface

In some other places like here https://www.mongodb.com/docs/atlas/resilient-application/, they still use it.
So it won't hurt making sure this option is set to true when we explicitly want to have it.

benzekrimaha (approved):
LGTM

francoisferrand:
The link you mention indicates it is not _needed_ anymore (which is a hint for "should not be used" or "is useless"); and looking at the (official) documentation [1] it is not present anymore, since 5.0 at least...

[1] https://mongodb.github.io/node-mongodb-native/5.0/interfaces/MongoClientOptions.html

PR #2164, opened 2023-09-11
Improvement/arsn 362 imp deny

KazToozs:
Opened after closed PR [here](https://github.com/scality/Arsenal/pull/2162)

Adds ImplicitDeny logic to policy checks, where an ImplicitDeny will be sent back in case no policy validates an action, but does not explicitly Deny it either, allowing for bucket policies and other authorization mechanisms to grant permission.

Part of the bucket policy redo epic: https://scality.atlassian.net/jira/software/c/projects/OS/boards/214?modal=detail&selectedIssue=S3C-7756

Green build in Vault and CS:
https://github.com/scality/Vault/actions/runs/6151692989/job/16692302075?pr=2135
https://github.com/scality/cloudserver/actions/runs/6157396066/job/16708032893?pr=5322

PR #2181, opened 2023-10-30
Improvement/arsn 362 implicit deny

benzekrimaha:
Opened after closed PR [here](https://github.com/scality/Arsenal/pull/2164)

Adds ImplicitDeny logic to policy checks, where an ImplicitDeny will be sent back in case no policy validates an action, but does not explicitly Deny it either, allowing for bucket policies and other authorization mechanisms to grant permission.

Part of the bucket policy redo epic: https://scality.atlassian.net/jira/software/c/projects/OS/boards/214?selectedIssue=S3C-7756

Green build in Vault and CS:
https://github.com/scality/Vault/actions/runs/6694175632/job/18186886853?pr=2135
https://github.com/scality/cloudserver/actions/runs/6693681174/job/18185346914?pr=5322

*Would appreciate reviews on Integration branches as well* 
*Arsenal version has been bumped lastly*

### DIFF
--- a/lib/policyEvaluator/evaluator.ts
+++ b/lib/policyEvaluator/evaluator.ts
@@ -310,6 +310,7 @@ export function evaluatePolicy(
 }
 
 /**
+ * @deprecated Upgrade to evaluateAllPoliciesNew
  * Evaluate whether a request is permitted under a policy.
  * @param requestContext - Info necessary to
  * evaluate permission
@@ -325,6 +326,16 @@ export function evaluateAllPolicies(
     allPolicies: any[],
     log: Logger,
 ): string {
+    return evaluateAllPoliciesNew(requestContext, allPolicies, log).verdict;
+}
+export function evaluateAllPoliciesNew(
+    requestContext: RequestContext,
+    allPolicies: any[],
+    log: Logger,
+): {
+    verdict: string;
+    isImplicit: boolean;
+} {
     log.trace('evaluating all policies');
     let allow = false;
     let allowWithTagCondition = false;
@@ -333,7 +344,10 @@ export function evaluateAllPolicies(
         const singlePolicyVerdict = evaluatePolicy(requestContext, allPolicies[i], log);
         // If there is any Deny, just return Deny
         if (singlePolicyVerdict === 'Deny') {
-            return 'Deny';
+            return {
+                verdict: 'Deny',
+                isImplicit: false,
+            };
         }
         if (singlePolicyVerdict === 'Allow') {
             allow = true;
@@ -344,6 +358,7 @@ export function evaluateAllPolicies(
         } // else 'Neutral'
     }
     let verdict;
+    let isImplicit = false;
     if (allow) {
         if (denyWithTagCondition) {
             verdict = 'NeedTagConditionEval';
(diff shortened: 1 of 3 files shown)


tests/unit/policyEvaluator.spec.js
```diff
@@ -1422,33 +1423,42 @@ describe('policyEvaluator', () => {
...
+                isImplicit: false,
+            });
         });
 
         it('should deny access if request action is not in any policy', () => {
             requestContext = new RequestContext({}, {},
                 'notVeryPrivate', undefined,
                 undefined, undefined, 'bucketDelete', 's3');
             requestContext.setRequesterInfo({});
-            const result = evaluateAllPolicies(requestContext,
```
KillianG, line 1436: `evaluateAllPolicies` is deprecated but should still work 'til its implemented, maybe it's worth to keep the tests for this functions ?

benzekrimaha:
ping

KillianG (approved):

benzekrimaha:
@bert-e create_pull_requests

welansari (approved):

benzekrimaha:
ping

benzekrimaha:
ping

lib/policyEvaluator/evaluator.ts
```diff
@@ -325,6 +326,16 @@ export function evaluateAllPolicies(
...
     log: Logger,
 ): string {
+    return standardEvaluateAllPolicies(requestContext, allPolicies, log).verdict;
+}
+export function standardEvaluateAllPolicies(
+    requestContext: RequestContext,
+    allPolicies: any[],
+    log: Logger,
+): {
+    verdict: string;
```
jonathan-gramain, line 336: After a discussion about this some time ago with William maybe in a PR (I can't find the ref) it was about returning a single state string like `ImplicitDeny` or `ExplicitDeny` rather than an object with a boolean, to make it simpler and more visual, forcing to handle all states explicitly. There could have been reasons to stick to using a boolean maybe for backward-compatibility, since I'm not aware of this just want to make sure it was a conscious decision.

benzekrimaha: I indeed found the comment you are refering to here : https://github.com/scality/citadel/pull/165#discussion_r1276904304
I'm looping @williamlardier in for more insight on this

ghost: Yes, this is for backward compatibility, that we use the boolean, given that we want to keep both flows. With a boolean, we can minimize the changes applied to our code base.

But our last discussion was about the request contexts responses (i.e., in Vault), not the internal logic. The request context API must be updated as stated in the design, but the evaluation logic _can_ stay like that, unless we see benefits of having a uniform authz result starting at Arsenal level (maybe by creating a new typescript type or enum). Then in cloudserver we can check the authz result using string comparisons, and update the functions here (and associated tests).

As discussed with Maha already, if we want to deviate from the design (with good reasons), we must update the design. Here, the current code looks compatible with it, but I think having a unified implementation would be better indeed. It will require creating the new authz types in arsenal, update the functions, and then reuse them in Vault/Cloudserver.

jonathan-gramain (approved):

anurag4DSB (approved):

ghost (approved):

benzekrimaha:
@bert-e approve

PR #2184, opened 2023-11-13
Improvement/arsn 372 remove mpu location from listing result to speedup result

KillianG:


lib/storage/metadata/mongoclient/readStream.js
```diff
@@ -85,7 +85,8 @@ class MongoReadStream extends Readable {
             Object.assign(query, searchOptions);
         }
 
-        this._cursor = c.find(query).sort({
+        const projection = { 'value.location': 0 };
+        this._cursor = c.find(query, { projection }).sort({
```
francoisferrand, line 89: LGTM, but would be best to add a functional test case : to verify we do indeed drop this attribute form listing
(maybe in https://github.com/scality/Arsenal/blob/development/8.1/tests/functional/metadata/mongodb/listObject.spec.js ?)

KazToozs (approved):

benzekrimaha (approved):

tests/functional/metadata/mongodb/listObject.spec.js
```diff
@@ -83,6 +113,7 @@ describe('MongoClientInterface::metadata.listObject', () => {
                         readPreference: 'primary',
                         database: DB_NAME,
                     },
+                    customListingParser,
```
francoisferrand: we don't want to use the custom parser always : otherwise, we do not cover the standard/default parser anymore... (which can be seen in the codecov report: https://app.codecov.io/github/scality/Arsenal/commit/52875fea66ecf04a76261592a4e825b833ce8858)

Also, (except for the "location" part), these tests validate that indeed the default parser does the proper mapping (which do not make sense to test with your change: since now we test that the custom parser does the proper mapping).

→ it should be more appropriate to create a separate test case, with a very simple custom parser (e.g. return parsed json object, no mapping!) : since listObject seems quite agnostic to the actual content of the parsed objects...
→ if that does not work, another approach may be to write a custom parser which "wraps" the default parser (`_parseListEntries`) after asserting the location field is not present. This could be used on all tests, but we would not verify that indeed this function is used as the default parser: hence it would be preferable to use the "separate test case" approach

tests/functional/metadata/mongodb/listObject.spec.js
```diff
@@ -523,6 +536,23 @@ describe('MongoClientInterface::metadata.listObject', () => {
                     return done();
                 });
             });
+
+            it('Should not include location in listing result and use custom listing parser', done => {
+                const defaultListingParser = metadata._listingParser;
```
francoisferrand: Instead of modifying the parser of the object (and having to reset it), you should create a new client, to keep the tests independents (and possibly allow running them in parallel):

```suggestion
                const metadata = new MetadataWrapper(IMPL_NAME, { opts..., customListingParser }, null, logger);
                metadata.setup(done);
```

tests/functional/metadata/mongodb/listObject.spec.js
```diff
@@ -523,6 +536,37 @@ describe('MongoClientInterface::metadata.listObject', () => {
...
+                    const params = {
+                        listingType: 'DelimiterMaster',
+                        maxKeys: 100,
+                    };
+                    return md.listObject(BUCKET_NAME, params, logger, (err, data) => {
+                        assert.ifError(err);
+                        assert.strictEqual(data.Contents.length, 3);
+                        assert.strictEqual(data.Contents[0].key, 'pfx1-test-object');
+                        assert.strictEqual(data.Contents[0].location, undefined);
+                        assert(parserSpy.calledOnce);
```
francoisferrand: nit: not sure the "semantics" of customListingParser is that it will be called only exactly once, so checking it here may be overly strict... (esp. for functional tests)

francoisferrand (approved):

PR #2185, opened 2023-11-20
ARSN-373 bump hdclient to 1.1.7

benoit-a:


ghost (approved):
You can also add a comit to bump arsenal in the same PR, if you want to release the fix in upper projects

francoisferrand (approved):

PR #2186, opened 2023-11-22
objectMD: add transitionTime when starting a transition

francoisferrand:
Store transition time when marking the object as ‘transition in progress’.
This is needed to compute lifecycle lateness & duration metrics.

Issue: ARSN-374

### DIFF
--- a/lib/models/ObjectMD.ts
+++ b/lib/models/ObjectMD.ts
@@ -58,6 +58,7 @@ export type ObjectMDData = {
     'x-amz-server-side-encryption-customer-algorithm': string;
     'x-amz-website-redirect-location': string;
     'x-amz-scal-transition-in-progress'?: boolean;
+    'x-amz-scal-transition-time'?: string;
     azureInfo?: any;
     acl: ACL;
     key: string;
@@ -649,10 +650,24 @@ export default class ObjectMD {
      * Set metadata transition in progress value
      *
      * @param inProgress - True if transition is in progress, false otherwise
+     * @param transitionTime - Date when the transition started
      * @return itself
      */
-    setTransitionInProgress(inProgress: boolean) {
+    setTransitionInProgress(inProgress: false): this
+    setTransitionInProgress(inProgress: true, transitionTime: Date|string|number): this
+    setTransitionInProgress(inProgress: boolean, transitionTime?: Date|string|number) {
         this._data['x-amz-scal-transition-in-progress'] = inProgress;
+        if (!inProgress || !transitionTime) {
+            delete this._data['x-amz-scal-transition-time'];
+        } else {
+            if (typeof transitionTime === 'number') {
+                transitionTime = new Date(transitionTime);
+            }
+            if (transitionTime instanceof Date) {
+                transitionTime = transitionTime.toISOString();
+            }
+            this._data['x-amz-scal-transition-time'] = transitionTime;
+        }
         return this;
     }
 
@@ -663,6 +678,14 @@ export default class ObjectMD {
      */
     getTransitionInProgress() {
         return this._data['x-amz-scal-transition-in-progress'];
+    }
+
+    /**
+     * Gets the transition time of the object.
+     * @returns The transition time of the object.
+     */
+    getTransitionTime() {
+        return this._data['x-amz-scal-transition-time'];
     }
 
     /**
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Initialize CodeQL
         uses: github/codeql-action/init@v2
--- a/.github/workflows/dependency-review.yaml
+++ b/.github/workflows/dependency-review.yaml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: 'Checkout Repository'
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: 'Dependency Review'
         uses: actions/dependency-review-action@v3
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -25,8 +25,8 @@ jobs:
           - 6379:6379
     steps:
     - name: Checkout
-      uses: actions/checkout@v2
-    - uses: actions/setup-node@v2
+      uses: actions/checkout@v4
+    - uses: actions/setup-node@v4
       with:
         node-version: '16'
         cache: 'yarn'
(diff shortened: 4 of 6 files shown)


benzekrimaha (approved):
LGTM

lib/models/ObjectMD.ts
```diff
@@ -58,6 +58,7 @@ export type ObjectMDData = {
     'x-amz-server-side-encryption-customer-algorithm': string;
     'x-amz-website-redirect-location': string;
     'x-amz-scal-transition-in-progress'?: boolean;
+    'x-amz-scal-transition-time'?: string;
```
welansari, line 61: (random line)
Should we update the `md-model-version` here https://github.com/scality/Arsenal/blob/4b642cf8b4de5a2d911a658884e3c2effbe79443/lib/constants.ts#L44

francoisferrand: that is a good point, but i actually don't know:
* the field itself is "kind of" optional, so not really needed
* it seems many such "optional" fields have been added without bumping the model version (which could also have been just a mistake)
* but it may be better indeed to allow tracking more precisely (though it could backfire, if we end having huge switch case to handle all versions)...

@nicolas2bert @jonathan-gramain Do you know what is the practical purpose/need of this field, and what is the policy/rationale for updating (or not) the field?

jonathan-gramain: To me it looks like it allows to handle incompatible changes in the metadata, to allow handling both older and newer formats correctly by checking the version first.

My opinion is that if the change can be handled gracefully and robustly without having to bump the version number (like adding an optional new field), then it's not necessary. Some other devs can have a different opinion on this.

lib/models/ObjectMD.ts
```diff
@@ -649,10 +650,24 @@ export default class ObjectMD {
...
+        if (!inProgress || !transitionTime) {
+            delete this._data['x-amz-scal-transition-time'];
+        } else {
+            if (typeof transitionTime === 'number') {
+                transitionTime = new Date(transitionTime);
+            }
+            if (transitionTime instanceof Date) {
+                transitionTime = transitionTime.toISOString();
+            }
+            this._data['x-amz-scal-transition-time'] = transitionTime;
```
nicolas2bert, line 669: Is no validation needed in case `transitionTime` is a string? (e.g. what if the string is not in an ISO format?)

lib/models/ObjectMD.ts
```diff
@@ -649,10 +650,24 @@ export default class ObjectMD {
      * Set metadata transition in progress value
      *
      * @param inProgress - True if transition is in progress, false otherwise
+     * @param transitionTime - Date when the transition started
      * @return itself
      */
-    setTransitionInProgress(inProgress: boolean) {
+    setTransitionInProgress(inProgress: false): this
+    setTransitionInProgress(inProgress: true, transitionTime: Date|string|number): this
+    setTransitionInProgress(inProgress: boolean, transitionTime?: Date|string|number) {
```
nicolas2bert, line 658: At compile time, (I think) overloaded function declarations do not raise an error.
Hence, do we need to handle the case where `inProgress` is `true` and `transitionTime` is `undefined`?

lib/models/ObjectMD.ts
francoisferrand: works for me, thanks!

lib/models/ObjectMD.ts
francoisferrand: As we can see on [Typescript playground](https://www.typescriptlang.org/play?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBBCcQQ0MwAUVWDEQC0gADVHBQGNJ3KAALygABEAAkYNAMOB4GRonzFBTBPB3AERcwABS0+mM9ysygc9wASnJjEEwhw+hoAG9FCUSgB9bh1Bj5MScPDBFQ4ZoeMheRDi5zA9g0L5uMzMdgMZiQfiwaDqRIc+B6ERePl0hlMwR8pLYvngByoADkwgAcpAJHygioAPJfMjvQQRrSQI6oRwKVzSPCgJzenTDLD6FAIRDQ4SXQMScXREQGDjYhZMVgcHhBOuoHBkMxyeqNACSnFobFY+kESoy+-gh8NDuSDGgdZ1DqGvxXa438ganF3Z4vx9PB4JBTQMh2HrVB10pN9GhwKFIAdAAROpIAAHypShOGYRDOHYRBpG+e8BCfRsX3ArcPz3P8jxPMjz3-B1gTlDxOH4UDXxIqDkAAfjghDkNA4YMKwnCyB1UBzUtfCAhtO0AG081WUoGQAL1KQRamgUpmOI99SgyUoCHIw08wAXR5RMv3-ABuC1LQRUAlQAQl-aij1ARDENAOyNM3d82MgYTRLEy1uBgRtbiECTbWbGS5MU5TVPUsCvMadToKMyyxIAX3BW8UH8gKRFSJVUHaIlPIgzgfJ5bleTzTDsO+PM-KsvKxNK1joJMzgS1AeCFEKhKyp8nU0ua9KmoCmy+pY7z2oyKkzAsIket8kSxuawCpsg9reVa6bkACdBt2ICtiF4tClSG1bLVGtbH3C6TZPkxAlJUm94o2zhkuQIyTJ2zbkGGkprstJxUHxTI1EESzRtG6VGipNAOq640qXOxRUACZ9+pIz99OPa9sr1dHMfenGnMNPrgP4TqJG6hDzsJjHCKx99Se-Cn6wARgAdgABg5gAWABmABODmBYZ4nNJ3Ki2dAyn+Q54XedKPnSgAJh5nAOYANnyQWufyRWAjF-mAC0+T1NHGdXZnpbMijhOalQaTINgyAdOkiFgVtQA59WxDIENh34aR-S0QZQB5q3JcS0j7fJuXlstZ3XfFB0KkDrCURoEr2lAPNE7zKd+gjhhDQoZhOG1SwMAIRwGRC7Fc6IfP8brPMgiAA), seem to report the expected errors:

```
t.setTransitionInProgress(false)
t.setTransitionInProgress(true, new Date())
t.setTransitionInProgress(true, 1701439114)
t.setTransitionInProgress(true, "1970-01-20T16:37:19.114Z")

t.setTransitionInProgress()        // Error: Expected 1-2 arguments, but got 0
t.setTransitionInProgress(true)    // Error: Argument of type 'true' is not assignable to parameter of type 'false'.

```

lib/models/ObjectMD.ts
francoisferrand: Not sure we need the validation: this is not a user-provided field, and only used by the code...

lib/models/ObjectMD.ts
nicolas2bert: My bad :)

nicolas2bert (approved):

PR #2192, opened 2023-12-13
ARSN-377 correctly handle null keys with common prefix

jonathan-gramain:
When encountering a null key, check for its common prefix before including it in either the Versions array or CommonPrefixes array, instead of always including it in the Versions array.

This commit refactors how `DelimiterVersions` works with null keys slightly: the null key is now inserted at its correct ordered position by the top-level `filter()` method, and the state machine handlers only have to deal with sorted versions. Previously the individual handlers would have to deal with the null key positioning themselves resulting in more complex state management.

### DIFF
--- a/tests/unit/algos/list/delimiterVersions.spec.js
+++ b/tests/unit/algos/list/delimiterVersions.spec.js
@@ -45,6 +45,8 @@ const valuePHD = '{"isPHD":"true","versionId":"1234567890abcdefg"}';
 const fooDM = '{"hello":"world","isDeleteMarker":"true","versionId":"foo"}';
 const barDM = '{"hello":"world","isDeleteMarker":"true","versionId":"bar"}';
 const quxDM = '{"hello":"world","isDeleteMarker":"true","versionId":"qux"}';
+const nullVersionMD = '{"hello":"world","isNull":true,"isNull2":true,"versionId":"vnull"}';
+const nullDMMD = '{"hello":"world","isNull":true,"isNull2":true,"isDeleteMarker":"true","versionId":"bar"}';
 const dataVersioned = {
     v0: [
         { key: 'Pâtisserie=中文-español-English', value: bar },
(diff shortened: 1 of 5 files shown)


jonathan-gramain:
ping

tests/unit/algos/list/delimiterVersions.spec.js
```diff
@@ -154,13 +167,11 @@ const receivedData = [
...
-    { key: 'notes/summer/4.txt', value: valueDeleteMarker, versionId: 'qux' },
-    { key: 'notes/summer/444.txt',
-        value: valueDeleteMarker, versionId: 'null' },
-    { key: 'notes/summer/44444.txt',
-        value: valueDeleteMarker, versionId: 'null' },
+    { key: 'notes/summer/4.txt', value: barDM, versionId: 'bar' },
+    { key: 'notes/summer/4.txt', value: fooDM, versionId: 'foo' },
+    { key: 'notes/summer/4.txt', value: quxDM, versionId: 'qux' },
+    { key: 'notes/summer/444.txt', value: fooDM, versionId: 'foo' },
+    { key: 'notes/summer/44444.txt', value: fooDM, versionId: 'foo' },
```
KillianG, line 174: I feel like these 2 arrays can be simplified into only one, as `receivedData` is most likely just using `dataVersioned` array values and adds `versionId` field.. What do you think ?

KillianG:
Have you tried running zenko tests and cloudserver tests to be sure we don't break anything with that change ?

jonathan-gramain:
@KillianG 

> Have you tried running zenko tests and cloudserver tests to be sure we don't break anything with that change ?

I haven't, and it's a good idea, I will do. I ran tests on Metadata to make sure it fixes the case described in the V1 test plan, but since the risk of regression is not negligible I think running the automated test suites beforehand (including S3C Integration) is useful.

jonathan-gramain:
@KillianG When running the Metadata and Cloudserver tests I found an issue with the `DelimiterNonCurrent` class that derives from `DelimiterVersions` and had to be modified. I fixed it, added a new unit test and re-ran the Metadata/Cloudserver tests, should be good now.

tests/unit/algos/list/delimiterVersions.spec.js
jonathan-gramain: It's actually not an exact 1-1, it depends on the keys, for example you can have a duplicate master/version key, or not (depending if it's V0 or V1 or if it's a current/noncurrent version, or delete marker). Hence I think it's best to keep them separate. There are actually 3 different arrays if you account for the V0 and V1 data.

nicolas2bert (approved):

KillianG (approved):

PR #2198, opened 2023-12-18
ARSN-284 fix and refactor Delimiter + DelimiterMaster

jonathan-gramain:
**Note for reviews**:

The original changes have already been reviewed, the main one being https://github.com/scality/Arsenal/pull/2054, so no need to review this logic again (but it can be useful to have a glimpse for context).

However, I had to adapt the 7.70 integration and 8.1 branches.

On **7.70** branch:
- `DelimiterVersions` was simplified, removing the duplicate code from `Delimiter`, and also had typescript check errors that needed to be addressed
 - `DelimiterCurrent` which inherits from `DelimiterMaster` that was significantly changed
 - `DelimiterNonCurrent` and `DelimiterOrphanDeleteMarker` classes which inherit from `DelimiterVersions` that had a slight change as explained above

Here are the changes added on top of the 7.70 integration branch, **please review them**: https://github.com/scality/Arsenal/compare/bfa51b926c6fcf67ed5df2776fd08abcbeea1944...w/7.70/bugfix/ARSN-379-cherry-pick-ARSN-284-and-ARSN-293

On **8.1** branch: Adapt skipping delete markers in DelimiterMaster

With the MongoDB implementation there may be delete markers in the
masters prefix to go through.

Replace the original implementation for this by a new implementation
compatible with the latest DelimiterMaster changes.

Note: changed the returned value from FILTER_SKIP to FILTER_ACCEPT:
this is the correct logic as there is no range to skip, only the key
shouldn't be added to the results.

Here are the changes added on top of the 8.1 integration branch, **please review them**:

https://github.com/scality/Arsenal/compare/1d06bbb59f378d929c88f6672cc4d0924c81d441...w/8.1/bugfix/ARSN-379-cherry-pick-ARSN-284-and-ARSN-293

**List of cherry-picks**:

This bugfix and refactor work has been re-integrated in the code by cherry-picking the following commits:

- f62c3d22 ARSN-252 - listing bug in DelimisterMaster
- 87b060f2 ARSN-269 - listing bug in versioned bucket edge cases.
- 4f0a8468 ARSN-284 [cleanup] remove unused test dependency
- 7b648962 ARSN-284 [rf] delimiterVersions.addCommonPrefix()
- 4d7eaee0 ARSN-284 fix and refactor Delimiter + DelimiterMaster
- 1c07618b ARSN-284 [doc] add state charts
- fbb62ef1 bugfix: ARSN-293 DelimiterMaster: default to vFormat=v0
- 6e5d8d14 bugfix: ARSN-294 use CommonPrefix for NextMarker

**Original changes description**:

Large refactor of Delimiter and DelimiterMaster classes to typescript, that fixes most known issues with the previous implementation.

The new implementation uses explicit states to manage various conditions, instead of relying on a bunch of internal variable values and maintaining their state. It allows a more robust code flow and fixes issues related to prefix skipping that were hard to fix by keeping the overall logic of the previous implementation.

This refactor brings the following bug fixes and enhancements:

- prefixes with delete markers and non-deleted objects are now always included in CommonPrefixes (S3C-7248)

- no more duplication of internal range listings when doing skip-scan over prefixes (discovered when analyzing regressions for S3C-4682)

- the skip-scan mecanism for prefixes and versions is no more disturbed by delete markers and PHD keys (S3C-2930)

- NextMarker is now always set to a valid, listed or listable key (that may still be hidden under a CommonPrefix), no more manipulation of next marker to avoid corner-cases with keys ending with a prefix (S3C-4682 and S3C-7274)

- deleting a delete marker immediately allows the new current version to be visible in the listing (S3C-7272)

- Expecting lower CPU usage overall, as the number of checks to do in each state is reduced (may help to reduce the load and reduce impact of cases such as S3C-3946)

- Uses typescript to allow more sanity checks


jonathan-gramain:
ping

nicolas2bert (approved):
The adaptation in `development/7.70` related to lifecycle listing looks good to me.

KillianG (approved):

PR #2202, opened 2023-12-29
ARSN-381 RPC command system between cluster workers

jonathan-gramain:
When using the cluster module, new processes are forked and are dispatched workloads, usually HTTP requests. The ClusterRPC module implements a RPC system to send commands to all cluster worker processes at once from any particular worker, and retrieve their individual command results, like a distributed map operation.

The existing cluster IPC channel is setup from the primary to each worker, but not between workers, so there has to be a hop by the primary.

How a command is treated:

- a worker sends a command message to the primary

- the primary then forwards that command to each existing worker (including the requestor)

- each worker then executes the command and returns a result or an error

- the primary gathers all workers results into an array

- finally, the primary dispatches the results array to the original requesting worker callback

The original use of this feature is in Metadata DBD (bucketd) to implement a global cache refresh across worker processes.

### DIFF
--- a/lib/clustering/ClusterRPC.ts
+++ b/lib/clustering/ClusterRPC.ts
@@ -1,0 +1,514 @@
+import cluster, { Worker } from 'cluster';
+import * as werelogs from 'werelogs';
+
+import { default as errors } from '../../lib/errors';
+
+const rpcLogger = new werelogs.Logger('ClusterRPC');
+
+/**
+ * Remote procedure calls support between cluster workers.
+ *
+ * When using the cluster module, new processes are forked and are
+ * dispatched workloads, usually HTTP requests. The ClusterRPC module
+ * implements a RPC system to send commands to all cluster worker
+ * processes at once from any particular worker, and retrieve their
+ * individual command results, like a distributed map operation.
+ *
+ * The existing nodejs cluster IPC channel is setup from the primary
+ * to each worker, but not between workers, so there has to be a hop
+ * by the primary.
+ *
+ * How a command is treated:
+ *
+ * - a worker sends a command message to the primary
+ *
+ * - the primary then forwards that command to each existing worker
+ *   (including the requestor)
+ *
+ * - each worker then executes the command and returns a result or an
+ *   error
+ *
+ * - the primary gathers all workers results into an array
+ *
+ * - finally, the primary dispatches the results array to the original
+ *   requesting worker
+ *
+ *
+ * Limitations:
+ *
+ * - The command payload must be serializable, which means that:
+ *   - it should not contain circular references
+ *   - it should be of a reasonable size to be sent in a single RPC message
+ *
+ * - The "toWorkers" parameter of value "*" targets the set of workers
+ *   that are available at the time the command is dispatched. Any new
+ *   worker spawned after the command has been dispatched for
+ *   processing, but before the command completes, don't execute
+ *   the command and hence are not part of the results array.
+ *
+ *
+ * To set it up:
+ *
+ * - On the primary:
+ *     if (cluster.isPrimary) {
+ *         setupRPCPrimary();
+ *     }
+ *
+ * - On the workers:
+ *     if (!cluster.isPrimary) {
+ *         setupRPCWorker({
+ *             handler1: (payload: object, uids: string, callback: HandlerCallback) => void,
+ *             handler2: ...
+ *         });
+ *     }
+ *     Handler functions will be passed the command payload, request
+ *     serialized uids, and must call the callback when the worker is done
+ *     processing the command:
+ *         callback(error: Error | null | undefined, result?: any)
+ *
+ * When this setup is done, any worker can start sending commands by calling
+ * the async function sendWorkerCommand().
+ */
+
+// exported types
+
+export type ResultObject = {
+    error: Error | null;
+    result: any;
+};
+
+/**
+ * saved Promise for sendWorkerCommand
+ */
+export type CommandPromise = {
+    resolve: (results?: ResultObject[]) => void;
+    reject: (error: Error) => void;
+    timeout: NodeJS.Timer | null;
+};
+export type HandlerCallback = (error: Error | null | undefined, result?: any) => void;
+export type HandlerFunction = (payload: object, uids: string, callback: HandlerCallback) => void;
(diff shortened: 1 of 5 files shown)


lib/clustering/ClusterRPC.ts
```diff
@@ -0,0 +1,478 @@
...
+export type CommandCallback = (error: Error | null, results?: ResultObject[]) => void;
+
+export type HandlerCallback = (error: Error | null | undefined, result?: any) => void;
+export type HandlerFunction = (payload: object, uids: string, callback: HandlerCallback) => void;
+export type HandlersMap = {
+    [index: string]: HandlerFunction;
+};
+
+// private types
+
```
tmacro, line 95: You could use a generic type instead of separate types for each message
```suggestion
type RPCMessage<T extends string, P> = {
    type: T;
    uids: string;
    payload: P;
}

type RPCCommandMessage = RPCMessage<'cluster-rpc:command', any> & {
    toWorkers: string;
    toHandler: string;
}

type MarshalledResultObject = {
    error: string | null;
    result: any;
};

type RPCCommandResultMessage = RPCMessage<'cluster-rpc:commandResult', { results: MarshalledResultObject }>;

type RPCCommandErrorMessage = RPCMessage<'cluster-rpc:commandError', { error: string }>;
```

tmacro (approved):

lib/clustering/ClusterRPC.ts
```diff
@@ -0,0 +1,469 @@
...
+    rpcLogger.info('sending command', { uids, payload });
+    uidsToCallback[uids] = callback;
+    const message: RPCCommandMessage = {
+        type: 'cluster-rpc:command',
+        toWorkers,
+        toHandler,
+        uids,
+        payload,
+    };
+    return process.send?.(message);
```
rachedbenmustapha: If `process.send` is undefined no worker will receive the call, so it might be worth failing early at init if the IPC channel is not available.

lib/clustering/ClusterRPC.ts
```diff
@@ -0,0 +1,469 @@
...
+ *
+ * - a worker sends a command message to the primary
+ *
+ * - the primary then forwards that command to each existing worker
+ *   (including the requestor)
+ *
+ * - each worker then executes the command and returns a result or an
+ *   error
+ *
+ * - the primary gathers all workers results into an array
```
rachedbenmustapha, line 31: I don't see a way to set a deadline or timeout. If one worker is stuck or dies for any reason, the command will stay in-flight and never return - leaving the requester to either hang or trigger its own timeout, and `uidsToWorkerId` would keep growing over time.

This might be acceptable, in which case a mention of that behavior in the jsdoc would be good.

lib/clustering/ClusterRPC.ts
```diff
@@ -0,0 +1,469 @@
...
+ *     processing the command:
+ *         callback(error: Error | null | undefined, result?: any)
+ *
+ * When this setup is done, any worker can start sending commands by calling
+ * sendWorkerCommand(), and get back results in the callback.
+ */
+
+// exported types
+
+export type ResultObject = {
```
rachedbenmustapha, line 75: I don't remember if `Promise`s are available within the target environment (I'd think so), but they would help with a nicer API even if `async`/`await` is not available: getting rid of nested callbacks and making the `ResultObject` wrapper type superfluous (even possibly the callback types but that may be going too far).

lib/clustering/ClusterRPC.ts
```diff
@@ -0,0 +1,469 @@
...
+ * by the primary.
+ *
+ * How a command is treated:
+ *
+ * - a worker sends a command message to the primary
+ *
+ * - the primary then forwards that command to each existing worker
+ *   (including the requestor)
+ *
+ * - each worker then executes the command and returns a result or an
```
rachedbenmustapha, line 28: Kind of obvious but a mention might be warranted of the type of messages that are allowed in/out - since they have to be serialized, any circular references or very big buffers would cause issues.

lib/clustering/ClusterRPC.ts
```diff
@@ -0,0 +1,469 @@
...
+        delete uidsToWorkerId[uids];
+
+        if (!toWorker) {
+            logger.warn('worker shut down while its command was executing', {
+                workerId: toWorkerId, uids,
+            });
+            return undefined;
+        }
+        // send back response to original worker
+        _dispatchCommandResultsToWorker(toWorker, uids, completeCommandResultsArray);
```
rachedbenmustapha, line 376: It may be a good idea to include a result slot for any new workers spawned after the command was first dispatched, that didn't get to execute the command.

I understand that it's not required for the initial purpose of just a cache flush, but since it's a more "generic" module, future users of this code might be interested to know that some processes  that didn't get to run the command are currently servicing requests. Or, just a mention in the jsdoc.

lib/clustering/ClusterRPC.ts
jonathan-gramain: I think it is a good idea indeed to add a worker timeout for the reason you mention (maybe configurable via an extra argument to `sendWorkerCommand`).

lib/clustering/ClusterRPC.ts
jonathan-gramain: Yes we could use promises. Actually Metadata can use promises and `async/await` even though most of the code is legacy, there are small pieces of code that do use them.

It's not clear to me why `ResultObject` would not be needed anymore though, but I can look at it I may figure it out :)

lib/clustering/ClusterRPC.ts
jonathan-gramain: OK, I can mention this fact in the docs. To check this I think we would have to check the `cluster.workers` array again before responding, or maybe by watching events when a new worker is spawned. Since it's not needed for now I think a comment will be fine.

lib/clustering/ClusterRPC.ts
jonathan-gramain: Ah I think I understand, replacing the "error" field by a rejected promise, and the "results" by a resolved one

rachedbenmustapha: Yes exactly!

rachedbenmustapha (approved):

PR #2203, opened 2024-01-02
ARSN-382: fix empty location when redirect to / 

BourgoisMickael:
If object has a redirect to / it is sliced out
and the function receives an empty string as redirectKey
Therefore if redirectLocation consists of a single character /
The Location header would be empty

Functional tests will be performed in cloudserver once it uses the new version of arsenal

### DIFF
--- a/lib/s3routes/routesUtils.ts
+++ b/lib/s3routes/routesUtils.ts
@@ -846,7 +846,7 @@ export function redirectRequest(
     }
     let redirectLocation = justPath ? `/${redirectKey}` :
         `${redirectProtocol}://${redirectHostName}/${redirectKey}`;
-    if (!redirectKey && redirectLocationHeader) {
+    if (!redirectKey && redirectLocationHeader && redirectLocation.length > 1) {
         // remove hanging slash
         redirectLocation = redirectLocation.slice(0, -1);
     }
--- a/lib/s3routes/routesUtils.ts
+++ b/lib/s3routes/routesUtils.ts
@@ -846,7 +846,7 @@ export function redirectRequest(
     }
     let redirectLocation = justPath ? `/${redirectKey}` :
         `${redirectProtocol}://${redirectHostName}/${redirectKey}`;
-    if (!redirectKey && redirectLocationHeader && redirectLocation.length > 1) {
+    if (!redirectKey && redirectLocationHeader && redirectLocation !== '/') {
         // remove hanging slash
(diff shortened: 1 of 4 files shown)


BourgoisMickael:
ping

KillianG:
> Functional tests will be performed in cloudserver once it uses the new version of arsenal

FYI, you can run functionnal tests without having to release arsenal by using commit hash and running tests in CS etc... before merging in arsenal

lib/s3routes/routesUtils.ts
```diff
@@ -846,7 +846,7 @@ export function redirectRequest(
     }
     let redirectLocation = justPath ? `/${redirectKey}` :
         `${redirectProtocol}://${redirectHostName}/${redirectKey}`;
-    if (!redirectKey && redirectLocationHeader) {
+    if (!redirectKey && redirectLocationHeader && redirectLocation.length > 1) {
```
jonathan-gramain: I think it would be more easily understood this way:

```suggestion
    if (!redirectKey && redirectLocationHeader && redirectLocation !== '/') {
```

jonathan-gramain (commented):
Could you add a test for this change?

NB if it's not easily testable, you may want to create a helper for the redirect location construction, and unit-test that helper.

BourgoisMickael:
Here is the functional test implemented in cloudserver using the commit hash: https://github.com/scality/cloudserver/pull/5511

The logs for the tests passing are at:
- https://github.com/scality/cloudserver/actions/runs/7389115405/job/20101554780?pr=5511#step:7:5430
- https://github.com/scality/cloudserver/actions/runs/7389115405/job/20101554780?pr=5511#step:7:5573

jonathan-gramain:
@BourgoisMickael it is good to have the cloudserver tests, there can still be value in unit-testing the specific code that builds a redirect location in Arsenal, to basically test input->output combinations.

But otherwise LGTM so approving.

jonathan-gramain (approved):

BourgoisMickael:
I just added some unit test in here @jonathan-gramain

jonathan-gramain (approved):
When I mentioned a unit test I was actually thinking of something simpler, testing a helper that would only create the appropriate redirect location from its components (taking into account the special case `/` and testing it as well). It was just a suggestion, not blocking for me.

nicolas2bert (approved):

benzekrimaha (approved):

PR #2206, opened 2024-01-05
ARSN-383: Dates must now be set to midnight for lifecycle configurations

fredmnl:
So this file was heavily refactored between `7.70` and `8.1`. For simplicity, I backported a little bit of the refacto here so that the changes don't look drastically different in the two branches.

### DIFF
--- a/lib/models/LifecycleConfiguration.ts
+++ b/lib/models/LifecycleConfiguration.ts
@@ -492,6 +492,29 @@ export default class LifecycleConfiguration {
         return { ...base, ruleStatus: status }
     }
 
+
+    /**
+     * Checks the validity of the given date
+     * @param date - The date the check
+     * @return Returns an error object or `null`
+     */
+    _checkDate(date: string) {
+        const isoRegex = new RegExp('^(-?(?:[1-9][0-9]*)?[0-9]{4})-' +
+            '(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9])' +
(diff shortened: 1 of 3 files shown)


fredmnl:
I've resolved the conflicts for the `8.1` version, [view the changes here](https://github.com/scality/Arsenal/compare/w/8.1/bugfix/ARSN-383-lifecycle-configuration-dates-must-be-set-to-midnight)

fredmnl:
So I had to redo the PR since the [previous one](https://github.com/scality/Arsenal/pull/2206). The content is the same, with the following addition:
- I spotted an issue with the regex, the decimal point after seconds (`HH:mm:ss.xxx`) was not escaped and thus treated as a wild card
- I changed the subsecond part of the midnight regex so that only zeros are allowed

ghost (approved):
LGTM, thanks for rebasing (feel free to add another commit before merging with a package.json version bump, if you need to release a new arsenal version right after, that will save you one round of reviews for another PR :slightly_smiling_face: )

fredmnl:
> LGTM, thanks for rebasing (feel free to add another commit before merging with a package.json version bump, if you need to release a new arsenal version right after, that will save you one round of reviews for another PR 🙂 )

Thanks! I mindlessly approved the PR while reading your comment then had to furiously send the commit before someone else approved!

nicolas2bert:
Shouldn't we target `development/7.10` since the fix version is needed for Fix versions **7.10.9** as well as 9.3.0.

ref: https://scality.atlassian.net/browse/S3C-8320?

benzekrimaha (approved):

PR #2207, opened 2024-01-08
ARSN-384: Add a website redirect with error and body

BourgoisMickael:
There are 2 cases where a redirect can contain error headers and a body:

- If the custom error document contains an object redirect location: the response should be a redirect with the custom error document in the body ([CLDSRV-485](https://scality.atlassian.net/browse/CLDSRV-485))
  - See PR: https://github.com/scality/cloudserver/pull/5518

- If a request is sent to a folder without trailing / that holds an index document: the response should be a redirect 302 to append the trailing / with a default html response in body ([CLDSRV-489](https://scality.atlassian.net/browse/CLDSRV-489))
  - See PR: https://github.com/scality/cloudserver/pull/5520

[CLDSRV-485]: https://scality.atlassian.net/browse/CLDSRV-485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLDSRV-489]: https://scality.atlassian.net/browse/CLDSRV-489?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/s3routes/routesUtils.ts
+++ b/lib/s3routes/routesUtils.ts
@@ -895,7 +895,7 @@ export function redirectRequestOnError(
 ) {
     response.setHeader('Location', routingInfo.location);
 
-    if (!dataLocations && err.name === errors.Found.name) {
+    if (!dataLocations && err.is.Found) {
         if (method === 'HEAD') {
             return errorHeaderResponse(err, response, corsHeaders, log);
         }
--- a/lib/s3routes/routesUtils.ts
+++ b/lib/s3routes/routesUtils.ts
@@ -664,6 +664,8 @@ export function streamUserErrorPage(
     log: RequestLogger,
 ) {
     setCommonResponseHeaders(corsHeaders, response, log);
+    response.setHeader('x-amz-error-code', err.message);
+    response.setHeader('x-amz-error-message', err.description);
     response.writeHead(err.code, { 'Content-type': 'text/html' });
     response.on('finish', () => {
         // TODO ARSN-216 Fix logger
@@ -695,7 +697,9 @@ export function errorHtmlResponse(
     log.trace('sending generic html error page',
         { err });
     setCommonResponseHeaders(corsHeaders, response, log);
-    response.writeHead(err.code, { 'Content-type': 'text/html' });
+    /** AWS uses old HTTP/1.0 statusMessage for 302 Found */
+    const statusMsg = err.code === 302 ? 'Moved Temporarily' : undefined;
+    response.writeHead(err.code, statusMsg, { 'Content-type': 'text/html' });
     const html: string[] = [];
     // response.statusMessage will provide standard message for status
     // code so much set response status code before creating html
@@ -763,7 +767,9 @@ export function errorHeaderResponse(
     setCommonResponseHeaders(corsHeaders, response, log);
     response.setHeader('x-amz-error-code', err.message);
     response.setHeader('x-amz-error-message', err.description);
-    response.writeHead(err.code);
+    /** AWS uses old HTTP/1.0 statusMessage for 302 Found */
+    const statusMsg = err.code === 302 ? 'Moved Temporarily' : undefined;
+    response.writeHead(err.code, statusMsg);
     return response.end(() => {
         // TODO ARSN-216 Fix logger
         // @ts-expect-error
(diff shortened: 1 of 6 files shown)


lib/s3routes/routesUtils.ts
```diff
@@ -695,7 +697,9 @@ export function errorHtmlResponse(
     log.trace('sending generic html error page',
         { err });
     setCommonResponseHeaders(corsHeaders, response, log);
-    response.writeHead(err.code, { 'Content-type': 'text/html' });
+    /** AWS uses old HTTP/1.0 statusMessage for 302 Found */
```
jonathan-gramain: Do we care about the message returned? I don't think it would cause compatibility issues to use the newer standard message instead of the deprecated one. Applications normally shouldn't rely on the message for their behavior, only the code, the message should be informational only.

lib/s3routes/routesUtils.ts
```diff
@@ -695,7 +697,9 @@ export function errorHtmlResponse(
...
+    const statusMsg = err.code === 302 ? 'Moved Temporarily' : undefined;
```
ghost: Maybe a good candidate as a new arsenal error with code 302?

Something like
```ts
export const MovedTemporarily: ErrorFormat = {
    code: 302,
    description:
        'Moved Temporarily',
};
```

Then this message can be reused.

tests/utils/HttpResponseMock.js
```diff
@@ -18,7 +20,13 @@ class HttpResponseMock {
...
 
     end(data, encoding, callback) {
-        this.write(data, encoding, callback);
+        let cb = callback;
+        if (!cb && typeof data === 'function') {
+            cb = data;
+            cb();
+        } else {
+            this.write(data, encoding, callback);
+        }
```
ghost: Why not just calling data?

```suggestion
        if (!callback && typeof data === 'function') {
            return data();
        }
        this.write(data, encoding, callback);
```

lib/s3routes/routesUtils.ts
BourgoisMickael: This `statusMessage` is used in the response body html. I used it to match exactly what AWS does, but I can ignore it

lib/s3routes/routesUtils.ts
BourgoisMickael: There is actually a new arsenal error that is used here:

https://github.com/scality/Arsenal/pull/2207/files#diff-86c7ea469aa817b8fffa051eac8166f8fb95aa85cd95f081381d5d03db920155R693-R696

But in HTTP aws doesn't use the new `Found` message but `Moved Temporarily`. 
I will ignore this case and just use the `Found` message

ghost: Yes, but you can create another error with the Moved Temporarily message I think, it's not a Set anyway, right? As long as we use it at the right time, it should work

BourgoisMickael: No actually the error `Found` is used to fill the headers `x-amz-error-code` that needs to be `Found` and not `Moved Temporarily`

lib/s3routes/routesUtils.ts
BourgoisMickael: I've removed the override, there will be `302 Found` in the body

lib/s3routes/routesUtils.ts
BourgoisMickael: I've removed the override, there will be `302 Found` in the body

package.json
ghost, line 1: Kind remainder to have this as the last commit before merging :slightly_smiling_face:

lib/s3routes/routesUtils.ts
```diff
@@ -863,6 +865,52 @@ export function redirectRequest(
...
+    },
+    dataLocations: { size: string | number }[] | null,
+    retrieveDataParams: any,
+    response: http.ServerResponse,
+    corsHeaders: { [key: string]: string },
+    log: RequestLogger,
+) {
+    response.setHeader('Location', routingInfo.location);
+
+    if (!dataLocations && err.name === errors.Found.name) {
```
ghost: I don't remember if we have that already in 7.x branches but there is a `err.is.[something]` available to ease this kind of checks

ghost (approved):

lib/s3routes/routesUtils.ts
BourgoisMickael: thanks

benzekrimaha (approved):

PR #2208, opened 2024-01-11
Bugfix/arsn 385/fully align with aws on lifecycle configuration dates

fredmnl:
Best reviewed as two commits. 

One change is about the parsing of dates for lifecycle configurations, we wanted to align closely to AWS.
The other change is what was needed to make the tests works in cloudserver, both against cloudserver and against AWS.

### DIFF
--- a/lib/models/LifecycleConfiguration.ts
+++ b/lib/models/LifecycleConfiguration.ts
@@ -499,17 +499,35 @@ export default class LifecycleConfiguration {
      * @return Returns an error object or `null`
      */
     _checkDate(date: string) {
-        const isoRegex = new RegExp('^(-?(?:[1-9][0-9]*)?[0-9]{4})-' +
-            '(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9])' +
-            ':([0-5][0-9]):([0-5][0-9])(\\.[0-9]+)?(Z)?$');
-        if (!isoRegex.test(date)) {
+        const isoRegex = new RegExp(
+            "^(-?(?:[1-9][0-9]*)?[0-9]{4})" + // Year
+            "-(1[0-2]|0[1-9])" + // Month
+            "-(3[01]|0[1-9]|[12][0-9])" + // Day
+            "T(2[0-3]|[01][0-9])" + // Hour
+            ":([0-5][0-9])" + // Minute
+            ":([0-5][0-9])" + // Second
+            "(\\.[0-9]+)?" + // Fractional second
+            "(Z|[+-][01][0-9]:[0-5][0-9])?$", // Timezone
+            "g"
+        );
+        const matches = [...date.matchAll(isoRegex)];
+        if (matches.length !== 1) {
             const msg = 'Date must be in ISO 8601 format';
             return errors.InvalidArgument.customizeDescription(msg);
         }
-        const midnightRegex = new RegExp('^(-?(?:[1-9][0-9]*)?[0-9]{4})-' +
-            '(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T00' +
-            ':00:00(\\.0+)?(Z)?$');
-        if (!midnightRegex.test(date)) {
+        // Check for a timezone in the last match group. If none, add a Z to indicate UTC.
+        if (!matches[0][matches[0].length-1]) {
+            date += 'Z';
+        }
+        const dateObj = new Date(date);
+        if (Number.isNaN(dateObj.getTime())) {
+            const msg = 'Date is not a valid date';
+            return errors.InvalidArgument.customizeDescription(msg);
+        }
+        if (dateObj.getUTCHours() !== 0
+            || dateObj.getUTCMinutes() !== 0
+            || dateObj.getUTCSeconds() !== 0
+            || dateObj.getUTCMilliseconds() !== 0) {
             const msg = '\'Date\' must be at midnight GMT';
             return errors.InvalidArgument.customizeDescription(msg);
         }
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=16"
   },
-  "version": "7.10.54",
+  "version": "7.10.55",
   "description": "Common utilities for the S3 project components",
   "main": "build/index.js",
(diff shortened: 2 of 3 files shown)


lib/auth/auth.ts
```diff
@@ -205,7 +206,8 @@ function generateV4Headers(
     request.setHeader('host', request._headers.host);
     request.setHeader('x-amz-date', amzDate);
     request.setHeader('x-amz-content-sha256', payloadChecksum);
```
nicolas2bert: Do we have already an existing ticket regarding the issue of missing payload SHA verification in CloudServer?

lib/auth/auth.ts
```diff
@@ -215,6 +217,7 @@ function generateV4Headers(
         .filter(headerName =>
             headerName.startsWith('x-amz-')
             || headerName.startsWith('x-scal-')
+            || headerName === 'content-md5'
```
nicolas2bert: Would this condition `|| headerName === 'content-md5'` always be false since we loop through `request._headers` keys but set the 'content-md5' header to `request.headers` with `request.setHeader`. I am probably forgetting something.

lib/auth/auth.ts
```diff
@@ -205,7 +206,8 @@ function generateV4Headers(
     request.setHeader('host', request._headers.host);
     request.setHeader('x-amz-date', amzDate);
     request.setHeader('x-amz-content-sha256', payloadChecksum);
-
+    request.setHeader('content-md5', crypto.createHash('md5')
```
nicolas2bert: Do we need to set 'content-md5' header even if no payload?

lib/auth/auth.ts
```diff
@@ -181,7 +181,8 @@ function generateV4Headers(
     secretKeyValue: string,
     awsService: string,
     proxyPath: string,
-    sessionToken: string
+    sessionToken: string,
+    payload: string,
```
jonathan-gramain:
```suggestion
    payload: string?,
```

Also can you add the JSDoc entry for this param

lib/models/LifecycleConfiguration.ts
```diff
@@ -501,15 +501,27 @@ export default class LifecycleConfiguration {
     _checkDate(date: string) {
         const isoRegex = new RegExp('^(-?(?:[1-9][0-9]*)?[0-9]{4})-' +
             '(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9])' +
-            ':([0-5][0-9]):([0-5][0-9])(\\.[0-9]+)?(Z)?$');
+            ':([0-5][0-9]):([0-5][0-9])(\\.[0-9]+)?(Z|[+-][0-5][0-9]:[0-5][0-9])?$');
```
jonathan-gramain: I may be nitpicking, I think the hours in the timezone cannot go beyond -12/12, so the first digit needs to be 0 or 1. (I would be fine with `[0-9]` instead but I think `[0-5]` looks particular and only seems relevant to count minutes)

```suggestion
            ':([0-5][0-9]):([0-5][0-9])(\\.[0-9]+)?(Z|[+-][0-1][0-9]:[0-5][0-9])?$');
```

lib/models/LifecycleConfiguration.ts
```diff
@@ -501,15 +501,27 @@ export default class LifecycleConfiguration {
...
             const msg = 'Date must be in ISO 8601 format';
             return errors.InvalidArgument.customizeDescription(msg);
         }
-        const midnightRegex = new RegExp('^(-?(?:[1-9][0-9]*)?[0-9]{4})-' +
-            '(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T00' +
-            ':00:00(\\.0+)?(Z)?$');
-        if (!midnightRegex.test(date)) {
+        // Extract the time portion of the date string to check if a timezone
+        // is specified. If not, add a Z to indicate UTC. We split to avoid
+        // then hyphens in the date portion.
```
jonathan-gramain:
```suggestion
        // the hyphens in the date portion.
```

lib/models/LifecycleConfiguration.ts
```diff
@@ -501,15 +501,27 @@ export default class LifecycleConfiguration {
...
+        const time = date.split('T')[1];
+        if (!time.includes('Z') && !time.includes('+') && !time.includes('-')) {
+            date += 'Z';
```
jonathan-gramain: Would it make more sense to use match groups in the previous regexp, so we can reuse them here?

lib/models/LifecycleConfiguration.ts
fredmnl: Agreed that it looks confusing. The range is actually -14/14 for some reason, If [0-1][0-9] is fine with you, we'll leave it at that. The thing is once we have checked that the format is indeed ISO-8601, further check will be done when parsing the date to ensure that the date is legal. For example 2024-02-31T00:00:00 is legal under this regex, we're not checking everything.

lib/auth/auth.ts
fredmnl: I'm not sure I totally understand the logic of the whole function (especially the `request.headers` vs `request._headers` shenanigans) but I just applied the same logic to our new content-md5 header and it indeed ends up as a signed header later in the request i.e. this code has an effect and it actually did not work without this line.

lib/auth/auth.ts
fredmnl: I'll create one

fredmnl: https://scality.atlassian.net/browse/S3C-8368

lib/models/LifecycleConfiguration.ts
fredmnl: Sounds much better, I'll look into that

fredmnl: Done, thanks!

fredmnl:
Reviewers, as I commented during the retro, I'm removing the signed headers stuff from this PR so it only tackles the lifecycle config date part of it. I'll pick up that work in a subsequent PR. 
Hence the PR got even simpler.

jonathan-gramain (approved):

nicolas2bert (approved):

benzekrimaha (approved):

PR #2209, opened 2024-01-18
Bugfix/arsn 386/fix generate v4 headers for put with body requests

fredmnl:
When attempting to send lifecycle configuration requests to AWS with this in-house client, I found that it did not include the necessary `content-md5` header. Unfortunately this header is also used when calling `backbeat`'s internal routes but the hash is expressed in hex form, not in base64.

Thus, I chose to go down the simplest path which is to add the header anyway but choose the encoding based on the route of the request. I am also adding the header all of the time, this does not have a negative impact on our tests in cloudserver (except for the one testing the behavior when the header is not present).

### DIFF
--- a/lib/auth/auth.ts
+++ b/lib/auth/auth.ts
@@ -162,6 +162,20 @@ function doAuth(
 }
 
 /**
+ * This function will generate a version 4 content-md5 header
+ * It looks at the request path to determine what kind of header encoding is required
+ *
+ * @param path - the request path
+ * @param payload - the request payload to hash
+ */
+function generateContentMD5Header(
+    path: string,
+    payload: string,
+) {
+    const encoding = path && path.startsWith('/_/backbeat/') ? 'hex' : 'base64';
+    return crypto.createHash('md5').update(payload, 'binary').digest(encoding);
+}
+/**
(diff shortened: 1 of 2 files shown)


lib/auth/auth.ts
```diff
@@ -161,6 +161,20 @@ function doAuth(
...
+ * It looks at the request path to determine what kind of header encoding is required
+ *
+ * @param path - the request path
+ * @param payload - the request payload to hash
+ */
+function generateContentMD5Header(
+    path: string,
+    payload: string,
+) {
+    const encoding = path && path.includes('backbeat') ? 'hex' : 'base64';
```
jonathan-gramain: The check looks too broad to me, although unlikely any path may contain `backbeat` without being targeted at a backbeat route. Would something like `path.startsWith('/_/backbeat')` work instead?

jonathan-gramain (changes requested):

lib/auth/auth.ts
fredmnl: Yep, sounds good!

jonathan-gramain (approved):

benzekrimaha (approved):

PR #2210, opened 2024-01-19
ARSN-387: check for forwarded proto header

KazToozs:
This fix is for the `aws:secureTransport` condition, related to the [TSKB](https://github.com/scality/tskb/pull/429).

With load balancers in front, the check for SSL in the request must be done on the x-forwarded-proto header. 

Tests have been updated accordingly.

The other necessary change for this condition to work correctly is for the nginx config to properly pass this header:
```
proxy_set_header X-Forwarded-Proto $scheme;
```

Edit:
Green CS and Vault builds:
https://github.com/scality/Vault/pull/2151
https://github.com/scality/cloudserver/pull/5546

### DIFF
--- a/lib/policyEvaluator/utils/variables.ts
+++ b/lib/policyEvaluator/utils/variables.ts
@@ -38,7 +38,7 @@ function findVariable(variable: string, requestContext: RequestContext): string 
     // aws:SecureTransport is boolean value that represents whether the
     // request was sent using SSL
     map.set('aws:SecureTransport',
-        requestContext.getSslEnabled() ? 'true' : 'false');
+        headers?.['x-forwarded-proto'] === 'https' ? 'true' : 'false');
     // aws:SourceIp is requester's IP address, for use with IP address
     // conditions
     map.set('aws:SourceIp', requestContext.getRequesterIp());
--- a/lib/policyEvaluator/utils/conditions.ts
+++ b/lib/policyEvaluator/utils/conditions.ts
@@ -61,7 +61,7 @@ export function findConditionKey(
     case 'aws:referer': return headers.referer;
     // aws:SecureTransport – Used to check whether the request was sent
     // using SSL (see Boolean Condition Operators).
-    case 'aws:SecureTransport': return requestContext.getSslEnabled() ? 'true' : 'false';
+    case 'aws:SecureTransport': return headers?.['x-forwarded-proto'] === 'https' ? 'true' : 'false';
     // aws:SourceArn – Used check the source of the request,
     // using the ARN of the source. N/A here.
     case 'aws:SourceArn': return undefined;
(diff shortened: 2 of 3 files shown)


lib/policyEvaluator/utils/variables.ts
```diff
@@ -38,7 +38,7 @@ function findVariable(variable: string, requestContext: RequestContext): string
     // aws:SecureTransport is boolean value that represents whether the
     // request was sent using SSL
     map.set('aws:SecureTransport',
-        requestContext.getSslEnabled() ? 'true' : 'false');
+        headers?.['x-forwarded-proto'] === 'https' ? 'true' : 'false');
```
rachedbenmustapha, line 41: Headers shouldn't be blindly trusted, as they could be forged if not explicitly overwritten by the http frontend - and since I'm not sure artesca is in scope for this PR, the forward port will automatically introduce the vulnerability.

This should be disabled by default and wrapped and enabled with explicit configuration, similar to how client IP conditions are handled.
https://github.com/scality/Arsenal/blob/development/7.10/lib/policyEvaluator/requestUtils.ts

fredmnl:
I may be wrong but I'm seeing that the `getSslEnabled` function has become dead code, it's not called anywhere anymore. We should either look into removing that if it's not needed, or perhaps the checking of the forwarded header should actually be put in that function. I'm not sure which one it is because I'm not sure what I'm looking at with regards to the requestcontext object.

lib/policyEvaluator/utils/variables.ts
KazToozs: Supposing I revert this Arsenal change, would something like this in Cloudserver work according to you?:
https://github.com/scality/cloudserver/pull/5546/commits/a66f293217fdcac716ef54c0e419a837250db05c

```
function _checkBucketPolicyConditions(request, conditions, log) {
    const ip = request ? requestUtils.getClientIp(request, config) : undefined;
    if (!conditions) {
        return true;
    }
    const viaProxy = config.requests.viaProxy;

    const sslEnabled = (viaProxy ?
        request.headers['x-forwarded-proto'] === 'https' : request.connection.encrypted);
    // build request context from the request!
    const requestContext = new RequestContext(request.headers, request.query,
        request.bucketName, request.objectKey, ip,
        sslEnabled, request.resourceType, 's3', null, null,
        null, null, null, null, null, null, null, null, null,
        request.objectLockRetentionDays);
    return evaluators.meetConditions(requestContext, conditions, log);
}
```

rachedbenmustapha: `x-forwarded-proto` should be a configurable, set by the same deployment infrastructure than the one that sets up the reverse proxy, so that no misguided cross-component assumptions are made. So for S3C that's Federation, and for artesca it's shared between artesca-installer and zenko-operator.

The reason I insist that we make it configurable and disabled by default even for RING is that I'm not sure 100% of customers go through the nginx we ship, some of them might use their own LB which may or may not overwrite `x-forwarded-proto` (or use a different header altogether). Or some might in the future and we don't want any surprises.

PR #2211, opened 2024-01-24
ft: ARSN-388 implement GapSet (caching of listing gaps)

jonathan-gramain:
The GapSet class is intended for caching listing "gaps", which are contiguous series of current delete markers in buckets, although the semantics can allow for other uses in the future.

The end goal is to increase the performance of listings on V0 buckets when a lot of delete markers are present, as a temporary solution until buckets are migrated to V1 format.

This data structure is intented to be used by a GapCache instance, which implements specific caching semantics (to ensure consistency wrt. DB updates for example).

### DIFF
--- a/lib/algos/cache/GapSet.ts
+++ b/lib/algos/cache/GapSet.ts
@@ -1,0 +1,362 @@
+import assert from 'assert';
+import { OrderedSet } from '@js-sdsl/ordered-set';
+
+import errors from '../../errors';
+
+export type GapSetEntry = {
+    firstKey: string,
+    lastKey: string,
+    weight: number,
+};
+
+export interface GapSetInterface {
+    maxWeight: number;
+    size: number;
+
+    setGap: (firstKey: string, lastKey: string, weight: number) => GapSetEntry;
+    removeOverlappingGaps: (overlappingKeys: string[]) => number;
+    lookupGap: (minKey: string, maxKey?: string) => Promise<GapSetEntry | null>;
+    [Symbol.iterator]: () => Iterator<GapSetEntry>;
+    toArray: () => GapSetEntry[];
+};
+
+/**
+ * Specialized data structure to support caching of listing "gaps",
+ * i.e. ranges of keys that can be skipped over during listing
+ * (because they only contain delete markers as latest versions)
+ */
+export default class GapSet implements GapSetInterface, Iterable<GapSetEntry> {
+    _gaps: OrderedSet<GapSetEntry>;
+    _maxWeight: number;
+
+    /**
+     * @constructor
+
+     * @param {number} maxWeight - weight threshold for each cached
+     * gap (unitless). Triggers splitting gaps when reached
+     */
+    constructor(maxWeight: number) {
+        this._gaps = new OrderedSet(
+            [],
+            (left: GapSetEntry, right: GapSetEntry) => (
+                left.firstKey < right.firstKey ? -1 :
+                left.firstKey > right.firstKey ? 1 : 0
+            )
+        );
+        this._maxWeight = maxWeight;
+    }
+
+    /**
+     * Create a GapSet from an array of gap entries (used in tests)
+     */
+    static createFromArray(gaps: GapSetEntry[], maxWeight: number): GapSet {
+        const gapSet = new GapSet(maxWeight);
+        for (const gap of gaps) {
+            gapSet._gaps.insert(gap);
+        }
+        return gapSet;
+    }
+
+    /**
+     * Record a gap between two keys, associated with a weight to limit
+     * individual gap sizes in the cache.
+     *
+     * The function handles splitting and merging existing gaps to
+     * maintain an optimal weight of cache entries.
+     *
+     * @param {string} firstKey - first key of the gap
+     * @param {string} lastKey - last key of the gap, must be greater
+     * or equal than 'firstKey'
+     * @param {number} weight - total weight between 'firstKey' and 'lastKey'
+     * @return {GapSetEntry} - existing or new gap entry
+     */
+    setGap(firstKey: string, lastKey: string, weight: number): GapSetEntry {
+        assert(lastKey >= firstKey);
+
+        // Step 1/4: Find the closest left-overlapping gap, and either re-use it
+        // or chain it with a new gap depending on the weights if it exists (otherwise
+        // just creates a new gap).
+        const curGapIt = this._gaps.reverseLowerBound(<GapSetEntry>{ firstKey });
+        let curGap;
+        if (curGapIt.isAccessible()) {
+            curGap = curGapIt.pointer;
+            if (curGap.lastKey >= lastKey) {
+                // return fully overlapping gap already cached
+                return curGap;
+            }
+        }
+        let remainingWeight = weight;
+        if (!curGap                         // no previous gap
+            || curGap.lastKey < firstKey    // previous gap not overlapping
+            || (curGap.lastKey === firstKey // previous gap overlapping by one key...
+                && curGap.weight + weight > this._maxWeight) // ...but we can't extend it
+           ) {
+            // create a new gap indexed by 'firstKey'
+            curGap = { firstKey, lastKey: firstKey, weight: 0 };
+            this._gaps.insert(curGap);
+        } else if (curGap.lastKey > firstKey && weight > this._maxWeight) {
+            // previous gap is either fully or partially contained in the new gap
+            // and cannot be extended: substract its weight from the total (heuristic
+            // in case the previous gap doesn't start at 'firstKey', which is the
+            // uncommon case)
+            remainingWeight -= curGap.weight;
+
+            // there may be an existing chained gap starting with the previous gap's
+            // 'lastKey': use it if it exists
+            const chainedGapIt = this._gaps.find(<GapSetEntry>{ firstKey: curGap.lastKey });
+            if (chainedGapIt.isAccessible()) {
+                curGap = chainedGapIt.pointer;
+            } else {
+                // no existing chained gap: chain a new gap to the previous gap
+                curGap = {
+                    firstKey: curGap.lastKey,
+                    lastKey: curGap.lastKey,
+                    weight: 0,
+                };
+                this._gaps.insert(curGap);
+            }
+        }
+        // Step 2/4: Cleanup existing gaps fully included in firstKey -> lastKey, and
+        // aggregate their weights in curGap to define the minimum weight up to the
+        // last merged gap.
+        let nextGap;
+        while (true) {
+            const nextGapIt = this._gaps.upperBound(<GapSetEntry>{ firstKey: curGap.firstKey });
+            nextGap = nextGapIt.isAccessible() && nextGapIt.pointer;
+            // stop the cleanup when no more gap or if the next gap is not fully
+            // included in curGap
+            if (!nextGap || nextGap.lastKey > lastKey) {
+                break;
+            }
+            this._gaps.eraseElementByIterator(nextGapIt);
+            curGap.lastKey = nextGap.lastKey;
+            curGap.weight += nextGap.weight;
+        }
+
+        // Step 3/4: Extend curGap to lastKey, adjusting the weight.
+        // At this point, curGap weight is the minimum weight of the finished gap, save it
+        // for step 4.
+        let minMergedWeight = curGap.weight;
+        if (curGap.lastKey === firstKey && firstKey !== lastKey) {
+            // extend the existing gap by the full amount 'firstKey -> lastKey'
+            curGap.lastKey = lastKey;
+            curGap.weight += remainingWeight;
+        } else if (curGap.lastKey <= lastKey) {
+            curGap.lastKey = lastKey;
+            curGap.weight = remainingWeight;
+        }
+
+        // Step 4/4: Find the closest right-overlapping gap, and if it exists, either merge
+        // it or chain it with curGap depending on the weights.
+        if (nextGap && nextGap.firstKey <= lastKey) {
+            // nextGap overlaps with the new gap: check if we can merge it
+            minMergedWeight += nextGap.weight;
+            let mergedWeight;
+            if (lastKey === nextGap.firstKey) {
+                // nextGap is chained with curGap: add the full weight of nextGap
+                mergedWeight = curGap.weight + nextGap.weight;
+            } else {
+                // strict overlap: don't add nextGap's weight unless
+                // it's larger than the sum of merged ranges (as it is
+                // then included in `minMergedWeight`)
+                mergedWeight = Math.max(curGap.weight, minMergedWeight);
+            }
+            if (mergedWeight <= this._maxWeight) {
+                // merge nextGap into curGap
+                curGap.lastKey = nextGap.lastKey;
+                curGap.weight = mergedWeight;
+                this._gaps.eraseElementByKey(nextGap);
+            } else {
+                // adjust the last key to chain with nextGap and substract the next
+                // gap's weight from curGap (heuristic)
+                curGap.lastKey = nextGap.firstKey;
+                curGap.weight = Math.max(mergedWeight - nextGap.weight, 0);
+                curGap = nextGap;
+            }
+        }
+        // return a copy of curGap
+        return Object.assign({}, curGap);
+    }
+
+    /**
+     * Remove gaps that overlap with one or more keys in a given array or
+     * OrderedSet. Used to invalidate gaps when keys are inserted or deleted.
+     *
+     * @param {OrderedSet<string> | string[]} overlappingKeys - remove gaps that overlap
+     * with any of this set of keys
+     * @return {number} - how many gaps were removed
+     */
+    removeOverlappingGaps(overlappingKeys: OrderedSet<string> | string[]): number {
+        // To optimize processing with a large number of keys and/or gaps, this function:
+        //
+        // 1. converts the overlappingKeys array to a OrderedSet (if not already a OrderedSet)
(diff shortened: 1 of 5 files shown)


lib/algos/cache/GapSet.ts
```diff
@@ -0,0 +1,352 @@
...
+     * retrieves an existing one depending on boundaries and weight
+     *
+     * @param {string} firstKey - first key of the gap
+     * @param {string} lastKey - last key of the gap
+     * @param {number} weight - total weight between 'firstKey' and 'lastKey'
+     * @return {object} - { gap: GapSetEntry, weightToMerge: number }
+     *   - gap: created or existing gap entry, to be possibly extended and merged
+     *   - weightToMerge: how much weight there is remaining to merge with the returned gap
+     */
+    _lookupOrCreateGap(firstKey: string, lastKey: string, weight: number): {
```
fredmnl: I have a bit of a hard time understanding this function's exact responsibility. Here's where I'm at: we're basically initializing the gap so that the firstKey is set (we actually don't care about it after this function call). Now we will either return an existing gap that already contains this key or a new gap that only contains this key. That choice depends on whether firstKey was already registered or not and whether we think the new gap will be too big to be merged with another one.

How about `initializeNewGap` (it's not great either, I'll keep thinking about it).

Not really a well thought of comment, I just don't quite grasp this logic yet.

lib/algos/cache/GapSet.ts
```diff
@@ -0,0 +1,352 @@
...
+    setGap(firstKey: string, lastKey: string, weight: number): GapSetEntry {
+        assert(lastKey >= firstKey);
+
+        const lookup = this._lookupOrCreateGap(firstKey, lastKey, weight);
+        let { gap: curGap, weightToMerge } = lookup;
+        let findNextGap;
+        if (curGap.firstKey === curGap.lastKey) {
+            // single-key gaps are never chained: find the first gap strictly after 'curGap'
+            // Note: using the 'OrEqual' method is not possible as it would yield 'curGap'
+            findNextGap = this._gaps.findLeastGreaterThan.bind(this._gaps);
```
fredmnl: Perhaps `findNextGap(gap)` could be a nice helper function: calls strictly `greaterThan` if `gap` is one single key?

lib/algos/cache/GapSet.ts
```diff
@@ -0,0 +1,352 @@
...
+            }
+            if (nextGap.firstKey === lastKey) {
+                // in this particular case the next gap is chained
+                // after the new gap, hence we add the full amount of
+                // the weight to merge as it doesn't overlap with the
+                // next gap
+                mergedWeightSum += weightToMerge;
+            }
+            // merge 'nextGap' into 'curGap'
+            curGap.lastKey = nextGap.lastKey;
```
fredmnl, line 132: Are we assuming here that the input gap is not too large? (specifically not larger than `max_weight`)

lib/algos/cache/GapSet.ts
```diff
@@ -0,0 +1,352 @@
...
+                gap = {
+                    firstKey: gap.lastKey,
+                    // split gap is temporarily single-key, will be extended by setGap()
+                    lastKey: gap.lastKey,
+                    // split gap only contains the weight not already included in the previous gap
+                    //weight: Math.max(weight - gap.weight, 0),
+                    weight: 0,
+                };
+                // there may be an existing gap starting with 'lastKey': delete it first
+                this._gaps.delete(gap);
```
fredmnl: Don't we risk deleting some data from the structure here? What about if the new gap inserted actually spans multiple known chained gaps? e.g.
```
NewGap: +1234+
GapSet: +123+456+789+
```
We would end up deleting the second gap in the GapSet and then go on to lose keys 5 and 6.

lib/algos/cache/GapSet.ts
```diff
@@ -0,0 +1,352 @@
...
+     * The function handles splitting and merging existing gaps to
+     * maintain an optimal weight of cache entries.
+     *
+     * @param {string} firstKey - first key of the gap
+     * @param {string} lastKey - last key of the gap, must be greater
+     * or equal than 'firstKey'
+     * @param {number} weight - total weight between 'firstKey' and 'lastKey'
+     * @return {GapSetEntry} - existing or new gap entry
+     */
+    setGap(firstKey: string, lastKey: string, weight: number): GapSetEntry {
```
fredmnl, line 73: I was thinking of an alternate logic for setGap (still similar though). 

In a for loop: 
- find the first gap in the GapSet that overlaps with the new gap and re-arrange the situation if needed (same as in `_lookupOrCreateGap()`)
- at this point, we have probably found part of the gap already inside the set, or a stretch of new keys that we have inserted, either way the new gap is being reduced
- continue with diminishing new gap until it's empty

I feel like it's a little bit more straightforward to reason about, what do you think?

lib/algos/cache/GapSet.ts
```diff
@@ -0,0 +1,352 @@
...
+     * Record a gap between two keys, associated with a weight to limit
+     * individual gap sizes in the cache.
+     *
+     * The function handles splitting and merging existing gaps to
+     * maintain an optimal weight of cache entries.
+     *
+     * @param {string} firstKey - first key of the gap
+     * @param {string} lastKey - last key of the gap, must be greater
+     * or equal than 'firstKey'
+     * @param {number} weight - total weight between 'firstKey' and 'lastKey'
```
fredmnl, line 70: I haven't looked at the implementation of the caller code but I'm guessing that it reduces all of the keys (the DMs and versions) into the `weight` somewhere before calling. What do you think of passing the entire list of keys instead, that way we can actually compute the exact weight of each of the gaps that are created even if the caller is unaware of how we are actually storing them.

lib/algos/cache/GapSet.ts
jonathan-gramain: You're right, I will add a lookup on the chained gap, and if it exists, return it instead of deleting it (it may still be merged with following gaps by `setGap`)

lib/algos/cache/GapSet.ts
jonathan-gramain: The general idea is that the caller needs to call `setGap` frequently enough not to have to provide the full list of keys. "Frequently enough" means at the very least as many times as there are real gaps to create in `GapSet` (including the ones maxing out their weight).

In practice, I added a parameter in the caller `triggerSaveGapWeight` (for lack of a better name!) that defaults to calling `setGap` every time half the max weight has been seen. It should be frequent enough to ensure all gaps are under but close enough to `maxWeight`. It also calls it whenever a gap stops and is considered large enough to be worth caching.

The caller doesn't maintain the entire list of keys encountered, it only keeps in memory what will actually be returned in the result, so I think it's best not to have to maintain and pass the full list of keys.

Here's the draft branch on the caller side: https://github.com/scality/Arsenal/tree/feature/ARSN-389-optimizeListingWithGapCache

lib/algos/cache/GapSet.ts
jonathan-gramain: Sorry it doesn't come clear to me, you may have to write some pseudo-code :)

The current implementation extends the new (or existing) gap until it reaches `lastKey`.

lib/algos/cache/GapSet.ts
jonathan-gramain: We actually don't rely on the existing weight here, only on the new weight. The reason is that the new weight is supposed to reflect the reality, which should also be reflected by the old weights but since we're extending the gaps (possibly filling in holes around the existing cached gaps - that's a lot of emptiness :smile: ) the existing cumulated weight must always be smaller than where we're extending to.

The envisioned use case is when there were a series of small-ish listings that cached individual small-ish gaps, in a large deleted gap that wasn't cached as a whole until now, we can just remove the old cached gaps without considering their weight.

lib/algos/cache/GapSet.ts
jonathan-gramain: Yes we can do this indeed since we're extending the gap at every loop iteration, it cannot be a single-key gap anymore after the first iteration, and it's simpler than relying on an extra variable that is updated.

I'll just fit the logic inside the loop body I think it's reasonably clear and compact not to warrant a helper.

jonathan-gramain (commented):
Thanks for the feedback @fredmnl !

I updated the PR with new commits that should address your comments.

lib/algos/cache/GapSet.ts
jonathan-gramain: > I have a bit of a hard time understanding this function's exact responsibility. Here's where I'm at: we're basically initializing the gap so that the firstKey is set (we actually don't care about it after this function call). Now we will either return an existing gap that already contains this key or a new gap that only contains this key. That choice depends on whether firstKey was already registered or not and whether we think the new gap will be too big to be merged with another one.

This is a good view of what the function does.

lib/algos/cache/GapSet.ts
```diff
@@ -145,20 +145,19 @@ export default class GapSet implements Iterable<GapSetEntry> {
...
-            // gaps with more than one key may be chained (g1.lastKey == g2.firstKey) and we
-            // need the 'OrEqual' method to find those chained gaps
-            findNextGap = this._gaps.findLeastGreaterThanOrEqual.bind(this._gaps);
-        }
         let mergedWeightSum = 0;
         // loop over existing mergeable gaps straddled by 'gap.lastKey' -> 'lastKey'
         while (lastKey > curGap.lastKey) {
-            const nextGap = findNextGap({ firstKey: curGap.lastKey })?.value;
+            let nextGap;
+            if (curGap.firstKey === curGap.lastKey) {
```
fredmnl: Actually, now that I think about it, getting the next gap could be `findLeastGreaterThan(gap.firstKey)` as long as the data store is healthy (no overlapping gap), what do you think?

fredmnl:
One question on parallelism, I'm still new to Node so maybe all potential race conditions are handled by the mono-threaded nature of node. Could we have the following situation for example:
```
Initial state of the storage is that there are delete markers on key 1 to 5. 

Listing operation: x starts listing  1 .. 2 .. 3 .. 4 .. 5 .. done, broadcast gap 1-5
Put operation:     x                         puts 2
GapSet:            x                         2 clears cache                    add 1-5 to cache
```

Basically a concurrent list and PUT operation will not update the data store in the right order. Is this possible?

lib/algos/cache/GapSet.ts
fredmnl: Sure! I'll put the weights aside for now. Basically, I'm thinking that we are adding a `newGap{first, last}` to a `GapSet`, let's call it `gapSet.mergeGap(newGap)`

First locate the first gap in `GapSet` that strictly overlaps with `newGap` (no chaining-style overlap), let's call it `gap1{first1, last1}`. Now there are three cases: 
- `gap1` completely encompasses `newGap`, we're done. 
- `gap1` starts before or on the same key as `newGap`, then we can do `gapSet.mergeGap(gap{last1, last})`
- `gap1` starts after `newGap`, then we insert `gap{first, first1}` and call `gapSet.mergeGap(gap{last1, last})`

I wrote it recursively only because I think it's easier to express. 

What I meant in my comment is that I feel like the handling of these three cases is kind of what you do in `_lookupOrCreateGap()`. There are more cases because I have hidden some of the cases behind wording like "first gap that strictly overlaps". 

I just thought it was easier to reason about this way but it might also just be personal taste, feel free to take it or leave it.

jonathan-gramain:
> One question on parallelism, I'm still new to Node so maybe all potential race conditions are handled by the mono-threaded nature of node. Could we have the following situation for example:
> 
> ```
> Initial state of the storage is that there are delete markers on key 1 to 5. 
> 
> Listing operation: x starts listing  1 .. 2 .. 3 .. 4 .. 5 .. done, broadcast gap 1-5
> Put operation:     x                         puts 2
> GapSet:            x                         2 clears cache                    add 1-5 to cache
> ```
> 
> Basically a concurrent list and PUT operation will not update the data store in the right order. Is this possible?

Yes you're totally correct, there is a race condition that we need to address in some way. I started thinking about this yesterday and potential solutions to it. Updating the cache after every listed key would not be enough as we can always have updates on or between listed keys (i.e. in the inclusive range between two listed keys).

My current idea is that the `GapSet` class somehow needs to be aware of the gaps that are in construction from the listings in progress, and either ignore or invalidate calls to `setGap` when those gaps overlap with PUT/DELETE updates that occurred while this gap construction was happening. In other words, if `GapSet.setGap(firstKey, lastKey)` is aware of when the `firstKey` was initialized, it can know whether there were new updates in the range `[firstKey, lastKey]` when the gap is set and ignore the call in such case.

Another idea can be to track all listings in progress, and push invalidation to each listing in progress when `removeOverlappingGaps` is called. Basically considering that the gaps in construction are already part of the gap set (but not yet usable by other listings).

The funny thing is originally I though of an API for `setGap` that takes a gap object returned from a prior call, rather than a first and last key, but I was concerned about such race conditions (in the sense that the gap may have been invalidated in the meantime), and decided to switch to first/last strings instead. Using a gap object could actually provide a way to deal with those race conditions by storing some state in the gap objects, although I'm not sure of the details. It could be a raft sequence number (cseq) that would be compared against the recent updates, for example.

lib/algos/cache/GapSet.ts
jonathan-gramain: Gaps do actually overlap (by a single key), that's why we need the `OrEqual` version to find them. But we also need to skip the single-key gap that we may start with hence the two different calls.

lib/algos/cache/GapSet.ts
jonathan-gramain: Ah I see, so combining the logic from `lookupOrCreateGap` and the merge process itself together, rather than forward-merging only from the result of the lookup. This could work too, I'll see how it goes and may change it. Indeed we need to either recurse or loop over all such gaps to complete the job.

jonathan-gramain:
So I came up with another approach to ensure atomicity, this PR should explain it: https://github.com/scality/Arsenal/pull/2213

lib/algos/cache/GapSet.ts
fredmnl: Sorry I meant strictly overlapping, as in intersecting on more than one key. We agree that the data store should never be in a state where we have strictly overlapping gaps?

While we're at it, can you confirm that we cannot have two gaps with the same `firstKey`? I inferred from the fact that it's a set but I just want confirmation. 

If that's the case, I think my logic works, there are four cases:
I'll write your approach as A#1 and my proposal as A#2:

```
If gap is a single key, and nextGap is chained:
ant..ant - ant..fox
I think this is not possible. 
---
If gap is a single key, and nextGap is not chained:
ant..ant - cat..fox
findLeastGreaterThan('ant') will work (both A#1 and A#2)
---
If gap spans multiple keys, and nextGap is chained:
ant..cat - cat..fox
A#1 - findLeastGreaterThanOrEqual('cat') will work
A#2 - findLeastGreaterThan('ant') will work
---
If gap spans multiple keys, and nextGap is not chained:
ant..cat - elk..fox
A#1 - findLeastGreaterThanOrEqual('cat') will work
A#2 - findLeastGreaterThan('ant') will work
```

jonathan-gramain: Ah I see what you mean, comparing against the first key rather than the last key of the current gap. Should work indeed!

Yes gaps can only overlap by a single key in any case, and there can be only one gap starting with a given key.

jonathan-gramain: updated

lib/algos/cache/GapSet.ts
fredmnl: Ok, so there is an implicit contract with the library user that the gaps sent should not be too large. I'm ok with that.

lib/algos/cache/GapSet.ts
fredmnl: I think I wanted to see it this way because I was missing the idea that the incoming gap is assumed to be of a weight smaller than the max weight. With that in mind this makes more sense

lib/algos/cache/GapSet.ts
```diff
@@ -0,0 +1,362 @@
...
+            }
+            if (nextGap.firstKey === lastKey) {
+                // in this particular case the next gap is chained
+                // after the new gap, hence we add the full amount of
+                // the weight to merge as it doesn't overlap with the
+                // next gap
+                mergedWeightSum += weightToMerge;
+            }
+            // merge 'nextGap' into 'curGap'
+            curGap.lastKey = nextGap.lastKey;
```
fredmnl, line 132: One thing that's still bugging me is that we are checking in `_lookupOrCreateGap` if there is a gap straddling `newGap.firstKey` and we are handling it depending on the weights. Yet, when we come to the gap that may be straddling `newGap.lastKey` we don't perform the same check. If I'm correct we could be extending a gap to an arbitrarily large weight if we are only extending it toward lower values (a workload that would delete objects with ever decreasing keys). 

If all of this is correct, the logic could be summarized as: 
- look for a gap straddling `newGap.firstKey` and make a decision as to whether we want to extend it, or split.
- look for a gap straddling `newGap.lastKey` and make a decision as to whether we want to extend it, or split (which may depend on the previous decision)
- with the boundaries handled, remove all of the gaps that are fully contained within `newGap` and insert `newGap`

What do you think?

jonathan-gramain: There is indeed the issue that you mentioned, it's probably what I observed too during my testing.

I am reworking this and should be able to finish on Monday, will let you know when it's done! I think the logic that you describe is good, there is some complexity in checking all cases and setting the weights accordingly but it should be achievable.

jonathan-gramain: I reworked this in a new commit

tmacro (approved):

fredmnl (approved):

PR #2214, opened 2024-02-02
ARSN-389 DelimiterMaster: v0 format gap skipping

jonathan-gramain:
**Note for reviews**: best to review commits individually.

- change contract of skipping() API

  Instead of returning a "prefix" for the listing task to skip over, directly return the key on which to skip and continue the listing.

  It is both more natural as well as needed to implement skipping over cached "gaps" of deleted objects.

  Note that it could even be more powerful to return the type of query param to apply for the next listing ('gt' or 'gte'), but it would be more complex to implement with little practical benefit, so instead we add a null byte at the end of the returned key to skip to, whenever we want a 'gt' behavior from the returned 'gte' key.

- Implement logic in DelimiterMaster to improve efficiency of listings of buckets in V0 format that have a lot of current delete markers.

  A GapCache instance can be attached to a DelimiterMaster instance which enables the following:

  - Lookups in the cache to be able to restart listing directly beyond the cached gaps. It is done by returning FILTER_SKIP code when listing inside a gap, which hints the caller (RepdServer) that it is allowed to restart a new listing from a specific later key.

  - Building gaps and cache them, when listing inside a series of current delete markers. This allows future listings to benefit from the gap information and skip over them.

  An important caveat is that there is a limited time in which gaps can be built from the current listing: it is a trade-off to guarantee the validity of cached gaps when concurrent operations may invalidate them. This time is set in the GapCache instance as `exposureDelayMs`, and is the time during which concurrent operations are kept in memory
to potentially invalidate future gap creations. Because listings use a snapshot of the database, they return entries that are older than when the listing started. For this reason, in order to be allowed to consistently build new gaps, it is necessary to limit the running time of listings, and potentially redo periodically new listings (based on time or number of listed keys), resuming from where the previous listing stopped, instead of continuing the current listing.

### DIFF
--- a/lib/algos/list/delimiterMaster.ts
+++ b/lib/algos/list/delimiterMaster.ts
@@ -5,18 +5,23 @@ import {
     DelimiterFilterStateId,
     DelimiterFilterState_NotSkipping,
     DelimiterFilterState_SkippingPrefix,
+    ResultObject,
 } from './delimiter';
 const Version = require('../../versioning/Version').Version;
 const VSConst = require('../../versioning/constants').VersioningConstants;
 const { BucketVersioningKeyFormat } = VSConst;
-const { FILTER_ACCEPT, FILTER_SKIP, FILTER_END, inc } = require('./tools');
+const { FILTER_ACCEPT, FILTER_SKIP, FILTER_END, SKIP_NONE, inc } = require('./tools');
+
+import { GapSetEntry } from '../cache/GapSet';
+import { GapCacheInterface } from '../cache/GapCache';
 
 const VID_SEP = VSConst.VersionId.Separator;
 const { DbPrefixes } = VSConst;
 
-const enum DelimiterMasterFilterStateId {
+export const enum DelimiterMasterFilterStateId {
     SkippingVersionsV0 = 101,
     WaitVersionAfterPHDV0 = 102,
+    SkippingGapV0 = 103,
 };
 
 interface DelimiterMasterFilterState_SkippingVersionsV0 extends FilterState {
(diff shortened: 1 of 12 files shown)


lib/algos/list/delimiterMaster.ts
```diff
@@ -93,6 +184,166 @@ export class DelimiterMaster extends Delimiter {
...
+    }
+
+    /**
+     * Trigger a lookup of the closest upcoming or already skippable gap.
+     *
+     * @param {string} fromKey - lookup a gap not before 'fromKey'
+     * @return {undefined} - the lookup is asynchronous and its
+     * response is handled inside this function
+     */
+    _triggerGapLookup(gapCaching: GapCachingInfo_NoCachedGap, fromKey: string): void {
```
fredmnl, line 292: For my own edification, this construct with the type `GapCachingInfo_NoCachedGap` is here to enforce the fact the the GapCache should be in a specific state for the function to be called? 

```
type GapCachingInfo_NoCachedGap = {
    state: GapCachingState.UnknownGap
        | GapCachingState.GapLookupInProgress
    gapCache: GapCacheInterface;
};
```

What happens if it is not in the proper state `UnknownGap | GapLookupInProgress`?

lib/algos/list/delimiterMaster.ts
```diff
@@ -177,14 +437,124 @@ export class DelimiterMaster extends Delimiter {
...
+            state: GapBuildingState.Building,
+            gapCache,
+            params,
+            gap: {
+                firstKey: gap.lastKey,
+                lastKey: gap.lastKey,
+                weight: 0,
+            },
+            gapWeight,
+        };
```
fredmnl, line 520: If I understand correctly, we are saving the gap that was under construction and we are setting the gap to start again from where we are currently at. Shouldn't we reset the `initTimestamp` as well? Otherwise any listing can only feed the `GapCache` for `exposureDelayMs` until it is gets refused all updates.

lib/algos/list/delimiterMaster.ts
```diff
@@ -177,14 +437,124 @@ export class DelimiterMaster extends Delimiter {
...
+        const { params, gap } = gapBuilding;
+        gap.lastKey = newKey;
+        gap.weight += 1;
+        gapBuilding.gapWeight += 1;
+        // the GapCache API requires updating a gap regularly because it can only split
+        // it once per update, by the known last key. In practice the default behavior
+        // is to trigger an update after a number of keys that is half the maximum weight.
+        // It is also useful for other listings to benefit from the cache sooner.
+        if (gapBuilding.gapWeight >= params.minGapWeight &&
+            gap.weight >= params.triggerSaveGapWeight) {
```
fredmnl, line 560: Would we benefit from triggering a save also when we are near the `exposureDelayMs` duration for the gap under construction? I'm not sure there's a realistic scenario where we are hitting that limit though.

fredmnl (approved):

lib/algos/list/delimiterMaster.ts
jonathan-gramain: This is essentially to get type errors from Typescript, normally by passing it in a function param, typescript can detect when the type is wrong, while using an object property doesn't usually provide the same type checks.

lib/algos/list/delimiterMaster.ts
jonathan-gramain: Yes it is on purpose, the `initTimestamp` can only be set when the listing starts, we're not allowed to change it until a new listing request is made. This is because the listing uses a snapshot on the DB so it returns the entries in the state they were in at that time.

This also means that we will need to restart the listing periodically to overcome this, that's a reason to have an `expirationDelayMs` not too small (thinking maybe 1 second to 10 seconds would be appropriate). Triggering listing requests is normally inexpensive, what gets expensive is listing thousands of entries. But the code in repd will need to be adapted for this, and call `refreshGapCache()` periodically every time a new listing is about to be triggered.

lib/algos/list/delimiterMaster.ts
jonathan-gramain: I'm not sure but it could help to trigger one in `refreshGapCache()` in case the current gap in progress is still valid, we could save it. The goal is to call `refreshGapCache()` before the expiration, to be able to continue chaining gaps that can be. Otherwise, hard to say, I'd rather keep it "simple" and not do it if not clearly helpful.

jonathan-gramain:
I spotted a couple issues in logic and tests so turning the PR back into draft, I should be able to get them fixed tomorrow.

lib/algos/list/delimiterMaster.ts
fredmnl: How does this work? How can the listing work on a snapshot of the DB? Does it rollback changes that happened during the listing? Is it related to the `WriteCache` you mentioned when we discussed this design? 

I'd like to understand this better so that I can take this into account when we will tackle metadata sharding.

jonathan-gramain: It's all handled by the LevelDB engine. LevelDB has also explicit snapshot support with an API, but for listings it automatically uses an internal snapshot of the DB.

jonathan-gramain:
Should be good now, removing the draft marker

lib/algos/list/delimiterMaster.ts
jonathan-gramain: done: the current gap is saved when `refreshGapCache()` gets called.

It is useful in the context of the caller periodically calling this function, to make sure that multiple listings can yield a chain of contiguous gaps if they are only listing current delete markers.

lib/algos/list/delimiterMaster.ts
fredmnl: That's pretty nifty! Makes sense that the LSM-structure easily allows such snapshot feature, is a snapshot created every time that a listing starts? Do we really need the listing to be done on a snapshot though? Sounds like it would be borderline impossible to do on a sharded metadata model, I was under the impression that the listing had to contain entries that represent all updates until the beginning of the listing but not necessary represent an instant in time. What do you think? 

Maybe we can keep this conversation going in Slack or somewhere more appropriate, I want to present to you an update on the designs we discussed for metadata sharding :)

nicolas2bert (approved):

PR #2215, opened 2024-02-12
ARSN-392 Put metadata to a standalone null master breaks CloudServer CRUD

nicolas2bert:
Functionally tested here: https://github.com/scality/cloudserver/pull/5531

**Notes:**

_NOTE1_: This change is only impacting the "file" metadata backend logic. Similar fixes will be implemented in scality/Metadata using `Arsenal.updateOrAppendNullVersionId()` method.
_NOTE2_: null version key is handled very differently from development/7.10 to development/7.70 where: 
- in development/7.10 only the **old** null key logic is supported whereas 
- in development/7.70 **both old and [new](https://scality.atlassian.net/browse/S3C-7352)** null key logic are supported for backward compatibility. 
**This change won't be forward-ported to development/7.70**. We will create a different PR: https://github.com/scality/Arsenal/pull/2216 targeting development/7.70.

**Description:**

cbe6a5e2d6a4cf4bb24dbacfde5b286a90ce9aea: File backend should be "mocking" the V0 metadata backend. To fix it, we imported the V0 [processVersionSpecificPut](https://github.com/scality/MetaData/blob/development/7.10/lib/server/VersioningRequestProcessor.js#L1134) from scality/Metadata.

46258bca746771fcde327e132dba2de2be5bba50: 
- Add the `nullVersionId` field into the master update. The `nullVersionId` is needed for listing, retrieving, and deleting null versions.

- Manage scenarios in which a version is marked with the `isNull` attribute set to `true`, but without a `version ID`.
It happens after `BackbeatClient.putMetadata()` is applied to a standalone null master.

### DIFF
--- a/lib/versioning/Version.ts
+++ b/lib/versioning/Version.ts
@@ -84,6 +84,33 @@ export class Version {
     }
 
     /**
+     * Appends a key-value pair to a JSON object represented as a string. It adds
+     * a comma if the object is not empty (i.e., not just '{}'). It assumes the input
+     * string is formatted as a JSON object.
+     *
+     * @param {string} stringifiedObject The JSON object as a string to which the key-value pair will be appended.
+     * @param {string} key The key to append to the JSON object.
+     * @param {string} value The value associated with the key to append to the JSON object.
+     * @returns {string} The updated JSON object as a string with the new key-value pair appended.
+     * @example
+     * _jsonAppend('{"existingKey":"existingValue"}', 'newKey', 'newValue');
+     * // returns '{"existingKey":"existingValue","newKey":"newValue"}'
+     */
+    static _jsonAppend(stringifiedObject: string, key: string, value: string): string {
+        // stringifiedObject value has the format of '{...}'
+        let index = stringifiedObject.length - 2;
+        while (stringifiedObject.charAt(index) === ' ') {
+            index -= 1;
+        }
+        const needComma = stringifiedObject.charAt(index) !== '{';
+        return (
+            `${stringifiedObject.slice(0, stringifiedObject.length - 1)}` +
+            (needComma ? ',' : '') +
+            `"${key}":"${value}"}`
+        );
+    }
+
+    /**
      * Put versionId into an object in the (cheap) way of string manipulation,
      * instead of the more expensive alternative parsing and stringification.
      *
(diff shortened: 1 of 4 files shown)


lib/versioning/Version.ts
```diff
@@ -83,6 +83,31 @@ export class Version {
...
+     *
+     * @param {string} stringifiedObject The JSON object as a string to which the key-value pair will be appended.
+     * @param {string} key The key to append to the JSON object.
+     * @param {string} value The value associated with the key to append to the JSON object.
+     * @returns {string} The updated JSON object as a string with the new key-value pair appended.
+     * @example
+     * _jsonAppend('{"existingKey":"existingValue"}', 'newKey', 'newValue');
+     * // returns '{"existingKey":"existingValue","newKey":"newValue"}'
+     */
+    static _jsonAppend(stringifiedObject: string, key: string, value: string): string {
```
jonathan-gramain, line 99: I see there are tests for `appendVersionId` but not for this internal helper, it could be worth adding some tests for it too as it is more generic and may be used to add other fields in the future. I'd say a handful of tests to cover the common cases would be enough like for `appendVersionId`.

lib/versioning/Version.ts
```diff
@@ -83,6 +83,31 @@ export class Version {
...
+        // stringifiedObject value has the format of '{...}'
+        let index = stringifiedObject.length - 2;
+        while (stringifiedObject.charAt(index--) === ' ');
```
jonathan-gramain: I see that this code is mainly moved from the original `appendVersionId` implementation, but there are style issues that could be improved on, hence suggesting a couple things in this PR.

I'd recommend not relying on the behavior of the decrement operator (prefix or postfix) as it is error-prone, doing `index--` or `index -= 1` inside the loop body would be safer.

lib/versioning/Version.ts
```diff
@@ -83,6 +83,31 @@ export class Version {
...
+        const comma = stringifiedObject.charAt(index + 1) !== '{';
```
jonathan-gramain:
```suggestion
        const needComma = stringifiedObject.charAt(index + 1) !== '{';
```

lib/versioning/Version.ts
```diff
@@ -83,6 +83,31 @@ export class Version {
...
+        return (
+            `${stringifiedObject.slice(0, stringifiedObject.length - 1)}` + // eslint-disable-line
```
jonathan-gramain: Instead of a bare `// eslint-disable-line`, having the exact check code afterwards so we know exactly what we disable would be better (both to understand and to avoid disabling other potentially useful rules):

`// eslint-disable-line the-check-that-fails`

You can probably run the tests without that line to figure out what it is.

jonathan-gramain (approved):
LGTM, just a few suggestions to improve the moved code

BourgoisMickael (approved):

nicolas2bert:
@bert-e approve

PR #2216, opened 2024-02-12
ARSN-392 [7.70] Put metadata to a standalone null master breaks CloudServer CRUD

nicolas2bert:
_NOTE_: This change is only impacting the "file" metadata backend logic. Similar fixes will be implemented in scality/Metadata using Arsenal.updateOrAppendNullVersionId() method.

**Description:**

1653fd2ad37b209f66bc995c2e9823813a504634: "File" metadata backend should be "mocking" the V0 scality/Metadata backend. To fix it, we imported the V0 [processVersionSpecificPut](https://github.com/scality/MetaData/blob/development/7.70/lib/server/VersioningRequestProcessor.js#L1628) from scality/Metadata.

58d06229383a692f073e75db2e6033217bcf29e9: 
- For "null key" backward compatibility (if `isNull` is undefined), add the `nullVersionId` field to the master update. The `nullVersionId` is needed for listing, retrieving, and deleting null versions. 
- For the new null key implementation (if `isNull` is defined): add the `isNull2` field and set it to true to specify that the new version is null AND has been put with a Cloudserver handling null keys (i.e., supporting [S3C-7352](https://scality.atlassian.net/browse/S3C-7352)). 
- Manage scenarios in which a version is marked with the `isNull` attribute set to true, but without a version ID. This happens after `BackbeatClient.putMetadata()` is applied to a standalone null master.

[S3C-7352]: https://scality.atlassian.net/browse/S3C-7352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/versioning/Version.ts
+++ b/lib/versioning/Version.ts
@@ -3,7 +3,7 @@ import { VersioningConstants } from './constants';
 const VID_SEP = VersioningConstants.VersionId.Separator;
 /**
  * Class for manipulating an object version.
- * The format of a version: { isNull, isDeleteMarker, versionId, otherInfo }
+ * The format of a version: { isNull, isNull2, isDeleteMarker, versionId, otherInfo }
  *
  * @note Some of these functions are optimized based on string search
  * prior to a full JSON parse/stringify. (Vinh: 18K op/s are achieved
@@ -13,24 +13,31 @@ const VID_SEP = VersioningConstants.VersionId.Separator;
 export class Version {
     version: {
         isNull?: boolean;
+        isNull2?: boolean;
         isDeleteMarker?: boolean;
         versionId?: string;
         isPHD?: boolean;
+        nullVersionId?: string;
     };
 
     /**
      * Create a new version instantiation from its data object.
      * @param version - the data object to instantiate
      * @param version.isNull - is a null version
+     * @param version.isNull2 - Whether new version is null or not AND has
+     * been put with a Cloudserver handling null keys (i.e. supporting
+     * S3C-7352)
      * @param version.isDeleteMarker - is a delete marker
      * @param version.versionId - the version id
      * @constructor
      */
     constructor(version?: {
         isNull?: boolean;
+        isNull2?: boolean;
         isDeleteMarker?: boolean;
         versionId?: string;
         isPHD?: boolean;
+        nullVersionId?: string;
     }) {
         this.version = version || {};
     }
@@ -167,6 +174,19 @@ export class Version {
     }
 
     /**
+     * Check if a version is a null version and has
+     * been put with a Cloudserver handling null keys (i.e. supporting
+     * S3C-7352).
+     *
+     * @return - stating if the value is a null version and has
+     * been put with a Cloudserver handling null keys (i.e. supporting
+     * S3C-7352).
+     */
+    isNull2Version(): boolean {
+        return this.version.isNull2 ?? false;
+    }
+
+    /**
      * Check if a stringified object is a delete marker.
      *
      * @param value - the stringified object to check
@@ -216,6 +236,15 @@ export class Version {
     }
 
     /**
+     * Get the nullVersionId of the version.
+     *
+     * @return - the nullVersionId
+     */
+    getNullVersionId(): string | undefined {
+        return this.version.nullVersionId;
+    }
+
+    /**
      * Mark a version as a delete marker.
      *
      * @return - the updated version
@@ -231,6 +260,19 @@ export class Version {
      * @return - the updated version
      */
     setNullVersion() {
+        this.version.isNull = true;
+        return this;
+    }
+
+    /**
+     * Mark that the null version has been put with a Cloudserver handling null keys (i.e. supporting S3C-7352)
+     * 
+     * If `isNull2` is set, `isNull` is also set to maintain consistency.
+     * Explicitly setting both avoids misunderstandings and mistakes in future updates or fixes.
+     * @return - the updated version
+     */
+    setNull2Version() {
+        this.version.isNull2 = true;
         this.version.isNull = true;
         return this;
     }
--- a/lib/versioning/VersioningRequestProcessor.ts
+++ b/lib/versioning/VersioningRequestProcessor.ts
@@ -1,6 +1,6 @@
 import errors, { ArsenalError } from '../errors';
 import { Version } from './Version';
-import { generateVersionId as genVID } from './VersionID';
+import { generateVersionId as genVID, getInfVid } from './VersionID';
 import WriteCache from './WriteCache';
 import WriteGatheringManager from './WriteGatheringManager';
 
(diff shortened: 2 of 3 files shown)


lib/versioning/Version.ts
```diff
@@ -83,6 +88,31 @@ export class Version {
...
+     *
+     * @param {string} stringifiedObject The JSON object as a string to which the key-value pair will be appended.
+     * @param {string} key The key to append to the JSON object.
+     * @param {string} value The value associated with the key to append to the JSON object.
+     * @returns {string} The updated JSON object as a string with the new key-value pair appended.
+     * @example
+     * _jsonAppend('{"existingKey":"existingValue"}', 'newKey', 'newValue');
+     * // returns '{"existingKey":"existingValue","newKey":"newValue"}'
+     */
+    static _jsonAppend(stringifiedObject: string, key: string, value: string): string {
```
nicolas2bert: If you have already reviewed https://github.com/scality/Arsenal/pull/2215, there's no need to do so again as it is a duplicate.

lib/versioning/Version.ts
```diff
@@ -93,14 +123,32 @@ export class Version {
...
+    * This function first checks if the `nullVersionId` property already exists within the input string.
+    * If it exists, the function updates the `nullVersionId` with the new value provided.
+    * If it does not exist, the function appends a `nullVersionId` property with the provided value.
+    *
+    * @static
+    * @param {string} value - The JSON-formatted string that may already contain a `nullVersionId` property.
+    * @param {string} nullVersionId - The new value for the `nullVersionId` property to be updated or appended.
+    * @returns {string} The updated JSON-formatted string with the new `nullVersionId` value.
+    */
+    static updateOrAppendNullVersionId(value: string, nullVersionId: string): string {
```
nicolas2bert: If you have already reviewed https://github.com/scality/Arsenal/pull/2215, there's no need to do so again as it is a duplicate.

tests/unit/versioning/Version.spec.js
```diff
@@ -0,0 +1,55 @@
+const { Version } = require('../../../lib/versioning/Version');
+
+describe('Version', () => {
```
nicolas2bert: If you have already reviewed https://github.com/scality/Arsenal/pull/2215, there's no need to do so again as it is a duplicate.

lib/versioning/VersioningRequestProcessor.ts
```diff
@@ -481,19 +481,93 @@ export default class VersioningRequestProcessor {
...
+                            // isNull === false means Cloudserver supports null keys,
+                            // so create a null key in this case, and a version key otherwise
+                            const masterKeyVersionId = request.options.isNull === false ?
+                                  '' : masterVersionId;
+                            const masterVersionKey = formatVersionKey(key, masterKeyVersionId);
+                            masterVersion.setNullVersion();
+                            // isNull === false means Cloudserver supports null keys,
+                            // so create a null key with the isNull2 flag
+                            if (request.options.isNull === false) {
+                                masterVersion.setNull2Version();
```
nicolas2bert, line 562: TODO: Delete the deprecated null versioned key for backward compatibility. 
 If metadata is added to a null master, a null key is created (or updated if exists), 
so delete the deprecated (compat) null versioned key to avoid having two null keys in the listing.

lib/versioning/Version.ts
```diff
@@ -190,6 +262,16 @@ export class Version {
         return this;
     }
 
+    /**
+     * Mark that the null version has been put with a Cloudserver handling null keys (i.e. supporting S3C-7352)
+     * 
+     * @return - the updated version
+     */
+    setNull2Version() {
+        this.version.isNull2 = true;
```
jonathan-gramain, line 275: I think we should also set `isNull` as we should never have `isNull2` without `isNull`. Even if we use this function only in the case `isNull` is already set, I think it doesn't hurt.

lib/versioning/VersioningRequestProcessor.ts
```diff
@@ -496,7 +496,26 @@ export default class VersioningRequestProcessor {
...
+                    // Deprecated null key gets deleted when the new CloudServer:
+                    // - updates metadata of a null master (options.isNull=true)
+                    // - puts metadata on top of a master null key (options.isNull=false)
+                    if (request.options.isNull !== undefined && // new null key behavior when isNull is defined.
+                        !masterVersion.isNull2Version() && // master does not support the new null key behavior yet.
+                        masterVersion.isNullVersion()) { // master is null
+                        const masterNullVersionId = masterVersion.getNullVersionId();
+                        // The deprecated null key is referenced in the "nullVersionId" property of the master key.
+                        if (masterNullVersionId) {
+                            const oldNullKey = formatVersionKey(key, masterNullVersionId);
```
jonathan-gramain: I chose to call "null key" the key that stores the null version under the new behavior, which key ends with a null byte, hence in this context it may be better to rename the variable for example `oldNullVersionKey` to avoid this ambiguity.

lib/versioning/VersioningRequestProcessor.ts
```diff
@@ -481,19 +481,120 @@ export default class VersioningRequestProcessor {
...
+                    // to avoid storing both deprecated and new null keys.
+                    // If master null version was put with an older Cloudserver (or in compat mode),
+                    // there is a possibility that it also has a null versioned key
+                    // associated, so we need to delete it as we write the null key.
+                    // Deprecated null key gets deleted when the new CloudServer:
+                    // - updates metadata of a null master (options.isNull=true)
+                    // - puts metadata on top of a master null key (options.isNull=false)
+                    if (request.options.isNull !== undefined && // new null key behavior when isNull is defined.
+                        !masterVersion.isNull2Version() && // master does not support the new null key behavior yet.
+                        masterVersion.isNullVersion()) { // master is null
```
jonathan-gramain: Could be slightly more logical to swap the two checks
```suggestion
                        masterVersion.isNullVersion() && // master is null
                        !masterVersion.isNull2Version()) { // master does not support the new null key behavior yet.

```

lib/versioning/VersioningRequestProcessor.ts
```diff
@@ -537,13 +556,31 @@ export default class VersioningRequestProcessor {
...
+                                // If metadata is added to a null master, a null key is created (or updated if exists), 
+                                // so remove the deprecated null versioned key to avoid having 
+                                // two null keys in the listing.
+                                // const masterNullVersionId = masterVersion.getNullVersionId();
+                                // if (masterNullVersionId) {
+                                //     const oldNullKey = formatVersionKey(key, masterNullVersionId);
+                                //     ops.push({ key: oldNullKey, type: 'del' });
+                                // }
+                            // else isNull === undefined means Cloudserver does not support null keys,
+                            // hence set/update the new master nullVersionId for backward compatibility
```
jonathan-gramain: Can we remove this block?

jonathan-gramain (approved):
LGTM (thanks for updating this version-specific PUT function by the way, it was not obvious at all!)

nicolas2bert:
ping

BourgoisMickael (approved):

nicolas2bert:
@bert-e approve

PR #2226, opened 2024-02-26
ARSN-401: Add PRIMARY communication in cluster RPC

BourgoisMickael:
Adds:

- Communication with `PRIMARY` only
- Propagate werelogs config (logLevel) from dependant project
- `errorCode` in error communication (for HTTP)

This is used by scuba https://github.com/scality/scuba/pull/79

### DIFF
--- a/lib/clustering/ClusterRPC.ts
+++ b/lib/clustering/ClusterRPC.ts
@@ -90,6 +90,8 @@ export type HandlerFunction = (payload: object, uids: string, callback: HandlerC
 export type HandlersMap = {
     [index: string]: HandlerFunction;
 };
+export type PrimaryHandlerFunction = (worker: Worker, payload: object, uids: string, callback: HandlerCallback) => void;
+export type PrimaryHandlersMap = Record<string, PrimaryHandlerFunction>;
 
 // private types
 
@@ -165,12 +167,17 @@ function _isRpcMessage(message) {
 /**
  * Setup cluster RPC system on the primary
  *
+ * @param {object} [handlers] - mapping of handler names to handler functions
+ *     handler function:
+ *         `handler({Worker} worker, {object} payload, {string} uids, {function} callback)`
+ *     handler callback must be called when worker is done with the command:
+ *         `callback({Error|null} error, {any} [result])`
  * @return {undefined}
  */
-export function setupRPCPrimary() {
+export function setupRPCPrimary(handlers?: PrimaryHandlersMap) {
     cluster.on('message', (worker, message) => {
         if (_isRpcMessage(message)) {
-            _handlePrimaryMessage(worker?.id, message);
+            _handlePrimaryMessage(worker, message, handlers);
         }
     });
 }
@@ -201,8 +208,9 @@ export function setupRPCWorker(handlers: HandlersMap) {
  * Send a command for workers to execute in parallel, and wait for results
  *
  * @param {string} toWorkers - which workers should execute the command
- *     Currently the only supported value is "*", meaning all workers will
- *     execute the command
+ *     Currently the supported values are:
+ *         - "*", meaning all workers will execute the command
+ *         - "PRIMARY", meaning primary process will execute the command
  * @param {string} toHandler - name of handler that will execute the
  * command in workers, as declared in setupRPCWorker() parameter object
  * @param {string} uids - unique identifier of the command, must be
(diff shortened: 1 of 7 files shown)


BourgoisMickael:
reset

BourgoisMickael:
@bert-e reset

ghost (commented):
LGTM but I was expecting some unit tests, can we add them?

lib/clustering/ClusterRPC.ts
```diff
@@ -74,6 +74,7 @@ const rpcLogger = new werelogs.Logger('ClusterRPC');
 
 export type ResultObject = {
     error: Error | null;
+    errorCode?: number;
```
jonathan-gramain: I'd prefer having the error code part of the `error` object itself to keep it tidy, where the `Error` can be reconstructed from something serialized. I'm pretty sure the `Error` type accepts a `code` attribute. The current serialization seems to be a simple string now but it could be instead an object with two fields for example (say `code` and `message`) if the goal is to keep the description of the error along with the code.

The serialization/deserialization should stay internal to ClusterRPC and not exposed to clients so it should be fine to change how it's made.

jonathan-gramain (commented):
LGTM with one comment about the error extension.

I also agree with William that there should be some tests, you can have a look at the test server that is spawned as part of the functional tests and add one or two test cases there. During development you can just start the server locally with a custom port and use `curl` to trigger the tests (each test is declared in the server https://github.com/scality/Arsenal/blob/development/7.70/tests/functional/clustering/ClusterRPC-test-server.js with a name and is invoked from the test code via an HTTP request URL containing the test name). Once it works you can integrate the test case to the CI in https://github.com/scality/Arsenal/blob/development/7.70/tests/functional/clustering/ClusterRPC.spec.js#L64

BourgoisMickael:
@bert-e reset

lib/clustering/ClusterRPC.ts
BourgoisMickael: I moved the code to the `error`

tests/functional/clustering/ClusterRPC.spec.js
```diff
@@ -46,13 +46,45 @@ function stopTestServer(done) {
...
     const req = http.request(`http://localhost:${TEST_SERVER_PORT}/${testUrl}`, res => {
+        let responseBody = '';
         res
-            .on('data', () => {})
+            .on('data', (chunk) => {
+                responseBody += chunk;
+            })
             .on('end', () => {
-                expect(res.statusCode).toEqual(200);
-                cb();
```
BourgoisMickael, line 55: This would cause an exception to be thrown in a callback, that makes an `uncaughtException`, the `cb` is never called and jest hangs a long time until timeout.

tests/functional/clustering/ClusterRPC.spec.js
```diff
@@ -46,13 +46,45 @@ function stopTestServer(done) {
     testServer.on('close', done);
 }
 
+/**
+ * Try to deserialize and recreate AssertionError with stackTrace from spawned server
+ * @param {string} responseBody maybe serialized AssertionError
+ * @throws {assert.AssertionError}
+ * @returns {undefined}
+ */
+function handleAssertionError(responseBody) {
```
BourgoisMickael, line 55: It produces this output, pointing at the actual assertion from the other file:

```js
  ● ClusterRPC › worker to primary › should succeed and return a result

    assert.strictEqual(received, expected)

    Expected value to strictly be equal to:
      2
    Received:
      1
    
    Message:
      I want to fail

      231 |
      232 |     const results = await sendWorkerCommand('PRIMARY', 'echoHandler', uids, payload);
    > 233 |     assert.strictEqual(1, 2, 'I want to fail');
          |            ^
      234 |     assert.strictEqual(results.length, 1, 'There is 1 and only 1 primary');
      235 |     assert.ifError(results[0].error);
      236 |     assert.deepStrictEqual(results[0].result, expected);

      at Object.workerToPrimaryEcho [as /worker-to-primary/echo] (tests/functional/clustering/ClusterRPC-test-server.js:233:12)

```

tests/functional/clustering/ClusterRPC.spec.js
```diff
@@ -46,13 +46,45 @@ function stopTestServer(done) {
...
+            })
             .on('end', () => {
-                expect(res.statusCode).toEqual(200);
-                cb();
+                try {
+                    handleAssertionError(responseBody);
+                    expect(res.statusCode).toEqual(200);
+                } catch (err) {
+                    if (!(err instanceof assert.AssertionError)) {
+                        err.message += `\n\nBody:\n${responseBody}`;
```
BourgoisMickael, line 83: Produces:
```js
Expected: 200
Received: 400

Body:
MissingParameter
```

jonathan-gramain (approved):

anurag4DSB (approved):
Scope of review: Reviewed from code point of view, still understanding MD
LGTM.
This technique is very interesting. Correct me if I understood wrong. This is useful in scenarios where a library or module (arsenal in this case) needs to accept a configuration object that is meant to be passed to another library or module (werelogs). By directly linking the type of the configuration object to the expected parameter type of the werelogs.configure method, it helps to ensure that the arsenal library will correctly propagate the configuration object to werelogs, even if future versions of werelogs change the expected configuration object structure.

jonathan-gramain:
> This is useful in scenarios where a library or module (arsenal in this case) needs to accept a configuration object that is meant to be passed to another library or module (werelogs). By directly linking the type of the configuration object to the expected parameter type of the werelogs.configure method, it helps to ensure that the arsenal library will correctly propagate the configuration object to werelogs, even if future versions of werelogs change the expected configuration object structure.

My understanding is that, but I may not have the full picture: if we fork a worker before we configure werelogs, we will need to apply the werelogs configuration to both the primary and the worker(s), because they're now different processes. But if we apply the config prior to the fork, then the worker will be already configured since the worker process memory maps the primary's memory (and copy-on-write when written to).

PR #2227, opened 2024-02-28
bf: ARSN-402 use local RequestLogger in batchDelete

jonathan-gramain:
**Note for reviews**: easier to review individual commits.

- bump werelogs dependency

  + typescript fixes to be compatible with the latest werelogs

- use local RequestLogger in DataWrapper.batchDelete

  Create a local RequestLogger in batchDelete(): this allows to track the elapsed time of the batch delete sub-request, and avoids being forced to create a new request logger before calling the function (due to the call to `log.end()`), which was error-prone and hardly maintainable.

- sanitize use of log object in DataWrapper.delete()

  Don't assume that we can safely call `end()` on the passed log object if there is no callback (separation of concerns). additionally, an error object was passed where `end()` expects a string as a message, causing implicit conversion.

  Since errors are already logged, there is no need to bind the `callback` object to `log.end` (there is no strong reason to log the elapsed time there, the only use I can see where we don't pass a callback in Cloudserver is to support deletion of old metadata with a string as location array. IMHO not worth the complexity of adding it there, as the rest of the API doesn't log elapsed time anyways except for `batchDelete`).

### DIFF
--- a/lib/models/BackendInfo.ts
+++ b/lib/models/BackendInfo.ts
@@ -1,3 +1,5 @@
+import { RequestLogger } from 'werelogs';
+
 import { legacyLocations } from '../constants';
 import escapeForXml from '../s3middleware/escapeForXml';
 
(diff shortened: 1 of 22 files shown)


jonathan-gramain:
ping

ghost (approved):

nicolas2bert (approved):

PR #2230, opened 2024-03-04
ARSN-403 Set nullVersionId to master when replacing a null version.

nicolas2bert:
- Should set `nullVersionId` to the master version if putting a version on top of a null version.

- Should not set `nullVersionId` to the master version if updating a null version.

### DIFF
--- a/tests/unit/versioning/VersioningRequestProcessor.spec.js
+++ b/tests/unit/versioning/VersioningRequestProcessor.spec.js
@@ -380,7 +380,7 @@ describe('test VSP', () => {
             const request = {
                 db: 'foo',
                 key: 'bar',
-                value: '{"qux":"quz2","isNull":true}',
+                value: `{"qux":"quz2","isNull":true,"versionId":"${nullVersionId}"}`,
                 options: {
                     versioning: true,
                     versionId: nullVersionId,
@@ -396,7 +396,7 @@ describe('test VSP', () => {
                 // NOTE: should not set nullVersionId to the master version if updating a null version.
                 {
                     key: 'bar',
-                    value: '{"qux":"quz2","isNull":true}',
+                    value: `{"qux":"quz2","isNull":true,"versionId":"${nullVersionId}"}`,
                 },
                 {
                     key: `bar\x00${nullVersionId}`,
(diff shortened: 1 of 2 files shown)


nicolas2bert:
@francoisferrand this change impacts `file` and `Metadata` metadata backend only.

tests/unit/versioning/VersioningRequestProcessor.spec.js
```diff
@@ -357,4 +358,67 @@ describe('test VSP', () => {
...
+            const expectedListing = [
+                // NOTE: should not set nullVersionId to the master version if updating a null version.
+                {
+                    key: 'bar',
+                    value: '{"qux":"quz2","isNull":true}',
+                },
+                {
+                    key: `bar\x00${nullVersionId}`,
+                    value: `{"qux":"quz","isNull":true,"versionId":"${nullVersionId}"}`,
+                },
```
jonathan-gramain, line 404: I don't understand why we have a second version here.

For me what should happen is:
- we create a null version on a suspended-versioning bucket -> creates a master key with `isNull` and an internal version ID
- we update that same null version -> master is updated with the new attributes and keeps same version ID

I don't understand why we have a version key in this case and why we end up with two different keys.

nicolas2bert: Because https://scality.atlassian.net/browse/MD-843 seems to have been reverted and so a new version gets created when we update a latest null version.

Also, it's a minor update, but I've upated the test to reflect the call made by putMetadata, including the addition of the versionId in the updated value.

nicolas2bert: About the fact that we end up with a version key with an outdated value.

I am not quite sure if it is by design or not but it does not seem to be a problem. 
Once PUT object on top of this latest null version key:
- the master gets updated with the new version value,
- the null version gets updated with the old master value.

So the null version has the correct value.

nicolas2bert: If it is a bug, I can create a following ticket for it.

jonathan-gramain: Ok good for me, indeed the duplicate key was not fixed on the 7.10 branch. We should make sure that we don't have a regression on 7.70 (I think the existing tests should cover this).

jonathan-gramain (approved):

BourgoisMickael (approved):

nicolas2bert:
@bert-e approve

PR #2232, opened 2024-03-18
ARSN-404: adding permission in BP and IAM action Map

benzekrimaha:
This pull request primarily focuses on adding new action maps related to bucket and account quotas. The changes allow for the handling of new actions such as getting, updating, and deleting bucket and account quotas.

Issue : ARSN-404

### DIFF
--- a/lib/policyEvaluator/utils/actionMaps.ts
+++ b/lib/policyEvaluator/utils/actionMaps.ts
@@ -52,6 +52,12 @@ const sharedActionMap = {
     objectPutVersion: 's3:PutObjectVersion',
 };
 
+const actionMapBucketQuotas = {
+    bucketGetQuota: 'scality:GetBucketQuota',
+    bucketUpdateQuota: 'scality:UpdateBucketQuota',
+    bucketDeleteQuota: 'scality:DeleteBucketQuota',
+};
+
 // action map used for request context
 const actionMapRQ = {
     bucketPut: 's3:CreateBucket',
@@ -79,6 +85,7 @@ const actionMapRQ = {
     objectPutLegalHoldVersion: 's3:PutObjectLegalHold',
     listObjectVersions: 's3:ListBucketVersions',
     ...sharedActionMap,
+    ...actionMapBucketQuotas,
 };
 
 // action map used for bucket policies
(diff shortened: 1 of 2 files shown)


lib/policyEvaluator/utils/actionMaps.ts
```diff
@@ -52,6 +52,12 @@ const sharedActionMap = {
     objectPutVersion: 's3:PutObjectVersion',
 };
 
+const actionMapBucketQuotas = {
+    getBucketQuota: 'quota:GetBucketQuota',
+    updateBucketQuota: 'quota:PutBucketQuota',
+    deleteBucketQuota: 'quota:PutBucketQuota',
```
ghost: I wonder if we should not just consider the APIs as the others, when we deal with specific permissions, like `Put*` when we actually delete it, in such case, I see no need to duplicate the value here, we can just pass the right permission: this also makes it clear in the code what the permission we look for is

benzekrimaha: As discussed we're separating the permissions now : Done here : https://github.com/scality/Arsenal/pull/2232/commits/b3a5ce5658803f54a2bb32c671d38be2a0d244c0#diff-7fb88a375e077c195e7a6345016d77a135949a6da340c76c6b57c0e3d05cccc9R164

ghost (approved):
Worth having the version bump in the PR

lib/policyEvaluator/utils/actionMaps.ts
```diff
@@ -52,6 +52,12 @@ const sharedActionMap = {
     objectPutVersion: 's3:PutObjectVersion',
 };
 
+const actionMapBucketQuotas = {
+    getBucketQuota: 'quota:GetBucketQuota',
```
francoisferrand: these `actions` will be user facing, right (e.g. in policies) ? so i think we should think a bit more about the naming here:
* should we use a `scality` prefix (instead of `quota`) ? → avoids any conflict, and make sure to isolate our own APIs vs AWS'
* or maybe do we need to keep separate prefix for each "section" of our api, so maybe `scuba:GetBucketQuota` or `sur:GetBucketQuota` ?

@williamlardier @rachedbenmustapha what do you think?

ghost: Latest discussions with Taylor also confirms `scality:` is the right choice. We do not want to have `scuba:` for this. This PR (and design) will be updated accordingly

benzekrimaha: Done here : https://github.com/scality/Arsenal/pull/2232/commits/3f995c5f5a2a8e328ef9db2231c73d363a97e34e

francoisferrand (approved):

anurag4DSB (approved):

benzekrimaha:
ping

PR #2233, opened 2024-04-03
ARSN-406: add APIs in the needQuotas checks list

ghost:
Necessary changes to support quotas in S3, and cover the missing APIs to consider when evaluating quotas.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,6 +1,9 @@
 import * as evaluators from './lib/policyEvaluator/evaluator';
 import evaluatePrincipal from './lib/policyEvaluator/principal';
-import RequestContext from './lib/policyEvaluator/RequestContext';
+import RequestContext, {
+    actionNeedQuotaCheck,
+    actionNeedQuotaCheckCopy,
+    actionWithDataDeletion } from './lib/policyEvaluator/RequestContext';
 import * as requestUtils from './lib/policyEvaluator/requestUtils';
 import * as actionMaps from './lib/policyEvaluator/utils/actionMaps';
 import { validateUserPolicy } from './lib/policy/policyValidator'
@@ -67,6 +70,9 @@ export const policies = {
     RequestContext,
     requestUtils,
     actionMaps,
+    actionNeedQuotaCheck,
+    actionWithDataDeletion,
+    actionNeedQuotaCheckCopy,
 };
 
 export const testing = {
--- a/lib/policyEvaluator/RequestContext.ts
+++ b/lib/policyEvaluator/RequestContext.ts
@@ -15,10 +15,35 @@ import {
     actionMapScuba,
 } from './utils/actionMaps';
 
-const _actionNeedQuotaCheck = {
+export const actionNeedQuotaCheck = {
     objectPut: true,
+    objectPutVersion: true,
     objectPutPart: true,
+    objectRestore: true,
 };
+
+/**
+ * This variable describes APIs that change the bytes
+ * stored, requiring quota updates
+ */
+export const actionWithDataDeletion = {
+    objectDelete: true,
+    objectDeleteVersion: true,
+    multipartDelete: true,
+    multiObjectDelete: true,
+};
+
+/**
+ * The function returns true if the current API call is a copy object
+ * and the action requires a quota evaluation logic, post retrieval
+ * of the object metadata.
+ * @param {string} action - the action being performed
+ * @param {string} currentApi - the current API being called
+ * @return {boolean} - whether the action requires a quota check
+ */
+export function actionNeedQuotaCheckCopy(action: string, currentApi: string) {
+    return action === 'objectGet' && (currentApi === 'objectCopy' || currentApi === 'objectPutCopyPart');
+}
 
 function _findAction(service: string, method: string) {
     switch (service) {
@@ -231,7 +256,8 @@ export default class RequestContext {
         this._securityToken = securityToken;
         this._policyArn = policyArn;
         this._action = action;
-        this._needQuota = _actionNeedQuotaCheck[apiMethod] === true;
+        this._needQuota = actionNeedQuotaCheck[apiMethod] === true
+            || actionWithDataDeletion[apiMethod] === true;
         this._requestObjTags = requestObjTags || null;
         this._existingObjTag = existingObjTag || null;
         this._needTagEval = needTagEval || false;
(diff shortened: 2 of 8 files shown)


benzekrimaha (approved):

lib/policyEvaluator/RequestContext.ts
```diff
@@ -15,9 +15,14 @@ import {
...
 
-const _actionNeedQuotaCheck = {
+export const actionNeedQuotaCheck = {
     objectPut: true,
+    objectPutVersion: true,
     objectPutPart: true,
+    objectRestore: true,
+    objectDelete: true,
+    multipartDelete: true,
+    objectDeleteVersion: true,
```
francoisferrand: not really a quota _check_ in that case : in the sense that these API must never fail due to the quota....
e.g. if quota is way higher than expected (i.e. due to overshoot or because it was lowever), the usage "after" delete may still be higher than quota, but should be accepted nonetheless...

→ we must make sure these 3 API should never return quotaExceeded

ghost: All delete APIs trigger quota evaluation logic - the logic, in case of quota exceeded, would first increase the inflights (scuba is not aware of the quota) and, upon quota exceeded detection, clear the last increase. In the case of deletes, we simply decrese everytime, and (in cloudserver), we explicitly ensure that any deletion always pass.

As this is something to be done in cloudserver, is this comment about the variable naming?

francoisferrand: Yes, making sure it cannot fail may indeed be in cloudserver. So:
* variable naming here may be misleading : and could cause confusion (or bugs) when people change (cloudserver) generic code in the future...
* checking the quota "check" may be expensive and we don't need the full evaluation in case of deletes (as you mentioned), only to track the "decrease" : so it would seem more appropriate to differentiate between these case, possibly returning 3 values (do not check, check quota, decrease usage)...

Basically, by bundling all APIs in the same "needQuotasCheck" we introduce coupling between this list of API and the details of the current implementation.... which should be avoided, to facilitate future maintenance.

ghost: Makes sense, we can split this variable into two: need quota check (increase), need quota check (decrease). Quota exceeded case will anyway require an additional call. In the case of deletes, we only issue a single call anyway.

ghost: Done in https://github.com/scality/Arsenal/pull/2233/commits/c5ab5d034c5b988c7a5c253427c4ce6237661604

ghost:
@ reviewers; added https://github.com/scality/Arsenal/pull/2233/commits/63cc54d3fa3c859773cabeef840fe8aad9e3ccae to compute the number of bucket with quota as part of the periodic report.

lib/policyEvaluator/RequestContext.ts
```diff
@@ -15,9 +15,19 @@ import {
...
-const _actionNeedQuotaCheck = {
+export const actionNeedQuotaCheck = {
     objectPut: true,
+    objectPutVersion: true,
     objectPutPart: true,
+    objectRestore: true,
+};
+
+// This variable describe APIs that change the bytes
+// stored, requiring quota updates
```
BourgoisMickael: It can be nice to use jsdoc `/** */` to have description in tooltip on hover in vscode

lib/policyEvaluator/RequestContext.ts
```diff
@@ -15,9 +15,19 @@ import {
...
+export const actionWithDataDeletion = {
```
BourgoisMickael, line 29: I think there is also the POST `multiObjectDelete`

lib/policyEvaluator/RequestContext.ts
```diff
@@ -15,9 +15,19 @@ import {
...
+    objectDelete: true,
+    objectDeleteVersion: true,
```
BourgoisMickael, line 31: It depends where you take the `apiMethod`, it seems you take it from `request.apiMethod`, in that case, I think there is no `objectDeleteVersion` (https://github.com/scality/Arsenal/blob/development/8.1/lib/s3routes/routes/routeDELETE.ts#L52) it is mixed in the `objectDelete` handler.

The `Version` suffix is added in the RequestContext for Vault https://github.com/scality/cloudserver/blob/development/8.7/lib/api/apiUtils/authorization/prepareRequestContexts.js#L61

lib/policyEvaluator/RequestContext.ts
ghost: this API will individually call the deleteObject internal name, so this is not needed

edit: multiObjectDelete does not delete anything, it's the subsequent (internal) calls that are deleting the objects provided as the input, after getting their metadata and authorizing with vault, for each object.

lib/policyEvaluator/RequestContext.ts
ghost: When evaluating the quotas, we do it for each of the returned actions from the authorization results. Today it's only the action provided as the input, but we might eventually have more actions (see below). These actions can have `Version` suffix.

We use the `request.apiMethod` only at the end, for the monitoring, where it will indeed not be `objectDeleteVersion`. So we need to support them.

Now, for the multi object delete, indeed I think I am not properly handling it in the internal additional calls, when we authorize one by one all the objects we want to delete. I will work on that.

lib/policyEvaluator/RequestContext.ts
ghost: See https://github.com/scality/Arsenal/pull/2233/commits/c412ada50646f284e9863d5a3f1b0adea951b91e

BourgoisMickael (approved):

lib/policyEvaluator/RequestContext.ts
ghost: Just an update: I added the inflights for the multi delete in cloudserver.

lib/policyEvaluator/RequestContext.ts
ghost: Finally added it for another reason, vault uses the _needQuota flag in the request context to output the info about the quota from the account; we thus need this value if we want to be able to evaluate account quotas in this API

francoisferrand (approved):

index.ts
```diff
@@ -67,6 +69,8 @@ export const policies = {
     RequestContext,
     requestUtils,
     actionMaps,
+    actionNeedQuotaCheck,
+    actionWithDataDeletion,
```
francoisferrand, line 74: `actionNeedQuotaRelease` ?

would be better to convey the relation with quota...

francoisferrand (approved):

PR #2238, opened 2024-04-26
ARSN-413 Replicating null version fails

nicolas2bert:
**This PR contains only cherry-picked code.**

_NOTE: Null versions are not being replicated unless triggered by CrrExistingObject s3utils script (or OOB Ingested in Zenko/Artesca)_

For "null" versions that are created before bucket versioning is configured, CloudServer/Arsenal generates a "null" version ID (`99999999999999999999RG001`) used **only** internally by metadata.

However, Backbeat tries to decode this internal version ID using `ObjectMD.getEncodedVersionId()` with no success and so fails to retrieve the "null" version.

### DIFF
--- a/lib/models/ObjectMD.ts
+++ b/lib/models/ObjectMD.ts
@@ -1,5 +1,6 @@
 import * as constants from '../constants';
 import * as VersionIDUtils from '../versioning/VersionID';
+import { VersioningConstants } from '../versioning/constants';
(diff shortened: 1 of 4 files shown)


anurag4DSB (approved):

BourgoisMickael (approved):

nicolas2bert:
@bert-e approve

PR #2249, opened 2024-07-02
Improvement/arsn 420 introduce put object hack

benzekrimaha:
Details on the implementation are within the commit description 

Issue : ARSN-420

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -899,35 +899,130 @@ class MongoClientInterface {
             return cb(errors.InternalError);
         });
     }
-
     /**
-     * Put object when versioning is not enabled
-     * @param {Object} c bucket collection
-     * @param {String} bucketName bucket name
-     * @param {String} objName object name
-     * @param {Object} objVal object metadata
-     * @param {Object} params params
-     * @param {Object} log logger
-     * @param {Function} cb callback
-     * @return {undefined}
+     * Puts an object into a MongoDB collection.
+     * Depending on the parameters, the object is either directly put into the collection
+     * or the existing object is marked as deleted and a new object is inserted.
+     *
+     * @param {Object} collection - The MongoDB collection to put the object into.
+     * @param {string} bucketName - The name of the bucket the object belongs to.
+     * @param {string} objName - The name of the object.
+     * @param {Object} value - The value of the object.
+     * @param {Object} params - Additional parameters.
+     * @param {string} params.vFormat - object key format.
+     * @param {boolean} params.needOplogUpdate - If true, the object is directly put into the collection
+     * with updating the operation log.
+     * @param {Object} log - The logger to use.
+     * @param {Function} cb - The callback function to call when the operation is complete. It is called with an error
+     * if there is an issue with the operation.
+     * @returns {Promise} A promise that resolves when the operation is complete. The promise is rejected with an error
+     * if there is an issue with the operation.
      */
-    putObjectNoVer(c, bucketName, objName, objVal, params, log, cb) {
-        const masterKey = formatMasterKey(objName, params.vFormat);
-        c.updateOne({
-            _id: masterKey,
-        }, {
+    putObjectNoVer(collection, bucketName, objName, value, params, log, cb) {
+        if (params?.needOplogUpdate) {
+            return this.putObjectNoVerWithOplogUpdate(collection, bucketName, objName, value, params, log, cb);
+        }
+        const key = formatMasterKey(objName, params.vFormat);
+        const putFilter = { _id: key };
+        return collection.updateOne(putFilter, {
             $set: {
-                _id: masterKey,
-                value: objVal,
+                _id: key,
+                value,
             },
         }, {
             upsert: true,
-        }).then(() => cb()).catch((err) => {
+        }).then(() => cb()).catch(err => {
             log.error('putObjectNoVer: error putting obect with no versioning', { error: err.message });
             return cb(errors.InternalError);
         });
     }
 
+    /**
+     * Updates an object in a MongoDB collection without changing its version.
+     * If the object doesn't exist, it will be created (upsert is true for the second update operation).
+     * The operation is logged in the oplog.
+     *
+     * @param {Object} collection - The MongoDB collection to update the object in.
+     * @param {string} bucketName - The name of the bucket the object belongs to.
+     * @param {string} objName - The name of the object.
+     * @param {Object} value - The new value of the object.
+     * @param {Object} params - Additional parameters.
+     * @param {string} params.vFormat - object key format
+     * @param {string} params.originOp - origin operation
+     * @param {Object} log - The logger to use.
+     * @param {Function} cb - The callback function to call when the operation is complete.
+     * It is called with an error if there is an issue with the operation.
+     * @returns {void}
+     */
+    putObjectNoVerWithOplogUpdate(collection, bucketName, objName, value, params, log, cb) {
+        const key = formatMasterKey(objName, params.vFormat);
+        const putFilter = { _id: key };
+        // filter used when finding and updating object
+        const findFilter = {
+            ...putFilter,
+            $or: [
+                { 'value.deleted': { $exists: false } },
+                { 'value.deleted': { $eq: false } },
+            ],
+        };
+        const updateDeleteFilter = {
+            ...putFilter,
+            'value.deleted': true,
+        };
+        return async.waterfall([
+            // Adding delete flag when getting the object
+            // to avoid having race conditions.
+            next => collection.findOneAndUpdate(findFilter, {
+                $set: updateDeleteFilter,
+            }, {
+                upsert: false,
+            }).then(doc => {
+                if (!doc.value) {
+                    log.error('internalPutObject: unable to find target object to update',
+                        { bucket: bucketName, object: key });
+                    return next(errors.NoSuchKey);
+                }
+                const obj = doc.value;
+                const objMetadata = new ObjectMD(obj.value);
+                objMetadata.setOriginOp(params.originOp);
+                objMetadata.setDeleted(true);
+                return next(null, objMetadata.getValue());
+            }).catch(err => {
+                log.error('internalPutObject: error getting object',
+                    { bucket: bucketName, object: key, error: err.message });
+                return next(errors.InternalError);
+            }),
+            // We update the full object to get the whole object metadata
+            // in the oplog update event
+            (objMetadata, next) => collection.bulkWrite([
+                {
+                    updateOne: {
+                        filter: updateDeleteFilter,
+                        update: {
+                            $set: { _id: key, value: objMetadata },
+                        },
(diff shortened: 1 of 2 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -899,35 +899,113 @@ class MongoClientInterface {
...
-            },
-        }, {
-            upsert: true,
-        }).then(() => cb()).catch((err) => {
-            log.error('putObjectNoVer: error putting obect with no versioning', { error: err.message });
-            return cb(errors.InternalError);
+    putObjectNoVer(collection, bucketName, objName, value, params, log, cb, originOp = 's3:LifecycleTransitionPut') {
+        const key = formatMasterKey(objName, params.vFormat);
+        const putFilter = { _id: key };
+        if (params && !params.doesNeedOplogUpdate) {
```
francoisferrand: if no param, should stick to the "default" behavior
i.e.
```suggestion
        if (!params || !params.needOplogUpdate) {
```
or better yet
```suggestion
        if (!params?.needOplogUpdate) {
```

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -899,35 +899,113 @@ class MongoClientInterface {
...
-     * @return {undefined}
+     * Puts an object into a MongoDB collection. Depending on the parameters,
+     * the object is either directly put into the collection
+     * or the existing object is marked as deleted and a new object is inserted.
+     *
+     * @param {Object} collection - The MongoDB collection to put the object into.
+     * @param {string} bucketName - The name of the bucket the object belongs to.
+     * @param {string} objName - The name of the object.
+     * @param {Object} value - The value of the object.
+     * @param {Object} params - Additional parameters. If params.doesNotNeedOpogUpdate is true, the object is directly
```
francoisferrand: should document the individual parameters:

```suggestion
     * @param {Object} params - Additional parameters
     * @param {bool} params.vFormat - ...
     * @param {bool} params.doesNotNeedOpogUpdate - fs true, the object is directly
```

lib/storage/metadata/mongoclient/MongoClientInterface.js
francoisferrand: somehow, the "structure" of the function looks a bit misleading to be: by having the negative condition, it makes it look like the "standard" case if to update log...

maybe worth reworking the overal organization of the change to make it more clear, something like:

```javascript
putObjectNoVer(...) {
    if (params.needOplogUpdate) {
        return putObjectNoVerWithOplogUpdate()
    }
    // [...] proceed with regular put object (old code)
}
```

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -899,35 +899,113 @@ class MongoClientInterface {
...
+        }, putFilter);
+
+        return async.waterfall([
+            // Adding delete flag when getting the object
+            // to avoid having race conditions.
+            next => collection.findOneAndUpdate(findFilter, {
+                $set: {
+                    '_id': key,
+                    'value.deleted': true,
+                },
```
francoisferrand: this is actually `updateDeleteFilter`...

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -899,35 +899,113 @@ class MongoClientInterface {
...
-            $set: {
-                _id: masterKey,
-                value: objVal,
-            },
-        }, {
-            upsert: true,
-        }).then(() => cb()).catch((err) => {
-            log.error('putObjectNoVer: error putting obect with no versioning', { error: err.message });
-            return cb(errors.InternalError);
+    putObjectNoVer(collection, bucketName, objName, value, params, log, cb, originOp = 's3:LifecycleTransitionPut') {
```
francoisferrand: why would it have such a default?
- this is an abstract metadata interface, there should not be any default related to a specific use case.
- in the scenario we are trying to solve, there is no "lifecycle" on this specific (user!) operation: this should be more something like `ReplaceArchivedObject`... 

* maybe just using `PutObject` would be enough to identify this case (e.g. `action=s3:PutObject` with `deleted=true`) without triggering unexpected event. This would allow getting the info cleanly from the object value, and precisely reflect what operation happened...
* if you need to introduce an extra `originOp`, it should come from cloudserver --> extra `params`

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -899,35 +899,113 @@ class MongoClientInterface {
...
+                {
+                    updateOne: {
+                        filter: putFilter,
+                        update: {
+                            $set: { _id: key, value },
+                        },
+                        upsert: true,
+                    },
+                },
+            ], { ordered: true }).then(() => next(null)).catch(err => next(err)),
```
francoisferrand:
```suggestion
            ], { ordered: true }).then(() => next(null)).catch(next),
```

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -973,8 +1051,7 @@ class MongoClientInterface {
                 return putObjectVer(c, bucketName, objName, objVal, _params, log,
                     cb);
             }
-            return this.putObjectNoVer(c, bucketName, objName, objVal,
-                _params, log, cb);
+            return this.putObjectNoVer(c, bucketName, objName, objVal, _params, log, cb);
```
francoisferrand, line 1071: no change here?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -899,35 +899,132 @@ class MongoClientInterface {
...
+                value,
             },
         }, {
             upsert: true,
-        }).then(() => cb()).catch((err) => {
-            log.error('putObjectNoVer: error putting obect with no versioning', { error: err.message });
+        }).then(() => cb()).catch(err => {
+            log.error('internalPutObject: error putting object',
+                { bucket: bucketName, object: key, error: err.message });
             return cb(errors.InternalError);
```
francoisferrand: probably no point changing this, old code works just as well ?
(and `putObjectNoVer: error putting obect with no versioning` error message was more appropriate, giving details about which part of the code fails...)

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -899,35 +899,132 @@ class MongoClientInterface {
...
+     * @param {Function} cb - The callback function to call when the operation is complete.
+     * It is called with an error if there is an issue with the operation.
+     * @returns {void}
+     */
+    putObjectNoVerWithOplogUpdate(collection, bucketName, objName, value, params, log, cb) {
+        const key = formatMasterKey(objName, params.vFormat);
+        const putFilter = { _id: key };
+        // filter used when finding and updating object
+        const findFilter = Object.assign({
+            _id: key,
```
francoisferrand: redundant with `, putFilter` a few lines below...

also, whenever possible, prefer to use `...` to "unpack" object instead of Object.assign, as it will be easier to migrate to typescript


```suggestion
    const findFilter = {
        ...putFilter,
        $or: [
            { 'value.deleted': { $exists: false } },
            { 'value.deleted': { $eq: false } },
        ],
    };
```

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -899,35 +899,132 @@ class MongoClientInterface {
...
+        }, putFilter);
+        const updateDeleteFilter = Object.assign({
+            '_id': key,
+            'value.deleted': true,
+        }, putFilter);
+        return async.waterfall([
+            // Adding delete flag when getting the object
+            // to avoid having race conditions.
+            next => collection.findOneAndUpdate(findFilter, {
+                $set: updateDeleteFilter,
```
francoisferrand, line 976: overall, I wonder if the introduction of all these filters actually helps with readability: because it reduces locality and decouples the function call, filter and "update". What do you think?

francoisferrand (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -899,35 +899,132 @@ class MongoClientInterface {
...
+            }, {
+                upsert: false,
+            }).then(doc => {
+                log.info('doc value', { doc: doc.value });
```
welansari: Is this a leftover from debugging ?

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -899,35 +899,132 @@ class MongoClientInterface {
...
+            _id: key,
+            $or: [
+                { 'value.deleted': { $exists: false } },
+                { 'value.deleted': { $eq: false } },
+            ],
+        }, putFilter);
+        const updateDeleteFilter = Object.assign({
+            '_id': key,
+            'value.deleted': true,
+        }, putFilter);
```
welansari: Not sure what's the point of `Object.assign` since we already set the `_id` key in the filters ?

lib/storage/metadata/mongoclient/MongoClientInterface.js
benzekrimaha: The fact that the filters are defined at the top may make it hard to follow indeed but it  reduces redundancy , and maintainability is easier than in the inline approach. We either make the code DRY (Don't Repeat Yourself) or easier to understand with the inline definition on each mongo op ( I don't really have reserves on both approaches this one just seemed cleaner to me)

lib/storage/metadata/mongoclient/MongoClientInterface.js
benzekrimaha: indeed , fixed here : https://github.com/scality/Arsenal/pull/2249/commits/043b64c3d17ad2bbcc3ea34ddc2b3f6590c436b1#diff-9c207befbd8f4d79a3b6e3d7d143444e804ca1684236b8fd9fa0f9e03c53477aR961-R971

lib/storage/metadata/mongoclient/MongoClientInterface.js
benzekrimaha: done here : https://github.com/scality/Arsenal/pull/2249/commits/043b64c3d17ad2bbcc3ea34ddc2b3f6590c436b1#diff-9c207befbd8f4d79a3b6e3d7d143444e804ca1684236b8fd9fa0f9e03c53477aR961-R971

lib/storage/metadata/mongoclient/MongoClientInterface.js
benzekrimaha: code was kept as is, the error message is fixed here : https://github.com/scality/Arsenal/pull/2249/commits/043b64c3d17ad2bbcc3ea34ddc2b3f6590c436b1#diff-9c207befbd8f4d79a3b6e3d7d143444e804ca1684236b8fd9fa0f9e03c53477aR935

ghost (approved):

PR #2251, opened 2024-07-08
bf: ARSN-425 listing crash if key contains "undefined"

jonathan-gramain:
Fix a crash in DelimiterMaster listing without a delimiter, when a key contains the string "undefined".

Note: a similar fix was done in ARSN-330 for DelimiterVersions. I ported the existing unit test there to the development/7.10 branch to enhance regression testing, even though this bug on DelimiterVersions
only existed on 7.70.

### DIFF
--- a/lib/algos/list/delimiter.ts
+++ b/lib/algos/list/delimiter.ts
@@ -196,6 +196,9 @@ export class Delimiter extends Extension {
     }
 
     getCommonPrefix(key: string): string | undefined {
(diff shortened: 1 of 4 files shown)


BourgoisMickael (approved):
I think it should be hotfixed in `hotfix/7.70.21` for metadata

https://github.com/scality/MetaData/blob/hotfix/7.70.21/package.json#L34

tmacro (approved):

PR #2256, opened 2024-08-28
Remove PII from log

ghost:
In case of error, the debug log would output PII.

### DIFF
--- a/lib/auth/Vault.ts
+++ b/lib/auth/Vault.ts
@@ -1,6 +1,6 @@
 import { Logger } from 'werelogs';
 import errors from '../errors';
-import AuthInfo from './AuthInfo';
+import AuthInfo, { AuthInfoType } from './AuthInfo';
 
 /** vaultSignatureCb parses message from Vault and instantiates
  * @param err - error from vault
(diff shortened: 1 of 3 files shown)


KillianG (approved):

lib/auth/Vault.ts
```diff
@@ -26,7 +33,20 @@ function vaultSignatureCb(
...
         return callback(err);
     }
-    log.debug('received info from Vault', { authInfo });
+
+    const { email, ...userInfoWithoutEmail } = authInfo.message.body.userInfo;
+
+    log.debug('received info from Vault', {
+        ...authInfo,
+        message: {
+            message: authInfo.message,
```
welansari:
```suggestion
            message: authInfo.message.message,
```

lib/auth/Vault.ts
```diff
@@ -12,7 +12,14 @@ import AuthInfo from './AuthInfo';
  */
 function vaultSignatureCb(
     err: Error | null,
-    authInfo: { message: { body: any } },
+    authInfo: {
+        message: {
+            message: string, body: {
+                userInfo: AuthInfoType,
```
welansari:
```suggestion
        message: {
            message: string,
            body: {
                userInfo: AuthInfoType,
```

ghost: Done + fixed the type for the quotas

welansari (approved):

PR #2257, opened 2024-09-13
Ensure getBucketInfos acces only mongo collections

francoisferrand:
This is especially needed after PRA bootstrap was executed, as we
started using views in that context.

Issue: ARSN-432

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -2099,7 +2099,7 @@ class MongoClientInterface {
         let bucketCount = 0;
         const bucketInfos = [];
 
-        this.db.listCollections().toArray().then(collInfos =>
+        this.db.listCollections({ filter: { type: 'collection' } }).toArray().then(collInfos =>
             async.eachLimit(collInfos, 10, (value, next) => {
                 if (this._isSpecialCollection(value.name)) {
                     // skip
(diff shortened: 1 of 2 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -2099,7 +2099,7 @@ class MongoClientInterface {
         let bucketCount = 0;
         const bucketInfos = [];
 
-        this.db.listCollections().toArray().then(collInfos =>
+        this.db.listCollections({ filter: { type: 'collection' } }).toArray().then(collInfos =>
```
ghost, line 2102: Is this function used in prod, not a dead code? It's used by arsenal from "countItems" and "scanItemCount", but we do not use these functions in s3utils anymore IIRC, we use https://github.com/scality/s3utils/blob/development/1.14/utils/S3UtilsMongoClient.js#L849 instead. I believe this code is here as a result of Armory's backport.
If we can confirm these are not used (beyond some tests in cloudserver), we should just drop them.

francoisferrand: It is used in Cloudserver's "metrics" pod

francoisferrand: (And the copy from s3utils should be removed to dedup' the code, relying on the single implementation here - i created a ticket for this already)

welansari (approved):

ghost (approved):

PR #2258, opened 2024-09-24
ARSN-430 Add new KMS backend with account-level default encryption key management

nicolas2bert:
Context:

- Introduce a new KMS backend that supports the AWS KMS protocol.
- Implement a more efficient approach to managing default encryption keys. Instead of creating a master encryption key per bucket, a single default master encryption key is associated with the account. This simplifies key management. The default key id will be stored in the account metadata.

Changes in this PR:
- `Vault.getOrCreateEncryptionKeyId()`: this method fetches the default encryption key ID for the account from Vault. If no key exists, it creates a new one.
- `isAccountEncryptionEnabled()`: this function checks if the account-level default encryption is enabled. This prevents the deletion of the account-level master encryption key when a bucket is deleted.
- A new KMS client has been added to interact with the AWS KMS compatible service and manage encryption operations.

### DIFF
--- a/lib/network/kmsAWS/Client.ts
+++ b/lib/network/kmsAWS/Client.ts
@@ -1,0 +1,276 @@
+'use strict'; // eslint-disable-line
+/* eslint new-cap: "off" */
+
+import errors from '../../errors';
+import { Agent } from "https";
+import { SecureVersion } from "tls";
+import * as werelogs from 'werelogs';
+import { KMSClient, CreateKeyCommand, ScheduleKeyDeletionCommand, EncryptCommand, DecryptCommand, GenerateDataKeyCommand, DataKeySpec } from "@aws-sdk/client-kms";
+import { NodeHttpHandler } from "@smithy/node-http-handler";
+import { AwsCredentialIdentity } from "@smithy/types";
+import assert from 'assert';
+
+/**
+ * Normalize errors according to arsenal definitions
+ * @param err - an Error instance or a message string
+ * @returns - arsenal error
+ *
+ * @note Copied from the KMIP implementation
+ */
+function _arsenalError(err: string | Error) {
+    const messagePrefix = 'AWS_KMS:';
+    if (typeof err === 'string') {
+        return errors.InternalError
+            .customizeDescription(`${messagePrefix} ${err}`);
+    } else if (
+        err instanceof Error ||
+        // INFO: The second part is here only for Jest, to remove when we'll be
+        //   fully migrated to TS
+        // @ts-expect-error
+        (err && typeof err.message === 'string')
+    ) {
+        return errors.InternalError
+            .customizeDescription(`${messagePrefix} ${err.message}`);
+    }
+    return errors.InternalError
+        .customizeDescription(`${messagePrefix} Unspecified error`);
+}
+
+export default class Client {
+    client: KMSClient;
+    options: any;
+
+    /**
+     * Construct a high level KMIP driver suitable for cloudserver
+     * @param options - Instance options
+     * @param options.kmsAWS - AWS client options
+     * @param options.kmsAWS.region - KMS region
+     * @param options.kmsAWS.endpoint - Endpoint URL of the KMS service
+     * @param options.kmsAWS.ak - Application Key
+     * @param options.kmsAWS.sk - Secret Key
+     * @param options.kmsAWS.tls.rejectUnauthorized - default to true, reject unauthenticated TLS connections (set to false to accept auto-signed certificates, useful in development ONLY)
+     * @param options.kmsAWS.tls.ca - override CA definition(s)
+     * @param options.kmsAWS.tls.cert - certificate or list of certificates
+     * @param options.kmsAWS.tls.minVersion - min TLS version accepted, One of 'TLSv1.3', 'TLSv1.2', 'TLSv1.1', or 'TLSv1' (see https://nodejs.org/api/tls.html#tlscreatesecurecontextoptions)
+     * @param options.kmsAWS.tls.maxVersion - max TLS version accepted, One of 'TLSv1.3', 'TLSv1.2', 'TLSv1.1', or 'TLSv1' (see https://nodejs.org/api/tls.html#tlscreatesecurecontextoptions)
+     * @param options.kmsAWS.tls.key - private key or list of private keys
+     * 
+     * This client also looks in the standard AWS configuration files (https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html).
+     * If no option is passed to this constructor, the client will try to get it from the configuration file.
+     * 
+     * TLS configuration options are those of nodejs, you can refere to https://nodejs.org/api/tls.html#tlsconnectoptions and https://nodejs.org/api/tls.html#tlscreatesecurecontextoptions
+     */
+    constructor(
+        options: {
+            kmsAWS: {
+                region?: string,
+                endpoint?: string,
+                ak?: string,
+                sk?: string,
+                tls?: {
+                    rejectUnauthorized?: boolean,
+                    ca?: [Buffer] | Buffer,
+                    cert?: [Buffer] | Buffer,
+                    minVersion?: string,
+                    maxVersion?: string,
+                    key?: [Buffer] | Buffer,
+                }
+            }
+        },
+    ) {
+        let requestHandler: {requestHandler: NodeHttpHandler} | null = null;
+        const tlsOpts = options.kmsAWS.tls;
+        if (tlsOpts) {
+            const agent = new Agent({
+                rejectUnauthorized: tlsOpts?.rejectUnauthorized,
+                ca: tlsOpts?.ca,
+                cert: tlsOpts?.cert,
+                minVersion: <SecureVersion>tlsOpts?.minVersion,
+                maxVersion: <SecureVersion>tlsOpts?.maxVersion,
+                key: tlsOpts?.key,
+            });
+
+            requestHandler = {requestHandler: new NodeHttpHandler({
+                httpAgent: agent,
+                httpsAgent: agent,
+            })}
+        }
+
+        let credentials: {credentials: AwsCredentialIdentity} | null = null;
+        if (options.kmsAWS.ak && options.kmsAWS.sk) {
+            credentials = {credentials: {
+                accessKeyId: options.kmsAWS.ak,
+                secretAccessKey: options.kmsAWS.sk,
+            }};
+        }
+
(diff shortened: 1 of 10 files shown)


nicolas2bert:
ping

lib/network/kmsAWS/Client.ts
```diff
@@ -0,0 +1,256 @@
...
+                logger.error("AWS KMS: failed to create master encryption key", { err });
+                cb(error);
+                return;
+            }
+            logger.debug("AWS KMS: master encryption key created", { KeyMetadata: data?.KeyMetadata });
+            cb(null, data?.KeyMetadata?.KeyId);
+        });
+    }
+
+    destroyBucketKey(bucketKeyId: string, logger: werelogs.Logger, cb: (err: Error | null) => void): void {
```
dvasilas, line 111: Maybe it's more appropriate to call the 1st argument `masterKeyId` as, if I understand correctly, this is what will be passed by the caller ?

lib/network/kmsAWS/Client.ts
```diff
@@ -0,0 +1,256 @@
...
+import assert from 'assert';
+
+type TLSVersion = 'TLSv1.3' | 'TLSv1.2' | 'TLSv1.1' | 'TLSv1';
+
+/**
+ * Normalize errors according to arsenal definitions
+ * @param err - an Error instance or a message string
+ * @returns - arsenal error
+ */
+function _arsenalError(err: string | Error) {
```
dvasilas: nitpick: Could be worth extracting the prefix as an argument and making this function common for `kmsAWS` and `kmip` to avoid duplication (if there is an obvious place to put it).

Could be used in other places eventually.

lib/network/kmsAWS/Client.ts
```diff
@@ -0,0 +1,256 @@
...
+
+    createBucketKey(bucketName: string, logger: werelogs.Logger, cb: (err: Error | null, keyId?: string) => void): void {
+        logger.debug("AWS KMS: creating encryption key managed at the bucket level", { bucketName });
+        this.createMasterKey(logger, cb);
+    }
+
+    createMasterKey(logger: werelogs.Logger, cb: (err: Error | null, keyId?: string) => void): void {
+        logger.debug("AWS KMS: creating master encryption key");
+        const params = {};
+        this.client.createKey(params, (err: AWSError, data) => {
```
BourgoisMickael:
```suggestion
        this.client.createKey({}, (err: AWSError, data) => {
```

lib/network/kmsAWS/Client.ts
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    generateDataKey(
+        cryptoScheme: number,
+        masterKeyId: string,
+        logger: werelogs.Logger,
+        cb: (err: Error | null, plainTextDataKey?: Buffer, cipheredDataKey?: Buffer) => void
+    ): void {
+        logger.debug("AWS KMS: generating data key", { cryptoScheme, masterKeyId });
+        assert.strictEqual(cryptoScheme, 1);
```
BourgoisMickael, line 148: What's this `cryptoScheme`, it's unused ?

lib/network/kmsAWS/Client.ts
```diff
@@ -0,0 +1,256 @@
...
+
+            if (!data) {
+                const error = _arsenalError("failed to generate data key: empty response");
+                logger.error("AWS KMS: failed to generate data key: empty response");
+                cb(error);
+                return;
+            }
+
+            logger.debug("AWS KMS: data key generated");
+            cb(null, Buffer.from(data.Plaintext as Uint8Array), Buffer.from(data.CiphertextBlob as Uint8Array));
```
BourgoisMickael: For `generateDataKey` and `decrypt`, the documentation says:

> Then erase the plaintext data key from memory.

https://docs.aws.amazon.com/kms/latest/APIReference/API_GenerateDataKey.html

We should probably clear the sensitive key from the received buffe like it's done herer before we lose reference to that buffer: https://github.com/aws/aws-encryption-sdk-javascript/blob/e248a96ef02e51baa00e1b0d71952406bd4cf78c/modules/kms-keyring/src/helpers.ts#L159-L173

lib/network/kmsAWS/Client.ts
```diff
@@ -0,0 +1,256 @@
...
+                return;
+            }
+
+            logger.debug("AWS KMS: data key deciphered");
+            cb(null, Buffer.from(data.Plaintext as Uint8Array));
+        });
+    }
+
+    healthcheck(logger: werelogs.Logger, cb: (err: Error | null) => void): void {
+        this.createMasterKey(logger, (err, masterKeyId) => {
```
BourgoisMickael: This seems excessive for health checking.

Is this going to be used by cloudserver for deep healthcheck ? https://github.com/aws/aws-encryption-sdk-javascript/blob/e248a96ef02e51baa00e1b0d71952406bd4cf78c/modules/kms-keyring/src/helpers.ts#L159-L173

And the cloudserver deep healthcheck is executed every 2 seconds by default by s3 frontend check scheduler:
https://github.com/scality/Federation/blob/63e05543d2b8995c40be796055925552bdab5fd1/roles/run-s3-frontend/tasks/main.yml#L267-L282

Is there no HEAD route that would be sufficient ?

lib/network/kmsAWS/Client.ts
nicolas2bert: for future use cases (multiple encryption scheme) and to enforce version control for the encryption scheme, but in the current version, it is simply a validation step.

lib/network/kmsAWS/Client.ts
nicolas2bert: I've added comments to the functions. Let me know if they make more sense now.

lib/network/kmsAWS/Client.ts
nicolas2bert: The KMS healthcheck was previously disabled in the [healthcheckHandler.js](https://github.com/scality/cloudserver/blob/development/7.70/lib/utilities/healthcheckHandler.js#L43). It is likely that the healthcheck will need to be enabled moving forward. I agree with the proposed changes and have updated the configuration accordingly.

lib/network/kmsAWS/Client.ts
nicolas2bert: We already zero out the plaintext key after [generateDataKey](https://github.com/scality/cloudserver/blob/development/7.70/lib/kms/wrapper.js#L204) and [decrypt](https://github.com/scality/cloudserver/blob/development/7.70/lib/kms/wrapper.js#L279). However, this approach is not sufficient because if the buffer is accessed or copied before fill(0) is called the plaintext key could be temporarily exposed in memory. For that reason, we introduced the `safePlaintext` method.

dvasilas (approved):

lib/network/kmsAWS/Client.ts
BourgoisMickael: Note that `Buffer.from` uses a nodejs shared buffer pool for perf reason for small (<4Kb) Buffer, so it's not really isolated.

Any other places in the process can access memory allocated this way by doing `Buffer.from('1').buffer # Access to the internal 8Kb ArrayBuffer pool`.

If you want to not use that shared pool you can use `Buffer.alloc` or `Buffer.allocUnsafeSlow` and copy the buffer.
It will be a little slower not using the shared buffer pool.

lib/network/kmsAWS/Client.ts
BourgoisMickael: I'm still not sure about using an actual endpoint. It means the AK SK used will need privileges to do a `ListKeys`.
And requests to KMS have quotas and pricing: 

- https://eu-north-1.console.aws.amazon.com/servicequotas/home/services/kms/quotas/L-8A08DAEA
- https://aws.amazon.com/kms/pricing/

The default quota is 500 req/s. If anything else on the account reach this quota on this endpoint, it might fail the healthcheck.

Pricing is `$0.03 per 10,000 requests` (after 20,000 free tier requests).

With our regular deep healthcheck, 1 cloudserver might incur a small cost.
Now if there are a lot of cloudservers deployed, this will incur a noticable usage for quota and pricing

___

> [!WARNING] 
> This means maybe cloudserver should cache the healthcheck result for quit some time to prevent spamming KMS, if we don't have another healthcheck method.
> 
> Maybe we don't need regular healthcheck as we don't manage KMS, and we can't do anything if it's down.
> So a very long cache could be good (at least daily, with a jitter on cloudserver so 80 instances won't send requests at the same time).
> And we should differentiate errors 500 (KMS down can't do anything) and 403 (missing permissions on the AK SK, customer needs to fix it)

lib/network/kmsAWS/Client.ts
nicolas2bert: updated there: https://github.com/scality/Arsenal/compare/5ec3eeec9b0fb0da4f6fc46241507fcdc27a1745..5d412c20a3e93bc2dfc69ee95a169e03386b2035

lib/network/kmsAWS/Client.ts
nicolas2bert: Yes, I'm not sure what the best approach is here yet; we need to think about it.
Meanwhile, I will comment it out to avoid any issues related to costs or permissions.

lib/network/kmsAWS/Client.ts
BourgoisMickael: `Buffer.alloc` is slower than `Buffer.allocUnsafeSlow` because it fills the buffer with `0`.

The naming is weird but:
- `Buffer.alloc`: alloc outside the shared pool, initialize with `0`.
- `Buffer.allocUnsafe`: alloc from the shared pool, not initialized.
- `Buffer.allocUnsafeSlow`: alloc outside the shared pool, not initialized.

You can pass the data to `Buffer.alloc` second param to fill it directly with data without filling with `0` before. (It's like `Buffer.allocUnsafeSlow`)

BourgoisMickael (approved):

lib/network/kmsAWS/Client.ts
nicolas2bert: Good catch! Sorry for the back and forth.

nicolas2bert: Latest update: https://github.com/scality/Arsenal/compare/6a74794998e635b52c2b48abca9735d66a336404..82a3420c5c59b68924c3968970cc73e755ccbb59

nicolas2bert:
@bert-e reset

nicolas2bert:
@bert-e approve

PR #2260, opened 2024-10-03
Fix listing mongodb collections

francoisferrand:
Issue: ARSN-432

### DIFF
--- a/lib/models/ObjectMDAmzRestore.ts
+++ b/lib/models/ObjectMDAmzRestore.ts
@@ -11,6 +11,7 @@
 export default class ObjectMDAmzRestore {
     'expiry-date': Date | string;
     'ongoing-request': boolean;
+    'content-md5': string;
 
     /**
      *
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -2099,7 +2099,7 @@ class MongoClientInterface {
         let bucketCount = 0;
         const bucketInfos = [];
 
-        this.db.listCollections({ filter: { type: 'collection' } }).toArray().then(collInfos =>
+        this.db.listCollections({ type: 'collection' }).toArray().then(collInfos =>
             async.eachLimit(collInfos, 10, (value, next) => {
                 if (this._isSpecialCollection(value.name)) {
                     // skip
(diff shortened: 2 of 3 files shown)


lib/models/ObjectMDAmzRestore.ts
```diff
@@ -11,6 +11,7 @@
 export default class ObjectMDAmzRestore {
     'expiry-date': Date | string;
     'ongoing-request': boolean;
+    'content-md5': string;
```
ghost, line 14: Can we add getter/setter functions (like other fields), and cover it with unit tests in `tests/unit/models/ObjectMDAmzRestore.spec.js`?

francoisferrand: Not worth the effort, this type is actually not used at all... at the moment, I would be more inclined to remove these accessors, which are akin to dead code

The only point of adding these here is for documentation, to ensure the fields can indeed be discovered by looking at these ˋmodels`

ghost: So I guess we can take advantage of the PR to remove these functions?

francoisferrand: we could ; but best do this separately: we have a real fix (on mongodb), and just adding the field for documentation.
Cleanup will be best done separately, after thinking through how we want to evolve the models (to make them more useful): remove unneeded fields, always use the models, ...

francoisferrand: we could ; but best do this separately: we have a real fix (on mongodb), and just adding the field for documentation.
Cleanup will be best done separately, after thinking through how we want to evolve the models (to make them more useful): remove unneeded fields, always use the models, ...

ghost: I don't mind but this change is already unrelated tothe PR objective (fix of the listing), so it would be logical to either follow the existing pattern, or do it separately as well?

ghost (approved):

lib/models/ObjectMDAmzRestore.ts
francoisferrand: ticket for followup : ARSN-434 (but it will be a big ticket, probably need breakdown to investigate and make a TAD)

(i can remove the field, but the goal was really only to keep up with the reality, and ensure we don't use a field which is not even mentioned in the models)

welansari (approved):

PR #2264, opened 2024-10-17
Memory leak during internal object listing 

ghost:
- Happens when the Skip algorithms tells the function to skip an entry.
- In this case, we recursively call the internalListObject function.
- However, we never properly close the stream created from the Transform object, that requires specific cleanup.
- Overall, the logic is too complex, so cleaning is now centralized and modular, so we clean anything that is still there, in a more clear fashion: we clean any of the three streams. If the stream was already destroyed, the operation is a no-op, but at least, all paths now lead to a consistent cleaning of the resources.

Issue: ARSN-437

For reference, some tests were started to automate memory leak detection, but not very stable so far, or require to run it for long times: https://github.com/scality/Arsenal/compare/bugfix/ARSN-437-ballooning?expand=1

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -1737,19 +1737,25 @@ class MongoClientInterface {
         let stream;
         let baseStream;
         let resolvePhdKey;
-        const cbOnce = jsutil.once((err, data) => {
+        const cleanupStreams = () => {
             if (resolvePhdKey) {
                 resolvePhdKey.removeAllListeners();
                 resolvePhdKey.destroy();
+                resolvePhdKey = null;
             }
-            if (baseStream) {
+            if (baseStream && baseStream !== stream) {
                 baseStream.removeAllListeners();
                 baseStream.destroy();
+                baseStream = null;
             }
             if (stream) {
                 stream.removeAllListeners();
                 stream.destroy();
+                stream = null;
             }
+        };
+        const cbOnce = jsutil.once((err, data) => {
+            cleanupStreams();
             return cb(err, data);
         });
         if (!params.secondaryStreamParams) {
@@ -1828,11 +1834,10 @@ class MongoClientInterface {
         });
         skip.setListingEndCb(() => {
             stream.emit('end');
-            stream.destroy();
         });
         skip.setSkipRangeCb(range => {
             // stop listing this key range
-            stream.destroy();
+            cleanupStreams();
 
             // update the new listing parameters here
             const newParams = params;
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -1737,6 +1737,21 @@ class MongoClientInterface {
         const cbOnce = jsutil.once(cb);
         let stream;
         let baseStream;
+        let resolvePhdKey;
+        const cleanupStreams = () => {
+            if (resolvePhdKey) {
+                resolvePhdKey.removeAllListeners();
+                resolvePhdKey.destroy();
+            }
+            if (baseStream) {
+                baseStream.removeAllListeners();
+                baseStream.destroy();
+            }
+            if (stream) {
+                stream.removeAllListeners();
+                stream.destroy();
+            }
+        };
         if (!params.secondaryStreamParams) {
             // listing masters only (DelimiterMaster)
             stream = new MongoReadStream(c, params.mainStreamParams, params.mongifiedSearch);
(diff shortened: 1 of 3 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1795,7 +1810,7 @@ class MongoClientInterface {
                     };
                     log.error(
                         'internalListObjectV1: error listing objects', logObj);
-                    baseStream.destroy();
+                    cleanupStreams();
```
francoisferrand: it seems we have already a mecanism to handle the propagation of events accross streams and associated cleanup (with a bug indeed):
```
                resolvePhdKey.on('end', () => {
                    baseStream.emit('end');
                });

	[...]

	skip.setListingEndCb(() => {
            stream.emit('end');
            stream.destroy();
        });

	[...]

	    .on('end', () => {
                const data = extension.result();
                // call explicitly the destroy method to clean the mongodb cursor
                stream.destroy();
                cbOnce(null, data);
            });
```

instead of adding a "global" cleanup called multiple times, making it difficult to understand where it will be called and thus if it will always be called exactly once: it seems better to let each transform handle its own cleanup AND propagate the signals/errors.

(i.e. we should preferably just fix the existing "layered" approach, which is consistent with the "stacking streams" approach used here ; and if not possible remove preexisting cleanup code, and do it just once at the end)

ghost: I fear "just" fixing the current code will likely let the issue come back someday if we miss another cleanup, which is the mistake we did last time. The current leak was already there, but since we did not see it, we forgot to add cleanup for this case. Given all the code paths here, I am definitely not confident that fixing only the current leak (line 1853) will fix everything for all code paths, unless we clean everything in a consistent way.

I vote for a wrapper at the end (cbOnce), so we clean every time we complete one function call. However I will need to spend time again to confirm this still works, due to the mix of recursive calls, events, and callbacks...

ghost: Is it what you had in mind? https://github.com/scality/Arsenal/pull/2264/commits/d2d5391bdcac821aad1d91f6240993ba3cd0af5c

francoisferrand: still some "duplicated" `stream.destroy`:

```
        skip.setListingEndCb(() => {
            stream.emit('end');
            stream.destroy();
        });
        skip.setSkipRangeCb(range => {
            // stop listing this key range
            stream.destroy();
```

(on the second one, maybe we do indeed want to destroy the stream before proceeding with the listing: but then should do the whole cleanup, and reset `stream` to a falsy value)

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -1734,9 +1734,24 @@ class MongoClientInterface {
...
+        const cbOnce = jsutil.once((err, data) => {
+            if (resolvePhdKey) {
+                resolvePhdKey.removeAllListeners();
+                resolvePhdKey.destroy();
+            }
+            if (baseStream) {
+                baseStream.removeAllListeners();
+                baseStream.destroy();
+            }
+            if (stream) {
```
francoisferrand, line 1751: in some case (Bucket v0 case), `baseStream` may be the same as `stream` : so we should handle this in the callback to avoid using the destroyed stream...

ghost: I don't see any issue, as in nodejs, destroying an already destroyed stream is a no-op...

> Once destroy() has been called any further calls will be a no-op

lib/storage/metadata/mongoclient/MongoClientInterface.js
ghost: I'm not sure we should remove this, I need to test first, sending "end" seem more appropriate, to keep the cleanup in a single place

ghost: Note: This PR will not be merged till we re-confirm the latest changes are stable on a real platform. Expecting no merge before next monday.

ghost: So we can remove the first destroy indeed (but keep the end event). For the second, we indeed must keep the cleanup, otherwise we may get more data from the stream in the parent function, making the whole thing never returning the callback.

- I also updated the logic to only destroy the baseStream if it is not the `stream` object.
- All streams are set to `null` -> not formally needed, but you suggested it and it makes it explicit for the garbage collector that we don't use the variable anymore, so likely to be freed faster than before.
- Setting to `null` also ensures we do not call `destroy` twice for the same stream, even if this is a no-op the second time

https://github.com/scality/Arsenal/pull/2264/commits/8961aa53c1a9b9ee637a5f12376e7ee15550299e

francoisferrand (approved):

welansari (approved):

PR #2265, opened 2024-10-18
ARSN-439: Vaut/authenticateV2(4)Request error handling

dvasilas:
Vaultclient's `verifySignatureV2(4)`  returns an error that contains a `code` and a `description` field.
This error is propagated to the callers of `authenticateV2(4)Request`.

However, callers expect an `ArsenalError`.
This change fixes this specifically for the case of `InvalidAccessKeyId` (when a non-existing access key is given).

Issue: ARSN-439

### DIFF
--- a/lib/auth/Vault.ts
+++ b/lib/auth/Vault.ts
@@ -1,5 +1,5 @@
 import { Logger } from 'werelogs';
-import errors from '../errors';
+import errors, { ArsenalError } from '../errors';
 import AuthInfo from './AuthInfo';
 
 /** vaultSignatureCb parses message from Vault and instantiates
@@ -11,10 +11,10 @@ import AuthInfo from './AuthInfo';
  * items used to calculate signature on chunks if streaming auth
  */
 function vaultSignatureCb(
-    err: Error | null,
+    err: Error & { code: string, description: string } | null,
     authInfo: { message: { body: any } },
     log: Logger,
-    callback: (err: Error | null, data?: any, results?: any, params?: any) => void,
+    callback: (err: Error | ArsenalError | null, data?: any, results?: any, params?: any) => void,
     streamingV4Params?: any
 ) {
     // vaultclient API guarantees that it returns:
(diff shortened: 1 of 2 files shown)


dvasilas:
ping

lib/auth/Vault.ts
```diff
@@ -24,6 +24,9 @@ function vaultSignatureCb(
     if (err) {
         log.debug('received error message from auth provider',
             { errorMessage: err });
+        if (err.code && err.code === 'InvalidAccessKeyId') {
+            return callback(errors.InvalidAccessKeyId);
+        }
```
ghost, line 29: Can we also try this change in cloudserver (I guess cloudserver CI should be enough, but because we don't really use vault there, maybe Integration/Zenko are needed) - just to be on the safe side, so we do not have any issue when the function returns an arsenal error

dvasilas: That's a good point. Not sure if we test this code path, I will check.

nicolas2bert: I feel like the issue we are facing is that we are sometimes managing Arsenal errors and other times 'request' errors, depending on the type of error and the backend being used. Should we consider translating 'request' errors into Arsenal errors so that we only need to manage one type of error?

ghost: Yes. Adding another note: we recently moved out of the arsenal errors in vaultclient, because we were importing the repo only to use these Arsenal errors (based on some metaprogramming logic, with perf impact). Maybe this is a more global thing to discuss here...

dvasilas:
> Yes. Adding another note: we recently moved out of the arsenal errors in vaultclient, because we were importing the repo only to use these Arsenal errors (based on some metaprogramming logic, with perf impact). Maybe this is a more global thing to discuss here...


Indeed there is maybe a discussion to be had before we do a fix.
I am closing this PR for now. I created https://scality.atlassian.net/browse/S3C-9452 to track this.

PR #2267, opened 2024-10-22
Fix service crash when non-arsenal errors is sent to delete route

ghost:
Fixes a crash when the error we receive from the rout is not an Arsenal error (which is a possible and valid case).

### DIFF
--- a/lib/s3routes/routes/routeDELETE.ts
+++ b/lib/s3routes/routes/routeDELETE.ts
@@ -1,7 +1,7 @@
 import { RequestLogger } from 'werelogs';
 
 import * as routesUtils from '../routesUtils';
-import errors from '../../errors';
+import errors, { ArsenalError } from '../../errors';
 import StatsClient from '../../metrics/StatsClient';
 import * as http from 'http';
 
@@ -54,11 +54,11 @@ export default function routeDELETE(
         api.callApiMethod('objectDelete', request, response, log,
             (err, corsHeaders) => {
                 /*
-              * Since AWS expects a 204 regardless of the existence of
-              the object, the errors NoSuchKey and NoSuchVersion should not
-              * be sent back as a response.
-              */
-                if (err && !err.is.NoSuchKey && !err.is.NoSuchVersion) {
+                 * Since AWS expects a 204 regardless of the existence of
+                 * the object, the errors NoSuchKey and NoSuchVersion should not
+                 * be sent back as a response.
+                 */
+                if (err instanceof ArsenalError && !err.is?.NoSuchKey && !err.is?.NoSuchVersion) {
                     return routesUtils.responseNoBody(err, corsHeaders,
                         response, undefined, log);
                 }
--- a/lib/s3routes/routes/routeDELETE.ts
+++ b/lib/s3routes/routes/routeDELETE.ts
@@ -58,7 +58,10 @@ export default function routeDELETE(
                  * the object, the errors NoSuchKey and NoSuchVersion should not
                  * be sent back as a response.
                  */
-                if (err instanceof ArsenalError && !err.is?.NoSuchKey && !err.is?.NoSuchVersion) {
+                if (err && (
+                    !(err instanceof ArsenalError) || 
+                    (!err.is.NoSuchKey && err.is.NoSuchVersion)
+                )) {
                     return routesUtils.responseNoBody(err, corsHeaders,
                         response, undefined, log);
                 }
--- a/lib/s3routes/routes/routeGET.ts
+++ b/lib/s3routes/routes/routeGET.ts
@@ -87,8 +87,7 @@ export default function routerGET(
                     if (resMetaHeaders && resMetaHeaders['Content-Length']) {
                         contentLength = resMetaHeaders['Content-Length'];
                     }
-                    // TODO ARSN-216 Fix logger
-                    // @ts-ignore
+                    // @ts-expect-error werelogs types are not up to date
                     log.end().addDefaultFields({ contentLength });
                     routesUtils.statsReport500(err, statsClient);
                     return routesUtils.responseStreamData(err, query,
--- a/lib/s3routes/routes/routeGET.ts
+++ b/lib/s3routes/routes/routeGET.ts
@@ -87,7 +87,6 @@ export default function routerGET(
                     if (resMetaHeaders && resMetaHeaders['Content-Length']) {
                         contentLength = resMetaHeaders['Content-Length'];
                     }
-                    // @ts-expect-error werelogs types are not up to date
                     log.end().addDefaultFields({ contentLength });
                     routesUtils.statsReport500(err, statsClient);
                     return routesUtils.responseStreamData(err, query,
--- a/lib/s3routes/routes/routeWebsite.ts
+++ b/lib/s3routes/routes/routeWebsite.ts
@@ -1,7 +1,7 @@
 import { RequestLogger } from 'werelogs';
 
 import * as routesUtils from '../routesUtils';
-import errors from '../../errors';
+import errors, { ArsenalError } from '../../errors';
 import * as http from 'http';
 import StatsClient from '../../metrics/StatsClient';
 
@@ -30,7 +30,7 @@ export default function routerWebsite(
                 // request being redirected
                 if (redirectInfo) {
                     if (err && redirectInfo.withError) {
-                        return routesUtils.redirectRequestOnError(err,
+                        return routesUtils.redirectRequestOnError(err as ArsenalError,
                             'GET', redirectInfo, dataGetInfo, dataRetrievalParams,
                             response, resMetaHeaders, log)
                     }
@@ -44,7 +44,7 @@ export default function routerWebsite(
                 }
                 // user has their own error page
                 if (err && dataGetInfo) {
-                    return routesUtils.streamUserErrorPage(err, dataGetInfo,
+                    return routesUtils.streamUserErrorPage(err as ArsenalError, dataGetInfo,
                         dataRetrievalParams, response, resMetaHeaders, log);
                 }
                 // send default error html response
@@ -65,7 +65,7 @@ export default function routerWebsite(
                 routesUtils.statsReport500(err, statsClient);
                 if (redirectInfo) {
                     if (err && redirectInfo.withError) {
-                        return routesUtils.redirectRequestOnError(err,
+                        return routesUtils.redirectRequestOnError(err as ArsenalError,
                             'HEAD', redirectInfo, null, dataRetrievalParams,
                             response, resMetaHeaders, log)
                     }
@@ -77,7 +77,7 @@ export default function routerWebsite(
                 }
                 // could redirect on err so check for redirectInfo first
                 if (err) {
-                    return routesUtils.errorHeaderResponse(err, response,
+                    return routesUtils.errorHeaderResponse(err as ArsenalError, response,
                         resMetaHeaders, log);
                 }
                 return routesUtils.responseContentHeaders(err, {}, resMetaHeaders,
--- a/lib/s3routes/routes/routeWebsite.ts
+++ b/lib/s3routes/routes/routeWebsite.ts
@@ -1,7 +1,7 @@
 import { RequestLogger } from 'werelogs';
 
 import * as routesUtils from '../routesUtils';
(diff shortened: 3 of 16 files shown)


lib/s3routes/routesUtils.ts
```diff
@@ -717,30 +709,37 @@ export function streamUserErrorPage(
...
     bucketName: string,
     response: http.ServerResponse,
     corsHeaders: { [key: string]: string } | null,
     log: RequestLogger,
 ) {
+    let error;
+    if (err instanceof ArsenalError) {
+        error = err;
+    } else if (!(err instanceof ArsenalError)) {
+        error = errors.InternalError.customizeDescription(err.message);
```
welansari:
```suggestion
        error = err;
    } else {
        error = errors.InternalError.customizeDescription(err.message);
```

lib/s3routes/routesUtils.ts
```diff
@@ -106,34 +104,36 @@ const XMLResponseBackend = {
...
         response: http.ServerResponse,
         log: RequestLogger,
         corsHeaders?: { [key: string]: string } | null,
     ) {
         setCommonResponseHeaders(corsHeaders, response, log);
+        let error;
+        if (errCode instanceof ArsenalError) {
+            error = errCode;
+        } else if (errCode instanceof Error) {
+            error = errors.InternalError.customizeDescription(errCode.message);
```
welansari: Is there a reason on why this check is inconsistent between the different functions of this file ?

lib/s3routes/routes/routeDELETE.ts
```diff
@@ -54,11 +54,11 @@ export default function routeDELETE(
...
-              the object, the errors NoSuchKey and NoSuchVersion should not
-              * be sent back as a response.
-              */
-                if (err && !err.is.NoSuchKey && !err.is.NoSuchVersion) {
+                 * Since AWS expects a 204 regardless of the existence of
+                 * the object, the errors NoSuchKey and NoSuchVersion should not
+                 * be sent back as a response.
+                 */
+                if (err instanceof ArsenalError && !err.is?.NoSuchKey && !err.is?.NoSuchVersion) {
                     return routesUtils.responseNoBody(err, corsHeaders,
```
welansari: Should the non arsenal errors also be sent as a response ?

ghost: No, because they would likely translate code-related error (with the stack), leaking information about the internals of the application. Better returning standard errors in this case. But the routes should still allow these non-arsenal errors to be received, before sending the final response

lib/s3routes/routesUtils.ts
ghost: I'm not sure I understand fully the question, as we also have this logic in the other `errorResponse` for the json mode, but we work with `err.code` in the functions, something that we do not have if the error is not an Arsenal (or S3 standard) error. In this case, we want to recreate an internal error, so we have a logical error code, and we do not alter the implementation, while allowing any kind of error. Because we cannot guarantee that the error we receive will always be an Arsenal error (the implementation is modular and we use projects like hdclient that do not use Arsenal error, yet hdclient errors can reach this function).

lib/s3routes/routes/routeDELETE.ts
```diff
@@ -54,11 +54,11 @@ export default function routeDELETE(
...
-              * Since AWS expects a 204 regardless of the existence of
-              the object, the errors NoSuchKey and NoSuchVersion should not
-              * be sent back as a response.
-              */
-                if (err && !err.is.NoSuchKey && !err.is.NoSuchVersion) {
+                 * Since AWS expects a 204 regardless of the existence of
+                 * the object, the errors NoSuchKey and NoSuchVersion should not
+                 * be sent back as a response.
+                 */
+                if (err instanceof ArsenalError && !err.is?.NoSuchKey && !err.is?.NoSuchVersion) {
```
francoisferrand: * why would it not be an `ArsenalError` ? (i.e. are we not just "masking" an underlying issue: which should be fixed as well)
* if we receive an error, but not an ArsenalError : we change the behavior here... We probably don't want to include too much information indeed, but we should probably not send 204 either and report some sort of error (maybe a generic 500?)

ghost: > why would it not be an ArsenalError ? (i.e. are we not just "masking" an underlying issue: which should be fixed as well)

Because we return errors from the underlying modules, such as hdclient, in our APIs. These errors are not Arsenal errors. It's possible for example in the "DeleteObject" API if hdclient returns an error. It does not happen most of the time, but it can happen, and we should not make assumption in Arsenal for the error type: this was changed when typescript was implemented in arsenal, but our S3 implementation was not updated to always return ArsenalErrors. And it actually not only returns ArsenalErrors. So this PR first fixes the bug causing a crash, and then improve the types, because even if for the other cases we do not crash, we may send back errors with invalid values or http code.

> if we receive an error, but not an ArsenalError : we change the behavior here... We probably don't want to include too much information indeed, but we should probably not send 204 either and report some sort of error (maybe a generic 500?)

I am not sure to understand what you mean by "we change the behavior here": it's the actual bug we are trying to fix. Reaching this line with a non-Arsenal error makes the process crash, because `is` is not a valid field from a normal error object. We do not want to return 204 because this is the S3 standard behavior for NoSuchKey errors (and yes, it's misleading, [see doc](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html#API_DeleteObjects_Examples)). We only enter this "if" if the error is an arsenal error and a NoSuchKey one. If the error is internal, we will return an InternalError.

Example:
```
 stderr F 2024-07-12T02:51:37.386Z: Uncaught exception: 
 stderr F TypeError: Cannot read properties of undefined (reading 'NoSuchKey')
 stderr F     at /usr/src/app/node_modules/arsenal/build/lib/s3routes/routes/routeDELETE.js:82:32
 stderr F     at /usr/src/app/lib/api/objectDelete.js:251:20
 stderr F     at /usr/src/app/node_modules/async/dist/async.js:421:16
 stderr F     at next (/usr/src/app/node_modules/async/dist/async.js:5302:29)
 stderr F     at /usr/src/app/node_modules/async/dist/async.js:906:16
 stderr F     at /usr/src/app/lib/api/objectDelete.js:211:41
 stderr F     at /usr/src/app/lib/services.js:372:36
 stderr F     at /usr/src/app/node_modules/hdclient/lib/hdcontroller.js:358:17
 stderr F     at /usr/src/app/node_modules/async/dist/async.js:421:16
```

lib/s3routes/routes/routeGET.ts
```diff
@@ -87,8 +87,7 @@ export default function routerGET(
                     if (resMetaHeaders && resMetaHeaders['Content-Length']) {
                         contentLength = resMetaHeaders['Content-Length'];
                     }
-                    // TODO ARSN-216 Fix logger
-                    // @ts-ignore
+                    // @ts-expect-error werelogs types are not up to date
                     log.end().addDefaultFields({ contentLength });
```
benzekrimaha, line 90: shouldn't this be log.addDefaultFields({ contentLength }); like what's in here : https://github.com/scality/Arsenal/pull/2267/commits/449be66b448ec7574eab8edd8377c9f97037ef40#diff-8357453c6abb1c2118b5ddb61026fcce2f18a329f56eef87a68c0a83c3782e27R228

ghost: It should actually be like in the get router, the end() returns "this" if no input is provided, but the types are not reporting it, so typescript thinks there is an error
https://github.com/scality/Arsenal/pull/2267/commits/8e830fe4b5f0dc34d243763a9bdf36f2cde40aa7

welansari (approved):

lib/s3routes/routes/routeGET.ts
ghost: I bumped werelogs and removed the ts-expect-error

lib/s3routes/routes/routeDELETE.ts
francoisferrand: > I am not sure to understand what you mean by "we change the behavior here": it's the actual bug we are trying to fix

the previous behavior was to response w/o body (i.e. take the `if` branch) on any error except NoSuchKey / NoSuchVersion.

with the change, non-arsenal errors will not take this code path anymore : which is changing the behavior, not just fixing the crash...

the following would fix the crash without changing the behavior:

```suggestion
                if (err && !err.is?.NoSuchKey && !err.is?.NoSuchVersion) {
```

francoisferrand: AFAIU, we can now have non-arsenal errors from Vault, right?
and we may want to have 204 in such case?

ghost: We always call `routesUtils.responseNoBody` in both cases, the only difference being the error code returned, but that does not change your suggestion indeed.
However we cannot do that, if we do not test that the error is an arsenal error we get TS errors, so I adapted it a bit.

-> Updated, I also removed the last bump commit as per our discussion

lib/s3routes/routes/routeWebsite.ts
```diff
@@ -30,7 +30,7 @@ export default function routerWebsite(
                 // request being redirected
                 if (redirectInfo) {
                     if (err && redirectInfo.withError) {
-                        return routesUtils.redirectRequestOnError(err,
+                        return routesUtils.redirectRequestOnError(err as ArsenalError,
```
francoisferrand: do you know why we need type coercion here?
`callApiMethod` already defines that the callback receives an `ArsenalError` error...

lib/s3routes/routes/routeDELETE.ts
```diff
@@ -54,11 +54,14 @@ export default function routeDELETE(
...
-              * be sent back as a response.
-              */
-                if (err && !err.is.NoSuchKey && !err.is.NoSuchVersion) {
+                 * Since AWS expects a 204 regardless of the existence of
+                 * the object, the errors NoSuchKey and NoSuchVersion should not
+                 * be sent back as a response.
+                 */
+                if (err && (
+                    !(err instanceof ArsenalError) || 
+                    (err.is?.NoSuchKey === false && err.is?.NoSuchVersion === false)
```
francoisferrand: nit: the `?` is not needed anymore in that case, since we are garanteed to have an `ArsenalError`.

(and best to use bool equivalence than strict equality to false, IMO...)

```suggestion
                    (!err.is.NoSuchKey && !err.is.NoSuchVersion)
```

lib/s3routes/routes/routeWebsite.ts
ghost: It's because the error we can currently get from APIs is defined in

```ts

export type CallApiMethod = (
    methodName: string,
    request: http.IncomingMessage,
    response: http.ServerResponse,
    log: RequestLogger,
) => void;
```

So it can be a non-arsenal error, this change was made on purpose so we benefit from ts checks everywhere and cover all paths that could lead to errors. The case of website is special, we already know it can only be arsenal errors. All of this will be removed once we have the "long term" solution in place

lib/s3routes/routes/routeDELETE.ts
ghost: done

lib/s3routes/routes/routeWebsite.ts
francoisferrand: indeed, I missed the alternate types 🤦‍♂️

but it seems a bit dangerous to just "expect" these will be arsenal errors and blindly trust this is the case : if we want to rely on typescript to detect errors, best do it all the way and add an early check that it is indeed an ArsenalError (or just don't cast, and allow `redirectRequestOnError` to handle any `Error`)

ghost: I updated the function so that it accepts non-arsenal errors - i kept the logic to dynamically recreate an arsenal error in this case, so we minimize the changes (we still need to return an Arsenal error today to the client).

francoisferrand: 👍, but then we should remove the type-coercion altogether

francoisferrand (approved):

lib/s3routes/routes/routeWebsite.ts
ghost: yes... thanks

PR #2268, opened 2024-11-05
do not initialize fields in ObjectMD that are not used in S3C

welansari:
We set fields that are not used in S3C to undefined to avoid serializing and unecessarilty store them in the metadata. This could affect sizing and the load on the platform.

Issue: ARSN-411

### DIFF
--- a/lib/models/ObjectMD.ts
+++ b/lib/models/ObjectMD.ts
@@ -32,7 +32,7 @@ export type ReplicationInfo = {
     role: string;
     storageType: string;
     dataStoreVersionId: string;
-    isNFS: boolean | null;
+    isNFS?: boolean;
 };
 
 export type ObjectMDData = {
@@ -40,7 +40,7 @@ export type ObjectMDData = {
     'owner-id': string;
     'cache-control': string;
     'content-disposition': string;
-    'content-language': string;
+    'content-language'?: string;
     'content-encoding': string;
     'creation-time'?: string;
     'last-modified'?: string;
@@ -84,12 +84,12 @@ export type ObjectMDData = {
     // Used for keeping object metadata in the oplog event
     // In case of a deletion the flag is first updated before
     // deleting the object
-    deleted: boolean;
+    deleted?: boolean;
     // PHD flag indicates whether the object is a temporary placeholder.
     // This is the case when the latest version of an object gets deleted
     // the master is set as a placeholder and gets updated with the new latest
     // version data after a certain amount of time.
-    isPHD: boolean;
+    isPHD?: boolean;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


KillianG (approved):

francoisferrand (approved):
* How can we make sure this does not break everything? If we used the model consistently, it would be easy to adress.... but all over the place we access the MD directly, so there may be cases where we don't properly handle the absence of a field...

* Should not be merge on 8.1 : best to target 8.2 to avoid regressions... (and this is getting merged in backbeat anyway, so we should be fine)

lib/models/ObjectMD.ts
```diff
@@ -183,7 +183,7 @@ export default class ObjectMD {
             'content-length': 0,
             'content-type': '',
             'content-md5': '',
-            'content-language': '',
+            'content-language': undefined,
```
francoisferrand, line 186: is it worth keeping these (undefiend) fields in the default value?
since:
- the model is defined above (so does not help with 'documenting')
- undefined will not be serialized
- when this type is used, methods should be used to access the data and handle the missing optional fields

lib/models/ObjectMD.ts
```diff
@@ -227,12 +227,12 @@ export default class ObjectMD {
...
                 dataStoreVersionId: '',
-                isNFS: null,
+                isNFS: undefined,
             },
             dataStoreName: '',
             originOp: '',
-            deleted: false,
-            isPHD: false,
+            deleted: undefined,
+            isPHD: undefined,
```
francoisferrand, line 235: we should esp. double check how these fields are used, and if we don't have mongodb queries/conditions on `delete: false` or `isPHD: false`

lib/models/ObjectMD.ts
```diff
@@ -197,7 +197,7 @@ export default class ObjectMD {
             'x-amz-server-side-encryption-aws-kms-key-id': '',
             'x-amz-server-side-encryption-customer-algorithm': '',
             'x-amz-website-redirect-location': '',
-            'x-amz-scal-transition-in-progress': false,
+            'x-amz-scal-transition-in-progress': undefined,
```
francoisferrand, line 200: should check this is not used in mongodb index/query/condition

anurag4DSB (approved):

anurag4DSB:
Is there a way we are testing this?

Edit: Code looks good, but I am not sure if it works fine, so can we test this e2e as well (if not already)

benzekrimaha (approved):

lib/models/ObjectMD.ts
welansari: This is not used in any index, query or condition

lib/models/ObjectMD.ts
welansari: `deleted: false`: To be retro-compatible, we always use to conditions: either delete doesn't exist or it's equal to false.
`isPHD: false`: We don't have mongo checks on this.
Conditions in the code also work when the fields are undefined.

PR #2269, opened 2024-11-08
ARSN-444: Add downsample policy actions

tmacro:


lib/policyEvaluator/utils/actionMaps.ts
```diff
@@ -233,6 +233,9 @@ const actionMapScuba = {
     AdminStopIngest: 'scuba:AdminStopIngest',
     AdminReadRaftCseq: 'scuba:AdminReadRaftCseq',
     AdminTriggerRepair: 'scuba:AdminTriggerRepair',
+    AdminStartDownsample: 'scuba:AdminStartDownsample',
+    AdminStopDownsample: 'scuba:AdminStopDownsample',
```
dvasilas, line 237: I think pause/resume is more representative than start/stop, but let's keep it consistent with {start|stop}ingest for now and we can decide at some point before release if we want to change both.

dvasilas (approved):

anurag4DSB (approved):

anurag4DSB (approved):
LGTM

PR #2272, opened 2024-11-27
Filter system.views in mongo.getBucketInfos

francoisferrand:
- Skip system collections in getBucketInfos
- Ignore bucket not found in `getBucketInfos()`
- Add getBucketInfos tests

Issue: ARSN-447

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.js
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.js
@@ -2128,7 +2128,6 @@ class MongoClientInterface {
                     // skip
                     return next();
                 }
-                bucketCount++;
                 const bucketName = value.name;
                 // FIXME: there is currently no way of distinguishing
                 // master from versions and searching for VID_SEP
@@ -2136,6 +2135,10 @@ class MongoClientInterface {
                 // in $regex
                 return this.getBucketAttributes(bucketName, log,
                     (err, bucketInfo) => {
+                        if (err?.is?.NoSuchBucket) {
+                            // Skip bucket if not found: can happen if bucket has just been removed
+                            return next();
+                        }
                         if (err) {
                             log.error('failed to get bucket attributes', {
                                 bucketName,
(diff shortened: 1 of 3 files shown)


tests/functional/metadata/mongodb/getBucketInfos.spec.js
```diff
@@ -0,0 +1,140 @@
...
+            .onSecondCall().callsArgWith(2, new Error('error'));
+
+        try {
+            await metadata.getBucketInfos(logger);
+            assert.fail('Expected an error');
+        } catch (err) {
+            assert.strictEqual(err.is?.InternalError, true);
+        }
+
+        sinon.restore();
```
rachedbenmustapha: should `sinon.restore()` be called in an `afterEach` in case the assert throws within the catch?

rachedbenmustapha (approved):

benzekrimaha (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.js
```diff
@@ -2107,34 +2107,55 @@ class MongoClientInterface {
...
 
     /*
-     * get bucket related information for count items, used by cloudserver
-     * and s3utils
+     * return true if it a system collection, reserved by Mongo for internal use: and which should
+     * not be used.
+     */
+    _isSystemCollection(name) {
+        return name.startsWith('system.');
+    }
```
ghost, line 2115: Shouldn't we also prevent creating bucket with such name? I believe that's specific to mongodb so to be handled in Arsenal. Not sure if it's already done...

francoisferrand: Somehow, this is already handled I think (i remember somebody tried when we first identified the issue)...
Si, I'd rather keep this separate, since we need to fix the bug for CNES and not sure how we want to fix it (e.g. introduce a way to "escape" names or something)

francoisferrand: → https://scality.atlassian.net/browse/ZENKO-4894

ghost: Thanks, as discussed, it will be done in a separate ticket

ghost (approved):

PR #2273, opened 2024-11-29
Fix route delete bug

ghost:
image

### DIFF
--- a/lib/s3routes/routes/routeDELETE.ts
+++ b/lib/s3routes/routes/routeDELETE.ts
@@ -60,7 +60,7 @@ export default function routeDELETE(
                  */
                 if (err && (
                     !(err instanceof ArsenalError) || 
-                    (!err.is.NoSuchKey && err.is.NoSuchVersion)
+                    (!err.is.NoSuchKey && !err.is.NoSuchVersion)
                 )) {
                     return routesUtils.responseNoBody(err, corsHeaders,
                         response, undefined, log);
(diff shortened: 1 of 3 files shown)


francoisferrand (approved):

lib/s3routes/routes/routeDELETE.ts
```diff
@@ -60,7 +60,7 @@ export default function routeDELETE(
                  */
                 if (err && (
                     !(err instanceof ArsenalError) || 
-                    (!err.is.NoSuchKey && err.is.NoSuchVersion)
+                    (!err.is.NoSuchKey && !err.is.NoSuchVersion)
```
francoisferrand, line 63: Can we add unit tests for these functions?

francoisferrand (approved):

benzekrimaha (approved):

lib/s3routes/routes/routeDELETE.ts
ghost: I added some but they don't cover all the paths, I will add it (while I am confirming in cloudserver the fix)

ghost: (we had 0 unit test for the routes before)

ghost: Done in https://github.com/scality/Arsenal/pull/2273/commits/24a87f59fd4959bfebdc94bf0b32fb1d86bd0b27

we already have it but it was not testing both cases...

PR #2274, opened 2024-12-07
Ensure needQuota flag is passed in stringified request context

ghost:
We were relying on the default value from the constructor, but for special API calls, the flag may be needed when the default value is false.

Issue: ARSN-450

### DIFF
--- a/lib/policyEvaluator/RequestContext.ts
+++ b/lib/policyEvaluator/RequestContext.ts
@@ -228,6 +228,7 @@ export default class RequestContext {
         existingObjTag?: string,
         needTagEval?: false,
         objectLockRetentionDays?: number,
+        needQuota?: boolean,
     ) {
         this._headers = headers;
         this._query = query;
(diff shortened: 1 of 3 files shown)


lib/policyEvaluator/RequestContext.ts
```diff
@@ -256,7 +257,7 @@ export default class RequestContext {
         this._securityToken = securityToken;
         this._policyArn = policyArn;
         this._action = action;
-        this._needQuota = actionNeedQuotaCheck[apiMethod] === true
+        this._needQuota = needQuota || actionNeedQuotaCheck[apiMethod] === true
```
francoisferrand, line 260: (Not for this PR)
In the cleanup to come, I'm thinking we should probably make this class dumber and leave the decision up to cloudserver

francoisferrand (approved):

benzekrimaha (approved):

PR #2275, opened 2024-12-11
remove supportedLifecycleRules from constants

welansari:
supportedLifecycleRules' value is different between the 7.x and 8.x branches. We should define the rules directly in the config of the Zenko components.

Issue: ARSN-435

### DIFF
--- a/lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
+++ b/lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
@@ -1,22 +1,19 @@
 import assert from 'assert';
 
 import LifecycleDateTime from './LifecycleDateTime';
-import { supportedLifecycleRules } from '../../constants';
 
 export default class LifecycleUtils {
     _supportedRules: string[];
     _datetime: LifecycleDateTime;
 
-    constructor(supportedRules?: string[], datetime?: LifecycleDateTime) {
-        if (supportedRules) {
-            assert(Array.isArray(supportedRules));
-        }
+    constructor(supportedRules: string[], datetime?: LifecycleDateTime) {
+        assert(Array.isArray(supportedRules));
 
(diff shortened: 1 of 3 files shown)


tests/unit/s3middleware/LifecycleUtils.spec.js
```diff
@@ -713,7 +713,13 @@ describe('LifecycleUtils::filterRules', () => {
     let lutils;
 
     beforeAll(() => {
-        lutils = new LifecycleUtils();
+        lutils = new LifecycleUtils([
```
ghost: You can keep a single const on top of this file to avoid duplication

lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
```diff
@@ -1,13 +1,12 @@
...
 
 import LifecycleDateTime from './LifecycleDateTime';
-import { supportedLifecycleRules } from '../../constants';
 
 export default class LifecycleUtils {
     _supportedRules: string[];
     _datetime: LifecycleDateTime;
 
-    constructor(supportedRules?: string[], datetime?: LifecycleDateTime) {
+    constructor(supportedRules: string[], datetime?: LifecycleDateTime) {
```
ghost, line 9: If we remove the `?` to say we want it to be defined, then we should add an assert in the constructor to ensure it is, the existing one only test if the variable is an array, if not a falsy value

francoisferrand (approved):

ghost (approved):
you may want to bump the project version if needed in cloudserver

welansari:
We probably want to wait for more meaningful changes (NodeJS 22 + dependency bumps) before releasing the 8.2 version. Cloudserver with it's current changes will still work with the previous version of Arsenal.

PR #1693, opened 2022-01-29
bf: ARSN-57 log correct client ip

naren-rajendran:
Log correct client ip with s3 requests. Check request header 'x-forwarded-for' if there is no request configuration.

PR #2277, opened 2024-12-23
Do not be case-sensitive when checking http headers

ghost:
This is affecting policies evaluation in both the S3 and IAM services, when there is a condition on the IP.
No impact for S3C because the header is already properly formatted, but not in Zenko.

Issue: ARSN-453

### DIFF
--- a/lib/policyEvaluator/requestUtils.ts
+++ b/lib/policyEvaluator/requestUtils.ts
@@ -8,9 +8,6 @@ export interface S3Config {
     }
 }
 
-// TODO
-//   I'm not sure about this behavior.
-//   Should it returns string | string[] | undefined or string ?
 /**
  * getClientIp - Gets the client IP from the request
  * @param request - http request object
@@ -20,7 +17,6 @@ export interface S3Config {
 export function getClientIp(request: IncomingMessage, s3config?: S3Config): string {
     const requestConfig = s3config?.requests;
     const remoteAddress = request.socket.remoteAddress;
-    // TODO What to do if clientIp === undefined ?
     const clientIp = remoteAddress?.toString() ?? '';
     if (requestConfig) {
         const { trustedProxyCIDRs, extractClientIPFromHeader } = requestConfig;
@@ -39,5 +35,5 @@ export function getClientIp(request: IncomingMessage, s3config?: S3Config): stri
             }
         }
     }
-    return clientIp?.toString() ?? '';
+    return clientIp;
 }
--- a/lib/policyEvaluator/requestUtils.ts
+++ b/lib/policyEvaluator/requestUtils.ts
@@ -30,7 +30,10 @@ export function getClientIp(request: IncomingMessage, s3config?: S3Config): stri
          * which header to be used to extract client IP
          */
         if (ipCheck.ipMatchCidrList(trustedProxyCIDRs, clientIp)) {
-            const ipFromHeader = request.headers[extractClientIPFromHeader]?.toString();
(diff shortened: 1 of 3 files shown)


KillianG (approved):

francoisferrand (dismissed):
do we really need the extra conversion to lowercase, can't we just fix the 'X-Forwarded-For' where it is defined instead?

This adds a (small) overhead to every check, slightly increasing latency... and small things add up, over time...

lib/policyEvaluator/requestUtils.ts
```diff
@@ -30,7 +30,10 @@ export function getClientIp(request: IncomingMessage, s3config?: S3Config): stri
          * which header to be used to extract client IP
          */
         if (ipCheck.ipMatchCidrList(trustedProxyCIDRs, clientIp)) {
-            const ipFromHeader = request.headers[extractClientIPFromHeader]?.toString();
+            // Request headers in nodejs are lower-cased, so we should not
+            // be case-sentive when looking for the header, as http headers
+            // are case-insensitive.
+            const ipFromHeader = request.headers[extractClientIPFromHeader.toLowerCase()]?.toString();
```
francoisferrand, line 32: looking a few lines above, we already retrieve the x-forwarded-for header:

```
const clientIp = (requestConfig ? remoteAddress : request.headers['x-forwarded-for'] || remoteAddress)?.toString() ?? '';
```

this seems inconsistent, please take the chance to review and clean this...
Esp. if `requestConfig` is not provided, we kind of bindly trust the `x-forwarded-for` header : is this really expected?

ghost: > do we really need the extra conversion to lowercase, can't we just fix the 'X-Forwarded-For' where it is defined instead?

We could do it but the headers are case-insensitive anyway, so the current code is not following the standard and we might have more issues in the future...

ghost: > looking a few lines above, we already retrieve the x-forwarded-for header:
> const clientIp = (requestConfig ? remoteAddress : request.headers['x-forwarded-for'] || remoteAddress)?.toString() ?? '';
this seems inconsistent, please take the chance to review and clean this...
Esp. if requestConfig is not provided, we kind of bindly trust the x-forwarded-for header : is this really expected?

I am not aware of any deployment today that does not set the `requestConfig`, but this header retrieval was introduced in https://scality.atlassian.net/browse/ARSN-57, so here: https://github.com/scality/Arsenal/pull/1693/files#diff-ca8ab692737b69e1ab977a7a6104be671972410cafadea45b119e8259070ad1bR11

If the question is whether we should trust the header blindly or not: as long as the proxy is not trusted we should not trust the header. I will remove this logic to be on the safe side.

Previously we were not using IP conditions so it was only about displaying the info in logs, hence the previous change.

ghost: > extractClientIPFromHeader probably does not work if not equal to x-forwarded-for

Please elaborate what you mean by "does not work": it's a string field, that can be any header: e.g., we can have a custom header set by nginx and use it.

francoisferrand: > extractClientIPFromHeader probably does not work if not equal to x-forwarded-for
> > Please elaborate what you mean by "does not work": it's a string field, that can be any header: e.g., we can have a custom header set by nginx and use it.

"does not work" was too quick, I missed the condition on `requestConfig`

francoisferrand (approved):

PR #2278, opened 2025-01-06
Support extracting the http scheme from the requests

ghost:
Currently, behind a proxy, we cannot detect the right scheme. Also, we use the `request.connection.encrypted` variable, that is (1) deprecated and (2) already removed since Node 16, so the whole thing end up being undefined.

We introduce a proper function to detect that the scheme provided by the trusted header is http or https, or we dynamically detect if the socket is a TLS instance, and check if it's secure.

### DIFF
--- a/lib/policyEvaluator/requestUtils.ts
+++ b/lib/policyEvaluator/requestUtils.ts
@@ -1,10 +1,12 @@
 import * as ipCheck from '../ipCheck'
 import { IncomingMessage } from 'http'
+import { TLSSocket } from 'tls'
 
 export interface S3Config {
     requests: {
         trustedProxyCIDRs: string[],
-        extractClientIPFromHeader: string
+        extractClientIPFromHeader: string,
+        extractProtocolFromHeader: string,
     }
 }
 
@@ -36,4 +38,21 @@ export function getClientIp(request: IncomingMessage, s3config?: S3Config): stri
         }
     }
     return clientIp;
+}
+
+/**
+ * getHttpProtocolSecurity - Dete²object
+ * @param s3config - s3 config
+ * @return {boolean} - returns true if the request is secure
+ */
+export function getHttpProtocolSecurity(request: IncomingMessage, s3config?: S3Config): boolean {
+    const requestConfig = s3config?.requests;
+    if (requestConfig) {
+        const { trustedProxyCIDRs } = requestConfig;
+        const clientIp = request.socket.remoteAddress?.toString() ?? '';
+        if (ipCheck.ipMatchCidrList(trustedProxyCIDRs, clientIp)) {
+            return request.headers[requestConfig.extractProtocolFromHeader.toLowerCase()] === 'https';
+        }
+    }
+    return request.socket instanceof TLSSocket && request.socket.encrypted;
 }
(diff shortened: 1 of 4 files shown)


tests/unit/policyEvaluator/requestUtils.spec.js
```diff
@@ -110,3 +111,79 @@ describe('requestUtils.getClientIp', () => {
...
+            },
+        });
+        const result = requestUtils.getHttpProtocolSecurity(request, configWithProxy);
+        assert.strictEqual(result, false);
+    });
+
+    it('should check TLS when request not from trusted proxy', () => {
+        const request = new DummyRequest({
+            headers: {
+                'x-forwarded-proto': 'https',
```
francoisferrand: we want to test here that this field is ignored : so it should be http...

either we have a TLS socket & x-forewarded-proto = http, or the opposite (kind of more logical, as it actually tests _do not trust x-forwarded-... field which do not come from a trusted proxy_)

tests/unit/policyEvaluator/requestUtils.spec.js
```diff
@@ -110,3 +111,79 @@ describe('requestUtils.getClientIp', () => {
...
+            socket: new TLSSocket(null),
+        });
+        request.socket.encrypted = true;
+        const result = requestUtils.getHttpProtocolSecurity(request, configWithoutProxy);
+        assert.strictEqual(result, true);
+    });
+
+    it('should return false for non-TLS socket', () => {
+        const request = new DummyRequest({
+            headers: {},
```
francoisferrand: should probably set `x-forewarded-proto = https` here as well, to explicitely test that we ignore it in that case
(or duplicate the test: i.e. both 'should return false for non-TLS socket' and 'ignore forwarded proto from non trusted proxy')

lib/policyEvaluator/requestUtils.ts
```diff
@@ -37,3 +39,20 @@ export function getClientIp(request: IncomingMessage, s3config?: S3Config): stri
...
+ * @param s3config - s3 config
+ * @return {boolean} - returns true if the request is secure
+ */
+export function getHttpProtocolSecurity(request: IncomingMessage, s3config?: S3Config): boolean {
+    const requestConfig = s3config?.requests;
+    if (requestConfig) {
+        const { trustedProxyCIDRs } = requestConfig;
+        const clientIp = request.socket.remoteAddress?.toString() ?? '';
+        if (ipCheck.ipMatchCidrList(trustedProxyCIDRs, clientIp)) {
+            return request.headers[requestConfig.extractProtocolFromHeader.toLowerCase()] === 'https';
```
francoisferrand, line 54: should we also lowercase the header value itself? i.e. to handle cases where protocol is `HtTpS`... ?

francoisferrand (approved):

lib/policyEvaluator/requestUtils.ts
ghost: The RFC 7239 does not document they can be uppercase, I'm not sure this is even possible: we can control the header itself, not the actual value...

benzekrimaha (approved):

PR #2279, opened 2025-01-17
Bucket quota should be treated and stored as BigInts

ghost:
Stored and treated as BigInt in the class. Accepted as string/number/Long (from a serialized object, or from an object), and outputted as a string for storage on the DB


welansari (approved):

KillianG (approved):

lib/auth/Vault.ts
```diff
@@ -18,7 +18,10 @@ function vaultSignatureCb(
             body: {
                 userInfo: AuthInfoType,
                 authorizationResults: { [key: string]: any },
-                accountQuota: number,
+                accountQuota: {
```
francoisferrand: API format changed?
this is not just passing a string instead of number (or ideally `number|string` for backwards compatibility), but a real compatibility break...

lib/models/BucketInfo.ts
```diff
@@ -188,7 +188,7 @@ export default class BucketInfo {
         notificationConfiguration?: any,
         tags?: Array<BucketTag> | [],
         capabilities?: Capabilities,
-        quotaMax?: number | 0,
+        quotaMax?: bigint,
```
francoisferrand: should we not be more leniant, to avoid issues: e.g. allow the api to take `number | bigint | 0`, and initalize the field with `this._quotaMax = BigInt(quotaMax || 0n)` ?
or is there a performance benefit to doing this?

(and same for setQuota api)

lib/auth/Vault.ts
ghost: No, the type was wrong from the beginning, it was already an object with the 2 fields, but numbers, now this is the same fields and the string type.

lib/models/BucketInfo.ts
ghost: The API accepts strings only since the beginning. Here, this is just how we parse/handle the value internally, it must be bigint here because we control how we create the BucketInfo class, no need to keep debt here.

Plus, the `0` is not needed, it's already a number.

lib/auth/Vault.ts
ghost: It would have created errors if we had TS in cloudserver.

lib/algos/list/skip.js
```diff
@@ -10,10 +10,10 @@ const MAX_STREAK_LENGTH = 100;
  */
 class Skip {
     /**
-     * @param {Object} params           - skip parameters
-     * @param {Object} params.extension - delimiter extension used (required)
-     * @param {String} params.gte       - current range gte (greater than or
-     *                                    equal) used by the client code
+     * @param {Object} params                - skip parameters
+     * @param {Object} params.extension      - delimiter extension used (required)
+     * @param {String | string[]} params.gte - current range gte (greater than or
```
ghost, line 15: This change is related to the fact that in MongoClientInterface we actually pass an array here: the function type was not properly defined here, leading to TS error. Having an array here is fine and expected.

lib/auth/backends/in_memory/Backend.ts
williamlardier, line 1: This file is not actually used (and not in production as well), but we want to have a similar behavior here as in vaultclient, because both are under the same abstraction layer.

lib/auth/backends/in_memory/Backend.ts
```diff
@@ -61,7 +70,15 @@ class InMemoryBackend extends BaseBackend {
...
         };
-        const vaultReturnObject = this.formatResponse(userInfoToSend);
+        const vaultReturnObject = this.formatResponse(
+            userInfoToSend,
+            [],
+            {
+                account: entity.canonicalID,
+                // account quota is not supported in this backend
+                accountQuota: 0,
+            }
```
francoisferrand: nit: would it not be simpler to just add a `cannonicalId` or `entity` parameter to `formatResponse`, and let that function handle the formatting of nested types and filling of missed fields just once?

```
function _formatResponse(userInfoToSend: AuthInfoType, account: string): { message: { body: AuthV4Results } } {
    return {
        message: {
            body: {
                userInfo: userInfoToSend,
                authorizationResults: [],
                accountQuota: {
                    account,
                    // account quota is not supported in this backend
                    accountQuota: 0,
                },
            },
        },
    };
}
```

```suggestion
const vaultReturnObject = this.formatResponse(userInfoToSent, entity.canonicalID);
```

lib/models/BucketInfo.ts
```diff
@@ -60,22 +63,83 @@ export type VeeamSOSApi = {
...
+        Capacity: bigint | number,
+        Available: bigint | number,
+        Used: bigint | number,
         LastModified?: string,
     },
 };
 
+// Bigints are exported as strings
+export type CapacityInfoExported = Omit<VeeamSOSApiSchema, 'CapacityInfo'> & {
+    CapacityInfo?: {
```
francoisferrand: what do you mean "exported" ?
it creates confusion with javascript export, and does help to understand where it would be in this stringified form: in mongo? when serialized to json? in "network" messages?...

lib/models/BucketInfo.ts
```diff
@@ -38,7 +38,10 @@ export type VersioningConfiguration = {
     MfaDelete: any;
 };
 
-export type VeeamSOSApi = {
+// The input VeeamSOSAPI type, where the capcity-related
+// field accept numbers, but will be treated as bigints
+// internally.
```
francoisferrand: this add a line to explain where this 'schema' applies : is it to the (javascript) API ? to the data we get/put in mongo ? ...

lib/models/BucketInfo.ts
```diff
@@ -60,22 +63,83 @@ export type VeeamSOSApi = {
...
 // our S3 implementation, at bucket level
 export type Capabilities = {
     VeeamSOSApi?: VeeamSOSApi,
 };
 
+// CapabilitiesInput is the input type for the Capabilities object
+// where the capacity-related fields can be numbers.
+export type CapabilitiesInput = {
+    VeeamSOSApi?: VeeamSOSApiSchema,
+};
```
francoisferrand: do we need a separate type here, or could we inline it inside `ParsableBucketMetadata` to reduce "noise" ?

lib/models/BucketInfo.ts
```diff
@@ -60,22 +63,83 @@ export type VeeamSOSApi = {
...
+    azureInfo: any | null,
+    objectLockEnabled: boolean | null,
+    objectLockConfiguration: any | null,
+    notificationConfiguration: any | null,
+    tags: Array<BucketTag>,
+    capabilities: Capabilities | undefined,
+    quotaMax: bigint | number,
+};
+
+export type ParsableBucketMetadata = Omit<Omit<BucketMetadata, 'quotaMax'>, 'capabilities'> & {
```
francoisferrand: probably same problem as `CapacityInfoExported` : should find and use a single convention, and better express "where" this is used --> maybe `BucketMetadataJson` and `CapacityInfoJson`, or other suffix (`Serializable`, `Safe`, ...)

lib/models/BucketInfo.ts
```diff
@@ -293,7 +379,7 @@ export default class BucketInfo {
         }
         assert.strictEqual(areTagsValid(tags), true);
         if (quotaMax) {
-            assert.strictEqual(typeof quotaMax, 'number');
+            assert(typeof quotaMax === 'bigint' || typeof quotaMax === 'number');
```
francoisferrand: duplicated & inconsistent check with line 343:
```
        if (quotaMax) {
            assert.strictEqual(typeof quotaMax, 'bigint', 'Quota must be a BigInt');
            assert(quotaMax >= 0, 'Quota cannot be negative');
        }
```

lib/models/BucketInfo.ts
```diff
@@ -323,8 +409,17 @@ export default class BucketInfo {
...
+
+        convertedCapacityInfo = BucketInfo.convertCapacityToBigInt(capabilities?.VeeamSOSApi?.CapacityInfo);
+        this._capabilities = capabilities ? {
+            ...capabilities,
+            VeeamSOSApi: {
+                ...capabilities?.VeeamSOSApi,
+                CapacityInfo: convertedCapacityInfo,
+            }
+        } : undefined;
+
```
francoisferrand: nit:

```suggestion
        this._capabilities = capabilities && {
            ...capabilities,
            VeeamSOSApi: {
                ...capabilities?.VeeamSOSApi,
                CapacityInfo: convertedCapacityInfo,
            }
        };
    
```

lib/models/BucketInfo.ts
```diff
@@ -359,8 +454,16 @@ export default class BucketInfo {
...
-            quotaMax: this._quotaMax,
+            capabilities: this._capabilities ? {
+                ...this._capabilities,
+                VeeamSOSApi: this._capabilities.VeeamSOSApi ? {
+                    ...this._capabilities.VeeamSOSApi,
+                    CapacityInfo: BucketInfo.serializeCapacityInfo(
+                        this._capabilities.VeeamSOSApi.CapacityInfo
+                    ),
+                } : undefined,
+            } : undefined,
```
francoisferrand: nit:
```suggestion
            capabilities: this._capabilities && {
                ...this._capabilities,
                VeeamSOSApi: this._capabilities.VeeamSOSApi && {
                    ...this._capabilities.VeeamSOSApi,
                    CapacityInfo: BucketInfo.serializeCapacityInfo(
                        this._capabilities.VeeamSOSApi.CapacityInfo
                    ),
                },
            },
```

lib/models/BucketInfo.ts
```diff
@@ -377,6 +480,13 @@ export default class BucketInfo {
      */
     static deSerialize(stringBucket: string) {
         const obj = JSON.parse(stringBucket);
+        const capabilities: CapabilitiesInput = {
+            ...obj.capabilities,
+            VeeamSOSApi: {
+                ...obj.capabilities?.VeeamSOSApi,
+                CapacityInfo: this.convertCapacityToBigInt(obj.capabilities?.VeeamSOSApi?.CapacityInfo),
+            },
+        };
```
francoisferrand: should we "create" the fields, if not provided; or keep the "status-quo", as before?

```suggestion
        const capabilities: CapabilitiesInput = obj.capabilities && {
            ...obj.capabilities,
            VeeamSOSApi: obj.capabilities.VeeamSOSApi && {
                ...obj.capabilities.VeeamSOSApi,
                CapacityInfo: this.convertCapacityToBigInt(obj.capabilities.VeeamSOSApi.CapacityInfo),
            },
        };
```

lib/models/BucketInfo.ts
```diff
@@ -405,6 +515,13 @@ export default class BucketInfo {
      * @return Return an BucketInfo
      */
     static fromObj(data: any) {
+        const capabilities: CapabilitiesInput = {
+            ...data._capabilities,
+            VeeamSOSApi: {
+                ...data._capabilities?.VeeamSOSApi,
+                CapacityInfo: this.convertCapacityToBigInt(data._capabilities?.VeeamSOSApi?.CapacityInfo),
+            },
+        };
```
francoisferrand: same:
```suggestion
        const capabilities: CapabilitiesInput = data._capabilities && {
            ...data._capabilities,
            VeeamSOSApi: data._capabilities.VeeamSOSApi && {
                ...data._capabilities.VeeamSOSApi,
                CapacityInfo: this.convertCapacityToBigInt(data._capabilities.VeeamSOSApi.CapacityInfo),
            },
        };
```

lib/models/BucketInfo.ts
```diff
@@ -414,8 +531,39 @@ export default class BucketInfo {
...
+            data._notificationConfiguration, data._tags, capabilities,
+            BigInt(data._quotaMax || 0n));
+    }
+
+    /**
+     * Converts the input Veeam capacity info to BigInts
+     * @param capacityInfo - capacity info to convert
+     * @returns - converted capacity info
+     */
+    static convertCapacityToBigInt(capacityInfo: VeeamSOSApiSchema['CapacityInfo']): VeeamSOSApi['CapacityInfo'] | undefined {
```
francoisferrand: * should this one not work when the values are actually formatted as `string` ?
i.e. so we can do a round-trip `convertCapacityToBigInt(serializeCapacityInfo(convertCapacityToBigInt(...)))`

* could we move these 2 functions are methods of `VeeamSOSApiSchema['CapacityInfo']` and `VeeamSOSApi['CapacityInfo']`, for better locally and readability?

* name should match the convention used for the differentiating the flavours of types (e.g. if we are `VeeamSOSApi` and `VeeamSOSApiSerializable` the method should be `toSerializable`), for better readability.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -70,25 +71,198 @@ function inc(str) {
...
+}
+
+export type ParsableCapabilitiesMongoDB = Capabilities & {
+    VeeamSOSApi?: Omit<VeeamSOSApi, 'CapacityInfo'> & {
+        CapacityInfo?: {
+            Capacity: string,
+            Available: string,
+            Used: string,
+        },
+    },
```
francoisferrand: CapacityInfoExported

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -221,26 +403,30 @@ class MongoClientInterface {
...
         const bucketMDStr = bucketInfo.serialize();
         const newBucketMD = JSON.parse(bucketMDStr);
-        const m = this.getCollection(METASTORE);
+        const m = this.getCollection<BucketMetastoreDocument>(METASTORE);
 
         const payload = {
             $set: {
                 _id: bucketName,
                 value: newBucketMD,
+                vFormat: BUCKET_VERSIONS.v0,
```
francoisferrand: technically, this has exactly the same result ; however reading this gives the false impression that the default is `v0` : when actually most of the time is the other way around...

--> would be more readable with either set `this.defaultBucketKeyFormat` and revert to v0 for usersBucket/pensieve/mpu, or create a function to "compute" the format (like before, but with `return`) and simply use it here...

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -221,26 +403,30 @@ class MongoClientInterface {
...
         };
         if (bucketName !== constants.usersBucket &&
             bucketName !== PENSIEVE &&
             !bucketName.startsWith(constants.mpuBucketPrefix)) {
             payload.$set.vFormat = this.defaultBucketKeyFormat;
-        } else {
-            payload.$set.vFormat = BUCKET_VERSIONS.v0;
+        }
+
+        if (!m) {
```
francoisferrand: nit: should go on line 413, right after we initialize `m`

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -252,30 +438,37 @@ class MongoClientInterface {
...
             .then(() => {
                 // caching bucket vFormat
-                this.bucketVFormatCache.add(bucketName, payload.vFormat);
-                this.lastItemScanTime = null;
+                this.bucketVFormatCache.add(bucketName, payload.$set.vFormat);
                 // NOTE: We do not need to create a collection for
                 // "constants.usersBucket" and "PENSIEVE" since it has already
                 // been created
                 if (bucketName !== constants.usersBucket && bucketName !== PENSIEVE) {
+                    if (!this.db) {
```
francoisferrand: this cannot happen :
* `this.db` is set only once on succesful connect()
* if we don't connect, we won't be able to get the connection (you added a check at the beginning)
* hence if we reach here, `this.db` is set already: no need to check...

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -252,30 +438,37 @@ class MongoClientInterface {
...
+                    }
                     return this.db.createCollection(bucketName)
                         .then(() => {
                             if (this.shardCollections) {
                                 const cmd = {
                                     shardCollection: `${this.database}.${bucketName}`,
                                     key: { _id: 1 },
                                 };
-                                return this.adminDb.command(cmd, {}).then(() => cb()).catch(err => {
+                                if (!this.adminDb) {
```
francoisferrand: same

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -301,7 +498,27 @@ class MongoClientInterface {
...
-                const bucketMDStr = JSON.stringify(doc.value);
+                // For "Long" values, convert to a string: the BucketInfo
+                // deSerialize function accepts a string for quotaMax
+                const parsedCapacityInfo = doc.value.capabilities.VeeamSOSApi?.CapacityInfo ? {
+                    ...doc.value.capabilities.VeeamSOSApi.CapacityInfo,
+                    Capacity: doc.value.capabilities.VeeamSOSApi.CapacityInfo.Capacity.toString(),
+                    Available: doc.value.capabilities.VeeamSOSApi.CapacityInfo.Available.toString(),
+                    Used: doc.value.capabilities.VeeamSOSApi.CapacityInfo.Used.toString(),
+                } : undefined;
+
```
francoisferrand: same as earlier comment from BucketInfo : can't we use `serializeCapacityInfo` instead?

or better yet: can't we fix the `FIXME` comment above, by removing both serialize JSON.stringify and BucketInfo.deSerialize and using `BucketInfo.fromObj()` instead? (may require that `Long` is supported in `convertCapacityToBigInt`)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -381,13 +608,21 @@ class MongoClientInterface {
         });
     }
 
-    putBucketAttributes(bucketName, bucketMD, log, cb) {
+    putBucketAttributes(bucketName: string, bucketMD: BucketInfo, log: werelogs.Logger, cb: NodeCallBack<void>) {
         // FIXME: there should be a version of BucketInfo.serialize()
         // that does not JSON.stringify()
         const bucketInfo = BucketInfo.fromObj(bucketMD);
```
francoisferrand, line 568: is this needed, e.g. is input not already of type `BucketInfo` ?

or maybe we want to keep it anyway, to ensure it works as before called with a plain object with the same fields (since we don't have typescript everywhere) ? then we may be conditional depending on type:

```suggestion
const bucketInfo = bucketMD instanceof BucketInfo ? bucketMD : BucketInfo.fromObj(bucketMD);
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1226,20 +1577,26 @@ class MongoClientInterface {
...
+            $gt: versionKey,
+            $lt: lastVersionKey,
+        };
         if (vFormat === BUCKET_VERSIONS.v0) {
             filter.$gt = masterKey;
             filter.$lt = lastVersionKey;
-        } else {
-            filter.$gt = versionKey;
-            filter.$lt = lastVersionKey;
         }
```
francoisferrand: nit:


```suggestion
        const filter = vFormat === BUCKET_VERSIONS.v0 ? {
            $gt: masterKey;
            $lt: lastVersionKey;
        } : {
            $gt: versionKey,
            $lt: lastVersionKey,
        };
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1353,16 +1726,24 @@ class MongoClientInterface {
...
+        mst: { versionId: string },
+        vFormat: string,
+        log: werelogs.Logger,
+        cb: NodeCallBack<void>,
+    ) {
         const masterKey = formatMasterKey(objName, vFormat);
         // Check if there are other versions available
         this.getLatestVersion(c, objName, vFormat, log, (err, version) => {
-            if (err && !err.is.NoSuchKey) {
+            if (err instanceof ArsenalError && !err.is.NoSuchKey) {
```
francoisferrand: not needed : can only return `errors.InternalError` or `errors.NoSuchKey``

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1114,7 +1448,7 @@ class MongoClientInterface {
                 // or last version is delete marker
                 if (!doc || doc.value.isPHD) {
                     this.getLatestVersion(c, objName, vFormat, log, (err, value) => {
-                        if (err && err.is.NoSuchKey) {
+                        if (err instanceof ArsenalError && err.is.NoSuchKey) {
```
francoisferrand: not needed, can only return `errors.InternalError` or `errors.NoSuchKey`

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -841,7 +1137,7 @@ class MongoClientInterface {
         }, {
             upsert: true,
         }).then(() => this.getLatestVersion(c, objName, params.vFormat, log, (err, mstObjVal) => {
-            if (err && err.is.NoSuchKey) {
+            if (err instanceof ArsenalError && err.is.NoSuchKey) {
```
francoisferrand: not needed, getLatestVersion can only return `errors.InternalError` or `erros.NoSuchKey`

(note: we could statically check this by change the callback type to accept an `ArsenalError` argument)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -361,7 +588,7 @@ class MongoClientInterface {
             }
             this.getObject(bucketName, objName, params, log, (err, obj) => {
                 if (err) {
-                    if (err.is.NoSuchKey) {
+                    if (err instanceof ArsenalError && err.is.NoSuchKey) {
```
francoisferrand: not needed, `getObject` only yields arsenal errors

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1353,16 +1726,24 @@ class MongoClientInterface {
...
         // Check if there are other versions available
         this.getLatestVersion(c, objName, vFormat, log, (err, version) => {
-            if (err && !err.is.NoSuchKey) {
+            if (err instanceof ArsenalError && !err.is.NoSuchKey) {
                 log.error('getLatestVersion: error getting latest version',
                     { error: err.message, bucket: bucketName, key: objName });
                 return cb(err);
             }
-            if ((err && err.is.NoSuchKey) || (version.isDeleteMarker && vFormat === BUCKET_VERSIONS.v1)) {
+            if ((err instanceof ArsenalError && err.is.NoSuchKey) || (version!.isDeleteMarker && vFormat === BUCKET_VERSIONS.v1)) {
```
francoisferrand: same

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1378,7 +1759,7 @@ class MongoClientInterface {
                         // the PHD master might get updated when a PUT is performed
                         // before the repair is done, we don't want to return an error
                         // in this case
-                        if (err.is.NoSuchKey) {
+                        if (err instanceof ArsenalError && err.is.NoSuchKey) {
```
francoisferrand: same

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1449,7 +1838,7 @@ class MongoClientInterface {
                     // we don't return an error in case we don't find
                     // a version as we expect this case when dealing with
                     // a versioning suspended object.
-                    if (err && err.is.NoSuchKey) {
+                    if (err instanceof ArsenalError && err.is.NoSuchKey) {
```
francoisferrand: same

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1847,7 +2291,7 @@ class MongoClientInterface {
             if (params.secondaryStreamParams) {
                 // eslint-disable-next-line no-param-reassign
                 newParams.mainStreamParams.gte = range[0];
-                newParams.secondaryStreamParams.gte = range[1];
+                newParams.secondaryStreamParams!.gte = range[1];
```
francoisferrand: same, but without any "trick" :

```suggestion
            if (newParams.secondaryStreamParams) {
                // eslint-disable-next-line no-param-reassign
                newParams.mainStreamParams.gte = range[0];
                newParams.secondaryStreamParams.gte = range[1];
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1971,7 +2429,7 @@ class MongoClientInterface {
 
     checkHealth(implName, log, cb) {
         const resp = {};
-        if (this.client && this.client.topology && this.client.topology.isConnected()) {
+        if (this.isConnected) {
```
francoisferrand, line 2358: we used to check the topology, now we rely on the `close` event : does this not change the behavior (or cause the risk there of) ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2020,11 +2486,11 @@ class MongoClientInterface {
      * we always try to generate a new UUID in order to be atomic in
      * case of concurrency. The write will fail if it already exists.
      */
-    getUUID(log, cb) {
+    getUUID(log: werelogs.Logger, cb: NodeCallBack<string | ObjectMDStats>) {
         const uuid = initialInstanceID || Uuid.v4();
         this.writeUUIDIfNotExists(uuid, log, err => {
             if (err) {
-                if (err.is.InternalError) {
+                if (err instanceof ArsenalError && err.is.InternalError) {
```
francoisferrand: not needed, can only be errors.KeyAlreadyExists or errors.InternalError

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2127,11 +2602,11 @@ class MongoClientInterface {
...
      * @return { undefined }
      */
-    getBucketInfos(log, cb) {
+    getBucketInfos(log: werelogs.Logger, cb: NodeCallBack<{ bucketCount: number, bucketInfos: BucketInfo[] }>) {
         let bucketCount = 0;
-        const bucketInfos = [];
+        const bucketInfos: BucketInfo[] = [];
 
-        this.db.listCollections({ type: 'collection' }).toArray().then(collInfos =>
+        this.db?.listCollections({ type: 'collection' }).toArray().then(collInfos =>
```
francoisferrand: if `db` is null, then we will call `<null>.toArray()`, and anyway not call the callback...
--> either early fail if !this.db (like in the other functions) or assert it is not null (`this.db!.`)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2144,7 +2619,7 @@ class MongoClientInterface {
                 // in $regex
                 return this.getBucketAttributes(bucketName, log,
                     (err, bucketInfo) => {
-                        if (err?.is?.NoSuchBucket) {
+                        if (err instanceof ArsenalError && err?.is?.NoSuchBucket) {
```
francoisferrand: not neeeded, can only be InternalError or NoSuchBucket

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2156,7 +2631,9 @@ class MongoClientInterface {
                             return next(errors.InternalError);
                         }
                         bucketCount++;
-                        bucketInfos.push(bucketInfo);
+                        if (bucketInfo) {
```
francoisferrand: not possible, value is passed unless errors... We could try to type the callback like this:

```
interface ArsenalCallback<T> {
    (err: null, result: T): void;
    (err: ArsenalError, result: undefined): void;
};
```

...but while it may ensure the callback is used (in the function) as expected, typescript may not be smart enough to understand that, or even stop type inferrence because of the overloads.

seems like a good use for null-assertion though...
```suggestion
                        bucketInfos!.push(bucketInfo);
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2282,10 +2777,14 @@ class MongoClientInterface {
                             this.getObjectMDStats(bucketName, bucketInfo,
                                 isTransient, log, next);
                         },
-                    ], (err, results) => {
+                    ], (err, results: any) => {
```
francoisferrand:
```suggestion
                    ], (err, results: ObjectMDStats) => {
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2282,10 +2777,14 @@ class MongoClientInterface {
...
                         if (err) {
                             return done(err);
                         }
+                        if (!results) {
```
francoisferrand: cannot happen if err is not null, would be best to type assert i think...

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2328,12 +2827,12 @@ class MongoClientInterface {
...
         async.waterfall([
-            next => this.getObject(PENSIEVE, overlayVersionId, {}, log, next),
+            next => this.getObject(PENSIEVE, overlayVersionId, null, log, next),
             (version, next) => {
                 const overlayConfigId = `configuration/overlay/${version}`;
-                return this.getObject(PENSIEVE, overlayConfigId, {}, log, next);
+                return this.getObject(PENSIEVE, overlayConfigId, null, log, next);
             },
-        ], (err, res) => {
+        ], (err, res: any) => {
```
francoisferrand:
```suggestion
        ], (err, res: ObjectMDData) => {
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -321,15 +538,19 @@ class MongoClientInterface {
      * @param {Function} cb callback
      * @return {undefined}
      */
-    getBucketVFormat(bucketName, log, cb) {
+    getBucketVFormat(bucketName: string, log: werelogs.Logger, cb: NodeCallBack<string>) {
```
francoisferrand: should we mark this function as yielding "a bucket format" (not a string!), i.e. either BUCKET_VERSIONS.v1 or BUCKET_VERSIONS.v0 ?

it would allow removing many type assertions `vFormat!`

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2577,29 +3090,33 @@ class MongoClientInterface {
...
+                    error: reshapeExceptionError(err as ErrorLike),
                 });
                 return cb(errors.InternalError);
             }
             return this.internalDeleteObject(c, bucketName, masterKey, filter, null, log,
                 err => {
                     if (err) {
                         // unable to find an object that matches the conditions
-                        if (err.is.NoSuchKey) {
+                        if (err instanceof ArsenalError && err.is.NoSuchKey) {
```
francoisferrand: can only be InternalError or NoSuchKey

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2708,13 +3232,17 @@ class MongoClientInterface {
...
+        if (!c) {
+            log.error('deleteBucketIndexes: error getting bucket collection', { bucketName });
+            return cb(errors.InternalError);
+        }
         async.each(indexSpecs,
             (spec, next) => c.dropIndex(spec.name).then(() => next()).catch(err => next(err)),
             err => {
                 if (err) {
-                    if (err.codeName === 'NamespaceNotFound') {
+                    if (err instanceof MongoServerError && err.codeName === 'NamespaceNotFound') {
```
francoisferrand, line 3153: do we need the instance test?
can we do something like `err.codeName? === 'NamespaceNotFound'`, since we don't really care about the type (only that it may have a codeName field, and in that case, if the value match)?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2734,8 +3262,18 @@ class MongoClientInterface {
...
      */
-    getBucketIndexes(bucketName, log, cb) {
-        const c = this.getCollection(bucketName);
+    getBucketIndexes(bucketName: string, log: werelogs.Logger, cb: NodeCallBack<{
+        name: any;
+        keys: {
+            key: any;
+            order: any;
+        }[];
+    }[]>) {
```
francoisferrand: these are `string`s, not `any` ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2753,14 +3291,22 @@ class MongoClientInterface {
 
     getIndexingJobs(log, cb) {
         // list active createIndexes jobs
-        this.adminDb.command({
+        this.adminDb?.command({
```
francoisferrand: if adminDb is null-ish, `this.adminDb?.command` will return null as well, and we will still crash and not call the callback, right?
-> either assert type (`this.adminDb!.`) or yield an error early?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2753,14 +3291,22 @@ class MongoClientInterface {
...
             currentOp: true,
             $or: [
                 { 'op': 'command', 'command.createIndexes': { $exists: true } },
                 { op: 'none', msg: /^Index Build/ },
             ],
         }).then(res => {
-            const jobs = [];
+            const jobs: {
```
francoisferrand, line 3205: maybe we should migrate `utils.js` to typescript as well, to avoid `any`...

tests/unit/models/BucketInfo.spec.js
```diff
@@ -292,8 +293,16 @@ Object.keys(acl).forEach(
...
-                    capabilities: dummyBucket._capabilities,
-                    quotaMax: dummyBucket._quotaMax,
+                    capabilities: dummyBucket._capabilities ? {
+                        ...dummyBucket._capabilities,
+                        VeeamSOSApi: dummyBucket._capabilities.VeeamSOSApi ? {
+                            ...dummyBucket._capabilities.VeeamSOSApi,
+                            CapacityInfo: BucketInfo.serializeCapacityInfo(
+                                dummyBucket._capabilities.VeeamSOSApi.CapacityInfo,
+                            ),
+                        } : undefined,
```
francoisferrand: we have this multiple times, probably should have another dedicated function `BucketInfo.serializeCapabilities()` or `Capabilities.serialize()`

tests/unit/models/BucketInfo.spec.js
```diff
@@ -292,8 +293,16 @@ Object.keys(acl).forEach(
                         dummyBucket._objectLockConfiguration,
                     notificationConfiguration: dummyBucket._notificationConfiguration,
                     tags: dummyBucket._tags,
-                    capabilities: dummyBucket._capabilities,
-                    quotaMax: dummyBucket._quotaMax,
+                    capabilities: dummyBucket._capabilities ? {
+                        ...dummyBucket._capabilities,
+                        VeeamSOSApi: dummyBucket._capabilities.VeeamSOSApi ? {
+                            ...dummyBucket._capabilities.VeeamSOSApi,
```
francoisferrand, line 300:
```suggestion
                    capabilities: dummyBucket._capabilities && {
                        ...dummyBucket._capabilities,
                        VeeamSOSApi: dummyBucket._capabilities.VeeamSOSApi && {
                            ...dummyBucket._capabilities.VeeamSOSApi,
```

lib/models/BucketInfo.ts
```diff
@@ -38,44 +39,54 @@ export type VersioningConfiguration = {
...
+    azureInfo?: any,
+    objectLockEnabled?: boolean,
+    objectLockConfiguration?: any,
+    notificationConfiguration?: any,
+    tags: Array<BucketTag>,
+    capabilities?: Capabilities,
+    quotaMax: bigint | number,
+};
+
+export type BucketMetadataJSON = Omit<Omit<BucketMetadata, 'quotaMax'>, 'capabilities'> & {
```
francoisferrand:
```suggestion
export type BucketMetadataJSON = Omit<BucketMetadata, 'quotaMax' | 'capabilities'> & {
```

lib/models/BucketInfo.ts
```diff
@@ -38,44 +39,54 @@ export type VersioningConfiguration = {
...
+    quotaMax: string;
+    capabilities: {
+        VeeamSOSApi?: VeeamSOSApiSchema,
```
francoisferrand: is this not the same as the "original" type, did you mean `VeeamSOSApiSerializable` ?

```suggestion
        VeeamSOSApi?: VeeamSOSApiSerializable,
```

lib/models/BucketInfo.ts
```diff
@@ -102,7 +113,7 @@ export default class BucketInfo {
     _azureInfo: any | null;
     _ingestion: { status: 'enabled' | 'disabled' } | null;
     _capabilities?: Capabilities;
-    _quotaMax: number | 0;
+    _quotaMax: bigint;
```
francoisferrand, line 116: quotaMax may not be set?

```suggestion
    _quotaMax?: bigint;
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -70,25 +71,181 @@ function inc(str) {
...
+        ingestion: boolean;
+    }[],
+    locations?: any;
+    buckets?: number;
+    bucketWithQuotaCount?: number;
+    stalled: number;
+};
+
+interface ArsenalCallback<T> {
+    (err: null, result: T): void;
```
francoisferrand, line 219: nit: could we add another overload with no paramter, when T is void ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -116,15 +274,15 @@ class MongoClientInterface {
 
         this.bucketVFormatCache = new LRUCache(constants.maxCachedBuckets);
         this.defaultBucketKeyFormat = [BUCKET_VERSIONS.v0, BUCKET_VERSIONS.v1]
-            .includes(process.env.DEFAULT_BUCKET_KEY_FORMAT) ? process.env.DEFAULT_BUCKET_KEY_FORMAT
-            : BUCKET_VERSIONS.v1;
+            .includes(<BucketVersioningFormat>process.env.DEFAULT_BUCKET_KEY_FORMAT) ?
```
francoisferrand: nit: semantically, this is not correct (we cannot assume the `env` is indeed one of the values), it should be the otherway: `[<string>BUCKET_VERSIONS.v0, <string>BUCKET_VERSIONS.v1].includes(process.env.DEFAULT_BUCKET_KEY_FORMAT)` ?

(probably does not make any difference in the end, but we never know...)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1132,7 +1391,12 @@ class MongoClientInterface {
                 MongoUtils.unserialize(doc.value);
                 return next(null, doc.value);
             },
-        ], cb);
+        ], (err: ArsenalError | null | undefined, result?: ObjectMDData) => {
+            if (err) {
+                return cb(err);
+            }
+            return cb(null, result!);
+        });
```
francoisferrand, line 1400: can't anymore pass `cb` directly here, it fails type-check?
maybe we could/should use `cb!` ?

lib/models/BucketInfo.ts
ghost: The quota is always set, with a default to 0 (means no quota)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2156,35 +2561,39 @@ class MongoClientInterface {
...
                 });
             })).catch(err => {
-            log.error('could not get list of collections', {
-                method: '_getBucketInfos',
-                error: err,
+                log.error('could not get list of collections', {
+                    method: '_getBucketInfos',
+                    error: err,
+                });
+                return cb(err);
```
francoisferrand: this one may not be an ArsenalError, e.g. in case `listCollections().toArray()` fails?
though i don't understand why it would pass type-check :-/ (esp. when a type assertion is required on line 2569)

benzekrimaha (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2342,6 +2760,8 @@ class MongoClientInterface {
                 return cb(err);
             }
             const isTransient =
+                // @ts-expect-error the locations type is not defined
+                // but is the historical check. Is this really needed?
```
francoisferrand, line 2775: indeed, the field is `location` 🤦

I guess either:
* it never really mattered because transient locations are not used in Zenko ;
* or this was working in "Zenko 1" with `locations` field, and at some point the field was renamed from locations (plural) to location, maybe to support seamless "upgrade" from 1 location to multiple?

we should create a ticket so we investigate this, both a product (is transient supposed to work) and eng (should we just look at the `location` field).

NOTE: looking at ObjectMD, `res.location` may also be a single, non-array field...

lib/models/BucketInfo.ts
francoisferrand: it is true even in case of "upgrade", i.e. if the bucket was created before quotas were introduced? (and so mongo may not have the field)
i.e. is this structure what we have in memory (where indeed it may also be set), or what we have in mongo (where it may be missing)

francoisferrand (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: Yes, the async type is not matching (`AsyncResultCallback<unknown, ArsenalError>`).

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: To be done in ARSN-459, I will update the comment

lib/models/BucketInfo.ts
ghost: I mean, the constructor accepts an empty / void / undefined / null quota, but will always default to 0 in the class. So whatever the starting state: either we have that quota already, either it will be set to 0 at the next update.

So yes, this is fully backward compatible this way.

lib/models/BucketInfo.ts
ghost: Updated and adapted the logic post type change

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: I added one overload that allows `cb()` instead of `cb(null)`.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: Updated

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: I updated to return the arsenal error if the error is of this instance, otherwise, InternalError.

PR #2280, opened 2025-01-17
Cache env variables

francoisferrand:
Reading env variables is costly, so do it only once on startup.

Issue: ARSN-442


KillianG (approved):

ghost (approved):
:+1:

PR #2281, opened 2025-01-24
ARSN-458: Add action scuba:GetMetricsBatch

BourgoisMickael:
GetMetricsBatch action is used by identisee

branch 7.70 for vault
and 8.1 for scuba

### DIFF
--- a/lib/policyEvaluator/utils/actionMaps.ts
+++ b/lib/policyEvaluator/utils/actionMaps.ts
@@ -203,6 +203,7 @@ const actionMapMetadata = {
 
 const actionMapScuba = {
     GetMetrics: 'scuba:GetMetrics',
+    GetMetricsBatch: 'scuba:GetMetricsBatch',
     AdminStartIngest: 'scuba:AdminStartIngest',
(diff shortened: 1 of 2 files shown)


BourgoisMickael:
ping

BourgoisMickael:
@bert-e reset

BourgoisMickael:
@bert-e force_reset

dvasilas (approved):
This will be used to differentiate `GetMetricsBatch` from `GetMetrics` in monitoring, right ?

Because otherwise I see no need for a different authentication.

anurag4DSB (approved):

BourgoisMickael:
> This will be used to differentiate `GetMetricsBatch` from `GetMetrics` in monitoring, right ?
> 
> Because otherwise I see no need for a different authentication.

@dvasilas Yes.

I put it there as the type of the action is checked.

But if we differentiate in monitoring it is also better to differentiate in the permissions as well, but we can document to rather use an action like `scuba:GetMetrics*` to handle both.

PR #2286, opened 2025-02-17
Fix error in during S3 GET routes (streamed data)

ghost:
- The error was fixed and sent to the client early with code 200 before the xcore data was retrieved.
- If the data is lost, unavailable, or there is an issue with pipe to send back the data bytes to the client, the response is abruptly destroyed, causing, in the reverse proxy, a 502 error, as the S3 client won't send back the correct response anymore. This leads to memory leak in the revers proxy, wrong logging, and wrong error code being returned (502 instead of 503).
    
The fix is twofold:
- We write the http code early only when applicable.
- When not applicable, the http code is written after we confirm the data is rerieved from the storage layer.

Issue: ARSN-467

### DIFF
--- a/lib/s3routes/routesUtils.ts
+++ b/lib/s3routes/routesUtils.ts
@@ -67,8 +67,7 @@ export function okHeaderResponse(
     log: RequestLogger,
 ) {
     setCommonResponseHeaders(headers, response, log);
-    log.debug('response http code', { httpCode });
-    response.writeHead(httpCode);
+    writeHttpCodeHeaders(response, httpCode, {}, log);
     return response.end(() => {
         log.end().info('responded to request', {
             httpCode: response.statusCode,
@@ -99,7 +98,7 @@ export const XMLResponseBackend = {
             bytesSent,
         });
         setCommonResponseHeaders(additionalHeaders, response, log);
-        response.writeHead(200, { 'Content-type': 'application/xml' });
+        writeHttpCodeHeaders(response, 200, { 'Content-type': 'application/xml' }, log);
         log.trace('xml response', { xml });
         return response.end(xml, 'utf8', () => {
             log.end().info('responded with XML', {
@@ -123,7 +122,7 @@ export const XMLResponseBackend = {
         }
         // early return to avoid extra headers and XML data
         if (error.code === 304) {
-            response.writeHead(error.code);
+            writeHttpCodeHeaders(response, error.code, {}, log);
             return response.end('', 'utf8', () => {
                 log.end().info('responded with empty body', {
                     httpCode: response.statusCode,
@@ -164,10 +163,10 @@ export const XMLResponseBackend = {
         const bytesSent = Buffer.byteLength(xmlStr);
         log.addDefaultFields({ bytesSent });
         response.removeAllListeners('finish');
-        response.writeHead(error.code, {
+        writeHttpCodeHeaders(response, error.code, {
             'Content-Type': 'application/xml',
             'Content-Length': bytesSent ,
-        });
+        }, log);
         return response.end(xmlStr, 'utf8', () => {
             log.end().info('responded with error XML', {
                 httpCode: response.statusCode,
@@ -196,7 +195,7 @@ export const JSONResponseBackend = {
         const bytesSent = Buffer.byteLength(json);
         log.addDefaultFields({ bytesSent });
         setCommonResponseHeaders(additionalHeaders, response, log);
-        response.writeHead(200, { 'Content-type': 'application/json' });
+        writeHttpCodeHeaders(response, 200, { 'Content-type': 'application/json' }, log);
         log.trace('sending success json response', { json });
         return response.end(json, 'utf8', () => {
             log.end().info('responded with JSON', {
@@ -244,10 +243,10 @@ export const JSONResponseBackend = {
         const bytesSent = Buffer.byteLength(data);
         log.addDefaultFields({ bytesSent });
         setCommonResponseHeaders(corsHeaders, response, log);
-        response.writeHead(error.code, {
+        writeHttpCodeHeaders(response, error.code, {
             'Content-Type': 'application/json',
             'Content-Length': bytesSent,
-        });
+        }, log);
         return response.end(data, 'utf8', () => {
             log.end().info('responded with error JSON', {
                 httpCode: response.statusCode,
@@ -258,7 +257,7 @@ export const JSONResponseBackend = {
 
 
 /**
- * Modify response headers for an objectGet or objectHead request
+ * Set response headers for an objectGet or objectHead request
  * @param overrideParams - parameters in this object override common
  * headers. These are extracted from the request's query object
  * @param resHeaders - object with common response headers
@@ -268,13 +267,11 @@ export const JSONResponseBackend = {
  * @param log - Werelogs logger
  * @return response - modified response object
  */
-function okContentHeadersResponse(
+function setOkContentResponseHeaders(
     overrideParams: { [key: string]: string },
     resHeaders: { [key: string]: string },
     response: http.ServerResponse,
-    range: [number, number] | undefined,
     log: RequestLogger,
-    shouldWriteHttpCode: boolean = true,
 ) {
     const addHeaders: { [key: string]: string } = {};
     if (process.env.ALLOW_INVALID_META_HEADERS) {
@@ -319,12 +316,21 @@ function okContentHeadersResponse(
             overrideParams['response-content-encoding'];
     }
     setCommonResponseHeaders(addHeaders, response, log);
-    if (shouldWriteHttpCode) {
-        const httpCode = range ? 206 : 200;
-        log.debug('response http code', { httpCode });
-        response.writeHead(httpCode);
-    }
     return response;
+}
+
+function writeHttpCodeHeaders(
+    response: http.ServerResponse,
+    httpCode: number,
+    headers: http.OutgoingHttpHeaders | http.OutgoingHttpHeader[],
+    log: RequestLogger
+) {
+    if (!response.headersSent) {
+        log.debug('response http code', { httpCode });
+        response.writeHead(httpCode, headers);
+    } else {
+        log.debug('response http code already sent', { httpCode });
+    }
 }
 
 function retrieveDataAzure(
@@ -395,6 +401,11 @@ export function retrieveData(
         _destroyReadable(currentStream);
     });
 
+    response.on('error', err => {
+        log.error('error piping data from source');
+        response.socket?.destroy();
+    });
+
     const {
         client,
         implName,
@@ -444,11 +455,7 @@ export function retrieveData(
                     return cbOnce(err);
                 });
                 currentStream = readable;
-                return readable.pipe(response, { end: false })
-                    .on('error', pipeErr => {
-                        log.error('error in pipe operation', { error: pipeErr });
-                        return cbOnce(pipeErr);
-                    });
+                return readable.pipe(response, { end: false });
             }), err => {
                 currentStream = null;
                 if (err) {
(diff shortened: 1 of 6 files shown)


lib/s3routes/routesUtils.ts
```diff
@@ -277,6 +278,7 @@ function okContentHeadersResponse(
     response: http.ServerResponse,
     range: [number, number] | undefined,
     log: RequestLogger,
+    shouldWriteHttpCode: boolean = true,
```
francoisferrand: will this not be a breaking change in javascript?

e.g. existing code will not pass the parameter, and typescript compiler will not see the default parameter value in javascript code : so calls to this function may suddenly stop sending http code, and we need to hope that we have upgraded & tested all places where it is used.

Or is this function only ever used internally here?

ghost: It's only used internally but I agree that the default value should be removed, for safety...

francoisferrand: alternatively, depending on how it is ysed maybe it would cleaner to rework the code to either have 2 functions for the different use case, or have one function for headers and another for the response code...

ghost: But the rationale here is more like: we don't want to change anything, so we set to true by default, except in one well-known place (streamed data). What do you think?

lib/s3routes/routesUtils.ts
```diff
@@ -445,23 +444,33 @@ function retrieveData(
...
-                    log.error('error piping data from source');
-                    _destroyResponse();
+                    log.error('error piping data from source', { error: err });
                     return cbOnce(err);
                 });
                 currentStream = readable;
-                return readable.pipe(response, { end: false });
+                return readable.pipe(response, { end: false })
+                    .on('error', pipeErr => {
+                        log.error('error in pipe operation', { error: pipeErr });
```
francoisferrand: is there actually cases where we go (and need to go) there: I expect piping would fail either when reading from `readable` (already handled on line 446) or failing to write (handling in line 398) ?

lib/s3routes/routesUtils.ts
```diff
@@ -445,23 +444,33 @@ function retrieveData(
...
+                        error: err.code, errMsg: err.message });
+                    // Only end if we haven't already destroyed the response
+                    if (!responseDestroyed) {
+                        return XMLResponseBackend.errorResponse(errors.ServiceUnavailable, response, log);
+                    }
+                }
+                if (codeOnSuccess && !response.headersSent) {
+                    response.writeHead(codeOnSuccess);
+                }
+                // call end for all cases (error/success) per node.js docs
```
francoisferrand: not clear to me why/how/when this is needed, quite confusing to send response only if headers not sent, but `okContentHeadersResponse` was changed to allow changing headers without setting response code...

I need to re-review this with a fresher mind, tomorrow morning

ghost: This part is unchanged, but in short we should always end the request whatever we did, otherwise the client won't get notified. With the changes all paths are setting the header code, but even if there is none, we can still end the request.

But I agree that the whole code is hard to follow, the mix of paths in the same function is not trivial, plus the steps before, where we used to even set the headers before the actual request to the data service...

lib/s3routes/routesUtils.ts
ghost: I added this because in case of response error we had no error handling (we have readable "error" and "end", but only "end" for response... We have it for Azure though...).
Maybe better to have it separately then

lib/s3routes/routesUtils.ts
ghost: PR was updated and default was removed. I confirmed this is not used anywhere else.

lib/s3routes/routesUtils.ts
ghost: A specific response.on('error' handling logic was added, which thus let this above line unchanged.

lib/s3routes/routesUtils.ts
ghost: So I reworked a bit how it works, and writing the http headers is now done in a dedicated function called at the right places. Then it becomes easier to ensure we always write the headers. But we keep the .end() anyway, as a nodejs-documented best practise.

ghost:
I am reworking the PR as I had some conflicts to fix following the last arsenal merge.

Edit: Done

tests/unit/s3routes/routes.spec.js
```diff
@@ -28,6 +28,7 @@ describe('routes', () => {
             setHeader: sinon.stub(),
             writeHead: sinon.stub(),
             on: sinon.stub(),
+            removeAllListeners: sinon.stub(),
```
francoisferrand, line 31: should probably add (or modify) a test to exercise the scenario where we actually need to remove the listener: i.e. have a test which actually fails when the whatever function is called twice...

(testing that we call `removeAllListener` is just code coverage, does not ensure we solve the actual problem : it would be better to have an actual EventSender and simulate the events)

or is this tested already in some functional tests in here?

lib/s3routes/routesUtils.ts
```diff
@@ -317,12 +316,23 @@ function okContentHeadersResponse(
...
+    response: http.ServerResponse,
+    httpCode: number,
+    headers: http.OutgoingHttpHeaders | http.OutgoingHttpHeader[],
+    log: RequestLogger
+) {
+    if (!response.headersSent) {
+        log.debug('response http code', { httpCode });
+        response.writeHead(httpCode, headers);
+    } else {
+        log.debug('response http code already sent', { httpCode });
```
francoisferrand, line 332: when can we reach this case? is it still the case with your latest changes?

If we set the headers twice, there may be a logic error somewhere (bad automaton, bad handling of some race condition or corner case...), and it creates the risk that we do not actually send the _expected_ result (e.g. if the first header & next ones do not match)

lib/s3routes/routesUtils.ts
```diff
@@ -395,6 +401,11 @@ function retrieveData(
         _destroyReadable(currentStream);
     });
 
+    response.on('error', err => {
+        log.error('error piping data from source');
+        response.socket?.destroy();
```
francoisferrand, line 406: should set `responseDestroyed = true;` already?

lib/s3routes/routesUtils.ts
```diff
@@ -441,23 +451,29 @@ function retrieveData(
...
+                if (err) {
+                    log.error('abort response due to error', {
+                        // @ts-expect-error
+                        error: err.code, errMsg: err.message });
+                    // Only end if we haven't already destroyed the response
+                    if (!responseDestroyed) {
+                        return XMLResponseBackend.errorResponse(errors.ServiceUnavailable, response, log);
+                    }
+                }
+                if (codeOnSuccess) {
```
francoisferrand, line 470: if `codeOnSuccess` can be null (as when called from `streamUserErrorPage`), then we are exposed to the same problem you are trying to fix : i.e. we may have set a result already, then fail to retrieve the data....and the error code will not be set properly (the earliest `writeHttpCodeHeaders` will take precedence)

lib/s3routes/routesUtils.ts
```diff
@@ -647,8 +663,7 @@ export function responseStreamData(
         }
     }
     if (!response.headersSent) {
-        okContentHeadersResponse(overrideParams, resHeaders, response,
-            range, log);
+        setOkContentResponseHeaders(overrideParams, resHeaders, response, log);
```
francoisferrand, line 666: should this not be set at the end of `retrieveData`, just like the result code?

One idea to consider: we may keep the existing logic, but set the headers only on success, maybe the parameter to add is really a callback : so the "caller" can specify exactly what headers should be set eventually, on success; we keep this config where it is already (in `responseContentHeaders` for exemple) but delay the acutal changes to the response object.

lib/s3routes/routesUtils.ts
```diff
@@ -441,23 +451,29 @@ function retrieveData(
...
+                    log.error('abort response due to error', {
+                        // @ts-expect-error
+                        error: err.code, errMsg: err.message });
+                    // Only end if we haven't already destroyed the response
+                    if (!responseDestroyed) {
+                        return XMLResponseBackend.errorResponse(errors.ServiceUnavailable, response, log);
+                    }
+                }
+                if (codeOnSuccess) {
+                    writeHttpCodeHeaders(response, codeOnSuccess, {}, log);
```
francoisferrand, line 471: since headers are now set at the end of the call, should we not set them as well in the early return cases of the `retrieveData` function?

i.e. 
```
    if (locations.length === 0) {
        // ....should set headers & result here
        return response.end();
    }
    if (locations[0].azureStreamingOptions) {
        return retrieveDataAzure(locations, retrieveDataParams, response, log); // ....should pass whatever info and let `retrieveDataAzure` set the headers & result eventually
    }
```

lib/s3routes/routesUtils.ts
```diff
@@ -441,23 +451,29 @@ function retrieveData(
...
-            // recommendation
-            response.end();
-        },
+                currentStream = null;
+                if (err) {
+                    log.error('abort response due to error', {
+                        // @ts-expect-error
+                        error: err.code, errMsg: err.message });
+                    // Only end if we haven't already destroyed the response
+                    if (!responseDestroyed) {
```
francoisferrand, line 466: this condition does not look very clear to me.
- this flag is set when the response is closed, so obviously we cannot send a response afterwards;
- the response is closed (actually destroyed) on error, so i guess it tackles a corner case in the aftermath of such event;
- but the socket may also be closed without error... we cannot send response anymore indeed, but we still have a logic error if we sent a response already before coming here

Basically I'd rather have a more "predictable" logic, something like we do all the processing in this 'final' callback:
- on success, send the 'success' response
- if the response is in `error`, do nothing indeed (or maybe we can destroy the socket here as well, for better locality?)
- on other error (typically on the input stream), send the xml error response

lib/s3routes/routesUtils.ts
francoisferrand: btw,

* why do we need to destroy the socket? Is it not handled automatically by node.js, when we send the response? is it about releasing resources early? (not specifically related to this PR, as it was there before; but still wondering...)

* when does such "response" error happen? what are the symptoms? It seems unrelated to the NGINX problem AFAIU?

lib/s3routes/routesUtils.ts
```diff
@@ -163,10 +162,11 @@ export const XMLResponseBackend = {
         const xmlStr = xml.join('');
         const bytesSent = Buffer.byteLength(xmlStr);
         log.addDefaultFields({ bytesSent });
-        response.writeHead(error.code, {
+        response.removeAllListeners('finish');
```
francoisferrand, line 165: similar to setting headers & result codes, can't we always set the finish handler (or `end()` callback) in a single place, when sending the response, instead of "pre-setting" it and somehow replacing it when it is not appropriate?

lib/s3routes/routesUtils.ts
francoisferrand: actually, I wonder if this can work in all cases : esp. for large (ranged) objects, the body will already have been sent.... and thus we will not set the expected http code.

as par HTTP protocol, header should be set before body, right?
so the required logic may actually be more complex:

- "start" the backend stream (but do not write any data to the response yet!)
- on error, send the appropriate error;
  on success, set the headers
- stream the data
- on error, the only way to 'report' an error is to abort the socket/stream?
  on success, end() the request to terminate cleanly

what do you think?

francoisferrand: (note: maybe indeed that is what the code does now, but then only thanks to multiple flags and indirect conditions...which make it difficult to review/ensure we really handle all cases and won't create more corner cases)

PR #2287, opened 2025-02-17
Remove count items dead code

ghost:
- MongoClientInterface still implements some duplicate (and outdated) logic implemented by count items in s3utils.
- Dependency functions are removed, and tests cleaned accordingly.
- Some functions were also wrong: checking the PENSIEVE collection to check for transient locations was relying on undefined field from a collection that is also not used anymore in Zenko 2. The logic itself is only related to statistics, currently overwritten by the count-items implementation. The Cloudserver logic around ops-metrics is unaffected, as it only uses the internal `countItems` function, used to read the count items results.
- Transient locations are still supported as usual: they are unrelated to these functions.

I am expecting a functional test to fail in cloudserver after this change is merged, as the `countItems` function is functionally tested there.

Issue: ARSN-459


lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2481,31 +2481,6 @@ class MongoClientInterface {
         });
     }
 
-    updateCountItems(value: ObjectMDStats, log: werelogs.Logger, cb: ArsenalCallback<void>) {
```
francoisferrand, line 2484: i wonder if this function should not stay here, to keep the whole _"metadata storage abstraction"_ in Arsenal?
(i.e. keep `updateCountItems()` here, and remove the duplicate from s3utils)

Though the actual "logic" of countItems (i.e. `consolidateData`, `scanItemCount`...) should indeed be removed from here, and keep only the code already in s3utils.

NOTE: same goes for `_getIsTransient`/`_pensieveLocationIsTransient` which should stay here... maybe also `getObjectMDStats`, but it seems to me it mixes a bit of both metadata storage and countItems logic, so not sure where it should preferably go...

What do you think?

ghost: I think the count-items related functions are meant to be permanently deleted soon, hence, better keeping them in s3utils indeed. Count items functions are related to count items, and are logical to have in s3utils, I do not see why we have them in the mongo client interface in Arsenal... The `updateCountItems` function does not exist in s3utils (we use `updateStorageConsumptionMetrics`), we use another function already: no reason to keep it, as it is not used anywhere, and not compatible with the current count items.

On the PENSIEVE part: can you elaborate on why you think some functions should stay here? There is no customer nor platform today where this code will do anything, as the PENSIEVE collection is not used anymore.

francoisferrand: Really a general approach : mongo client interface provide an abstraction of the metadata layer ("how metadata is stored"), so it's good to keep it centralized this way I think (as much as we can anyway, since the other side of pensieve is in go).

The processing based is not part of the storage, and can indeed go in count-items.

So based on such criteria, part if it makes a mongo access, may stay in MongoClientInterface/arsenal ; and if it implements some processing (e.g. processing & aggregating the content of the metadatas) it shoud be moved out.

(but the more complex part is when one function does both mongo access & processing....)

> I think the count-items related functions are meant to be permanently deleted soon

it will in due time, hopefully as early as possible ; but for the moment it is not... and we don't know when exactly this will land, so let's not over-rush things. I don't mean we should not clean, only that the current state is we have count-items, and we should make changes based on this.

ghost: I disagree with the abstraction you want (I believe this is not a truth but an opinion), because from what I see in the code, MongoClientInterface is an abstraction below the MetadataWrapper, for S3 object metadata, similar to `file`, `mem`, `scality` (metadata) or other backends we use from the MetadataWrapper. What `MongoClientInterface` should **not** be is:

- A place to have functions like `getDiskUsage`, totally unrelated to MongoDB.
- A place to have count items logic: this is a specific use case, and similarly to Vault or Cloudserver operations requiring specific handling/reads of specific collections or documents, such logic should be in a dedicated place. We don't implement all types of reads from Vault here, for example, because this class is for S3 objects. These count items functions are not something we access from the metadata wrapper either (or we have dirty checks like `if (!this.client.countItems) {`, a good proof that it should not be here). S3utils implements `S3UtilsMongoClient` for this reason. It has a lot of read-only (with mongodb) logic, that should, if I follow the logic, be moved here. I can list `_getUsersBucketCreationDates`, `readStorageConsumptionMetrics`, `readStorageConsumptionInflights`, and more... This has zero value given the context with count items.
- [A place to keep dead code that we will never need anymore](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it): these functions are not used anywhere, they bring no value, only maintenance cost. If we need themback, we can always implement it using the previous code as we have historization, but we already know we will never need them (Zenko 1 is gone, Count Items does not use any of these functions and will not be maintained anymore, then removed). Keeping it is equal to keeping debt. Moving code between s3utils and arsenal is equal to adding even more debt: we know count-items is to be removed, so we should keep all the code there (in s3utils) so we don't add complexity where it's not needed. Better closing this PR than adding more debt IMHO.

francoisferrand (approved):

welansari (approved):

PR #2288, opened 2025-02-18
Enabling option passing to the authV4

benzekrimaha:
Issue : ARSN-469

-**_Note_** : the version of the project will be bumped

PR #2299, opened 2025-02-21
Typescript improvements

ghost:
Improve TS:

- Auth types are fixed as they were inconsistent. We still have a mix of 4 possible types as output whtn doing auth (authinfo, authv4 info, authv2 or public user).
- Old style `_` prefix to define private variable in TS classes was removed: this is not following the TS standard. More file might still be affected.
- Data models for both Bucket and Objects were standardized to avoid using `any`.
  - This allowed to find a limitation in the createBucket function in the mongo client interface.

Issue: ARSN-462

### DIFF
--- a/lib/auth/Vault.ts
+++ b/lib/auth/Vault.ts
@@ -152,11 +152,11 @@ export default class Vault {
      */
     authenticateV2Request(
         params: AuthV2RequestParams,
-        requestContexts: any[] | null,
+        requestContexts: RequestContext[] | null,
         callback: (err: Error | null, data?: any) => void
     ) {
         params.log.debug('authenticating V2 request');
-        let serializedRCsArr: any;
+        let serializedRCsArr;
         if (requestContexts) {
             serializedRCsArr = requestContexts.map(rc => rc.serialize());
         }
@@ -204,12 +204,12 @@ export default class Vault {
     */
     authenticateV4Request(
         params: AuthV4RequestParams,
-        requestContexts: any[] | null,
+        requestContexts: RequestContext[] | null,
         options: AuthenticationOptions = {},
         callback: (err: Error | null, data?: any) => void,
     ) {
         params.log.debug('authenticating V4 request');
-        let serializedRCs: any;
+        let serializedRCs;
         if (requestContexts) {
             serializedRCs = requestContexts.map(rc => rc.serialize());
         }
--- a/lib/auth/Vault.ts
+++ b/lib/auth/Vault.ts
@@ -69,14 +69,14 @@ export type AuthV2RequestParams = {
     version: 2;
     log: RequestLogger;
     data: {
-        securityToken: string;
         accessKey: string;
-        signatureFromRequest: string;
-        stringToSign: string;
         algo: string;
         authType: 'query' | 'header' | 'REST-HEADER' | 'REST-QUERY-STRING';
-        signatureVersion: string;
+        securityToken: string;
         signatureAge?: number;
+        signatureFromRequest: string;
+        signatureVersion: string;
+        stringToSign: string;
     };
 };
 
@@ -88,7 +88,6 @@ export type AuthV4RequestParams = {
         algo?: string;
         authType?: 'query' | 'header' | 'REST-HEADER' | 'REST-QUERY-STRING';
         credentialScope?: string;
-        log: RequestLogger;
         region: string;
         scopeDate: string;
         securityToken?: string;
@@ -97,7 +96,7 @@ export type AuthV4RequestParams = {
         signatureAge?: number;
         signatureFromRequest: string;
         stringToSign: string;
-        timestamp: number;
+        timestamp?: string;
     };
 };
 
@@ -153,7 +152,7 @@ export default class Vault {
      */
     authenticateV2Request(
         params: AuthV2RequestParams,
-        requestContexts: any[],
+        requestContexts: any[] | null,
         callback: (err: Error | null, data?: any) => void
     ) {
         params.log.debug('authenticating V2 request');
--- a/lib/auth/Vault.ts
+++ b/lib/auth/Vault.ts
@@ -70,7 +70,7 @@ export type AuthV2RequestParams = {
     log: RequestLogger;
     data: {
         accessKey: string;
-        algo: string;
+        algo: 'sha1' | 'sha256';
         authType: 'query' | 'header' | 'REST-HEADER' | 'REST-QUERY-STRING';
         securityToken: string;
         signatureAge?: number;
@@ -85,7 +85,7 @@ export type AuthV4RequestParams = {
     log: RequestLogger;
     data: {
         accessKey: string;
-        algo?: string;
+        algo?: 'sha1' | 'sha256';
         authType?: 'query' | 'header' | 'REST-HEADER' | 'REST-QUERY-STRING';
         credentialScope?: string;
         region: string;
@@ -293,7 +293,7 @@ export default class Vault {
     getEmailAddresses(
         canonicalIDs: string[],
         log: RequestLogger,
-        callback: (err: Error | null, data?: { [key: string]: any }) => void
+        callback: (err: Error | null, data?: Record<string, any>) => void
     ) {
         log.trace('getting emailAddresses from Vault based on canonicalIDs',
             { canonicalIDs });
@@ -331,7 +331,7 @@ export default class Vault {
     getAccountIds(
         canonicalIDs: string[],
         log: RequestLogger,
-        callback: (err: Error | null, data?: { [key: string]: string }) => void
+        callback: (err: Error | null, data?: Record<string, string>) => void
     ) {
         log.trace('getting accountIds from Vault based on canonicalIDs',
             { canonicalIDs });
--- a/lib/auth/Vault.ts
+++ b/lib/auth/Vault.ts
@@ -65,24 +65,39 @@ export function vaultSignatureCb(
         accountQuota: info.accountQuota || {},
     });
 }
+export type AuthV2RequestParams = {
+    version: 2;
+    log: RequestLogger;
+    data: {
+        securityToken: string;
+        accessKey: string;
+        signatureFromRequest: string;
+        stringToSign: string;
+        algo: string;
+        authType: 'query' | 'header' | 'REST-HEADER' | 'REST-QUERY-STRING';
+        signatureVersion: string;
+        signatureAge?: number;
+    };
+};
 
 export type AuthV4RequestParams = {
     version: 4;
     log: RequestLogger;
     data: {
         accessKey: string;
-        signatureFromRequest: string;
-        region: string;
-        stringToSign: string;
-        scopeDate: string;
-        authType: 'query' | 'header';
-        signatureVersion: string;
-        signatureAge?: number;
-        timestamp: number;
-        credentialScope: string;
-        securityToken: string;
-        algo: string;
+        algo?: string;
+        authType?: 'query' | 'header' | 'REST-HEADER' | 'REST-QUERY-STRING';
+        credentialScope?: string;
         log: RequestLogger;
+        region: string;
+        scopeDate: string;
+        securityToken?: string;
+        service?: string;
+        signatureVersion?: string;
+        signatureAge?: number;
+        signatureFromRequest: string;
+        stringToSign: string;
+        timestamp: number;
     };
 };
 
@@ -137,21 +152,7 @@ export default class Vault {
      * @param callback - callback with either error or user info
      */
     authenticateV2Request(
-        params: {
-            version: 2;
-            log: RequestLogger;
-            data: {
-                securityToken: string;
-                accessKey: string;
-                signatureFromRequest: string;
-                stringToSign: string;
-                algo: string;
-                authType: 'query' | 'header';
-                signatureVersion: string;
-                signatureAge?: number;
-                log: RequestLogger;
-            };
-        },
+        params: AuthV2RequestParams,
         requestContexts: any[],
         callback: (err: Error | null, data?: any) => void
     ) {
--- a/lib/auth/auth.ts
+++ b/lib/auth/auth.ts
@@ -17,8 +17,11 @@ import validateAuthConfig from './backends/in_memory/validateAuthConfig';
 import AuthLoader from './backends/in_memory/AuthLoader';
 import Vault, { AuthV2RequestParams, AuthV4RequestParams } from './Vault';
 
+export type AuthResult<T> = { err: ArsenalError } | { err: null, params: T };
+
 let vault: Vault | null = null;
 const auth = {};
+
 const checkFunctions = {
     v2: {
         headers: v2.header.check,
@@ -61,7 +64,7 @@ function extractParams(
     log: Logger,
     awsService: string,
     data: { [key: string]: string }
-): { err: null | Error | ArsenalError, params?: AuthV2RequestParams | AuthV4RequestParams | AuthInfo } {
+): AuthResult<AuthV2RequestParams | AuthV4RequestParams | AuthInfo> {
     log.trace('entered', { method: 'Arsenal.auth.server.extractParams' });
     const authHeader = request.headers.authorization;
     let version: 'v2' |'v4' | null = null;
@@ -127,7 +130,8 @@ function doAuth(
     const res = extractParams(request, log, awsService, request.query);
     if (res.err) {
         return cb(res.err);
-    } else if (res.params instanceof AuthInfo) {
+    }
+    if (res.params instanceof AuthInfo) {
         return cb(null, res.params);
     }
     if (requestContexts) {
@@ -146,11 +150,11 @@ function doAuth(
     }
 
     // Corner cases managed, we're left with normal auth
-    res.params!.log = log;
-    if (res.params!.version === 2) {
-        return vault!.authenticateV2Request(res.params!, requestContexts, cb);
+    res.params.log = log;
+    if (res.params.version === 2) {
+        return vault!.authenticateV2Request(res.params, requestContexts, cb);
     }
-    if (res.params!.version === 4) {
+    if (res.params.version === 4) {
         // @ts-ignore
         return vault!.authenticateV4Request(res.params, requestContexts, options, cb);
     }
--- a/lib/auth/auth.ts
+++ b/lib/auth/auth.ts
@@ -1,6 +1,7 @@
 import * as crypto from 'crypto';
-import { RequestLogger } from 'werelogs';
-import errors, { ArsenalError } from '../errors';
+import type { RequestLogger } from 'werelogs';
+import errors from '../errors';
+import type { ArsenalError } from '../errors';
 import * as queryString from 'querystring';
 import AuthInfo from './AuthInfo';
 import * as v2 from './v2/authV2';
@@ -73,7 +74,7 @@ function extractParams(
     request: any,
     log: RequestLogger,
     awsService: string,
-    data: { [key: string]: string }
+    data: Record<string, string>
 ): AuthResult<AuthV2RequestParams | AuthV4RequestParams | AuthInfo> {
     log.trace('entered', { method: 'Arsenal.auth.server.extractParams' });
     const authHeader = request.headers.authorization;
@@ -145,7 +146,7 @@ function doAuth(
         return cb(null, res.params);
     }
     const data = res.params.data;
-    if (requestContexts) {
+    if (data && requestContexts) {
         requestContexts.forEach(requestContext => {
             if (data.authType) {
                 requestContext.setAuthType(data.authType);
@@ -156,9 +157,7 @@ function doAuth(
             if (data.securityToken) {
                 requestContext.setSecurityToken(data.securityToken);
             }
-            if (data.signatureAge) {
-                requestContext.setSignatureAge(data.signatureAge);
-            }
+            requestContext.setSignatureAge(data.signatureAge || 0);
         });
     }
 
@@ -208,7 +207,7 @@ function generateContentMD5Header(
  */
 function generateV4Headers(
     request: any,
-    data: { [key: string]: string },
+    data: Record<string, string>,
     accessKey: string,
     secretKeyValue: string,
     awsService: string,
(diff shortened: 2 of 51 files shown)


lib/auth/auth.ts
```diff
@@ -145,16 +145,12 @@ function doAuth(
     }
 
     // Corner cases managed, we're left with normal auth
-    // TODO What's happening here?
-    // @ts-ignore
-    res.params.log = log;
-    if (res.params.version === 2) {
-        // @ts-ignore
-        return vault!.authenticateV2Request(res.params, requestContexts, cb);
+    res.params!.log = log;
```
francoisferrand: do you know why Typescript is not able to infer the type here?
As far as I understand, `extractParams()` return either AuthV2RequestParams, AuthV4RequestParams or AuthInfo. Since the AuthInfo case is handle, TS should be able to deduce this will be either AuthV2RequestParams or AuthV4RequestParams...

Maybe this is due to the "optional" params (`params?`) in result of extractParams?
And the return type should be better type to either return an error without params, or params with null error: so that type deduction can work its magic, and we don't need type coercion...

lib/auth/auth.ts
```diff
@@ -61,7 +61,7 @@ function extractParams(
     log: Logger,
     awsService: string,
     data: { [key: string]: string }
-) {
+): { err: null | Error | ArsenalError, params?: AuthV2RequestParams | AuthV4RequestParams | AuthInfo } {
```
francoisferrand: can we really return an `Error` ?

lib/auth/v2/headerAuthCheck.ts
```diff
@@ -1,11 +1,16 @@
...
 import checkRequestExpiry from './checkRequestExpiry';
 import algoCheck from './algoCheck';
+import { AuthV2RequestParams } from '../Vault';
 
-export function check(request: any, log: Logger, data: { [key: string]: string }) {
+export function check(
+    request: any,
+    log: Logger,
+    data: { [key: string]: string },
+): { err: null | ArsenalError | Error, params?: AuthV2RequestParams } {
```
francoisferrand: can we really return an `Error`, are not just missing some typing below?

lib/auth/v2/queryAuthCheck.ts
```diff
@@ -1,14 +1,19 @@
...
 export const PRE_SIGN_URL_EXPIRY = process.env.PRE_SIGN_URL_EXPIRY ?
     Number.parseInt(process.env.PRE_SIGN_URL_EXPIRY, 10) :
     constants.defaultPreSignedURLExpiry * 1000;
 
-export function check(request: any, log: Logger, data: { [key: string]: string }) {
+export function check(
+    request: any,
+    log: Logger,
+    data: { [key: string]: string },
+): { err: null | ArsenalError | Error, params?: AuthV2RequestParams } {
```
francoisferrand: can we really return an Error, are not just missing some typing below?

lib/auth/v4/constructStringToSign.ts
```diff
@@ -40,14 +40,6 @@ export default function constructStringToSign(params: {
         service: params.awsService,
     });
 
-    // TODO Why that line?
```
francoisferrand, line 43: why indeed?
and why can we now remove it?

lib/auth/v4/headerAuthCheck.ts
```diff
@@ -26,7 +27,7 @@ export function check(
     log: Logger,
     data: { [key: string]: string },
     awsService: string
-) {
+): { err: null | ArsenalError | Error, params?: AuthV4RequestParams } {
```
francoisferrand: It seems this cannot return an `Error`.

The case would be a failure to utf8.encode:
```
    const stringToSign = constructStringToSign({ ... });
    if (stringToSign instanceof Error) {
        return { err: stringToSign };
    }
```

But according to the docs this function does not return an error, it throws on error:

> Encodes any given JavaScript string (string) as UTF-8, and returns the UTF-8-encoded version of the string. It throws an error if the input string contains a non-scalar value, i.e. a lone surrogate. (If you need to be able to encode non-scalar values as well, use WTF-8 instead.)

lib/auth/v4/constructStringToSign.ts
ghost: I removed it because after reviewing the function and checking all the types, the function cannot return an error, so this code cannot be reached.

lib/auth/v4/queryAuthCheck.ts
```diff
@@ -1,18 +1,23 @@
...
  * @param request - HTTP request object
  * @param log - logging object
  * @param data - Contain authentification params (GET or POST data)
  */
-export function check(request: any, log: Logger, data: { [key: string]: string }) {
+export function check(
+    request: any,
+    log: Logger,
+    data: { [key: string]: string },
+): { err: null | ArsenalError | Error, params?: AuthV4RequestParams } {
```
francoisferrand: same

lib/models/BucketPolicy.ts
```diff
@@ -48,64 +48,77 @@ const objectActions = [
...
+        Resource: string | string[];
+        Condition?: { [key: string]: string | string[] };
+    }[],
+};
+
 export default class BucketPolicy {
-    _json: string;
-    _policy: any;
+    private json: string;
+    private policy: BucketPolicyMetadata | null = null;
```
francoisferrand: would this not be more appropriate (and consistent with the changes above)?
```suggestion
    private policy?: BucketPolicyMetadata;
```

lib/auth/auth.ts
ghost: Through the `checkFunctions`, yes: the `validateCredentials` function in the header+query v4 check can return such error. V2 isn't returning `Error`s indeed.

lib/models/WebsiteConfiguration.ts
```diff
@@ -77,10 +77,10 @@ export type RedirectAllRequestsTo = {
...
 
 export type WebsiteConfigurationParams = {
-    indexDocument: string;
-    errorDocument: string;
-    redirectAllRequestsTo: RedirectAllRequestsTo;
-    routingRules: RoutingRule[] | any[],
+    indexDocument?: string;
+    errorDocument?: string;
+    redirectAllRequestsTo?: RedirectAllRequestsTo;
+    routingRules?: RoutingRule[] | any[],
```
francoisferrand: does not seem really typed....

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -407,7 +407,11 @@ class MongoClientInterface {
         const payload = {
             $set: {
                 _id: bucketName,
-                value: newBucketMD,
+                value: {
+                    ...newBucketMD,
+                    quotaMax: new Long(newBucketMD.quotaMax || '0'),
```
francoisferrand, line 407: should have an associated test, to ensure we can indeed properly create buckets with very large quotas

lib/auth/auth.ts
francoisferrand: `validateCredentials()` function is typed to return an `Error`, but implementation looks like it is always returning an `ArsenalError`

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: Creating a bucket with a quota is not possible today, this is only to enforce the right type in the DB, as it fails the type validation. Having large quotas is already tested in the proper routes, and this `Long` conversion is (indirectly) tested when we create buckets, when we test that we can properly pass a `BucketInfo` object.

Do we still want such test?

lib/auth/auth.ts
ghost: I updated the code (and all other related comments).

lib/models/WebsiteConfiguration.ts
ghost: I updated it, due to a wrong type being used in BucketInfo (class type instead of the "Params" version)

lib/auth/auth.ts
ghost: I updated the PR, I created a new type that allows TS to understand that params is defined if `err` is `null`.

ghost:
I added a commit to fix the types of the loggers and fix a conflict with Vault.ts.

lib/models/WebsiteConfiguration.ts
ghost: Actually, the whole website implementation is a mess: js code uses private members of the website class and we eneded up storing that in mongodb, so we need to keep this behavior... Working on it

ghost: Done...

lib/auth/v2/headerAuthCheck.ts
```diff
@@ -1,11 +1,17 @@
...
 import checkRequestExpiry from './checkRequestExpiry';
 import algoCheck from './algoCheck';
+import { AuthV2RequestParams } from '../Vault';
+import { AuthResult } from '../auth';
 
-export function check(request: any, log: Logger, data: { [key: string]: string }) {
+export function check(
+    request: any,
+    log: RequestLogger,
+    data: { [key: string]: string },
```
BourgoisMickael: Don't you prefer:

```suggestion
    data: Record<string, string>,
```

lib/auth/auth.ts
```diff
@@ -1,6 +1,6 @@
 import * as crypto from 'crypto';
-import { Logger } from 'werelogs';
-import errors from '../errors';
+import { RequestLogger } from 'werelogs';
```
BourgoisMickael:
```suggestion
import type { RequestLogger } from 'werelogs';
```

For type only import this can be interesting as the import is discarded after transpiling

lib/models/BucketAzureInfo.ts
```diff
@@ -3,32 +3,34 @@ export type DeleteRetentionPolicy = {
...
-        networkACL: any[];
-        cname: string;
-        azureFilesAADIntegration: boolean;
-        hnsEnabled: boolean;
-        logging: any;
-        hourMetrics: any;
-        minuteMetrics: any;
-        serviceVersion: string;
-    };
+    private data: AzureInfoMetadata;
```
BourgoisMickael:
```suggestion
    private readonly data: AzureInfoMetadata;
```

And this type can be used in the constructor as well ?

lib/models/BucketPolicy.ts
```diff
@@ -48,64 +48,77 @@ const objectActions = [
     's3:PutObjectTagging',
 ];
 
+export type BucketPolicyMetadata = {
+    Id: string;
+    Version: string;
+    Statement: {
+        Sid: string;
+        Effect: string;
+        Principal: string | { AWS: string[] };
```
BourgoisMickael:
```suggestion
        Principal: string | { AWS: string | string[] };
```

This can be a string as well like `Principal: { AWS: '*' }`

lib/models/BucketPolicy.ts
```diff
@@ -48,64 +48,77 @@ const objectActions = [
     's3:PutObjectTagging',
 ];
 
+export type BucketPolicyMetadata = {
+    Id: string;
+    Version: string;
+    Statement: {
+        Sid: string;
```
BourgoisMickael: `Id` and `Sid` should be optional

lib/models/BucketPolicy.ts
```diff
@@ -48,64 +48,77 @@ const objectActions = [
...
+        Effect: string;
```
BourgoisMickael:
```suggestion
        Effect: 'Allow' | 'Deny';
```

lib/models/BucketPolicy.ts
```diff
@@ -48,64 +48,77 @@ const objectActions = [
...
+        Principal: string | { AWS: string[] };
+        Action: string | string[];
+        Resource: string | string[];
+        Condition?: { [key: string]: string | string[] };
```
BourgoisMickael: There is a missing intermediate object and value can be number
```suggestion
        Condition?: Record<string, Record<string, string | string[] | number>>;
```

Example 

```json

{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Deny",
            "Principal": "*",
            "Action": "s3:PutObject",
            "Resource": [
                "arn:aws:s3:::amzn-s3-demo-bucket1",
                "arn:aws:s3:::amzn-s3-demo-bucket1/*"
            ],
            "Condition": {
                "NumericLessThan": {
                    "s3:TlsVersion": 1.2
                }
            }
        }
    ]
}
```

lib/models/WebsiteConfiguration.ts
```diff
@@ -22,7 +22,7 @@ export type Condition = {
     httpErrorCodeReturnedEquals?: string;
 };
 
-export type RoutingRuleParams = { redirect: Redirect; condition?: Condition };
+export type RoutingRuleParams = { redirect?: Redirect; condition?: Condition };
```
BourgoisMickael: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-sdk2-types/clients/s3.d.ts#L6279

I don't see the redirect being optional

ghost:
Note for reviewers: all comments adressed & the PR was updated after #2288 merge, adding more types for request contexts and auth options.

lib/policyEvaluator/RequestContext.ts
```diff
@@ -154,7 +154,31 @@ export type RequesterInfo = {
     username: string;
     keycloakGroup: string;
     keycloakRole: string;
-}
+};
+
+export type RequestContextType = {
+    headers: { [key: string]: string | string[] };
```
BourgoisMickael:
```suggestion
    headers: Record<string, string | string[]>;
```

BourgoisMickael (approved):

lib/policyEvaluator/RequestContext.ts
ghost: thanks! I added https://github.com/scality/Arsenal/pull/2299/commits/5542f1ed0649c2509aa6c0756445a3923003064c to do it everywhere in Arsenal.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
francoisferrand: > it fails the type validation

what type validation? payload is untyped (for typescript), right?

> Do we still want such test?

from MongoClientInterface standpoint, it is possible to set a quota (i.e. it uses the `newBucketMD.quotaMax`) which is provided: so i think it would be better indeed to add one (or more: no quota/small quota/large bigint) test case in MongoClientInterface tests

lib/models/ObjectLockConfiguration.ts
```diff
@@ -89,23 +100,18 @@ export default class ObjectLockConfiguration {
...
+    private parseTime(dr: DefaultRetention): ParsedRetention {
+        if (dr.Days && dr.Years) {
             const msg = 'request xml contains both Days and Years';
             const error = errors.MalformedXML.customizeDescription(msg);
             return { error };
         }
-        const timeType = 'Days' in dr ? 'Days' : 'Years';
+        const timeType = dr.Days ? 'Days' : 'Years';
         if (!dr[timeType] || !dr[timeType][0]) {
             const msg = 'request xml does not contain Days or Years';
```
francoisferrand: looking at this code, we should push the cleanup a bit further I think: the variable we actually need is `dr[timeType]`, and we can avoid accessing field by name (and later type assertion)

```
if (dr.Days && dr.Years) {
    return 'request xml contains both Days and Years';
}
if (!dr.Days && !dr.Years) {
    return 'request xml does not contain Days or Years';
}
const {timeRaw, timeMax, timeType} = func() {
    if (dr.Days) {
        return {dr.Days, 36500, 'days'};
    } else {
        return {dr.Years, 100, 'years'};
    }
}(); 
const timeValue = Number.parseInt(timeRaw[0], 10);
```

lib/models/ObjectLockConfiguration.ts
```diff
@@ -89,23 +100,18 @@ export default class ObjectLockConfiguration {
...
             const error = errors.MalformedXML.customizeDescription(msg);
             return { error };
         }
-        if (dr[timeType].length > 1) {
```
francoisferrand, line 104: why remove this check?

it is consistent with the array indexing we do (`dr[timeType][0]`), though I don't understand this indexing with regard to the typing: it seems to me either the type is not correct (we can an array of string, not a string), or the indexing is just taking the first digit?

lib/models/BucketInfo.ts
```diff
@@ -36,7 +37,7 @@ export type SSE = {
 
 export type VersioningConfiguration = {
     Status: string;
```
francoisferrand: should be typed as well (can be Enabled, Disabled or Suspended I think) ?

lib/auth/auth.ts
```diff
@@ -1,6 +1,6 @@
 import * as crypto from 'crypto';
-import { Logger } from 'werelogs';
-import errors from '../errors';
+import type { RequestLogger } from 'werelogs';
+import errors, { ArsenalError } from '../errors';
```
francoisferrand: only need type of `ArsenalError` ?

lib/auth/auth.ts
```diff
@@ -118,39 +131,38 @@ function extractParams(
...
+            if (data.signatureVersion) {
                 requestContext.setSignatureVersion(data.signatureVersion);
+            }
+            if (data.securityToken) {
                 requestContext.setSecurityToken(data.securityToken);
-                if ('signatureAge' in data) {
-                    requestContext.setSignatureAge(data.signatureAge);
-                }
             }
+            requestContext.setSignatureAge(data.signatureAge || 0);
```
francoisferrand, line 160: - this will crash if `res.params.data` is not set, which was not the case before
- we used not to call `setSignatureAge` when `data.signatureAge` was not set

```suggestion
            if (data?.authType) {
                requestContext.setAuthType(data.authType);
            }
            if (data?.signatureVersion) {
                requestContext.setSignatureVersion(data.signatureVersion);
            }
            if (data?.securityToken) {
                requestContext.setSecurityToken(data.securityToken);
            }
            if (data?.signatureAge) {
                requestContext.setSignatureAge(data.signatureAge);
            }
```
or
```suggestion
    const data = res.params.data;
    if (data && requestContexts) {
        requestContexts.forEach(requestContext => {
            if (data.authType) {
                requestContext.setAuthType(data.authType);
            }
            if (data.signatureVersion) {
                requestContext.setSignatureVersion(data.signatureVersion);
            }
            if (data.securityToken) {
                requestContext.setSecurityToken(data.securityToken);
            }
            if (data.signatureAge) {
                requestContext.setSignatureAge(data.signatureAge);
            }
        }
    }
```

lib/auth/Vault.ts
```diff
@@ -63,24 +63,38 @@ export function vaultSignatureCb(
         accountQuota: info.accountQuota || {},
     });
 }
+export type AuthV2RequestParams = {
+    version: 2;
+    log: RequestLogger;
+    data: {
+        accessKey: string;
+        algo: string;
```
francoisferrand: can we type `algo`, with the possible values?

lib/auth/Vault.ts
```diff
@@ -63,24 +63,38 @@ export function vaultSignatureCb(
...
+    };
+};
 
 export type AuthV4RequestParams = {
     version: 4;
     log: RequestLogger;
     data: {
         accessKey: string;
-        signatureFromRequest: string;
+        algo?: string;
```
francoisferrand: can we type algo, with the possible values?

lib/auth/Vault.ts
```diff
@@ -63,24 +63,38 @@ export function vaultSignatureCb(
...
+        authType?: 'query' | 'header' | 'REST-HEADER' | 'REST-QUERY-STRING';
+        credentialScope?: string;
```
francoisferrand, line 90: can we type `credentialScope`, with the possible values?

lib/auth/v4/validateInputs.ts
```diff
@@ -1,5 +1,5 @@
-import { Logger } from 'werelogs';
-import errors from '../../../lib/errors';
+import type { RequestLogger } from 'werelogs';
+import errors, { ArsenalError } from '../../../lib/errors';
```
francoisferrand, line 2: only type needed for ArsenalError ?

lib/models/ObjectMDAzureInfo.ts
```diff
@@ -1,22 +1,24 @@
+export type ObjectMDAzureInfoMetadata = {
```
francoisferrand, line 1: can we type these fields (containerPublicAccess, containerStoredAccessPolicies, containerImmutabilityPolicy, blobType, blobCopyInfo) ?
should we? (not sure they are still used, it may be a remnant of BlobServer...)

francoisferrand (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: > what type validation? payload is untyped (for typescript), right?

We enforce types in the mongodb driver, o we cannot actually create a bucket MD if the quota is not a Long (which is expected). I'll add a test

lib/models/ObjectLockConfiguration.ts
ghost: This variable is a number and not a string. I will review this, there is something off.

lib/auth/Vault.ts
ghost: We cannot, this is dynamically extracted from the string to sign.

lib/models/ObjectLockConfiguration.ts
ghost: So it's indeed a string array, and it's due to how the xml parser works. The old code is correct, but the type wasn't. I'll update and add your suggestions.

lib/models/ObjectMDAzureInfo.ts
ghost: I think we can but as you said this is not used today in cloudserver. I also see several fields with values I cannot easily type (no information, no usage beyond this class). Thinking from a higher level, I think a debt we have here is to potentially remove useless fields from metadata objects (and potentially don't write fully the fields like `'x-amz-server-side-encryption-customer-algorithm'` but use a dict so we greatly reduce the size of MD, in turn reducing storage, RAM consumption, network usage, impact of oplog listening, etc).

I created OS-932.

PR #2300, opened 2025-02-26
Improve logging

ghost:
- Change the "received request" log level
- Better define the default logging fields
- Remove useless logs
- Log in case of error in the mongoclientinterface, as InternalError hides all the underlying information
- Pass the right error in case of error during deletion, to actually fix a bug in case of mongodb erroor leading to no error log AND deletion of data from the storage layer, but not in mongodb
- In the dataWrapper some logs were in trace, meaning no info available in case of error when handling data, and harder debugging
- Adding multiple tests missing today, affecting code coverage due to the logging updates
- Updated the in memory data store, as we need the `head` method to properly unit test backbeat routes.

Issue: ARSN-471

### DIFF
--- a/lib/auth/Vault.ts
+++ b/lib/auth/Vault.ts
@@ -1,4 +1,4 @@
-import { Logger } from 'werelogs';
+import { Logger, RequestLogger } from 'werelogs';
 import errors from '../errors';
 import AuthInfo, { AccountInfos, AuthInfoType, AuthorizationResults, AuthV4Results } from './AuthInfo';
 
@@ -18,7 +18,7 @@ function vaultSignatureCb(
             body: AuthV4Results,
         },
     },
-    log: Logger,
+    log: RequestLogger,
     callback: (
         err: Error | null,
         data?: AuthInfoType,
@@ -41,13 +41,10 @@ function vaultSignatureCb(
     const { email, ...userInfoWithoutEmail } = authInfo.message.body.userInfo;
 
     log.debug('received info from Vault', {
-        ...authInfo,
-        message: {
-            message: authInfo.message.message,
-            body: {
-                ...authInfo.message.body,
-                userInfo: userInfoWithoutEmail,
-            },
+        message: authInfo.message.message,
+        body: {
+            ...authInfo.message.body,
+            userInfo: userInfoWithoutEmail,
         },
     });
 
@@ -69,7 +66,7 @@ function vaultSignatureCb(
 
 export type AuthV4RequestParams = {
     version: 4;
-    log: Logger;
+    log: RequestLogger;
     data: {
         accessKey: string;
         signatureFromRequest: string;
@@ -83,7 +80,7 @@ export type AuthV4RequestParams = {
         credentialScope: string;
         securityToken: string;
         algo: string;
-        log: Logger;
+        log: RequestLogger;
     };
 };
 
@@ -131,7 +128,7 @@ export default class Vault {
     authenticateV2Request(
         params: {
             version: 2;
-            log: Logger;
+            log: RequestLogger;
             data: {
                 securityToken: string;
                 accessKey: string;
@@ -141,7 +138,7 @@ export default class Vault {
                 authType: 'query' | 'header';
                 signatureVersion: string;
                 signatureAge?: number;
-                log: Logger;
+                log: RequestLogger;
             };
         },
         requestContexts: any[],
@@ -238,14 +235,13 @@ export default class Vault {
     */
     getCanonicalIds(
         emailAddresses: string[],
-        log: Logger,
+        log: RequestLogger,
         callback: (
             err: Error | null,
             data?: { canonicalID: string; email: string }[]
         ) => void
     ) {
-        log.trace('getting canonicalIDs from Vault based on emailAddresses',
-            { emailAddresses });
+        log.trace('getting canonicalIDs from Vault based on emailAddresses');
         this.client.getCanonicalIds(emailAddresses,
             // @ts-ignore
             { reqUid: log.getSerializedUids() },
@@ -282,7 +278,7 @@ export default class Vault {
     */
     getEmailAddresses(
         canonicalIDs: string[],
-        log: Logger,
+        log: RequestLogger,
         callback: (err: Error | null, data?: { [key: string]: any }) => void
     ) {
         log.trace('getting emailAddresses from Vault based on canonicalIDs',
@@ -320,13 +316,12 @@ export default class Vault {
     */
     getAccountIds(
         canonicalIDs: string[],
-        log: Logger,
+        log: RequestLogger,
         callback: (err: Error | null, data?: { [key: string]: string }) => void
     ) {
         log.trace('getting accountIds from Vault based on canonicalIDs',
             { canonicalIDs });
         this.client.getAccountIds(canonicalIDs,
-            // @ts-expect-error
             { reqUid: log.getSerializedUids() },
             (err: Error | null, info?: any) => {
                 if (err) {
@@ -365,7 +360,7 @@ export default class Vault {
     checkPolicies(
         requestContextParams: any[],
         userArn: string,
-        log: Logger,
+        log: RequestLogger,
         callback: (err: Error | null, data?: any[]) => void
     ) {
         log.trace('sending request context params to vault to evaluate' +
@@ -384,7 +379,7 @@ export default class Vault {
         });
     }
 
-    checkHealth(log: Logger, callback: (err: Error | null, data?: any) => void) {
+    checkHealth(log: RequestLogger, callback: (err: Error | null, data?: any) => void) {
         if (!this.client.healthcheck) {
             const defResp = {};
             defResp[this.implName] = { code: 200, message: 'OK' };
@@ -411,7 +406,7 @@ export default class Vault {
         });
     }
 
-    report(log: Logger, callback: (err: Error | null, data?: any) => void) {
+    report(log: RequestLogger, callback: (err: Error | null, data?: any) => void) {
         // call the report function of the client
         if (!this.client.report) {
             return callback(null, {});
@@ -445,7 +440,7 @@ export default class Vault {
     */
     getOrCreateEncryptionKeyId(
         canonicalID: string,
-        log: Logger,
+        log: RequestLogger,
         callback: (err: Error | null, data?: { 
             canonicalId: string, 
             encryptionKeyId: string, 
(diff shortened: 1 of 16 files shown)


tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
```diff
@@ -846,6 +846,80 @@ describe('MongoClientInterface, tests', () => {
...
+        afterEach(async () => {
+            await new Promise(resolve =>
+                client.deleteBucket(bucketName, logger, err => {
+                    if (err) {
+                        logger.error('Failed to delete bucket in cleanup', { error: err });
+                    }
+                    resolve();
+                }),
+            );
+        });
```
welansari: should use callbacks like we do in the beforeEach or make both async ?

tests/unit/s3middleware/prepareStream.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    const stream = {
+        headers: {},
+        pipe: sinon.stub(),
+    };
+
+    beforeEach(() => {
+        stream.headers['x-amz-content-sha256'] = 'STREAMING-AWS4-HMAC-SHA256-PAYLOAD';
+    });
+
+    it('should return the stream is sha256 header is not streaming', () => {
```
welansari:
```suggestion
    it('should return the stream if sha256 header is not streaming', () => {
```

tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
ghost: I kept cb indeed.

Note: I also added a commit for the error handling in mongoclientinterface, it was in the same commit as logging, while its is a logic change...

welansari (approved):

lib/auth/Vault.ts
```diff
@@ -1,4 +1,4 @@
-import { Logger } from 'werelogs';
+import { Logger, RequestLogger } from 'werelogs';
```
francoisferrand: `Logger` not needed anymore

lib/auth/Vault.ts
```diff
@@ -445,7 +440,7 @@ export default class Vault {
     */
     getOrCreateEncryptionKeyId(
         canonicalID: string,
-        log: Logger,
+        log: RequestLogger,
```
francoisferrand, line 443: should update the comment as well (line 429: `@param {Logger} log - logger`)

lib/auth/v4/streamingV4/V4Transform.ts
```diff
@@ -1,7 +1,7 @@
 import { Transform } from 'stream';
 import async from 'async';
 import errors from '../../../errors';
-import { Logger } from 'werelogs';
+import { Logger, RequestLogger } from 'werelogs';
```
francoisferrand: `Logger` not used anymore

lib/s3routes/routes.ts
```diff
@@ -221,17 +235,18 @@ export default function routes(
 
     if (!req.url!.startsWith('/_/healthcheck') &&
         !req.url!.startsWith('/_/report')) {
-        log.info('received request', clientInfo);
+        log.debug('received request', clientInfo);
     }
 
-    log.end().addDefaultFields(clientInfo);
+    log.addDefaultFields(clientInfo);
```
francoisferrand, line 241: removing `end()` means we may not see any log until the request is complete: which may make it harder to troubleshoot requests which somehow do not terminate. Is this really what we want?

lib/s3routes/routesUtils.ts
```diff
@@ -195,12 +193,10 @@ export const JSONResponseBackend = {
         additionalHeaders?: { [key: string]: string } | null,
     ) {
         const bytesSent = Buffer.byteLength(json);
-        log.trace('sending success json response');
         log.addDefaultFields({ bytesSent });
         setCommonResponseHeaders(additionalHeaders, response, log);
         response.writeHead(200, { 'Content-type': 'application/json' });
-        log.debug('response http code', { httpCode: 200 });
-        log.trace('json response', { json });
+        log.trace('sending success json response', { json });
```
francoisferrand, line 199: * should we not keep the `httpCode` field? (may be consistent with other error logs)
* my understanding of the "duplicated" traces is that we want(ed) a debug log in anycase, but include JSON only when log level is trace : should we try to keep such behavior (i.e. changing the behavior based on `logger.config.level()`), or do we really _want_ to remove the debug log?

tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
```diff
@@ -846,6 +846,78 @@ describe('MongoClientInterface, tests', () => {
...
+            const versionId = 'test-version-id';
+            const objMD = new ObjectMD()
+                .setKey(objName)
+                .setDataStoreName('us-east-1')
+                .setContentLength(100)
+                .setLastModified(new Date());
+
+            // Mock findOne to throw an error
+            const originalFindOne = collection.findOne;
+            collection.findOne = () => Promise.reject(new Error('Simulated MongoDB error'));
```
francoisferrand, line 885: working on mongodb driver update, @benzekrimaha had issues when Promises where used to catch errors in mongodb driver, which AFAIU could only be solved by actually using try/catch block instead... and which could not be seen with such mocking of mongodb, but only with real mongodb integration tests.

not really related to this PR, but may be best to confirm ASAP that we are safe on that aspect (i.e. make sure we handle errors properly, and that we are not adding false confidence by adding tests where the mock does not behave like the real driver)

ghost: The mongodb driver will reject the promise if there is an error. See https://www.mongodb.com/docs/drivers/node/current/fundamentals/promises/. As far as the test structure is concerned, we can simulate such rejection like that, and it will enter the right `.catch` methods. I do not see why try/catch would be required, as it does the same thing.

Try/catch however has a drawback: when we enter the "try" and success, then we might call another function and continue. If in this case we have an exception, the `catch` in the mongo client interface will catch the error, leading to another callback call, which is usually the root cause of "callback already called" errors, plus it complexifies a lot the debugging, as we have no trace and no info about the real root cause.

@benzekrimaha, can you confirm if there is something that forces us to test with try/catch and not promises in tests?

francoisferrand:
as a side note / complementary effort : we are changing some log levels here, but we should remember the logs (esp. debug/trace) are used in 2 contexts:

* Either when the user activate more verbose logging
  → display all the time, and indeed we may need to streamline these
* In case of error, every log related to the request is dumped again
  → in that specific context, having more (and separate lines) logs may/could be acceptable;
  → yet there may be an additional improvement in werelogs, to ensure these logs (some of which may already have been logged!) are clearly isolated: maybe by including them in the "final" json line or adding some sort of prefix/..., instead of just printing them like regular logs

That is probably a separate improvement, but sharing as I think it may bring a different point of view on the updating of logs.

lib/s3routes/routesUtils.ts
ghost: We have the debug log when the response is complete just below, with the real status code (while the raw 200 is incorrect some times, could be 2xx). Meaning that, in case of debug log level, we end up with both the 200 and the real code sent to the client, which is not useful. This is removed on purpose, this brings 0 benefit as a log.

Then we keep a trace as before with the json info.

If we have an error log we should not enter the okResponse anyway (or it's a multi delete).

lib/s3routes/routes.ts
ghost: The `end` function in werelogs should only be used to end the RequestLogger. It returns an `EndLogger` that should only be used to log one last time. It also allows to have the "elapsed time" info. But as long as we log using the current logging level, the logs will be flushed in real time. When the dump level is reached, we might flush those logs again, but with more info (like, debug logs).

I am not sure to understand what you mean by "we may not see any log until the request is complete"?

ghost:
> → yet there may be an additional improvement in werelogs, to ensure these logs (some of which may already have been logged!) are clearly isolated: maybe by including them in the "final" json line or adding some sort of prefix/..., instead of just printing them like regular logs

Indeed today we might log twice the same logs, in particular those with logging level >= to the log level, if a dump occurs.

Note: there is also a performance impact of having logs under the levels we listen to, because we are still processing them "in case of error", and that might lead to a lot of object copy/parsing, even if not useful.

lib/s3routes/routes.ts
francoisferrand: > as long as we log using the current logging level, the logs will be flushed in real time

this answers the "we may not see any log until the request is complete" issue, so no problem actually :-)

tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
benzekrimaha: When working on the bump of the driver on vault I indeed was able to see that without a try a catch that awaits for the operation , it was not working as expected as the operations were failing without being fulfilled.

ghost: > the operations were failing without being fulfilled.

Do you mean they are failing but we never get the error? Because if they fail, that's expected that the operations are not fulfilled, no?

benzekrimaha: it was failing with no error thrown as it was not waiting for the operation , I only had the operations really failing / being processed when using a try/catch and an await on the operations

ghost: Ok, I will try to review if we are missing something here, as this is not consistent with MongoDB documentation.

This is not blocking for this PR however.

ghost: After reviewing Vault code, the issue is more on the lack of synchronous error handling: when it happens, we don't call mongodb indeed, because we are not ensured that the db is available before we call mongodb methods. However, this is only in Vault. Independently to how we implement, the mongodb driver will always call its promises properly.

Here, the tests are mocking a mongodb error, which is indeed not a synchronous error, but still tests that we properly handle mongodb errors. This is also confirmed by all CIs: if it was not working, nothing would pass. Arsenal already has functional tests using mongodb in memory, and we test error scenarios already, e.g. in `tests/functional/metadata/mongodb/getObject.spec.js`.

The problem in vault is more a migration issue, where we switch from all-callbacks to promises, but the architecture is different and we are likely missing `catch` for some types of errors, or misconfigured how we use the promises.

I am closing this comment: in Arsenal, the migration was already done, and everything works as expected.

francoisferrand (approved):

ghost:
Removing the last commit as we need to merge #2286  before releasing.

PR #2301, opened 2025-03-03
ARSN-473 bump node fcntl with node 22 support

tcarmet:
- **ARSN-473 bump node-fcntl with node 22 support**
- **ARSN-473 bump package version**

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "bson": "^6.8.0",
     "debug": "^4.3.7",
     "diskusage": "^1.2.0",
-    "fcntl": "github:scality/node-fcntl#0.2.2",
+    "fcntl": "github:scality/node-fcntl#0.3.0",
     "httpagent": "scality/httpagent#1.1.0",
     "https-proxy-agent": "^7.0.5",
     "ioredis": "^5.4.1",
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=20"
   },
-  "version": "8.2.3",
+  "version": "8.2.4",
   "description": "Common utilities for the S3 project components",
   "main": "build/index.js",
   "repository": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -1651,11 +1651,6 @@
   dependencies:
     levn "^0.4.1"
 
-"@gar/promisify@^1.0.1":
-  version "1.1.3"
-  resolved "https://registry.yarnpkg.com/@gar/promisify/-/promisify-1.1.3.tgz#555193ab2e3bb3b6adc3d551c9c030d9e860daf6"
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -45,6 +45,7 @@
     "level-sublevel": "~6.6.5",
     "mongodb": "^6.11.0",
     "node-forge": "^1.3.1",
+    "path-scurry": "^2.0.0",
```
BourgoisMickael: what's this ?

BourgoisMickael (approved):

package.json
tcarmet: some dependency needed to build `node-fcntl`, otherwise it would fail:
```
gyp ERR! configure error 
gyp ERR! stack Error: Cannot find module 'path-scurry'
gyp ERR! stack Require stack:
gyp ERR! stack - /home/runner/work/Arsenal/Arsenal/node_modules/cacache/node_modules/glob/dist/commonjs/glob.js
gyp ERR! stack - /home/runner/work/Arsenal/Arsenal/node_modules/cacache/node_modules/glob/dist/commonjs/index.js
gyp ERR! stack - /home/runner/work/Arsenal/Arsenal/node_modules/cacache/lib/util/glob.js
```

jonathan-gramain (approved):
I'm good with the bump itself, but wondering why we don't bump in `development/7.70` since this is the dependency that Metadata will use, do we plan on depending on arsenal 8.x in Metadata too? The same question applies for other components that depend on arsenal 7.70.x and will use node 22.

tcarmet:
> I'm good with the bump itself, but wondering why we don't bump in `development/7.70` since this is the dependency that Metadata will use, do we plan on depending on arsenal 8.x in Metadata too? The same question applies for other components that depend on arsenal 7.70.x and will use node 22.

@jonathan-gramain I'm actually using arsenal 8.2 in the node upgrade for Metadata, and got a green CI with it. I'm avoiding touching development/7.70 in my path (while I can), see how far this takes me, cause if we manage to "unify" arsenal as well, that would be nice for us I think.

package.json
BourgoisMickael: oh maybe it should be in node-fcntl then

BourgoisMickael:
And metadata has a branch dev/8 that we are going to use for 9.5

package.json
tcarmet: let me make a PR over there then.

tcarmet: https://github.com/scality/node-fcntl/pull/16

BourgoisMickael: some problem happened with the lockfile, it's old can you rather

```
yarn remove fcntl
yarn add github:scality/node-fcntl#0.3.0
```

it should fix the issue without the need to add the dependency yourself

tcarmet: indeed, thanks!

ghost (approved):

BourgoisMickael (approved):

PR #2303, opened 2025-03-06
When data is sent to client, send http code after storage call

ghost:
- In case of error, we can then call the error.
- We do not need to destroy the response before we started to send data to the client: we just need to destroy the stream, if set, and return the right error to the client.
- No change when we started to receive bytes: we destroy the stream.
- When there is no error, we send the expected http header and potential headers before starting to send data.
- We centralize the logic to end the request and log.
- To better test, the http response mock is improved so we can really test streams.

Issue: ARSN-467

### DIFF
--- a/lib/s3routes/routesUtils.ts
+++ b/lib/s3routes/routesUtils.ts
@@ -40,8 +40,10 @@ export function setCommonResponseHeaders(
                     response.setHeader(key, headers[key]);
                 } catch (e: any) {
                     log.debug('header can not be added ' +
-                      'to the response', { header: headers[key],
-                        error: e.stack, method: 'setCommonResponseHeaders' });
+                        'to the response', {
+                            header: headers[key],
+                        error: e.stack, method: 'setCommonResponseHeaders'
+                    });
                 }
             }
         });
@@ -145,8 +147,8 @@ export const XMLResponseBackend = {
         xml.push(
             '<?xml version="1.0" encoding="UTF-8"?>',
             '<Error>',
-                `<Code>${error.message}</Code>`,
-                `<Message>${error.description}</Message>`,
+            `<Code>${error.message}</Code>`,
+            `<Message>${error.description}</Message>`,
         );
         const invalidArguments = error.metadata.get('invalidArguments') || [];
         invalidArguments.forEach((invalidArgument, index) => {
@@ -156,8 +158,8 @@ export const XMLResponseBackend = {
             xml.push(`<ArgumentValue${counter}>${ArgumentValue}</ArgumentValue${counter}>`);
         });
         xml.push(
-                '<Resource></Resource>',
-                `<RequestId>${log.getSerializedUids()}</RequestId>`,
+            '<Resource></Resource>',
+            `<RequestId>${log.getSerializedUids()}</RequestId>`,
             '</Error>',
         );
         const xmlStr = xml.join('');
@@ -165,7 +167,7 @@ export const XMLResponseBackend = {
         log.addDefaultFields({ bytesSent });
         response.writeHead(error.code, {
             'Content-Type': 'application/xml',
-            'Content-Length': bytesSent ,
+            'Content-Length': bytesSent,
         });
         return response.end(xmlStr, 'utf8', () => {
             log.end().info('responded with error XML', {
@@ -310,16 +312,13 @@ function okContentHeadersResponse(
     }
     if (overrideParams['response-content-disposition']) {
         addHeaders['Content-Disposition'] =
-        overrideParams['response-content-disposition'];
+            overrideParams['response-content-disposition'];
     }
     if (overrideParams['response-content-encoding']) {
         addHeaders['Content-Encoding'] =
             overrideParams['response-content-encoding'];
     }
     setCommonResponseHeaders(addHeaders, response, log);
-    const httpCode = range ? 206 : 200;
-    log.debug('response http code', { httpCode });
-    response.writeHead(httpCode);
     return response;
 }
 
@@ -365,6 +364,12 @@ function retrieveData(
     locations: any[],
     retrieveDataParams: any,
     response: http.ServerResponse,
+    // When data is streamed, we can have errors
+    // when retrieving the data from the storage.
+    // As such, we must send the header as soon as
+    // we confirmed data backend availability for
+    // the requested key.
+    httpCodeOnSucess: number,
     log: RequestLogger,
 ) {
     if (locations.length === 0) {
@@ -408,15 +413,18 @@ function retrieveData(
         client, implName, config, kms, metadata, locStorageCheckFn, vault);
     return eachSeries(locations,
         (current, next) => data.get(current, response, log,
-            (err: any, readable: http.IncomingMessage) => {
+            (err: Error, readable: http.IncomingMessage) => {
                 const cbOnce = jsutil.once(next);
                 // NB: readable is of IncomingMessage type
                 if (err) {
+                    // In this case, we do not destroy the response:
+                    // nothing was sent to the client yet, so we
+                    // can return a proper error.
                     log.error('failed to get object', {
                         error: err,
                         method: 'retrieveData',
                     });
-                    _destroyResponse();
+                    _destroyReadable(readable);
                     return cbOnce(err);
                 }
                 // response.isclosed is set by the S3 server. Might happen if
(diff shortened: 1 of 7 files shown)


lib/s3routes/routesUtils.ts
```diff
@@ -441,23 +456,38 @@ function retrieveData(
...
-                    log.error('error piping data from source');
+                    log.error('error piping data from source', { error: err });
                     _destroyResponse();
                     return cbOnce(err);
                 });
+                // At this point, we assume that the data backend is available.
+                // We write the http headers before piping the data to the
+                // response object.
+                // Any subsequent error leads to destror deletion.
+                if (!response.headersSent) {
```
francoisferrand, line 460: this should always be `false`, right?

lib/s3routes/routesUtils.ts
```diff
@@ -441,23 +456,38 @@ function retrieveData(
...
-            // recommendation
-            response.end();
-        },
+                currentStream = null;
+                if (err) {
+                    log.error('abort response due to error', {
+                        // @ts-expect-error
+                        error: err.code, errMsg: err.message });
+                    // Only end if we haven't already destroyed the response
+                    if (!responseDestroyed) {
```
francoisferrand: should we not just check `!response.headersSent` instead?

lib/s3routes/routesUtils.ts
```diff
@@ -408,15 +420,18 @@ function retrieveData(
...
-                    log.error('failed to get object', {
+                    // In this case, we do not destroy the response:
+                    // nothing was sent to the client yet, so we
+                    // can return a proper error.
+                    log.debug('failed to get object', {
                         error: err,
                         method: 'retrieveData',
                     });
-                    _destroyResponse();
+                    _destroyReadable(readable);
```
francoisferrand, line 427: we used to destroy the `response`, now we destroy the `request` : can't we just "not destroy anything", and just let the readable be destroyed normally? (the readable will be closed once the response is closed anyway)
or is there a reason/need for this explicit "early" destroy?

lib/s3routes/routesUtils.ts
```diff
@@ -317,9 +318,11 @@ function okContentHeadersResponse(
             overrideParams['response-content-encoding'];
     }
     setCommonResponseHeaders(addHeaders, response, log);
-    const httpCode = range ? 206 : 200;
-    log.debug('response http code', { httpCode });
-    response.writeHead(httpCode);
+    if (shouldWriteHeaders) {
```
francoisferrand: instead of adding a condition here, I would rather remove `writeHead()` from here, and move it to the one place where `okContentHeadersResponse` is called with the default value shouldWriterHeaders=true (i.e. in `responseContentHeaders()`)

lib/s3routes/routesUtils.ts
```diff
@@ -689,16 +718,15 @@ export function streamUserErrorPage(
...
-    response.writeHead(error.code, { 'Content-type': 'text/html' });
-    response.on('finish', () => {
-        log.end().info('responded with streamed content', {
-            httpCode: response.statusCode,
-        });
-    });
-    return retrieveData(dataLocations, retrieveDataParams, response, log);
+    return retrieveData(dataLocations, retrieveDataParams, response, {
+        statusCode: error.code,
+        headers: { 'Content-type': 'text/html' },
```
francoisferrand: nit: `'Content-type': 'text/html'` can be set *here* with `response.setHeader` ?
(and since it is a header, it can be "overwritten" later if we actually send an error?)

(so we don't need to pass the headers and complexify retrieveData)

lib/s3routes/routesUtils.ts
ghost: Yes. As discussed this is more a security (lack of types where we do not have full confidence if we don't have a hidden route that sent the headers already). I can confirm that I see none currently. However, it will be handled in the long term ticket; ARSN-474 that will follow up this ticket with the long term rework.

lib/s3routes/routesUtils.ts
ghost: Yes, code is updated.

lib/s3routes/routesUtils.ts
ghost: I added a comment and a todo, to be done in ARSN-474

lib/s3routes/routesUtils.ts
ghost: Done.

lib/s3routes/routesUtils.ts
ghost: It was moved to a `setHeader`: it is basically the same at the end, but at least it simplifies how we pass the http code in the stream data function.

lib/s3routes/routesUtils.ts
```diff
@@ -441,23 +447,42 @@ function retrieveData(
...
-                    // @ts-expect-error
-                    error: err.code, errMsg: err.message });
-            }
-            // call end for all cases (error/success) per node.js docs
-            // recommendation
-            response.end();
-        },
+                currentStream = null;
+                if (err) {
+                    log.error('abort response due to error', {
```
francoisferrand: `abort` is only in case the header was already sent, and we don't send an errorResponse. When we reach line 475, we are just sending a "regular" error.

lib/s3routes/routesUtils.ts
```diff
@@ -441,23 +447,42 @@ function retrieveData(
...
+                        // @ts-expect-error
+                        error: err.code, errMsg: err.message });
+                    if (!response.headersSent) {
+                        return XMLResponseBackend.errorResponse(errors.ServiceUnavailable, response, log);
+                    }
```
francoisferrand, line 474: should we call `_destroyResponse` explicitely in the `else` case (to force aborting the response), or is it enough to call `response.end()` (i.e. no risk of closing the request cleanly but 'truncating' the body) ?

lib/s3routes/routesUtils.ts
```diff
@@ -441,23 +447,42 @@ function retrieveData(
...
+                }
+                // call end for all cases (error/success) per node.js docs
+                // recommendation
+                response.end(() => {
+                    log.end().info('responded with streamed content', {
```
francoisferrand, line 482: when this happens after an error, the httpCode is not relevant anymore : the log does not match in that case...
It is also somewhat redundant with the error logs posted above.

lib/s3routes/routesUtils.ts
```diff
@@ -408,15 +411,18 @@ function retrieveData(
...
-            (err: any, readable: http.IncomingMessage) => {
+            (err: Error, readable: http.IncomingMessage) => {
                 const cbOnce = jsutil.once(next);
                 // NB: readable is of IncomingMessage type
                 if (err) {
-                    log.error('failed to get object', {
+                    // In this case, we do not destroy the response:
+                    // nothing was sent to the client yet, so we
+                    // can return a proper error.
+                    log.debug('failed to get object', {
```
francoisferrand: nit: not sure about the log level change. This "inner" log has all the context of the error, so may be more accurate... But at the same time indeed the final log is sure to be called, in every situation.

not sure which way to go, but if we change this error then the other logs must be consistent: i.e. `log.error('error piping data from source'` should also be changed to lower log level.

lib/s3routes/routesUtils.ts
francoisferrand: we _expect_ the response has been destroyed in other error case, but it is not clear or enforced: at least a comment should be added to explain  it.

Actually, I wonder if maybe we should just move this line, i.e.
```
- _destroyReadable(readable);
+ if (!response.headersSent) {
+     _destroyReadable(readable);
+     XMLResponseBackend.errorResponse(errors.ServiceUnavailable, response, log);
+ } else {
      _destroyResponse();
+ }
```

what do you think?

tests/unit/s3routes/routesUtils/redirectRequestOnError.spec.js
```diff
@@ -70,36 +70,26 @@ describe('routesUtils.redirectRequestOnError', () => {
...
             assert.strictEqual(responseMock.statusCode, 301);
-            assertHeaders(responseMock, corsHeaders);
-            assertHeaders(responseMock, errorHeaders);
             assert.strictEqual(responseMock._headers.Location, routing.location);
-            assert.strictEqual(dataWrapperGetStub.callCount, 1);
-            assert.strictEqual(dataWrapperGetStub.getCall(0).args[0],
-                mockedDataLocations[0]);
-            assert.strictEqual(dataWrapperGetStub.getCall(0).args[1],
-                responseMock);
+            dataWrapperGetStub.restore();
```
francoisferrand: best to put this in a cleanup step, to ensure `restore()` even in case the test fails (so it does not affect other tests)

lib/s3routes/routesUtils.ts
ghost: The readable is for each location, hence we do not have it in the (error) callback, its handling should be implemented before, as might might have multiple locations...

francoisferrand: (really not sure it is best : it makes the case more self-contained, but still requires to tweak error handling to skip the `end()`, and changes responses to be sent from multiple places...)

francoisferrand: > The readable is for each location, hence we do not have it in the (error) callback, its handling should be implemented before, as might might have multiple locations...

yes, the "suggestion" (bad or good) was to send the response as well from within the loop (~line 425), were we close the readable

ghost: So calling response.end is indeed not necessary, and would actually lead to an error:

> Destroy the stream. Optionally emit an 'error' event, and emit a 'close'event (unless emitClose is set to false). After this call, the writable stream has ended and subsequent calls to write() or end() will result in an ERR_STREAM_DESTROYED error.

lib/s3routes/routesUtils.ts
ghost: I kept the error log, better dump twice than missing an error... And it remains consistent with other logs.

lib/s3routes/routesUtils.ts
ghost: The logs were moved out of the `err` case, and consistently log once based on the error. Also we don't call `end` if the response was destroyed.

lib/s3routes/routesUtils.ts
ghost: See first comment's answser

lib/s3routes/routesUtils.ts
```diff
@@ -441,23 +449,42 @@ function retrieveData(
...
-                    log.error('error piping data from source');
+                    log.error('error piping data from source', { error: err });
                     _destroyResponse();
                     return cbOnce(err);
                 });
+                // At this point, we assume that the data backend is available.
+                // We write the http headers before piping the data to the
+                // response object.
+                // Any subsequent error leads to destror deletion.
+                // todo: need to check if headers were already set
```
francoisferrand: not "todo", already done
```suggestion
```

francoisferrand (approved):

lib/s3routes/routesUtils.ts
ghost: done

welansari (approved):

benzekrimaha (approved):

PR #2311, opened 2025-03-13
ARSN-476: Perf improvement around arsenal error

BourgoisMickael:
## Best to review by commit

See Jira ticket (https://scality.atlassian.net/browse/ARSN-476) for profilings in cloudserver, vault, bucketd, metadata.

- Instanciate the `Is` object only once at startup
- Instanciate all errors to use for cases like `errors.AccessDenied.code` or `errors.AccessDenied.customizeDescription` and avoid an useless intermediate instance when instance fields like stack trace are not needed.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -19,7 +19,7 @@ import SubStreamInterface from './lib/s3middleware/azureHelpers/SubStreamInterfa
 import * as processMpuParts from './lib/s3middleware/processMpuParts';
 import * as retention from './lib/s3middleware/objectRetention';
 import * as lifecycleHelpers from './lib/s3middleware/lifecycleHelpers';
-export { default as errors } from './lib/errors';
+export { default as errors, errorInstances } from './lib/errors';
 export { default as Clustering } from './lib/Clustering';
 export * as ClusterRPC from './lib/clustering/ClusterRPC';
 export * as ipCheck from './lib/ipCheck';
--- a/lib/errors/index.ts
+++ b/lib/errors/index.ts
@@ -36,6 +36,16 @@ const createIs = (type: Name): Is => {
     return new Proxy(final, { get });
 };
 
+/**
+ * Is helper object for each of the 189 errors.
+ * It makes 189 * 189 = 35721 booleans in heap at startup.
+ * Adds +1.3 MB of js objects but greatly increase speed when using errors
+ * And avoid gc.
+ */
+const isPreCreated = Object.fromEntries(
+    (Object.keys(rawErrors) as Name[]).map(key => [key, createIs(key)])
+) as Record<Name, Is>;
+
 export class ArsenalError extends Error {
     /** HTTP status code. Example: 401, 403, 500, ... */
     #code: number;
@@ -52,7 +62,7 @@ export class ArsenalError extends Error {
         this.#code = code;
         this.#description = description;
         this.#type = type;
-        this.#is = createIs(type);
+        this.#is = isPreCreated[type];
 
         // This restores the old behavior of errors, to make sure they're now
         // backward-compatible. Fortunately it's handled by TS, but it cannot
--- a/lib/errors/index.ts
+++ b/lib/errors/index.ts
@@ -149,7 +149,17 @@ export class ArsenalError extends Error {
             const error = value[1];
             const { code, description } = error;
             const get = () => new ArsenalError(name, code, description);
-            Object.defineProperty(errors, name, { get });
+            // Perf improvement for the ok error.
+            // As it's used mainly for HTTP response using the code and message only.
+            // Return the same instance as no stack trace or other property are needed
+            if (name === 'ok') {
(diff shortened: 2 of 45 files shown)


lib/errors/index.ts
```diff
@@ -139,12 +149,46 @@ export class ArsenalError extends Error {
...
 /** Mapping of all possible Errors.
  * Use them with errors[error].customizeDescription for any customization. */
 export default ArsenalError.errors();
+
+/**
+ * Same as errors but already instanciated.
+ * To be used like errorInstances[error].customizeDescription or errorInstances.ok.code.
+ * Avoid intermediate instance
+ */
+export const errorInstances = ArsenalError.errorInstances();
```
BourgoisMickael, line 200: I'll try to use this where it's possible (in metadata, bucket, vault, cloudserver).

But if you think it's too much changes, I can drop this commit

lib/errors/index.ts
```diff
@@ -139,12 +149,46 @@ export class ArsenalError extends Error {
...
         });
         return errors as Errors
     }
+
+        /**
+         * Like errors but generate singleton instances of errors
+         * To use before .customizeDescription to avoid intermediate instance.
+         * Adds +0.2 MB in heap at start
+         */
+        static errorInstances() {
```
nicolas2bert, line 176: Should we enforce it? Or warn the contributor?

BourgoisMickael: Ideally I'd prefer we enforce it, and when I bump arsenal I'd replace any occurence of of `errors.xxx.customizeDescription` or `errors.xxx.code` with this so we have the good pattern.

nicolas2bert: How to enforce it?

BourgoisMickael: Right now there is no automated way of doing it. Just during review.

index.ts
ghost, line 1: Should we target the 8.2 branch? As this is breaking all implementations, this comes with risks. We had bugs in the past after altering how errors work.

lib/errors/index.ts
```diff
@@ -139,12 +149,46 @@ export class ArsenalError extends Error {
...
+            const errors = {}
+            Object.entries(rawErrors).forEach((value) => {
+                const name = value[0] as Name;
+                const error = value[1];
+                const { code, description } = error;
+                const instance = new ArsenalError(name, code, description);
+                Object.defineProperty(errors, name, { value: instance });
+            });
+            return errors as Errors
```
ghost:
```suggestion
            return errors as Errors;
```

lib/errors/index.ts
ghost: Do you think there is a way to keep the same variable name in the projects, to not require any code change? I believe no, but that would be the best way to be backward compatible. From what I understand this is mandatory unless you drop the commit, right?

lib/errors/index.ts
```diff
@@ -139,12 +149,46 @@ export class ArsenalError extends Error {
...
+            if (name === 'ok') {
+                Object.defineProperty(errors, name, { value: get() });
+            } else {
+                // Ideally for other errors caching the instance outside request handling
+                // would be beneficial if .customizeDescription is used or if
+                // no stack trace or additional field is required
+                Object.defineProperty(errors, name, { get });
+            }
         });
         return errors as Errors
```
ghost:
```suggestion
        return errors as Errors;
```

index.ts
BourgoisMickael: I put in 7.70 so we can include it in the next S3C `9.4.3` and then `9.5` that doesn't use the unification yet and still need arsenal 7.70.

It should not break most implementation, changes should be trivial (search and replace for each integration branch), but maybe we could look into having components use the latest versions of arsenal first to avoid big difference between branches

lib/errors/index.ts
BourgoisMickael: We can't use the same variable, we need to keep the `errors` helper for places where we want a real new error instance with stacktrace and possibility to add fields on it.

If I drop this commit, we would have no code change to do but keep a less performant intermediate instanciation when doing `errors.xx.customizeDescription`

ghost (approved):

dvasilas (approved):

lib/errors/index.ts
BourgoisMickael: Actually, I just noticed branches development/8 copy the stack from the error: https://github.com/scality/Arsenal/blob/56e3db709b1202908d95f6ce1c0e5f7148064501/lib/errors/index.ts#L116

So I can't bring this pre instanciated errors into branches 8...

BourgoisMickael: I've added https://github.com/scality/Arsenal/pull/2311/commits/cb8250b273cc5fe50a9d979a64dca5f9abfec7dd to handle the stack trace of pre instanciated error correctly for branches dev/8

PR #2318, opened 2025-03-14
support getCanonicalIdsByAccountIds in vault helper class

welansari:
Support `getCanonicalIdsByAccountIds` in Vault helper class. This API is needed for Cloudserver to replace Account info in the metadata of CRR replicated objects.

In S3C this used to be called directly by Backbeat of the replication source, but in Zenko the Vault admin API is not exposed so we delegate the task to the replication destination's Cloudserver.

Linked PR: https://github.com/scality/cloudserver/pull/5762

Issue: ARSN-478

### DIFF
--- a/lib/auth/AuthInfo.ts
+++ b/lib/auth/AuthInfo.ts
@@ -32,6 +32,18 @@ export type AuthV4Results = {
     accountQuota: AccountQuota,
 };
 
+export type AccountCanonicalInfo = {
+    accountId: string;
+    canonicalId: string;
+    name: string;
+}
+
+export type AccountCanonicalInfoResults = {
+    message: {
+        body: AccountCanonicalInfo[],
+    },
+};
+
 /**
  * Class containing requester's information received from Vault
  * @param {object} info from Vault including arn, canonicalID,
--- a/lib/auth/Vault.ts
+++ b/lib/auth/Vault.ts
@@ -1,6 +1,8 @@
 import { RequestLogger } from 'werelogs';
 import errors from '../errors';
-import AuthInfo, { AccountInfos, AuthInfoType, AuthorizationResults, AuthV4Results } from './AuthInfo';
+import AuthInfo, { AccountInfos, AuthInfoType, AuthorizationResults,
+    AuthV4Results, AccountCanonicalInfo } from './AuthInfo';
+import { ArsenalCallback } from '../types';
 import RequestContext from '../policyEvaluator/RequestContext';
 
 /** vaultSignatureCb parses message from Vault and instantiates
@@ -354,6 +356,36 @@ export default class Vault {
                 });
                 return callback(null, result);
             });
+    }
+
+    /**
+     * A getter for account canonical IDs given a list of account IDs
+     * @param accountIds - list of account IDs
+     * @param log - log object
+     * @param callback - callback function
+     * @returns callback with either error or an object from Vault
+     * containing canonicalID and display name for each account ID
+     */
+    getCanonicalIdsByAccountIds(
+        accountIds: string[],
+        log: RequestLogger,
+        callback: ArsenalCallback<AccountCanonicalInfo[]>,
+    ) {
+        log.trace('getting canonicalIDs from Vault based on accountIDs');
+        const options = {
+            reqUid: log.getSerializedUids(),
+            logger: log,
+        };
+        this.client.getCanonicalIdsByAccountIds(accountIds, options, (err, res) => {
+            if (err) {
+                log.debug('received error message from vault', {
+                    error: err,
+                    accountIds,
+                });
+                return callback(err);
+            }
+            return callback(null, res.message.body);
+        });
     }
 
     /** checkPolicies -- call Vault to evaluate policies
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -1,0 +1,7 @@
+import { ArsenalError } from './errors';
+
+export interface ArsenalCallback<T> {
+    (err: null, result: T): void;
+    (err: void): void;
+    (err: ArsenalError): void;
+};
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=20"
   },
-  "version": "8.2.7",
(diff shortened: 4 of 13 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=20"
   },
-  "version": "8.2.6",
+  "version": "8.2.7",
```
francoisferrand: please be careful here:
- a few arsenal PRs in progres, should only bump "right before" merging
- do we need a version for every arsenal PR, or can/should we "mutualize" and release once the last PR in the batch is merged?

lib/auth/Vault.ts
```diff
@@ -356,6 +357,35 @@ export default class Vault {
...
+     * @param accountIds - list of account IDs
+     * @param log - log object
+     * @param callback - callback function
+     * @returns callback with either error or an object from Vault
+     * containing canonicalID and display name for each account ID
+     */
+    getCanonicalIdsByAccountIds(
+        accountIds: string[],
+        log: RequestLogger,
+        callback: (err: Error | null, data?: AccountCannonicalInfo[]) => void
```
francoisferrand: nit: this callback is not precisely typed, it cannot have both null error and no data...

type of the callback is more something like `ArsenallBack` (in MongoClientInterface):

```
interface AccountCannonicalInfoCallback {
    (err: null, data: AccountCannonicalInfo[]): void;
    (err: Error): void;
};
```

In any case we should define this type, instead of copying all over the place (or maybe we can use a "standard" generic node error callback...)

francoisferrand (approved):

package.json
welansari: > do we need a version for every arsenal PR, or can/should we "mutualize" and release once the last PR in the batch is merged?

Should probably just bump at the last one unless we know the review cycle will take time.

lib/auth/AuthInfo.ts
```diff
@@ -32,6 +32,18 @@ export type AuthV4Results = {
     accountQuota: AccountQuota,
 };
 
+export type AccountCannonicalInfoResults = {
+    message: {
+        body: AccountCannonicalInfo[],
+    },
+};
+
+export type AccountCannonicalInfo = {
```
ghost: Should ideally be defined before `AccountCannonicalInfoResults` for coherence

lib/types.ts
```diff
@@ -0,0 +1,4 @@
+export interface GenericCallback<T> {
+    (err: null, data: T): void;
+    (err: Error): void;
+};
```
ghost: We have a very similar type defined in `lib/storage/metadata/mongoclient/MongoClientInterface.ts`, that now makes sense to move here:

```ts
interface ArsenalCallback<T> {
    (err: null, result: T): void;
    (err: void): void;
    (err: ArsenalError): void;
};
```

It also has the benefit of enforcing that we either have an error and no data, or data but no error.

lib/auth/Vault.ts
```diff
@@ -356,6 +358,35 @@ export default class Vault {
...
+     * @param callback - callback function
+     * @returns callback with either error or an object from Vault
+     * containing canonicalID and display name for each account ID
+     */
+    getCanonicalIdsByAccountIds(
+        accountIds: string[],
+        log: RequestLogger,
+        callback: GenericCallback<AccountCannonicalInfo[]>,
+    ) {
+        log.addDefaultFields({ accountIds, method: 'getCanonicalIdsByAccountIds' });
```
ghost: I am not sure we want to add a default field on the `method`, as the logger object will be shared with the S3 API call, that might also use this method field. If the field becomes default, then it might make debugging harder.

If the list of account id is large, this might also have an impact, why not only logging this in case of an error?

lib/auth/Vault.ts
```diff
@@ -356,6 +358,35 @@ export default class Vault {
...
+        log.trace('getting canonicalIDs from Vault based on accountIDs');
+        const options = {
+            reqUid: log.getSerializedUids(),
+            logger: log,
+        };
+        this.client.getCanonicalIdsByAccountIds(accountIds, options,
+            (err: Error | null, res?: AccountCannonicalInfoResults) => {
+                if (err) {
+                    log.error('received error message from vault', { error: err });
```
ghost: Usually I prefer debug logs in case of errors in "low-level" methods like that: this function alone does not know if the error is expected or not. Maybe we expect from above that none of the provided ID are found. So better to let the upper layers log with `error`: this will dump the debug log here.

```suggestion
                    log.debug('received error message from vault', { error: err });
```

lib/auth/Vault.ts
```diff
@@ -356,6 +358,35 @@ export default class Vault {
...
+                    return callback(err);
+                }
+                return callback(null, res?.message?.body || []);
```
ghost: By using the type from MongoClient Interface, we can avoid the `?`

lib/auth/AuthInfo.ts
```diff
@@ -32,6 +32,18 @@ export type AuthV4Results = {
     accountQuota: AccountQuota,
 };
 
+export type AccountCannonicalInfoResults = {
```
ghost: Typo
```suggestion
export type AccountCanonicalInfoResults = {
```

benzekrimaha (approved):

lib/auth/Vault.ts
```diff
@@ -356,6 +358,36 @@ export default class Vault {
...
+        callback: ArsenalCallback<AccountCanonicalInfo[]>,
+    ) {
+        log.trace('getting canonicalIDs from Vault based on accountIDs');
+        const options = {
+            reqUid: log.getSerializedUids(),
+            logger: log,
+        };
+        this.client.getCanonicalIdsByAccountIds(accountIds, options, (err, res) => {
+            if (err) {
+                log.debug('received error message from vault', {
```
benzekrimaha, line 381: shouldn't it be a log.error instead ?

welansari: see https://github.com/scality/Arsenal/pull/2318#discussion_r2005706112

PR #2337, opened 2025-03-28
impr(ARSN-475): Add bucketOwnerId field to ObjectMD model

tmacro:


ghost (approved):
You might want to bump arsenal as well before merging 🙂

lib/constants.ts
```diff
@@ -41,7 +41,9 @@ export const emptyFileMd5 = 'd41d8cd98f00b204e9800998ecf8427e';
 // any metadata of the object.
 // version 6 also adds the "isPHD" flag that is used to indicate that the master
 // object is a placeholder and is not up to date.
-export const mdModelVersion = 6;
+// Version 7 adds the "bucketOwnerId" attribute to the object metadata.
+// This is set when the owner of the bucket is different from the owner-id of the object.
```
ghost, line 45: Shall we add a bit more context here? To explain why we need it.

BourgoisMickael: at least describe that owner-id is the requester account id for cross account put object

lib/models/ObjectMD.ts
```diff
@@ -96,6 +96,7 @@ export type ObjectMDData = {
     // the master is set as a placeholder and gets updated with the new latest
     // version data after a certain amount of time.
     isPHD?: boolean;
+    bucketOwnerId?: string;
```
BourgoisMickael, line 104: document here the bucketOwnerId, and document the `owner-id` as well.

Should we also have a `bucket-owner-display-name` ?

tmacro: I think we should be fine without `bucket-owner-display-name`.

BourgoisMickael (approved):

PR #2345, opened 2025-04-07
Use SUR instead of SCUBA for the service name

ghost:
- Scuba name is not something we want to expose to our clients.
- Use the SUR name.

The actual service name in the ARN is still `scuba:`, because that's how the service is named internally.

Issue: ARSN-484

### DIFF
--- a/lib/policyEvaluator/utils/actionMaps.ts
+++ b/lib/policyEvaluator/utils/actionMaps.ts
@@ -228,15 +228,15 @@ const actionMapMetadata = {
 };
 
 const actionMapScuba = {
-    GetMetrics: 'scuba:GetMetrics',
-    GetMetricsBatch: 'scuba:GetMetricsBatch',
-    AdminStartIngest: 'scuba:AdminStartIngest',
-    AdminStopIngest: 'scuba:AdminStopIngest',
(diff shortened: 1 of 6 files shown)


francoisferrand (approved):

lib/policyEvaluator/utils/actionMaps.ts
```diff
@@ -228,15 +228,15 @@ const actionMapMetadata = {
 };
 
 const actionMapScuba = {
```
BourgoisMickael: rename this as well for consistency and to notice easily where it is used ?

BourgoisMickael:
To be sure you want to keep the name `scuba` for the service in arn like `arn:scality:scuba::JOHNACCOUNT:general-resource/specific-resource`

But replace the scuba only in the action ?

lib/policyEvaluator/utils/actionMaps.ts
ghost: Let's rename everything. We were not planning to use these ARNs in Zenko for now as we use roles but indeed at some point it will be needed and exposed, so better use `SUR` everywhere...

benzekrimaha (approved):

PR #2349, opened 2025-04-10
ARSN-485: Prefix KMS Keys with an arn (HF cldsrv 9.2.0.12)

BourgoisMickael:
To keep track of backend provider

For AWS use key Arn instead of KeyId by default
to allow for cross-account

___

See https://scality.atlassian.net/browse/S3C-9949
This hotfix to handle SSE with both internal and external KMS. Needed to built a custom image of cloudserver & vault.

Cherry-pick for vault: https://github.com/scality/Arsenal/pull/2350

It will then be cherry-picked into latest development for 9.5

### DIFF
--- a/lib/network/KMSInterface.ts
+++ b/lib/network/KMSInterface.ts
@@ -1,0 +1,167 @@
+import errors from "../errors";
+
+export const SCAL_KMS_ARN = 'arn:scality:kms:';
+
+export enum KmsType {
+    /** Internal scality provider */
+    int = 'int',
+    /** External provider */
+    ext = 'ext',
+};
+
+export enum KmsProtocol {
+    /** For tests & dev */
+    mem = 'mem',
+    /** Default if no provider */
+    file = 'file',
+    /** (SafeNet from Gemalto/Thales using scality-kms) */
+    scality = 'scality',
+    aws_kms = 'aws_kms',
+    kmip = 'kmip',
+};
+
+const kmsTypeProtocolMapping = {
+    [KmsType.int]: [KmsProtocol.mem, KmsProtocol.file] as const,
+    [KmsType.ext]: [KmsProtocol.scality, KmsProtocol.aws_kms, KmsProtocol.kmip] as const,
+} as const;
+export type KmsProtocolByType<T extends KmsType> = typeof kmsTypeProtocolMapping[T][number];
+
+/**
+ * Possible arn are:
+ *
+ * - `arn:scality:kms:int:mem:scality:key/$keyId` (tests & dev)
+ * - `arn:scality:kms:int:file:scality:key/$keyId` (default if no provider)
+ * - `arn:scality:kms:ext:scality:safenet:key/$keyId` (deprecated provider)
+ * - `arn:scality:kms:ext:aws_kms:aws:key/$keyIdOrArn`
+ * - `arn:scality:kms:ext:aws_kms:custom:alias/$keyIdOrArn`
+ * - `arn:scality:kms:ext:kmip:thales:key/$keyId`
+ * - `arn:scality:kms:ext:kmip:hashicorp:key/$keyId`
+ * - `arn:scality:kms:ext:kmip:hytrust:key/$keyId`
+ */
+export type AwsLikeKmsArnPrefix<T extends KmsType = KmsType> =
+    `${typeof SCAL_KMS_ARN}${T}:${KmsProtocolByType<T>}:${string}:${'key' | 'alias'}/`;
+/** Scality arn prefixing KMS KeyId to keep track of KMS backend */
+export type AwsLikeKmsArn<T extends KmsType = KmsType> = `${AwsLikeKmsArnPrefix<T>}${string}`;
+
+export interface KmsBackend<T extends KmsType> {
+    readonly type: T;
+    readonly protocol: KmsProtocolByType<T>;
+    readonly provider: string;
+    readonly arnPrefix: AwsLikeKmsArnPrefix<T>;
+}
+
+export function makeScalityArnPrefix<T extends KmsType>(
+    type: T, protocol: KmsProtocolByType<T>, provider: string
+): AwsLikeKmsArnPrefix<T> {
+    return `${SCAL_KMS_ARN}${type}:${protocol}:${provider}:key/`;
+}
+
+export function makeBackend<T extends KmsType>(
+    type: T, protocol: KmsProtocolByType<T>, provider: string
+): KmsBackend<T> {
+    return {
+        type,
+        protocol,
+        provider,
+        arnPrefix: makeScalityArnPrefix(type, protocol, provider),
+    }
+}
+
+export function isScalityKmsArn(kmsKey: string) {
+    return kmsKey.startsWith(SCAL_KMS_ARN);
+}
+
+/**
+ * Drop scality arn prefix from a KMS Key so it can be sent to the backend KMS.
+ * This is used in KMS clients as a safety but the key should be passed
+ * without arnPrefix as it should be extract by a KMS wrapper using `extractDetailFromArn`.
+ * For old key without arnPrefix this does nothing.
+ * Can be used to return KeyId in HTTP headers as well.
+ * @param kmsKeyIdOrArn KMS key coming from object metadata
+ * @returns KeyId extracted from scality arn (for AWS can still be an aws arn)
+ */
+export function getKeyIdFromArn(kmsKeyIdOrArn: AwsLikeKmsArn | string) {
+    if (isScalityKmsArn(kmsKeyIdOrArn)) {
+        // Do not use full arnPrefix to allow for different providerName
+        return kmsKeyIdOrArn.substring(kmsKeyIdOrArn.indexOf('/') + 1)
+    }
+    return kmsKeyIdOrArn;
+}
+
+// Functions used to validate configuration
+const kmsTypes = Object.values(KmsType);
+export function isValidType(type: KmsType | string): type is KmsType {
+    return kmsTypes.includes(type as KmsType);
+}
+export function isValidProtocol<T extends KmsType = KmsType>(
+    type: T, protocol: KmsProtocol | string
+): protocol is KmsProtocolByType<T> {
+    return (kmsTypeProtocolMapping[type] as unknown as KmsProtocol[] | undefined)
+        ?.includes(protocol as KmsProtocol) || false;
+}
+/** must be lowercase alphanumeric */
+export function isValidProvider(provider: string) {
+    return /^[a-z0-9]+$/.test(provider);
+}
+// end of functions to validate configuration
+
+export interface KmsArnDetail {
+    type: KmsType,
+    protocol: KmsProtocol,
+    provider: string,
+    arnType: 'key' | 'alias', // we don't support alias yet
+    id: string,
+}
+type NotValidatedKmsArnDetail = { [Key in keyof KmsArnDetail]?: KmsArnDetail[Key] | string };
+
+/**
+ * Split the scality arn to extract KMS backend identifiers.
+ * @param keyArn KMS Key prefixed with scality arn
+ * @returns KMS Key id with KMS backend identifiers
+ */
+export function extractDetailFromArn(keyArn: AwsLikeKmsArn): NotValidatedKmsArnDetail {
+    if (!isScalityKmsArn(keyArn)) {
+        return { id: keyArn };
+    }
+
+    const [arnPrefix, ...keyIdOrArn] = keyArn.split('/');
+    const [,,,type, protocol, provider, arnType] = arnPrefix.split(':');
+
+    return {
+        type,
+        protocol,
+        provider,
+        arnType,
+        id: keyIdOrArn?.join('/')
+    }
+}
+
+/**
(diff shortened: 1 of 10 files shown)


lib/network/kmsAWS/Client.ts
```diff
@@ -101,20 +114,24 @@ export default class Client {
                 return;
             }
             logger.debug("AWS KMS: master encryption key created", { KeyMetadata: data?.KeyMetadata });
-            cb(null, data?.KeyMetadata?.KeyId);
+            const keyId = this.noAwsArn ? data?.KeyMetadata?.KeyId : data?.KeyMetadata?.Arn ?? data?.KeyMetadata?.KeyId;
+            const arn = `${this.arnPrefix}${keyId}`;
```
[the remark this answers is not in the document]
BourgoisMickael: Yes, this is expected, I've added a comment here to describe the use case and workaround

tcarmet (approved):

lib/network/KMSInterface.ts
```diff
@@ -0,0 +1,68 @@
...
+
+interface KmsTypeProtocolMapping {
+    [KmsType.int]: KmsProtocol.scality | KmsProtocol.mem | KmsProtocol.file,
+    [KmsType.ext]: KmsProtocol.aws_kms | KmsProtocol.kmip,
+};
+
+/**
+ * Possible arn are:
+ * 
+ * - `arn:scality:kms:int:scality:safenet:key/$keyId` (deprecated provider)
```
BourgoisMickael: I consider the deprecated `Safenet / Gemalto/Thales` using the `scality-kms` backend as internal provider with a protocol = `scality` and provider name = `safenet`.

Is this ok ?

lib/network/KMSInterface.ts
```diff
@@ -0,0 +1,68 @@
...
+ * - `arn:scality:kms:ext:aws_kms:custom:alias/$keyIdOrArn`
+ * - `arn:scality:kms:ext:kmip:thales:key/$keyId`
+ * - `arn:scality:kms:ext:kmip:hashicorp:key/$keyId`
+ */
+export type AwsLikeKmsArnPrefix<T extends KmsType = KmsType> = `arn:scality:kms:${T}:${KmsTypeProtocolMapping[T]}:${string}:${'key' | 'alias'}/`;
+/** Scality arn prefixing KMS KeyId to keep track of KMS backend */
+export type AwsLikeKmsArn<T extends KmsType = KmsType> = `${AwsLikeKmsArnPrefix<T>}${string}`;
+
+/**
+ * Drop scality arn prefix from a KMS Key so it can be sent to be backend KMS
```
jonathan-gramain:
```suggestion
 * Drop scality arn prefix from a KMS Key so it can be sent to the backend KMS
```

lib/network/kmsAWS/Client.ts
```diff
@@ -101,20 +114,27 @@ export default class Client {
                 return;
             }
             logger.debug("AWS KMS: master encryption key created", { KeyMetadata: data?.KeyMetadata });
-            cb(null, data?.KeyMetadata?.KeyId);
+            const keyId = this.noAwsArn ? data?.KeyMetadata?.KeyId : data?.KeyMetadata?.Arn ?? data?.KeyMetadata?.KeyId;
```
jonathan-gramain: This looks quite cryptic, could you spread it on multiple lines with explicit `if`s instead?

jonathan-gramain (approved):

lib/network/kmsAWS/Client.ts
```diff
@@ -23,6 +26,11 @@ interface KMSOptions {
         maxVersion?: TLSVersion;
         key?: Buffer | Buffer[];
     };
+    /**
+     * Do not use AWS KMS Key arn but id (to keep backward compatibility)
+     * Not set or false by default to use arn to be future-proof (multi account, region)
+     */
+    noAwsArn?: boolean;
```
jonathan-gramain, line 40: Wondering:
- can existing customers use the ARN when they already have KMS-encrypted keys or internal keys?
- if not, how do we ensure they will not run into issues? Can we enforce the correct configuration automatically in such case?

BourgoisMickael: - for internal keys, they will never reach this kmsAws client. Existing key or non kmsAws key will be sent to the correct client by cloudserver that parse our own scality arn prefix  (different from the kmsAws client's aws arn) to know the key provider.

- if they already have kmsAws keys (using only the KeyId) it's not a problem, for aws KeyId and KeyArn are interchangeable (except for cross-acount). We should still verify with custom if they do handle KeyArn as AWS does.

- Correct configuration would be to use Arn by default, except if customer's own kmsAws doesn't handle Arn. This case should be written in documentation. The only difference for now is we don't plan to migrate aws KeyId to include the aws KeyArn

lib/network/kmsAWS/Client.ts
BourgoisMickael:
```javascript
            const keyMd = data?.KeyMetadata;

            const keyId = this.noAwsArn
                ? keyMd?.KeyId
                : keyMd?.Arn || keyMd?.KeyId;

```
Does this looks better ?

lib/network/KMSInterface.ts
BourgoisMickael: After discussion with david, this one should rather be an external

lib/network/KMSInterface.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @param kmsKeyIdOrArn KMS key coming from object metadata
+ * @returns KeyId extracted from scality arn (for AWS can still be an aws arn)
+ */
+export function getKeyIdFromArn(kmsKeyIdOrArn: AwsLikeKmsArnPrefix | string) {
+    if (kmsKeyIdOrArn.startsWith('arn:scality:kms:')) {
+        // Do not use full arnPrefix to allow for different providerName
+        return kmsKeyIdOrArn.substring(kmsKeyIdOrArn.indexOf('/') + 1)
+    }
+    return kmsKeyIdOrArn;
```
nicolas2bert, line 88: Is the condition introduced to handle old kmsKeyIdOrArn? Can you add it in a comment?

lib/network/kmsAWS/Client.ts
nicolas2bert: with if/else? Could it be more readable?

 ```
const keyMetadata = data?.KeyMetadata;

let keyId;

if (this.noAwsArn) {
    // Use KeyId when ARN is not required
    keyId = keyMetadata?.KeyId;
} else {
    // Prefer ARN, but fall back to KeyId if ARN is missing
    keyId = keyMetadata?.Arn ?? keyMetadata?.KeyId;
}
```

File: lib/network/kmip/Client.ts [hunk 7]
```diff
@@ -384,19 +391,20 @@ export default class Client {
             if (!this.options.compoundCreateActivate) {
                 return this._activateBucketKey(uniqueIdentifier, logger, cb);
             }
-            return cb(null, uniqueIdentifier);
+            return cb(null, uniqueIdentifier, `${this.arnPrefix}${uniqueIdentifier}`);
```
nicolas2bert: Why do we still need to pass the `uniqueIdentifier`? 
`${this.arnPrefix}${uniqueIdentifier}` is not enough?

lib/network/KMSInterface.ts
```diff
@@ -0,0 +1,68 @@
...
+/** Scality arn prefixing KMS KeyId to keep track of KMS backend */
+export type AwsLikeKmsArn<T extends KmsType = KmsType> = `${AwsLikeKmsArnPrefix<T>}${string}`;
+
+/**
+ * Drop scality arn prefix from a KMS Key so it can be sent to be backend KMS
+ * @param kmsKeyIdOrArn KMS key coming from object metadata
+ * @returns KeyId extracted from scality arn (for AWS can still be an aws arn)
+ */
+export function getKeyIdFromArn(kmsKeyIdOrArn: AwsLikeKmsArnPrefix | string) {
+    if (kmsKeyIdOrArn.startsWith('arn:scality:kms:')) {
```
nicolas2bert: NIT:

we could move the prefix to a constant since it is used later on.
We could also move this simple logic to a function to explain what we are doing.
exemple:
`isScalityKmsArn(...)`
// The Scality KMS ARN prefix identifying the new Scality KMS ARN format.

tests/functional/kmsAWS/highlevel.spec.js
```diff
@@ -56,9 +57,58 @@ describe('KmsAWSClient', () => {
...
+        client.createMasterKey(logger, (err, keyId, keyArn) => {
             assert.ifError(err);
             assert.strictEqual(keyId, 'mock-key-id');
+            assert.strictEqual(keyArn, 'arn:scality:kms:ext:aws_kms:tests:key/mock-key-id');
+            assert(createKeyStub.calledOnce);
+            done();
+        });
+    });
+
+    it('should create a new master encryption key with aws arn', done => {
```
nicolas2bert, line 100: Can we add a test when the KeyMetadata.Arn is missing from the response?

lib/network/kmsAWS/Client.ts
nicolas2bert: Can you provide an example of what a keyId looks like with noAwsArn set to false or true.

`arn:scality:kms:ext:aws_kms:custom:key/arn:aws:kms:region:accountId:key/<KEY_ID>` vs 
`arn:scality:kms:ext:aws_kms:custom:key/<KEY_ID>`

lib/network/KMSInterface.ts
BourgoisMickael: Fixed https://github.com/scality/Arsenal/pull/2349/files#diff-1c24d37a19bc08405b548ef9b5ac12a2ef83b7edb3a2ee62cc8426c7df8e5b95R23-R26

lib/network/KMSInterface.ts
BourgoisMickael: Fixed https://github.com/scality/Arsenal/pull/2349/files#diff-1c24d37a19bc08405b548ef9b5ac12a2ef83b7edb3a2ee62cc8426c7df8e5b95R75

lib/network/kmsAWS/Client.ts
BourgoisMickael: Fixed https://github.com/scality/Arsenal/pull/2349/files#diff-fb4f257ae2f25bd280160e3cdec22c57fdd9aaa0ee35637e056146ce2b35e6ecR124-R131

lib/network/kmsAWS/Client.ts
BourgoisMickael: Fixed https://github.com/scality/Arsenal/pull/2349/files#diff-fb4f257ae2f25bd280160e3cdec22c57fdd9aaa0ee35637e056146ce2b35e6ecR29-R39

lib/network/KMSInterface.ts
BourgoisMickael: Yes, updated the doc: https://github.com/scality/Arsenal/pull/2349/files#diff-1c24d37a19bc08405b548ef9b5ac12a2ef83b7edb3a2ee62cc8426c7df8e5b95R75-R78

lib/network/kmip/Client.ts
BourgoisMickael: I've added another parameter on each KMS client to keep backwards compatibility and avoid breaking anything, but this is not used anymore.

The KMS wrapper will always pick the 3rd arg with arnPrefix. I could remove the 2nd arg without arnPrefix if needed

lib/network/KMSInterface.ts
BourgoisMickael: Added there: https://github.com/scality/Arsenal/pull/2349/files#diff-1c24d37a19bc08405b548ef9b5ac12a2ef83b7edb3a2ee62cc8426c7df8e5b95R70-R72

tests/functional/kmsAWS/highlevel.spec.js
BourgoisMickael: There was already a test above doing that, as the flag `noAwsArn` is falsy by default and the Arn was no provided, but I added a test by passing tha flag `noAwsArn` to false explicitly: https://github.com/scality/Arsenal/pull/2349/files#diff-beda77b39ab0827a93fdfc3954bd6e3232f38f334dec6851b3699262214ca633R73-R98

lib/network/kmip/Client.ts
```diff
@@ -302,11 +308,12 @@ export default class Client {
      * Activate a cryptographic key managed by the server,
      * for a specific bucket. This is a required action to perform after
      * the key creation.
-     * @param keyIdentifier - The bucket key Id
+     * @param keyIdentifierOrArn - The bucket key Id
      * @param logger - Werelog logger object
      * @param cb - The callback(err: Error)
```
dvasilas: minor: Maybe also update `callback(err: Error)`.

Same in `createBucketKey`.

dvasilas (approved):

anurag4DSB (approved):

lib/network/kmip/Client.ts
BourgoisMickael: Fixed

PR #2350, opened 2025-04-10
ARSN-485: Prefix KMS Keys with an arn (HF vault 9.2.0.12) Cherry-pick

BourgoisMickael:
Cherry-pick https://github.com/scality/Arsenal/pull/2349

### DIFF
--- a/lib/network/KMSInterface.ts
+++ b/lib/network/KMSInterface.ts
@@ -1,0 +1,167 @@
+import errors from "../errors";
+
+export const SCAL_KMS_ARN = 'arn:scality:kms:';
+
+export enum KmsType {
+    /** Internal scality provider */
+    int = 'int',
+    /** External provider */
+    ext = 'ext',
+};
+
+export enum KmsProtocol {
+    /** For tests & dev */
+    mem = 'mem',
+    /** Default if no provider */
+    file = 'file',
+    /** (SafeNet from Gemalto/Thales using scality-kms) */
+    scality = 'scality',
+    aws_kms = 'aws_kms',
+    kmip = 'kmip',
+};
+
+const kmsTypeProtocolMapping = {
+    [KmsType.int]: [KmsProtocol.mem, KmsProtocol.file] as const,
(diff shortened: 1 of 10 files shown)


tcarmet (approved):

dvasilas (approved):

anurag4DSB (approved):
Cherry pick, approving.

lib/network/KMSInterface.ts
```diff
@@ -0,0 +1,167 @@
...
+    [KmsType.int]: [KmsProtocol.mem, KmsProtocol.file] as const,
+    [KmsType.ext]: [KmsProtocol.scality, KmsProtocol.aws_kms, KmsProtocol.kmip] as const,
+} as const;
+export type KmsProtocolByType<T extends KmsType> = typeof kmsTypeProtocolMapping[T][number];
+
+/**
+ * Possible arn are:
+ *
+ * - `arn:scality:kms:int:mem:scality:key/$keyId` (tests & dev)
+ * - `arn:scality:kms:int:file:scality:key/$keyId` (default if no provider)
```
anurag4DSB, line 33: Is this from before? Can we use `internal`?

lib/network/KMSInterface.ts
```diff
@@ -0,0 +1,167 @@
...
+ * - `arn:scality:kms:ext:scality:safenet:key/$keyId` (deprecated provider)
+ * - `arn:scality:kms:ext:aws_kms:aws:key/$keyIdOrArn`
+ * - `arn:scality:kms:ext:aws_kms:custom:alias/$keyIdOrArn`
+ * - `arn:scality:kms:ext:kmip:thales:key/$keyId`
+ * - `arn:scality:kms:ext:kmip:hashicorp:key/$keyId`
+ * - `arn:scality:kms:ext:kmip:hytrust:key/$keyId`
```
anurag4DSB, line 39: Same for this `external` is preferable

anurag4DSB: May be not possible as this was from before, but if we are introducing something new, we should use full name for clarity.

BourgoisMickael: It's new and it was shipped like that but it will be documented for user to understand every value possible

PR #2351, opened 2025-04-14
Support latest MongoDB driver

ghost:
There is actually no change to make, but I seized the opportunity to cleanup a bit related logic, and the lib we use for testing.

The main contribution is to use MongoDB stats to get the disk usage (although not very useful as we have the metrics independently, might be reused later...).

We also update the LogConsumer class, that was not migrated.

Issue: ARSN-454


lib/storage/metadata/mongoclient/LogConsumer.js
```diff
@@ -89,18 +91,17 @@ class LogConsumer {
...
                 tailable: true,
                 awaitData: true,
                 noCursorTimeout: true,
                 numberOfRetries: Number.MAX_VALUE,
-            }, (err, cursor) => {
-                const recordStream = new ListRecordStream(
-                    cursor, this._logger, startSeq.uniqID, latestOplogID);
-                return cb(null, { log: recordStream, tailable: true });
             });
+            const recordStream = new ListRecordStream(
```
francoisferrand, line 102: should `ListRecordStream` not be updated as well, to use the new custer?
looking at https://github.dev/scality/Arsenal/blob/dd9283d156a5d7fa57987e48314aa3f904f4508a/lib/storage/metadata/mongoclient/ListRecordStream.js#L47, we see it is still using the callback-based API...

ghost: Yes, I can include it in this PR as well...  I will add some commits.

ghost: Done in https://github.com/scality/Arsenal/pull/2351/commits/44f4beaa759ef9d62968c9e137a23e74927535e3

ghost: This whole logic will be removed in ARSN-486.

francoisferrand (approved):

benzekrimaha (approved):

PR #2363, opened 2025-04-24
Add mongodb driver checks

ghost:
Catching errors might not be enough: in some cases, a call can succeed yet the result is not what we expect: e.g. no document matching the filters. This leads to inconsistent behaviors where the API thinks the operation succeed while it didn't...

In this PR we identify and fix multiple places where we are ignoring MongoDB return flags. We add the error handling and multiple tests are added, as today some methods are not unit tested at all.

Issue: ARSN-488

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -417,7 +417,11 @@ class MongoClientInterface {
         }, payload, {
             upsert: true,
         })
-            .then(() => {
+            .then(result => {
+                if (result.matchedCount === 0 && result.modifiedCount === 0 && result.upsertedCount === 0) {
+                    log.debug('createBucket: failed to create bucket', { bucketName, result });
+                    return cb(errors.InternalError);
+                }
                 // caching bucket vFormat
                 this.bucketVFormatCache.add(bucketName, payload.$set.vFormat);
                 // NOTE: We do not need to create a collection for
@@ -583,7 +587,13 @@ class MongoClientInterface {
         }, {
             upsert: true,
         })
-            .then(() => cb(null))
+            .then(result => {
+                if (result.matchedCount === 0 && result.modifiedCount === 0 && result.upsertedCount === 0) {
+                    log.debug('putBucketAttributes: failed to update bucket', { bucketName });
+                    return cb(errors.NoSuchBucket);
+                }
+                return cb(null);
+            })
             .catch(err => {
                 log.error(
                     'putBucketAttributes: error putting bucket attributes',
@@ -623,7 +633,13 @@ class MongoClientInterface {
             },
         }, {
             upsert: true,
-        }).then(() => cb(null)).catch(err => {
+        }).then(result => {
+            if (result.matchedCount === 0 && result.modifiedCount === 0 && result.upsertedCount === 0) {
+                log.error('putBucketAttributesCapabilities: failed to update bucket', { bucketName });
+                return cb(errors.InternalError);
+            }
+            return cb(null);
+        }).catch(err => {
             log.error(
                 'putBucketAttributesCapabilities: error putting bucket attributes',
                 { error: err.message });
@@ -657,7 +673,14 @@ class MongoClientInterface {
             $unset: {
                 [updateString]: '',
             },
-        }).then(() => cb(null)).catch(err => {
+        }).then(result => {
+            if (result.matchedCount === 0 && result.modifiedCount === 0) {
+                log.debug('deleteBucketAttributesCapability: bucket not found or capability not present',
+                    { bucketName });
+                return cb(errors.NoSuchBucket);
+            }
+            return cb(null);
+        }).catch(err => {
             if (err) {
                 log.error(
                     'deleteBucketAttributesCapability: error deleting bucket attributes',
@@ -1967,7 +1990,19 @@ class MongoClientInterface {
         if (params && params.doesNotNeedOpogUpdate) {
             // If flag is true, directly delete object
             return collection.deleteOne(deleteFilter)
-                .then(() => cb(null, undefined))
+                .then(result => {
+                    // In case of race conditions (e.g. two concurrent deletes), or invalid state
+                    // (e.g. object is already deleted), the result.deletedCount will be 0
+                    // without error, and we need to catch it because the API might continue and
+                    // delete the actual data, leading to an invalid state, where MongoDB references
+                    // and object not in the data layers anymore.
+                    if (!result || result?.deletedCount != 1) {
+                        log.debug('internalDeleteObject: object not found or already deleted',
+                            { bucket: bucketName, object: key });
+                        return cb(errors.NoSuchKey);
+                    }
+                    return cb(null, undefined);
+                })
                 .catch(err => {
                     log.error('internalDeleteObject: error deleting object',
                         { bucket: bucketName, object: key, error: err.message });
@@ -2030,10 +2065,24 @@ class MongoClientInterface {
                         filter: updateDeleteFilter,
                     },
                 },
-            ], { ordered: true }).then(() => next(null)).catch(err => next(err)),
+            ], { ordered: true }).then(result => {
+                // in case of race conditions, the bulk operation might fail
+                // in this case we return a DeleteConflict error
+                if (!result || !result.ok) {
+                    log.debug('internalDeleteObject: bulk operation failed', 
+                        { bucket: bucketName, object: key });
+                    return next(errors.DeleteConflict);
+                }
+                if (result.deletedCount === 0) {
+                    log.debug('internalDeleteObject: object not found or already deleted', 
+                        { bucket: bucketName, object: key });
+                    return next(errors.DeleteConflict);
+                }
+                return next(null);
+            }).catch(err => next(err)),
         ], (err, res) => {
             if (err) {
-                if (err instanceof ArsenalError && err.is.NoSuchKey) {
+                if (err instanceof ArsenalError) {
                     return cb(err);
                 }
                 log.error('internalDeleteObject: error deleting object',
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=16"
   },
-  "version": "8.1.152",
+  "version": "8.1.153",
```
ghost: Bump is only for the 8.1 branch. The 8.2 will be released in another PR.

francoisferrand: do we actually need this in 8.1 ? (not sure which branch of cloudserver is still using 8.1...)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -417,7 +417,11 @@ class MongoClientInterface {
         }, payload, {
             upsert: true,
         })
-            .then(() => {
+            .then(result => {
+                if (result.modifiedCount === 0 && result.upsertedCount === 0) {
+                    log.debug('createBucket: failed to create bucket', { bucketName });
+                    return cb(errors.InternalError);
```
francoisferrand, line 423: * error should not be `InternalError`, but `errors.BucketAlreadyExists`
* detecting such case may be needed, to better handle the situation in upper layer : but it also will break things as some race conditions which were handled here (somewhat implicitely) will now have to be managed... → we may want to double-check if we really want to detect the error here "systematically", or if in some cases it is best to (explicitely) handle the race condition seamlessly here.... This analysis should be done case-by-case for each API, as different behavior may be needed in each case.
* we should also look at how this is handled in metadata, to ensure we have consistent behavior accross backends...

----
note: thinking about race condition on bucket creation, we may need more in here to handle it: if we exit early (and deal with it in the upper layers), the "second" call will return while the rest of the code (createCollection, enable sharding) has not executed yet....and we may thus start to write before sharding is enabled :-/

package.json
ghost: All Artesca versions in prod are currently using it: I believe that we'll want to have the fixes around the ghost issue for previous versions as well for some customers... So we should add the changes in 8.1.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -583,7 +587,13 @@ class MongoClientInterface {
         }, {
             upsert: true,
         })
-            .then(() => cb(null))
+            .then(result => {
+                if (result.modifiedCount === 0 && result.upsertedCount === 0) {
```
francoisferrand: is it really an error?
the update was useless, but we "reached" the expected state, and upper layer can proceed...

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2030,7 +2065,19 @@ class MongoClientInterface {
                         filter: updateDeleteFilter,
                     },
                 },
-            ], { ordered: true }).then(() => next(null)).catch(err => next(err)),
+            ], { ordered: true }).then(result => {
+                if (!result || !result.ok) {
+                    log.debug('internalDeleteObject: bulk operation failed', 
+                        { bucket: bucketName, object: key });
+                    return next(errors.InternalError);
+                }
```
francoisferrand: when does this happen?
is it the same kind of race condition on concurrent delete, which should return a 409 / `errors.DeleteConflict` ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2030,7 +2065,19 @@ class MongoClientInterface {
...
+                if (result.deletedCount === 0) {
```
francoisferrand, line 2076: same comment (and ideally code?) as above...

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2030,7 +2065,19 @@ class MongoClientInterface {
...
+                    log.debug('internalDeleteObject: object not deleted',
```
francoisferrand: `object not found or already deleted`

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -3124,7 +3177,13 @@ class MongoClientInterface {
     putBucketIndexes(bucketName: string, indexSpecs, log: werelogs.Logger, cb: ArsenalCallback<void>) {
         const c = this.getCollection<ObjectMetastoreDocument>(bucketName);
         const indexes = MongoUtils.indexFormatObjectToMongoArray(indexSpecs);
-        c.createIndexes(indexes).then(() => cb(null)).catch(err => {
+        c.createIndexes(indexes).then(result => {
```
francoisferrand: same, may be OK if the (same) index already exists, and we just have a race condition on creating the same index....

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -3147,7 +3206,13 @@ class MongoClientInterface {
     deleteBucketIndexes(bucketName: string, indexSpecs: { name: string }[], log: werelogs.Logger, cb: ArsenalCallback<void>) {
         const c = this.getCollection<ObjectMetastoreDocument>(bucketName);
         async.each(indexSpecs,
-            (spec, next) => c.dropIndex(spec.name).then(() => next()).catch(err => next(err)),
+            (spec, next) => c.dropIndex(spec.name).then(result => {
```
francoisferrand, line 3205: probably OK if the index is already deleted : could be handled here, or return an error `NoSuchIndex` or `DeleteConflict` so caller can manage...

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: There should be an additional `result.matchedCount` here, meaning that we really missed the update...
But actually, why do we accept upserts in this case? We should already have a bucketMD when we update its attributes...

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: In case of race condition yes, the bulk might conflict and return an error. I agree with the DeleteConflict error.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: the rationale was the same as the other similar check below, so it is missing the .matchedCount. What I wanted to detect here is when we failed to match a document, modified none and upserted none... Which is not an existing bucket in this case. For existing buckets, this is already handled as you mentioned.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: I went with the second approach: we return a DeleteConflict.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: I ignore errors here, it's true that in this case we can ignore, as the index is already there.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: I added the DeleteConflict error in this case, which is what we want: if we missed to delete the document then it means something went wrong, or a race condition happened: we do not want to continue the API call.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -623,7 +633,13 @@ class MongoClientInterface {
             },
         }, {
             upsert: true,
-        }).then(() => cb(null)).catch(err => {
+        }).then(result => {
+            if (result.modifiedCount === 0 && result.upsertedCount === 0) {
```
benzekrimaha: we can have a case where we try to update with the same value , then we can have the operation succeeding but the upserted, and modified counts to 0 but it's still not an internal error ?  ( the case where the bucket is created with capabilities already... )

benzekrimaha: we can add a check for the acknowledged in this case 
```suggestion
            if (result.acknowledged && result.modifiedCount === 0 && result.upsertedCount === 0) {
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -657,7 +673,14 @@ class MongoClientInterface {
             $unset: {
                 [updateString]: '',
             },
-        }).then(() => cb(null)).catch(err => {
+        }).then(result => {
+            if (result.modifiedCount === 0) {
```
benzekrimaha: For me here we can't rely on the modifiedCount to actually return the noSuch bucket, since if the bucket exists and has no capabilities , deleting it would actually result in a modified count to 0 even if the bucket exists? 
To check the bucket existence we should go with a findOne instead before trying to delete based on which IMO the error would be more relevant

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: I went with the .matchedCount, which is what we want here: the acknowledged should always be true because we specify it in the writeConcern. If it's not true we end up in the `catch`.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -583,7 +587,13 @@ class MongoClientInterface {
         }, {
             upsert: true,
         })
-            .then(() => cb(null))
+            .then(result => {
+                if (result.matchedCount === 0 && result.modifiedCount === 0 && result.upsertedCount === 0) {
+                    log.debug('putBucketAttributes: failed to update bucket', { bucketName });
+                    return cb(errors.NoSuchBucket);
```
francoisferrand: nit:
- this is really a case which "should not" (but does) happen: may be best to keep it as InternalError, possibly customized with an indication of the "mongo" issue (`errors.InternalError.customize("mongo failed to match, update and upsert")`)
- maybe we can dedup this case of error detection after an update operation, by introducing a `processUpdateResult(result, cb)` function?

package.json
```diff
@@ -3,7 +3,7 @@
...
   "description": "Common utilities for the S3 project components",
```
francoisferrand: 8.1.153 already released, pleaase rebase

francoisferrand (approved):

welansari (approved):

PR #2371, opened 2025-04-30
Improvement/Fixup on buckets creations

benzekrimaha:
with the migration to typescript the bucket creation has been changed always passing the capabilities as `undefined` , which is now breaking s3utils as we’re not able to retrieve the capabilities in order to update them

Issue: ARSN-492

### DIFF
--- a/lib/models/Veeam.ts
+++ b/lib/models/Veeam.ts
@@ -4,6 +4,7 @@
  * The capcity-related field accept numbers, but will be treated as
  * bigints internally.
  */
+import { Long } from 'mongodb';
 export type VeeamSOSApiSchema = {
     SystemInfo?: {
         ProtocolVersion: string,
@@ -26,9 +27,9 @@ export type VeeamSOSApiSchema = {
         LastModified?: string,
     },
     CapacityInfo?: {
-        Capacity: bigint | number,
-        Available: bigint | number,
-        Used: bigint | number,
+        Capacity: bigint,
+        Available: bigint,
+        Used: bigint,
         LastModified?: string,
     },
 };
@@ -52,7 +53,12 @@ export type VeeamSOSApiSerializable = Omit<VeeamSOSApiSchema, 'CapacityInfo'> & 
  * ability to use the proprietary SOSAPI feature.
  */
 export class VeeamCapacityInfo {
-    static serialize(capacity: VeeamSOSApiSchema['CapacityInfo']): VeeamSOSApiSerializable['CapacityInfo'] {
+    static serialize(capacity: {
+        Capacity: bigint | number | Long,
+        Available: bigint | number | Long,
+        Used: bigint | number | Long,
+        LastModified?: string,
+    }): VeeamSOSApiSerializable['CapacityInfo'] {
         return {
             Capacity: capacity?.Capacity?.toString() || '0',
             Available: capacity?.Available?.toString() || '0',
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -19,7 +19,7 @@ import errors, { ArsenalError, errorInstances } from '../../../errors';
 import BucketInfo, { BucketMetadata, Capabilities } from '../../../models/BucketInfo';
 import ObjectMD, { ObjectMDData } from '../../../models/ObjectMD';
 import * as jsutil from '../../../jsutil';
-import { ArsenalCallback } from '../../../types';
+import { ArsenalCallback, NestedOmit } from '../../../types';
 
 import { MongoClient, UpdateFilter, Long, MongoServerError } from 'mongodb';
 import type {
@@ -45,7 +45,7 @@ import { Transform } from 'stream';
 import { Version } from '../../../versioning/Version';
 
 import { formatMasterKey, formatVersionKey } from './utils';
-import { VeeamCapacityInfo, VeeamSOSApiSchema } from '../../../models/Veeam';
+import { VeeamCapacityInfo } from '../../../models/Veeam';
 import { BucketVersioningFormat } from '../../../versioning/constants';
 
 const VID_NONE = '';
@@ -108,20 +108,19 @@ export type MongoDBClientInterfaceParameters = {
     shardCollections: boolean,
 };
 
-export type CapabilitiesMongoDB = Capabilities & {
-    VeeamSOSApi?: Omit<VeeamSOSApiSchema, 'CapacityInfo'> & {
-        CapacityInfo?: {
-            Capacity: Long,
-            Available: Long,
-            Used: Long,
-        },
-    },
-}
-
 export type BucketMetadataMongoDB = Omit<Omit<BucketMetadata, 'quotaMax'>, 'capabilities'> & {
-    // Old buckets might not have a quotaMax
-    quotaMax?: Long,
-    capabilities?: CapabilitiesMongoDB,
+    // Old buckets might not have a quotaMax 
+    quotaMax?: Long;
+    capabilities?: NestedOmit<Capabilities, 'VeeamSOSApi.CapacityInfo'> & {
+        VeeamSOSApi?: {
+            CapacityInfo?: {
+                Capacity: Long;
+                Available: Long;
+                Used: Long;
+                LastModified?: string;
+            };
+        };
+    };
 };
 
 export interface BucketMetastoreDocument extends Document {
@@ -411,7 +410,18 @@ class MongoClientInterface {
                 value: {
                     ...newBucketMD,
                     quotaMax: new Long(newBucketMD.quotaMax || '0'),
-                    capabilities: undefined,
+                    capabilities: newBucketMD.capabilities && {
+                        ...newBucketMD.capabilities,
+                        VeeamSOSApi: newBucketMD.capabilities.VeeamSOSApi && {
+                            ...newBucketMD.capabilities.VeeamSOSApi,
+                            CapacityInfo: newBucketMD.capabilities.VeeamSOSApi.CapacityInfo && {
+                                Capacity: new Long(newBucketMD.capabilities.VeeamSOSApi.CapacityInfo.Capacity || '0'),
+                                Available: new Long(newBucketMD.capabilities.VeeamSOSApi.CapacityInfo.Available || '0'),
+                                Used: new Long(newBucketMD.capabilities.VeeamSOSApi.CapacityInfo.Used || '0'),
+                                LastModified: (newBucketMD.capabilities.VeeamSOSApi.CapacityInfo as any).LastModified,
+                            },
+                        },
+                    },
                 },
                 vFormat: this.defaultBucketKeyFormat,
             },
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -5,3 +5,18 @@ export interface ArsenalCallback<T> {
     (err: void): void;
     (err: ArsenalError): void;
 };
+
+type Primitive = string | number | boolean | symbol | null | undefined | bigint;
+
+type StripKey<K extends PropertyKey, P extends PropertyKey> =
+    K extends `${Exclude<P, symbol>}.${infer R}` ? R : never;
+
+export type NestedOmit<T, K extends PropertyKey> =
+    T extends Primitive | Function
+        ? T
+        : T extends Array<infer U>
+            ? Array<NestedOmit<U, K>>
+            : {
+                [P in keyof T as P extends Extract<K, string> ? never : P]:
+                    NestedOmit<T[P], StripKey<K, P>>;
+            };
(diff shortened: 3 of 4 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -411,7 +411,7 @@ class MongoClientInterface {
                 value: {
                     ...newBucketMD,
                     quotaMax: new Long(newBucketMD.quotaMax || '0'),
-                    capabilities: undefined,
+                    capabilities: newBucketMD.capabilities as  CapabilitiesMongoDB|| undefined,
```
ghost:
```suggestion
                    capabilities: newBucketMD.capabilities as CapabilitiesMongoDB || undefined,
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -479,19 +479,31 @@ class MongoClientInterface {
                 if (!doc) {
                     return cb(errors.NoSuchBucket);
                 }
+                log.info('getBucketAttributes: bucket found', {
```
francoisferrand: seems way too complex for logs, logs should only include what we have...
esp. on `getBucketAttributes` (or any metadata getter), i don't think we want to log every time...

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -479,19 +479,31 @@ class MongoClientInterface {
...
+                        CapacityInfo: doc.value.capabilities?.VeeamSOSApi?.CapacityInfo &&
+                            VeeamCapacityInfo.serialize(doc.value.capabilities?.VeeamSOSApi?.CapacityInfo),
+                    }
+                });
                 const bucketMetadata = {
                     ...doc.value,
                     quotaMax: doc.value.quotaMax?.toString() || '0',
                     capabilities: {
-                        ...doc.value.capabilities,
+                        ...(doc.value.capabilities || {}),
```
francoisferrand: what i the point? `...undefined` is exactly the same as `...{}`

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -479,19 +479,31 @@ class MongoClientInterface {
...
+                            ...doc.value.capabilities?.VeeamSOSApi,
                             // Long values are automatically serialized to strings
-                            CapacityInfo: doc.value.capabilities.VeeamSOSApi.CapacityInfo &&
-                                VeeamCapacityInfo.serialize(doc.value.capabilities.VeeamSOSApi.CapacityInfo),
+                            CapacityInfo: doc.value.capabilities?.VeeamSOSApi?.CapacityInfo &&
+                                VeeamCapacityInfo.serialize(doc.value.capabilities?.VeeamSOSApi?.CapacityInfo),
                         },
                     },
                 };
+                log.info('getBucketAttributes: bucket metadata', {
```
francoisferrand: debug log, should be removed

package.json
```diff
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=20"
   },
-  "version": "8.2.13",
+  "version": "8.2.14",
```
francoisferrand: not sure we want to release just yet, we got a few more that also need to reach 9.0 very soon.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -411,7 +411,8 @@ class MongoClientInterface {
                 value: {
                     ...newBucketMD,
                     quotaMax: new Long(newBucketMD.quotaMax || '0'),
-                    capabilities: undefined,
+                    // We're passing the capabilities value here as it's used for testing purposes
+                    capabilities: newBucketMD.capabilities as CapabilitiesMongoDB|| undefined,
```
ghost:
```suggestion
                    capabilities: newBucketMD.capabilities as CapabilitiesMongoDB || undefined,
```

ghost (approved):

package.json
benzekrimaha: we will actually need to bump, for the s3utils PR

francoisferrand: yes we need to bump, but not on every PR. We currently have 3 PRs open in arsenal, all of which need to be bump'ed : so better to merge as many PRs as possible, and release/bump just once, after the last one of the batch.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -411,7 +411,8 @@ class MongoClientInterface {
                 value: {
                     ...newBucketMD,
                     quotaMax: new Long(newBucketMD.quotaMax || '0'),
-                    capabilities: undefined,
+                    // We're passing the capabilities value here as it's used for testing purposes
+                    capabilities: newBucketMD.capabilities as CapabilitiesMongoDB || undefined,
```
francoisferrand: that is very weird:
* `newBucketMD` is the "serializable" variant of `bucketMD`: if typing is needed, it should be done in `BucketInfo.makeSerializable`
* adding the type in here will only affect this specific function, as `payload` is only used here
* Passing `bucketMD.capabilities` to createBucket() is not just for testing: today, maybe we do not do it "in prod" and only do it in tests, but since it is part of bucketInfo it should be supported, like all other fields... → if a fix is needed, it is not just for tests...

benzekrimaha: Actually like for the quotaMax , here we don't really need the makeSerializable to make capabilities typed the way it's needed to be set in mongodb as it's only used here thus in the payload used to create the bucket

lib/models/Veeam.ts
```diff
@@ -72,9 +73,9 @@ export class VeeamCapacityInfo {
 
     static toBigInt(capacity: VeeamSOSApiSchema['CapacityInfo']): VeeamSOSApiSchema['CapacityInfo'] {
```
francoisferrand: the return type is not correct : it now has only `bigint` only

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -107,21 +107,34 @@ export type MongoDBClientInterfaceParameters = {
...
-    VeeamSOSApi?: Omit<VeeamSOSApiSchema, 'CapacityInfo'> & {
-        CapacityInfo?: {
-            Capacity: Long,
-            Available: Long,
-            Used: Long,
-        },
-    },
-}
+type StripKey<K extends PropertyKey, P extends PropertyKey> =
+    K extends `${Exclude<P, symbol>}.${infer R}` ? R : never;
```
francoisferrand: should go to a dedicated file...

lib/models/Veeam.ts
```diff
@@ -26,9 +27,9 @@ export type VeeamSOSApiSchema = {
         LastModified?: string,
     },
     CapacityInfo?: {
-        Capacity: bigint | number,
-        Available: bigint | number,
-        Used: bigint | number,
+        Capacity: bigint | number | Long,
+        Available: bigint | number | Long,
+        Used: bigint | number | Long,
```
francoisferrand: why is it needed?
AFAIK we want to convert to/from Long inside MongoClientInterface : so the type handled throughout the code should not have `Long` anywhere...

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -411,7 +424,18 @@ class MongoClientInterface {
...
+                        ...newBucketMD.capabilities,
+                        VeeamSOSApi: newBucketMD.capabilities.VeeamSOSApi ? {
+                            ...newBucketMD.capabilities.VeeamSOSApi,
+                            CapacityInfo: newBucketMD.capabilities.VeeamSOSApi.CapacityInfo ? {
+                                Capacity: new Long(newBucketMD.capabilities.VeeamSOSApi.CapacityInfo.Capacity || '0'),
+                                Available: new Long(newBucketMD.capabilities.VeeamSOSApi.CapacityInfo.Available || '0'),
+                                Used: new Long(newBucketMD.capabilities.VeeamSOSApi.CapacityInfo.Used || '0'),
+                                LastModified: (newBucketMD.capabilities.VeeamSOSApi.CapacityInfo as any).LastModified,
+                            } : undefined,
+                        } : undefined,
```
francoisferrand: instead of ternary operator, prefer `newBucketMD.capabilities.VeeamSOSApi && { .... }` syntax

```suggestion
                    capabilities: newBucketMD.capabilities && {
                        ...newBucketMD.capabilities,
                        VeeamSOSApi: newBucketMD.capabilities.VeeamSOSApi && {
                            ...newBucketMD.capabilities.VeeamSOSApi,
                            CapacityInfo: newBucketMD.capabilities.VeeamSOSApi.CapacityInfo && {
                                Capacity: new Long(newBucketMD.capabilities.VeeamSOSApi.CapacityInfo.Capacity || '0'),
                                Available: new Long(newBucketMD.capabilities.VeeamSOSApi.CapacityInfo.Available || '0'),
                                Used: new Long(newBucketMD.capabilities.VeeamSOSApi.CapacityInfo.Used || '0'),
                                LastModified: (newBucketMD.capabilities.VeeamSOSApi.CapacityInfo as any).LastModified,
                            },
                        },
                    },
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -411,7 +424,18 @@ class MongoClientInterface {
...
+                    } : undefined,
```
francoisferrand: missing tests for this case

lib/models/Veeam.ts
benzekrimaha: Actually this was done for the VeeamCapacityInfo serialize function here to be able to proceed , the thing is that we're storing this as a long in mongodb then expecting the variable to be typed as `bigint | number` on the function as it's taking the` capacity: VeeamSOSApiSchema['CapacityInfo']` as a param , which will not be the case as we're using the doc.value.x.x which is a long , this is why long was added in order to make the serialization possible. This will not affect the calculous as we're making the variables bigInts later on for calculous

lib/models/Veeam.ts
benzekrimaha: As we also have the lastModified which is not a big int => keeping the VeeamSOSApiSchema['CapacityInfo'] seems fair...

francoisferrand: the point of typing is to allow the compiler to do the work : if we type too broadly, then typing becomes useless, we have no benefit anymore... The whole point of this function is to convert to bigint, so best to specify the actual type...
```suggestion
    static toBigInt(capacity: VeeamSOSApiSchema['CapacityInfo']): { Capacity: BigInt; Available: BigInt; Used: BigInt; LastModified?: string } {
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -19,7 +19,7 @@ import errors, { ArsenalError, errorInstances } from '../../../errors';
 import BucketInfo, { BucketMetadata, Capabilities } from '../../../models/BucketInfo';
 import ObjectMD, { ObjectMDData } from '../../../models/ObjectMD';
 import * as jsutil from '../../../jsutil';
-import { ArsenalCallback } from '../../../types';
+import { ArsenalCallback , NestedOmit } from '../../../types';
```
francoisferrand: typo:
```suggestion
import { ArsenalCallback, NestedOmit } from '../../../types';
```

lib/models/Veeam.ts
francoisferrand: this however means that "user" of Capacity need (according to the compiler) to handle the `Long` type : which is exactly what we don't want, we want the `Long` type to stay in Mongo...

a better way seems to specify `bigint` here only (as it is what we want the code to deal with)
```suggestion
        Used: bigint,
```

and change the type of VeeamCapacityInfo.serialize() to be more permissive, like this:

```
static serialize(capacity: {
        Capacity: bigint | number | Long,
        Available: bigint | number | Long,
        Used: bigint | number | Long,
        LastModified?: string,
    }): VeeamSOSApiSerializable['CapacityInfo'] {
```

francoisferrand (commented):
Also please update the PR with proper title and description

lib/models/Veeam.ts
```diff
@@ -72,9 +78,9 @@ export class VeeamCapacityInfo {
 
     static toBigInt(capacity: VeeamSOSApiSchema['CapacityInfo']): VeeamSOSApiSchema['CapacityInfo'] {
         return {
-            Capacity: BigInt(capacity?.Capacity || 0),
-            Available: BigInt(capacity?.Available || 0),
-            Used: BigInt(capacity?.Used || 0),
+            Capacity: BigInt(capacity?.Capacity.toString() || 0),
+            Available: BigInt(capacity?.Available.toString() || 0),
+            Used: BigInt(capacity?.Used.toString() || 0),
```
francoisferrand: why do we _need_ to convert to string?
- if will actually crash if capacity were null/undefined (`capacity?.`)
- according to the type (and typescript compiler), Capacity/Available/Used can now only be bigint already...

We can keep the `BigInt` call for extra caution, but we should avoid the systematic double conversion to and from string...

francoisferrand (approved):

PR #2373, opened 2025-05-02
Fixes for cloudserver unification

francoisferrand:
- accept lifecycle config only if all rules are supported
- return arsenal error of same type when available
- Fix crash on buckets created with 7.70 branch
- Fix validation of unsupported rules
- Fix compute ContentLength from location

Issue: ARSN-490

### DIFF
--- a/lib/models/LifecycleConfiguration.ts
+++ b/lib/models/LifecycleConfiguration.ts
@@ -94,6 +94,7 @@ export default class LifecycleConfiguration {
     _ruleIDs: string[];
     _tagKeys: string[];
     _storageClasses: string[];
+    _supportedLifecycleRules: string[];
     _config: {
         error?: ArsenalError;
         rules?: any[];
@@ -105,13 +106,23 @@ export default class LifecycleConfiguration {
      * @param config - the CloudServer config
      * @return - LifecycleConfiguration instance
      */
-    constructor(xml: any, config: { replicationEndpoints: { site: string }[] }) {
+    constructor(xml: any, config: { replicationEndpoints: { site: string }[], supportedLifecycleRules: string[] }) {
         this._parsedXML = xml;
         this._storageClasses =
             config.replicationEndpoints.map(endpoint => endpoint.site);
+        this._supportedLifecycleRules = this._buildSupportedLifecycleRules(config.supportedLifecycleRules);
         this._ruleIDs = [];
         this._tagKeys = [];
         this._config = {};
+    }
+
+    _buildSupportedLifecycleRules(ruleNames) {
+        return ruleNames.map(rule => {
+            if (rule === 'noncurrentVersionTransition') {
+                return 'NoncurrentVersionTransitions';
+            }
+            return rule.charAt(0).toUpperCase() + rule.slice(1);
+        });
     }
 
     /**
@@ -229,6 +240,13 @@ export default class LifecycleConfiguration {
      * }
      */
     _parseRule(rule: XMLRule) {
+        const isRuleSupported = Object.keys(rule)
+            .some(key => this._supportedLifecycleRules.includes(key));
+        if (!isRuleSupported) {
+            const msg = 'lifecycle action not implemented';
+            const error = errorInstances.NotImplemented.customizeDescription(msg);
+            return { error };
+        }
         // Either Prefix or Filter must be included, but can be empty string
         if ((!rule.Filter && rule.Filter !== '') &&
         (!rule.Prefix && rule.Prefix !== '')) {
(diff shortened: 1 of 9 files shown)


lib/models/LifecycleConfiguration.ts
```diff
@@ -229,6 +240,13 @@ export default class LifecycleConfiguration {
      * }
      */
     _parseRule(rule: XMLRule) {
+        const isRuleSupported = Object.keys(rule)
+            .some(key => this._supportedLifecycleRules.includes(key));
+        if (!isRuleSupported) {
+            const msg = 'lifecycle action not implemented';
+            const error = errorInstances.NotImplemented.customizeDescription(msg);
+            return { error };
```
ghost:
```suggestion
            return {
                error: errorInstances.NotImplemented
                    .customizeDescription('lifecycle action not implemented');
            };
```

lib/s3routes/routesUtils.ts
```diff
@@ -22,6 +22,22 @@ export type CallApiMethod = (
     callback: (err: ArsenalError | Error | null, ...data: any[]) => void,
 ) => void;
 
+/**
+ * Convert an error to an ArsenalError instance
+ * @param err - Error to convert
+ * @returns ArsenalError instance
+ */
+export function toArsenalError(err: ArsenalError | Error): ArsenalError {
```
ghost, line 30: Maybe a candidate for `lib/errorUtils.ts`? Or a static member of `ArsenalError`?

lib/models/LifecycleConfiguration.ts
```diff
@@ -89,11 +88,24 @@ export type LifecycleConfigurationMetadata = {
     rules: Rule[],
 };
 
+export type XMLRule = {
```
ghost: To differentiate between the XMLRule for replication, shall we use a more specific name?

```suggestion
export type XMLLifecycleRule = {
```


... and change the replication rule

tests/unit/models/LifecycleConfiguration.spec.js
```diff
@@ -1289,3 +1347,55 @@ describe('LifecycleConfiguration::getConfigJson', () => {
...
+        assert.notStrictEqual(result1, result2); // Should be different instances
+    });
+
+    it('should handle mixed rule names including special cases', () => {
+        const names = ['expiration', 'noncurrentVersionTransition', 'transition'];
+        const result = LifecycleConfiguration._buildSupportedLifecycleRules(names);
+        assert.deepStrictEqual(result, [
+            'Expiration',
+            'NoncurrentVersionTransitions',
+            'Transition'
```
ghost:
```suggestion
            'Transition',
```

ghost (approved):
Some minor comments, not blocking

lib/models/LifecycleConfiguration.ts
```diff
@@ -917,11 +937,20 @@ export default class LifecycleConfiguration {
...
-        });
+
+            const isRuleSupported = this._supportedLifecycleRules.includes(action);
+            if (!isRuleSupported) {
+                const msg = `${action} lifecycle action not implemented`;
+                const error = errorInstances.NotImplemented.customizeDescription(msg);
+                return { ...actionsObj, error };
+            }
+
+            actionsObj.actions.push({ actionName: `${action}` });
```
jonathan-gramain: Do we use `${...}` to copy the string? I don't think it's needed since we don't modify the contents of `action`, and strings are immutable anyways.

jonathan-gramain (approved):

lib/models/LifecycleConfiguration.ts
```diff
@@ -105,15 +117,42 @@ export default class LifecycleConfiguration {
...
             config.replicationEndpoints.map(endpoint => endpoint.site);
+        this._supportedLifecycleRules = LifecycleConfiguration._buildSupportedLifecycleRules(
+            config.supportedLifecycleRules,
+        );
         this._ruleIDs = [];
         this._tagKeys = [];
         this._config = {};
     }
 
+    // Memoize the supported lifecycle rules to avoid recalculating them
```
benzekrimaha, line 132:
```suggestion
    // Memorize the supported lifecycle rules to avoid recalculating them
```

benzekrimaha (approved):

lib/models/LifecycleConfiguration.ts
francoisferrand: No typo, [Memoization](https://en.wikipedia.org/wiki/Memoization) is the right term here

lib/s3routes/routesUtils.ts
francoisferrand: may be generalized as part of the (future, heopfully soon) error rework effort; but at the moment moving it creates more quesitons (should it support other types, possibly using `ArsenalError.deserialize`...)

PR #2387, opened 2025-05-15
Ensure we detect properly bulk error when creating versions

ghost:
## this was tested successfully with the script that reproduces the problem - no more 502/503 nor InternalErrors returned.

We used to rely on a non-working detection algorithm that always
detected that the versionID operation failed, even if it was
the master version. As a result, the retry would recreate the
versionID a second time, and lead to more race conditions and
invalid states.

Now, we properly detect the index of the failing operation.
For `version`: we retry. For `master`: we ignore, it will be
repaired.

Issue: ARSN-491


lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -859,45 +871,71 @@ class MongoClientInterface {
...
-                 * We check here than than the MongoDB error is related to the
-                 * second operation, the master version update and than the error
-                 * code is the one related to mentionned issue.
                  */
                 if (err.code === 11000) {
                     log.debug('putObjectVerCase1: error putting object version', {
                         code: err.code,
                         error: err.errmsg,
-                        isRetry: isRetry ? true : false, // eslint-disable-line no-unneeded-ternary
+                        isRetry: isRetry ? true : false,
```
francoisferrand: nit: the ternary is only needed (is it really?) for the case where isRetry is undefined [based on the type], so we may do this instead
```suggestion
                        isRetry: isRetry || false,
```
or if we want to be even more resilient:
```suggestion
                        isRetry: !!isRetry,
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -859,45 +871,71 @@ class MongoClientInterface {
...
-                    if (typeof count === 'number' && count !== 1) {
-                        // This may be a race condition, when two different S3 Connector try to put the same
-                        // version id
-                        if (!isRetry) {
-                            // retrying with a new version id
-                            return process.nextTick(() =>
-                                this.putObjectVerCase1(c, bucketName, objName, objVal, params, log, cb, true));
+                    
+                    const writeErrors = err.writeErrors || [];
+                    const firstError: WriteError = writeErrors[0];
```
francoisferrand: are errors guaranteed to be "ordered" ?
i.e. could we have instead `writeErrors[1].index === 0` and `writeErrors[0].index === 1` ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -859,45 +871,71 @@ class MongoClientInterface {
...
+                        
+                        // Master operation failed but version succeeded - this is OK
+                        if (isMasterOpError) {
+                            return cb(null, `{"versionId": "${versionId}"}`);
+                        }
+                    }
+                    
+                    // If we can't determine which operation failed from writeErrors,
+                    // fall back to checking result counts
+                    const upsertedCount = err.result?.upsertedCount || err.result?.nUpserted || 0;
```
francoisferrand: can this really happen, or is it just an happen to cover all possible "code path" ?
is there any information in mongodb regarding the possible outcomes of batch operation?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -859,45 +871,71 @@ class MongoClientInterface {
...
+                            this.putObjectVerCase1(c, bucketName, objName, objVal, params, log, cb, true));
                     }
-                    // Otherwise this error is expected, it means that two differents version was put at the
-                    // same time
-                    return cb(null, `{"versionId": "${versionId}"}`);
+                    
+                    log.error('putObjectVerCase1: race condition upserting versionId', {
+                        error: err.errmsg,
+                    });
+                    return cb(errors.InternalError);
```
francoisferrand, line 926: overal logic sounds ok, but the code feels too complicated, with nested conditions and duplicated "exits"...
can we rework it in a simpler way, maybe something like:

```
   if (write_error_on_master || (no_write_error && upsertedCount === 1)) {
      // error on master → success
      return cb(null, `{"versionId": "${versionId}"}`);
   }
   if (!isRetry) {
      return nextTick(...);
   }
   return internalError;
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -859,45 +871,71 @@ class MongoClientInterface {
...
+                            return cb(null, `{"versionId": "${versionId}"}`);
+                        }
+                    }
+                    
+                    // If we can't determine which operation failed from writeErrors,
+                    // fall back to checking result counts
+                    const upsertedCount = err.result?.upsertedCount || err.result?.nUpserted || 0;
+                    
+                    // If we successfully upserted one document (the version), but had an error on master,
+                    // consider this a success
```
francoisferrand: I think the original comment gave a clearer context:
```
Otherwise this error is expected, it means that two differents version was put at the same time
```

It is not just that we "consider" it a success, it is really that we expect this to happen (and it's fine) on simultaneous version put : so may be better to somewhat "merge" both comments

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -859,45 +871,71 @@ class MongoClientInterface {
             ordered: true,
         })
             .then(() => cb(null, `{"versionId": "${versionId}"}`))
-            .catch((err) => {
+            .catch((err: MongoBulkWriteError) => {
                 /*
                  * Related to https://jira.mongodb.org/browse/SERVER-14322
                  * It happens when we are pushing two versions "at the same time"
                  * and the master one does not exist. In MongoDB, two threads are
```
francoisferrand, line 880: > and the master one does not exist

- can this only happen when the master does not exist? (i.e. trying to "create" the object twice at the same time)
- don't we have a similar issue (maybe not in this "case") when updating the master with 2 new revisions pushed at the same time? Can we add tests as for this case as well, if not already covered?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -859,45 +871,71 @@ class MongoClientInterface {
...
                  * trying to create the same key, the master version, and one of
                  * them, the one with the highest versionID (less recent one),
                  * fails.
-                 * We check here than than the MongoDB error is related to the
-                 * second operation, the master version update and than the error
-                 * code is the one related to mentionned issue.
```
francoisferrand, line 872: this comment looks somewhat good (except for typos)?
I think it would be good to keep such very high-level overview (i.e. we ignore error on master document, if the version document was successfully written)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: No, that's why we rely on the `index` property. the "first" here meant the `[0]` index of the array: if defined, we have at least one error...
I'll rename it better.
In any case we cannot have two errors, only one, as we are doing thing with ordering (see https://www.mongodb.com/docs/manual/core/bulk-write-operations/#ordered-vs-unordered-operations).

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: Logcally it could happen but the documentation (see https://www.mongodb.com/docs/manual/reference/method/db.collection.bulkWrite/#std-label-bulkwrite-error-handling-txn) does not really state it's possible. I looked at the code and `writeErrors` is a `OneOrMore`, meaning if we enter the catch, we are sure to have at least one element... I can remove this logic to simplify.
I don't think a write could silently fail...

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: The discussed mongodb defect was already fixed in MongoDB v5.0.
But otherwise yes, any write conflict will create this error. In our case it should not be possible, given that versionIDs are properly implemented.

> when updating the master with 2 new revisions pushed at the same time

If by revision you mean updating the latest version ID concurrently, then yes, it's done in https://github.com/scality/Arsenal/pull/2387/files#diff-743cf32ff59c21d31aec6a69642b8af09c99bed0329c5b5a8772a6bd440e4e6fR3149, where we "ignore" the error here, because it will be repaired

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: I adapted it a bit, because it was really discussing a mongodb defect already fixed. The new version will present the risk of version ID conflict and the two cases: we can fail either at the first, or the second operation, for different reasons.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: Note: I first kept this check "just in case" the isMasterOpError is not true but we really inserted something. But I wonder if that's wise, because then if there was no master version and the operation failed, we end up passing this check and returning a version ID, even if the master was not created, leading to an inconsistent state (if no master, then we cannot list/get the versionID...). So I removed it completely.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: Logic was simplified.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
francoisferrand: would be worth pointing out (in the comments) that we can only have a single error, as the way we access it (including index) hints that there could be more...

welansari (approved):

tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
```diff
@@ -3051,3 +3051,207 @@ describe('MongoClientInterface, deleteBucketIndexes', () => {
...
+                    assert.strictEqual(bulkWriteStub.callCount, 2, 'Expected bulkWrite to be called twice');
+
+                    // Verify putObjectVerCase1 was called with isRetry=true for the retry
+                    assert(putObjectVerCase1Spy.calledTwice, 'Expected putObjectVerCase1 to be called twice');
+                    assert(putObjectVerCase1Spy.secondCall.args[7], 'Expected second call to have isRetry=true');
+
+                    done();
+                } catch (assertionError) {
+                    done(assertionError);
+                }
```
welansari: nit: do we really need to catch the assertion error ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: I added a comment.

tests/unit/storage/metadata/mongoclient/putObject.spec.js
```diff
@@ -221,28 +221,12 @@ describe('MongoClientInterface:putObjectVerCase1', () => {
         }, false);
     });
 
-    it('should not fail when error code is 11000 and upsertedCount is 1', done => {
+    it('should fail when error code is 11000, master operation failed', done => {
```
francoisferrand: - `index === 0` means the "version" document failed, not master 
- should we add a test case here for the "master operation failed" case indeed (which should not fail the operation)

tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
```diff
@@ -3051,3 +3051,195 @@ describe('MongoClientInterface, deleteBucketIndexes', () => {
         });
     });
 });
+
+describe('MongoClientInterface, putObjectVerCase1', () => {
```
francoisferrand: should these not go in `tests/unit/storage/metadata/mongoclient/putObject.spec.js`, which already has a `MongoClientInterface:putObjectVerCase1` test suite?

tests/unit/storage/metadata/mongoclient/putObject.spec.js
ghost: The tests were updated, i also merged the two similar test suites into one, and it covers all cases, in the right file. I think we should either split the MongoClientInterface accordingly, or use a single test file though, as it doesn't follow the usuql pattern to have different names for subparts of one (big) ts file...

francoisferrand (approved):

PR #2397, opened 2025-05-19
ARSN-495: KMIP retries

BourgoisMickael:
- Retry internally for thales session token revoked
- Blacklist unhealthy hosts (hosts that return error) and add retry mechanism in kmip cluster

### DIFF
--- a/lib/network/kmip/Client.ts
+++ b/lib/network/kmip/Client.ts
@@ -339,7 +339,8 @@ export default class Client implements KMSInterface {
                 kmipMsg(operation, keyIdentifier,
                     'Server did not return the expected identifier')
             );
-            logger.error(`KMIP::${operation}`,
+            // warn level to avoid dumping debug and trace logs on retryable errors
+            logger.warn(`KMIP::${operation}`,
                 { error, uniqueIdentifier, keyIdentifier });
             return error;
         }
@@ -360,7 +361,8 @@ export default class Client implements KMSInterface {
             KMIP.TextString('Unique Identifier', keyIdentifier),
         ], (error, response) => {
             if (error) {
-                logger.error('KMIP::_activateBucketKey',
+                // warn level to avoid dumping debug and trace logs on retryable errors
+                logger.warn('KMIP::_activateBucketKey',
                     { error,
                         serverInformation: this.serverInformation });
                 return cb(error);
@@ -404,7 +406,8 @@ export default class Client implements KMSInterface {
             KMIP.Structure('Template-Attribute', attributes),
         ], (error, response) => {
             if (error) {
-                logger.error('KMIP::createBucketKey',
+                // warn level to avoid dumping debug and trace logs on retryable errors
+                logger.warn('KMIP::createBucketKey',
                     { error,
                         serverInformation: this.serverInformation });
                 return cb(error);
@@ -419,7 +422,8 @@ export default class Client implements KMSInterface {
                     kmipMsg('Create', bucketName,
                         'Server created an object of wrong type')
                 );
-                logger.error('KMIP::createBucketKey',
+                // warn level to avoid dumping debug and trace logs on retryable errors
+                logger.warn('KMIP::createBucketKey',
                     { error, createdObjectType });
                 return cb(error);
             }
@@ -451,7 +455,8 @@ export default class Client implements KMSInterface {
             ]),
         ], (error, response) => {
             if (error) {
-                logger.error('KMIP::_revokeBucketKey',
+                // warn level to avoid dumping debug and trace logs on retryable errors
+                logger.warn('KMIP::_revokeBucketKey',
                     { error,
                         serverInformation: this.serverInformation });
                 return cb(error);
@@ -472,7 +477,8 @@ export default class Client implements KMSInterface {
         return this._revokeBucketKey(bucketKeyId, logger, err => {
             if (err) {
                 const error = arsenalErrorKMIP(err);
-                logger.error('KMIP::destroyBucketKey: revocation failed',
+                // warn level to avoid dumping debug and trace logs on retryable errors
+                logger.warn('KMIP::destroyBucketKey: revocation failed',
                     { error,
                         serverInformation: this.serverInformation });
                 return cb(error);
@@ -482,7 +488,8 @@ export default class Client implements KMSInterface {
             ], (err, response) => {
                 if (err) {
                     const error = arsenalErrorKMIP(err);
-                    logger.error('KMIP::destroyBucketKey',
+                    // warn level to avoid dumping debug and trace logs on retryable errors
+                    logger.warn('KMIP::destroyBucketKey',
                         { error,
                             serverInformation: this.serverInformation });
                     return cb(error);
@@ -524,7 +531,8 @@ export default class Client implements KMSInterface {
             KMIP.ByteString('IV/Counter/Nonce', CRYPTOGRAPHIC_DEFAULT_IV),
         ], (error, response) => {
             if (error) {
-                logger.error('KMIP::cipherDataKey',
+                // warn level to avoid dumping debug and trace logs on retryable errors
+                logger.warn('KMIP::cipherDataKey',
                     { error,
                         serverInformation: this.serverInformation });
                 return cb(error);
@@ -566,7 +574,8 @@ export default class Client implements KMSInterface {
             KMIP.ByteString('IV/Counter/Nonce', CRYPTOGRAPHIC_DEFAULT_IV),
         ], (error, response) => {
             if (error) {
-                logger.error('KMIP::decipherDataKey',
+                // warn level to avoid dumping debug and trace logs on retryable errors
+                logger.warn('KMIP::decipherDataKey',
                     { error,
                         serverInformation: this.serverInformation });
                 return cb(error);
@@ -582,7 +591,7 @@ export default class Client implements KMSInterface {
         // the bucket does not have to exist, just passing a common bucket name here
         this.createBucketKey('kmip-healthcheck-test-bucket', logger, (err, bucketKeyId) => {
             if (err) {
-                logger.error('KMIP::healthcheck: failure to create a test bucket key', {
+                logger.warn('KMIP::healthcheck: failure to create a test bucket key', {
                     error: err, kmip: kmipLog,
                 });
                 return cb(err);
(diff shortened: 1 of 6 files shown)


dvasilas (approved):

lib/network/kmip/error-comparison.jsonc
```diff
@@ -34,7 +34,9 @@
                 "resultMessage": [
                     "[NCERRInsufficientPermissions]: ",
                     // TLS session idle timeout (24h) - needs a new connection or no idle
-                    "[NCERRUnauthorizedAccess]: Token has been revoked"
+                    "[NCERRUnauthorizedAccess]: Token has been revoked",
+                    // write after receiving above token revoked error
+                    "[NCERRUnauthorizedAccess]: Invalid token"
```
anurag4DSB, line 41: for my own edification and as this is different from S3, worth commenting documentation link for these errors

lib/network/kmip/index.ts
```diff
@@ -28,6 +28,23 @@ function _PrimitiveType(tagName: string, type: string, value: any) {
     return { [tagName]: { type, value } };
 }
 
+/** Thales has a 24h idle tls session error. We can close and reopen new connection */
```
anurag4DSB: worth commenting the Thales version so future devs know if something has changed and documentation link

lib/network/kmip/index.ts
```diff
@@ -28,6 +28,23 @@ function _PrimitiveType(tagName: string, type: string, value: any) {
...
+const thalesInternalTokenError = {
+    status: 'Operation Failed',
+    reason: 'General Failure',
+    messages: [
+        '[NCERRUnauthorizedAccess]: Token has been revoked',
+        '[NCERRUnauthorizedAccess]: Invalid token'
+    ],
+};
+
+//Could test on the kmip conf providerName as well, Should test other providers too
```
anurag4DSB:
```suggestion
// Could test on the kmip conf providerName as well, should test other providers too
```

anurag4DSB: whats the goal of this comment?

lib/network/kmip/Client.ts
```diff
@@ -339,7 +339,8 @@ export default class Client implements KMSInterface {
                 kmipMsg(operation, keyIdentifier,
                     'Server did not return the expected identifier')
             );
-            logger.error(`KMIP::${operation}`,
+            // warn level to avoid dumping debug and trace logs on retryable errors
```
anurag4DSB, line 342: so do we not consider these as errors, seems like an unhealthy server and valid error even if we can retry.
I am not asking to change this back, but trying to understand more as this is not just about debug and trace logs but also about the error log file.

lib/network/kmip/ClusterClient.ts
```diff
@@ -56,25 +90,168 @@ export default class ClusterClient implements KMSInterface {
...
     }
 
+    checkUnhealthyClient(unhealthyClient: UnhealthyClient) {
+        const client = unhealthyClient.client;
+        const healthyIndex = this.clients.indexOf(client);
+        if (healthyIndex === -1) {
+            this.clients.push(client);
+        } else {
+            // should not happen
+            this.logger.warn('checkUnhealthyClient: already moved in healthy', { unhealthyHost: client.host });
```
anurag4DSB, line 100: What does mean, "already moved in healthy"

lib/network/kmip/ClusterClient.ts
```diff
@@ -56,25 +90,168 @@ export default class ClusterClient implements KMSInterface {
         return this.clients[this.roundRobinIndex++];
     }
 
+    checkUnhealthyClient(unhealthyClient: UnhealthyClient) {
+        const client = unhealthyClient.client;
+        const healthyIndex = this.clients.indexOf(client);
+        if (healthyIndex === -1) {
+            this.clients.push(client);
+        } else {
+            // should not happen
```
anurag4DSB: If this should not happen, why do we have a condition for this?

lib/network/kmip/ClusterClient.ts
```diff
@@ -56,25 +90,168 @@ export default class ClusterClient implements KMSInterface {
...
+            this.logger.warn('checkUnhealthyClient: already moved in healthy', { unhealthyHost: client.host });
+        }
+
+        const unhealthyIndex = this.unhealthyClients.indexOf(unhealthyClient);
+        if (unhealthyIndex === -1) {
+            // should not happen
```
anurag4DSB: Same for this, may be I am understanding "should not happen" wrong.

anurag4DSB (commented):
Should we add a blacklist mechanism so for an unhealthy node further requests are not forwarded.

lib/network/kmip/error-comparison.jsonc
BourgoisMickael: no documentation link, it's only something we've been told (that's why we kept an healthcheck running regularly to avoid a 24h session revoked in prod).

I had to ask thales to tweak my thales lab to reduce the session idle timeout to 10m to be able to test this

lib/network/kmip/index.ts
BourgoisMickael: no documentation about this, that's their internal, it's something they told us, it's not written and not part of KMIP specs.

lib/network/kmip/index.ts
BourgoisMickael: This is a thales specific behavior that we discovered in prod when some clients had some problems.

But we also support hashicorp or hytrust for kmip providers and I have no idea if they have such a thing as well. So it was more like a todo but I can drop the comment from the code

lib/network/kmip/error-comparison.jsonc
anurag4DSB: told by whom
We should add that context to the code(without customer name ofcourse)

lib/network/kmip/index.ts
anurag4DSB: We should add this context to the comments(without customer name)

lib/network/kmip/index.ts
anurag4DSB: If its todo, a ticket(product ie. Jira S3C project) should be linked to this.

lib/network/kmip/Client.ts
BourgoisMickael: When testing I see it dumps unrelated logs in debug and trace (mostly related to auth) and we have no debug or traces related to KMIP. The log can also be self contained, it has the operation and keys that differ.

And in cluster mode, with retry this become even less critical as another host can succeed on the request.

So it helps avoid many other unrelated logs, even for a simple non HA cluster kmip, we don't need to bump the debug logs.

I see we almost never use this `warn` level.

lib/network/kmip/ClusterClient.ts
BourgoisMickael: the unhealthy client has already been pushed into the list of healthy clients that can be used in the round robin to send KMIP requests

lib/network/kmip/Client.ts
anurag4DSB: yes we do not use warn as it adds to normal s3 logs and we want to keep normal logs as lean as possible. If its an error it should go to error logs.

lib/network/kmip/ClusterClient.ts
BourgoisMickael: There are 2 list of clients.

The healthy ones `this.clients` and the unhealthy `this.unhealthyClients`.

I move the clients between those 2 arrays depending on their state (here a failing request makes the client unhealthy).

The movement should be synchronous between the 2 lists: leave one list, join the other.

There should be no case where a client is in the 2 lists or 0 lists.

This is more a defensive programing where if this condition ever happen I don't duplicate the client in a list (as it might put more weight on a host in the round robin) and I get a log to warn about something that should not happen.

lib/network/kmip/Client.ts
BourgoisMickael: but those error logs go in the same log file

lib/network/kmip/ClusterClient.ts
anurag4DSB: so why do we need this?

anurag4DSB: Okay got my answer in the next comment,.

lib/network/kmip/ClusterClient.ts
anurag4DSB: Okay, wondering why we need defensive programming here, if it will never happen.
Do you see cases where it might happen(for my understanding). I am okay with this generally.

lib/network/kmip/Client.ts
anurag4DSB: [Discussed over slack](https://scality.slack.com/archives/C01N31LU4HW/p1747748084977989?thread_ts=1747736902.944369&cid=C01N31LU4HW)

anurag4DSB (approved):
Approving but some comments improvement is needed for context.

lib/network/kmip/error-comparison.jsonc
BourgoisMickael: idk by thales when we saw that error

lib/network/kmip/ClusterClient.ts
BourgoisMickael: This is a basic check, we want later to improve it with actual healthcheck running to find healthy / unhealthy hosts.

So for any modifications later, and avoid any potential error later.

As I was waiting to test the full integration, but I'll actually add functional tests here in arsenal for kmip cluster

lib/network/kmip/error-comparison.jsonc
anurag4DSB: Let's add that context in comments

lib/network/kmip/ClusterClient.ts
anurag4DSB: Let's add the ticket number for the"later" part so people have context from comments

lib/network/kmip/index.ts
BourgoisMickael: Added the doc about thales: [89a4a1d](https://github.com/scality/Arsenal/pull/2397/commits/89a4a1d801a0be07eea2188f0d13909a8f7d14ba)

lib/network/kmip/ClusterClient.ts
BourgoisMickael: Added a ticket number [a59c1cd](https://github.com/scality/Arsenal/pull/2397/commits/a59c1cdd234430eaa685563c94c11127fd109dd3)

PR #2398, opened 2025-05-19
Fix handling of supported lifeycle type

francoisferrand:
- Fix typing of lifecycle rule/config
- Fix handling of supportedRules: properly convert between config & fields
- Dedup & simplify supported lifecycle rules: use S3 names in the config

Issue: ARSN-496


lib/models/LifecycleRule.ts
```diff
@@ -14,6 +15,31 @@ export type NoncurrentExpiration = {
...
+export type AbortIncompleteMultipartUpload = {
+    DaysAfterInitiation: number;
+};
+export type Transition = {
+    Date: number;
+    Days: number;
+    StorageClass: string;
+};
+export type NoncurrentVersionTransition = {
+    NoncurrentDays: number | null;
```
ghost:
```suggestion
    DaysAfterInitiation: number;
};

export type Transition = {
    Date: number;
    Days: number;
    StorageClass: string;
};

export type NoncurrentVersionTransition = {
    NoncurrentDays: number | null;
```

lib/models/LifecycleRule.ts
```diff
@@ -14,6 +15,31 @@ export type NoncurrentExpiration = {
...
+export type LifecycleRuleData = {
+    ID: string;
+    Status: Status;
+    Expiration?: Expiration;
+    NoncurrentVersionExpiration?: NoncurrentExpiration;
+    AbortIncompleteMultipartUpload?: AbortIncompleteMultipartUpload;
+    Transitions?: Transition[];
+    NoncurrentVersionTransitions?: NoncurrentVersionTransition[];
+    Filter?: Filter;
+    Prefix?: '';
```
ghost, line 45: We only support empty strings as prefix?

lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
```diff
@@ -244,11 +245,11 @@ export default class LifecycleUtils {
     *  i.e. { Expiration: { Date: <DateObject>, Days: 10 },
     *         NoncurrentVersionExpiration: { NoncurrentDays: 5 } }
     */
-    getApplicableRules(rules: any[], metadata: any) {
+    getApplicableRules(rules: LifecycleRuleData[], metadata: any) {
         // Declare the current date before the reducing function so that all
         // rule comparisons use the same date.
         const currentDate = new Date();
-        const applicableRules = rules.reduce((store, rule) => {
+        const applicableRules = rules.reduce((store: any, rule) => {
```
ghost: Would this type work?

```ts
type Store = {
    Expiration?: {
        ID?: string;
        Days?: number;
        Date?: string;
        ExpiredObjectDeleteMarker?: boolean;
    };
    NoncurrentVersionExpiration?: {
        ID?: string;
        NoncurrentDays?: number;
        NewerNoncurrentVersions?: number;
    };
    AbortIncompleteMultipartUpload?: {
        ID?: string;
        DaysAfterInitiation?: number;
    };
    Transition?: {
        Days?: number;
        Date?: string;
        StorageClass: string;
    };
    NoncurrentVersionTransition?: {
        NoncurrentDays: number;
        StorageClass: string;
    };
}
```

lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
```diff
@@ -305,21 +308,19 @@ export default class LifecycleUtils {
                     store[aimu][dai] = rule[aimu][dai];
                 }
             }
-            const hasTransitions = Array.isArray(rule.Transitions) && rule.Transitions.length > 0;
-            if (hasTransitions && this._supportedRules.includes('transitions')) {
+            const transitions = rule.Transitions;
+            if (transitions && transitions.length > 0 && this._supportedRules.includes('transitions')) {
```
ghost:
```suggestion
            if (transitions.length > 0 && this._supportedRules.includes('transitions')) {
```

it should be enough to test that it's a truthy falue and > 0

lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
```diff
@@ -305,21 +308,19 @@ export default class LifecycleUtils {
...
                     store,
                     currentDate,
                 });
             }
-
-            const ncvt = 'NoncurrentVersionTransitions';
-            const hasNoncurrentVersionTransitions = Array.isArray(rule[ncvt]) && rule[ncvt].length > 0;
-            if (hasNoncurrentVersionTransitions && this._supportedRules.includes('noncurrentVersionTransition')) {
+            const ncvt = rule.NoncurrentVersionTransitions;
+            if (ncvt && ncvt.length > 0 && this._supportedRules.includes('noncurrentVersionTransition')) {
```
ghost: same

lib/models/LifecycleRule.ts
francoisferrand: this was already present, the type was just move out of the `build()` function.

overall, this type (and class) is just for rendering a LifeycycleConfiguration to json, it does reflect what we support (or not). But looking at the code, yes : we will try hard to fill the `Filter` param instead, and use `Prefix` with an empty string as fallback when we can find any filter...

lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
francoisferrand: this works, and can be simplified using the already introduced Expiration/Transition/... types.
However, this fulls quite a bit more changes, as functions called were not properly typed.

ghost (approved):

benzekrimaha (approved):

PR #2399, opened 2025-05-21
Fix serverSideEncryption handling in copyObject

francoisferrand:
`serverSideEncryption` was changed from null to undefined in BucketInfo
in [1] so we may get undefined instead of null.

[1] https://github.com/scality/Arsenal/pull/2299

Issue: ARSN-497


ghost (approved):

ghost (commented):
Thank you for taking care of this change! We should add a test, as with it it wouldn't have been missed...

benzekrimaha (approved):

PR #2402, opened 2025-05-27
ARSN-498: Test KMIP retry and cluster with pykmip

BourgoisMickael:
You can review by commit

Picked pykmip from cloudserver, updated it and test kmip with pykmip


lib/network/kmip/transport/TransportTemplate.ts
```diff
@@ -221,11 +223,16 @@ export default class TransportTemplate {
      * Gracefuly interrupt the conversation. If the caller keeps sending
      * message after calling this function, the conversation won't
      * converge to its end.
+     * @param onEndCallback - callback called once the socket is ended and pipelines are drained
      */
-    end() {
+    end(onEndCallback?: any) {
```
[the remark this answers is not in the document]
BourgoisMickael: Yes it is actually called...

.github/workflows/tests-kmip.yaml
```diff
@@ -0,0 +1,49 @@
...
+on:
+  push:
+    branches-ignore:
+    - 'development/**'
+    paths:
+    - .github/workflows/tests-kmip.yaml
+    - .github/docker-compose.yaml
+    - .github/pykmip/**
+    - lib/network/kmip/**
+    - tests/functional/kmip/**
```
tcarmet: `paths` does not work as you would expect with `push` when working on the feature branches, and does not work well with gitwaterflow integration branches as well (`w/**`).

As a quick fix I can recommend you to use a combination of `push` and `pull_request` trigger, even though it won't be perfect, it will work better:

```suggestion
  pull_request:
    branches:
    - development/*
    - hotfix/*
    paths:
    - .github/workflows/tests-kmip.yaml
    - .github/docker-compose.yaml
    - .github/pykmip/**
    - lib/network/kmip/**
    - tests/functional/kmip/**
  push:
    branches:
    - q/*
    - w/**
    paths:
    - .github/workflows/tests-kmip.yaml
    - .github/docker-compose.yaml
    - .github/pykmip/**
    - lib/network/kmip/**
    - tests/functional/kmip/**
```

For a more accurate way I recommend you checking out the use of https://github.com/tj-actions/changed-files/, we do it in Artesca for the documentation build, if you want to see a real life example of it. 

But honestly, if it doesn't take too long, I would just leave it. I see no issue in having it around:

```suggestion
  push:
    branches-ignore:
    - 'development/**'
```

.github/docker-compose.yaml
```diff
@@ -0,0 +1,13 @@
+services:
+  pykmip:
+    network_mode: "host"
+    profiles: ['pykmip']
+    #image: ${PYKMIP_IMAGE:-ghcr.io/scality/cloudserver/pykmip}
```
tcarmet:
```suggestion
```

.github/pykmip/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+        libressl && \
+    apk add --no-cache --virtual .build-deps \
+        python3-dev \
+        libffi-dev \
+        libressl-dev \
+        sqlite-dev \
+        build-base \
+        curl \
+        git && \
+    git clone https://github.com/openkmip/pykmip.git && \
```
tcarmet, line 16: Was gonna say to specify a version here, but I see that we havent got releases in a while and there has been some modifications to the repo a while ago. Is that why you're not specifying a tag here?

BourgoisMickael: yep I want the latest changes of the repo to include any latest fix as it almost never receive any updates.

But maybe I should lock on the commit hash?

tcarmet: Yeah, all things considered (the repo not having much updates and all) either is fine I think. Mostly wanted to make sure this was on purpose.

tests/functional/kmip/highlevel.spec.js
```diff
@@ -120,4 +122,130 @@ describe('KMIP High Level Driver', () => {
...
+        const kmsKey = await util.promisify(kmipClient.createBucketKey.bind(kmipClient))('mybucket', logger);
+
+        const encrypted = await util.promisify(kmipClient.cipherDataKey.bind(kmipClient))(
+            1, kmsKey, Buffer.from('abc'), logger);
+
+        const reason = 'General Failure';
+        const message = '[NCERRUnauthorizedAccess]: Token has been revoked';
+
+        // the next call should return a thales session revoked error only once
+        const stub = sinon.stub(kmipClient.kmip.transport.channel, '_transform')
```
anurag4DSB: I see that there is no cleanup, I'd recommend cleaning up spies/stubs via `afterEach()` to reduce test flakiness if extended or parallelized.
Something like

```
afterEach(() => {
  sinon.restore();
});
```

tests/functional/kmip/highlevel.spec.js
```diff
@@ -120,4 +122,130 @@ describe('KMIP High Level Driver', () => {
...
+            `tls connection should have been ended once instead of ${tlsEndSpy.callCount}`);
+
+        assert.strictEqual(tlsConnectSpy.callCount, 1,
+            `tls connect should have been called once instead of ${tlsConnectSpy.callCount}`);
+
+        assert.strictEqual(tlsWriteSpy.callCount, 2,
+            `tls write should have been called twice because of retry instead of ${tlsWriteSpy.callCount}`);
+        assert.strictEqual(Buffer.compare(tlsWriteSpy.getCall(0).args[0], tlsWriteSpy.getCall(1).args[0]), 0,
+            'tls write should have twice the same argument');
+        assert.strictEqual(stub.callCount, 2,
```
anurag4DSB: Do we plan to increase this count in the future?

tests/functional/kmip/highlevel.spec.js
```diff
@@ -120,4 +122,130 @@ describe('KMIP High Level Driver', () => {
...
+
+        assert.strictEqual(tlsEndSpy.callCount, 1,
+            `tls connection should have been ended once instead of ${tlsEndSpy.callCount}`);
+
+        assert.strictEqual(tlsConnectSpy.callCount, 1,
+            `tls connect should have been called once instead of ${tlsConnectSpy.callCount}`);
+
+        assert.strictEqual(tlsWriteSpy.callCount, 2,
+            `tls write should have been called twice because of retry instead of ${tlsWriteSpy.callCount}`);
+        assert.strictEqual(Buffer.compare(tlsWriteSpy.getCall(0).args[0], tlsWriteSpy.getCall(1).args[0]), 0,
```
anurag4DSB: Shouldn't retry logic should preserve the original payload for idempotency

tests/functional/kmip/highlevel.spec.js
```diff
@@ -120,4 +122,130 @@ describe('KMIP High Level Driver', () => {
...
+                cb(null, kmipClient.kmip.transport.channel.errorResponse(reason, message));
+                kmipClient.kmip.transport.channel.stubbedResult = false;
+            })
+            .callThrough();
+        const tlsWriteSpy = sinon.spy(kmipClient.kmip.transport.channel, 'write');
+        const tlsEndSpy = sinon.spy(kmipClient.kmip.transport.channel, 'end');
+        const tlsConnectSpy = sinon.spy(kmipClient.kmip.transport.channel, 'connect');
+
+
+        const encrypted2 = await util.promisify(kmipClient.cipherDataKey.bind(kmipClient))(
```
anurag4DSB: I would rename this from `encrypted2` to `retriedEncrypted` to better reflect intent

tests/functional/kmip/highlevel.spec.js
```diff
@@ -120,4 +122,130 @@ describe('KMIP High Level Driver', () => {
             done();
         });
     });
+
+    it('should retry without kmipHandshake on thales session revoked', async () => {
```
anurag4DSB: Both tests use nearly identical setup (options, client init, healthcheck). could we factor out common setup into a shared helper or `beforeEach()` block to improve readability and reduce duplication. Example: `initTestClient()`.

tests/functional/pykmip/ClusterClient.spec.js
```diff
@@ -0,0 +1,508 @@
...
+    let client;
+    let healthyClients;
+    let unhealthyClients;
+    let promises;
+    let spys;
+
+    let kmsKey;
+
+    // Make sure to always disconnect any client
+    afterEach(async () => {
```
anurag4DSB: Can we factor out repeated client teardown + key deletion into a shared helper to keep the `afterEach` blocks DRY and easier to maintain.

tests/functional/pykmip/ClusterClient.spec.js
```diff
@@ -0,0 +1,508 @@
...
+    let client;
+    let promises;
+    let kmsKey;
+
+    beforeEach(() => {
+        client = new KmipClientClass(options);
+        promises = promisify(client);
+    });
+
+    afterEach(async () => {
```
anurag4DSB: We should add `sinon.restore()` here as well, like in the cluster tests, to ensure future added spies don't leak between test cases.

tests/functional/pykmip/ClusterClient.spec.js
```diff
@@ -0,0 +1,508 @@
...
+const sinon = require('sinon');
+const util = require('util');
+
+const KmipClient = require('../../../lib/network/kmip/Client').default;
+const ClusterClient = require('../../../lib/network/kmip/ClusterClient').default;
+const {
+    logger,
+} = require('../../utils/kmip/ersatz');
+
+jest.setTimeout(5000);
```
anurag4DSB, line 16: Please consider making the timeout configurable or environment-dependent.
In CI under load, 5s might not be enough and could cause flaky tests.

tests/functional/pykmip/ClusterClient.spec.js
```diff
@@ -0,0 +1,508 @@
...
+        await assert.doesNotReject(async () => {
+            ciphered = await promises.cipherDataKey(1, kmsKey, dataKey, logger);
+        }, showArsenalErr);
+
+        let deciphered;
+        await assert.doesNotReject(async () => {
+            deciphered = await promises.decipherDataKey(1, kmsKey, ciphered, logger);
+        }, showArsenalErr);
+
+        assert.ok(Buffer.compare(dataKey, deciphered) === 0,
```
anurag4DSB: `Buffer.equals` if supported, could help for better readability.

tests/functional/pykmip/ClusterClient.spec.js
```diff
@@ -0,0 +1,508 @@
...
+        });
+    });
+
+    describe.each(hostsToTest)('for %i hosts + 1 unhealthy at start', (nb) => {
+        const transports = new Array(nb).fill().map((_, i) =>
+            Object.assign({}, transport,
+                { tls: Object.assign({}, transport.tls, { host: `127.0.0.${i + 1}` }) }),
+        );
+        transports.push(Object.assign({}, transport,
+            // not in certificate, will fail connection
```
anurag4DSB: This works for simulating host failure, but please consider being explicit:
“Host not in TLS cert, used to simulate healthcheck failure and trigger fallback logic.”

tests/functional/pykmip/ClusterClient.spec.js
```diff
@@ -0,0 +1,508 @@
...
+            promises = promisify(client);
+            spys = clusterSpys(client.clients);
+            clock = sinon.useFakeTimers({ shouldAdvanceTime: true });
+        });
+
+        afterEach(async () => {
+            if (clock) {
+                clock.reset();
+            }
+        });
```
anurag4DSB: What about using clock.restore() after reset to fully clean up timers and restore global state.
Just resetting time might leave patched globals active.

tests/functional/pykmip/ClusterClient.spec.js
```diff
@@ -0,0 +1,508 @@
...
+                nb + 1,
+                { healthy: healthyClients, unhealthy: unhealthyClients },
+                { healthy: nb + 1, unhealthy: 0 },
+                { spysCipherHealthy: spys.cipherDataKeys },
+                false,
+            );
+
+            spys.cipherDataKeys.forEach(spy => spy.resetHistory());
+
+            await stopSocat(socatProcess);
```
anurag4DSB, line 455: Consider wrapping stopSocat in try/finally or putting it in `afterEach` to avoid port leakage in case the test fails midway. Helpful in CI stability.

tests/functional/pykmip/ClusterClient.spec.js
```diff
@@ -0,0 +1,508 @@
...
+
+            // extract unhealthy spy from healthy
+            const spysCipherUnhealthy = spys.cipherDataKeys.splice(spys.cipherDataKeys.length - 1, 1);
```
anurag4DSB, line 458: Is it always true that always last client is the one that fails

tests/functional/kmip/highlevel.spec.js
BourgoisMickael: No, at this place we don't expect any configurable retry. It's only 1 specific message we know we can retry immediately otherwise any other error message should be sent back to client

tests/functional/kmip/highlevel.spec.js
BourgoisMickael: Yes, this is was test checks

> Compares buf with target and returns a number indicating whether buf comes before, after, or is the same as target in sort order. Comparison is based on the actual sequence of bytes in each Buffer.
>
> 0 is returned if target is the same as buf
> 1 is returned if target should come before buf when sorted.
> -1 is returned if target should come after buf when sorted.


https://nodejs.org/api/buffer.html#bufcomparetarget-targetstart-targetend-sourcestart-sourceend

BourgoisMickael: Or you mean we should check the buffer is actually the same instance instead of checking the content of the buffer ?

tests/functional/pykmip/ClusterClient.spec.js
BourgoisMickael: in that test yes because I pushed the client that I use to simulate disconnection at the end of the transport list

tests/functional/pykmip/ClusterClient.spec.js
BourgoisMickael: This is configured for CI, it runs alone in a separate test CI for kmip with pykmip.

I tested with many more hosts in cluster than 10 and more encryption than 100.

5s is way more than enough for a cluster of 10 hosts

tests/functional/pykmip/ClusterClient.spec.js
BourgoisMickael: Indeed forgot to include it in the `afterEach`

tests/functional/kmip/highlevel.spec.js
anurag4DSB: No, not the same instance, I was referring to buffer content equality, not reference equality. The current check using `Buffer.compare(...) === 0` is exactly what I was expecting: validating that the payload is byte-for-byte identical on retry to ensure idempotency.

So all good here, just wanted to confirm that this behavior was intentional. 🙌

As a side note: if we’re targeting a modern Node.js version, [Buffer.equals()](https://nodejs.org/api/buffer.html#bufequalsotherbuffer) might make the intent a bit more readable here, semantically expresses equality instead of comparing sort order.

tests/functional/pykmip/ClusterClient.spec.js
anurag4DSB: Why not add it in package.json then?

BourgoisMickael: The test is not in package.json, it's executed directly in its own CI workflow `yarn jest tests/functional/pykmip/ClusterClient.spec.js` separate from the other jest tests

So there is no other config to add, everything is self contained for this test

tests/functional/pykmip/ClusterClient.spec.js
BourgoisMickael: yep with `restore` I can skip the `reset` as the stub is removed

tests/functional/kmip/highlevel.spec.js
BourgoisMickael: Fixed this file in [945dbac](https://github.com/scality/Arsenal/pull/2402/commits/945dbac1e9f510b39a7b9e61643a3bab2eae4cc8) and [b11e602](https://github.com/scality/Arsenal/pull/2402/commits/b11e6020627cf0321681b1d348a7194e0a3ec701)

tests/functional/pykmip/ClusterClient.spec.js
BourgoisMickael: Refactored this file here [716b69c](https://github.com/scality/Arsenal/pull/2402/commits/716b69cb3f7a901b88a948d1ea39d5f434de5d80)

.github/workflows/tests-kmip.yaml
BourgoisMickael: let's try with both push and pull_request

tcarmet (approved):

anurag4DSB (approved):

tests/functional/pykmip/ClusterClient.spec.js
anurag4DSB: In that case we should add it to package.json scripts as thats expected for our convention.

BourgoisMickael: added a script in package.json for pykmip tests

anurag4DSB (approved):
If there is any change in integration branches, please do message me on Slack, I would like to review them too.

tmacro (approved):

PR #2403, opened 2025-05-28
Fix version ID generation conflicts in Kubernetes

ghost:
- Today two instances in k8s will share the same process ID, as deployed in distinct Docker containers.
- If two puts on the same key happen concurrently (with ms resolution), we end up with the same key.
- As a result we override one of the two versions.

- In Clustered S3 mode, each instance will use a distinct PID.
- With kubernetes, we now detect, thanks to standard env variables, that we are in a pod. Then, we rely on the HOSTNAME
  standard environment variable, set by kube, that contains the pod name, as `<name>-<randomstring>`. The last part is usually 5-6 characters long and is used as the unique key in the k8s case, ensure no conflict in vID generation.
- We seize this opportunity to standardize the vID-related functions, that should not be mongodb-specific, to put them in the right file.
- Tests are added.

Issue: [ARSN-489](https://scality.atlassian.net/browse/ARSN-489)

[ARSN-489]: https://scality.atlassian.net/browse/ARSN-489?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/versioning/VersionID.ts
+++ b/lib/versioning/VersionID.ts
@@ -21,6 +21,10 @@ const TEMPLATE_TS = new Array(LENGTH_TS + 1).join('0');
 const TEMPLATE_SEQ = new Array(LENGTH_SEQ + 1).join('0');
 const TEMPLATE_RG = new Array(LENGTH_RG + 1).join(' ');
 
+// Counter that is increased after each call to generateUniqueVersionId
+export let uidCounter = 0;
+export const versionIdSeed = getVersionIdSeed();
+
 /**
  * Left-pad a string representation of a value with a given template.
  * For example: pad('foo', '00000') gives '00foo'.
@@ -83,6 +87,23 @@ function wait(span: number) {
     while (getspan(process.hrtime(start)) < span) {
         // do nothing
     }
+}
+
+export function getVersionIdSeed(): string {
+    // The HOSTNAME environment variable is set by default by Kubernetes
+    // and populated with the pod name, containing a suffix with a unique id
+    // as a string.
+    // By default, we rely on the pid, to account for multiple workers in
+    // cluster mode. As a result, the unique id is either <pod-suffix>.<pid>
+    // or <pid>.
+    // If unique vID are needed in a multi cluster mode architecture (i.e.,
+    // multiple server instances, each with multiple workers), the
+    // HOSTNAME environment variable can be set.
+    return `${process.env.HOSTNAME?.split('-').pop() || ''}${process.pid}`;
+}
+
+export function generateUniqueVersionId(replicationGroupId: string): string {
+    return generateVersionId(`${versionIdSeed}.${uidCounter++}`, replicationGroupId);
 }
 
 /**
--- a/tests/unit/versioning/VersionID.spec.js
+++ b/tests/unit/versioning/VersionID.spec.js
@@ -23,6 +23,29 @@ function generateRandomVIDs(count) {
 const count = 1000000;
 
 describe('test generating versionIds', () => {
+    describe('getVersionIdSeed', () => {
+        it('should return the correct versionIdSeed', () => {
+            const versionIdSeed = VID.getVersionIdSeed();
+            assert.strictEqual(versionIdSeed, process.pid.toString());
+        });
+
+        it('should return the correct versionIdSeed when HOSTNAME is set', () => {
+            process.env.HOSTNAME = 'test-pod-123';
+            const versionIdSeed = VID.getVersionIdSeed();
+            assert.strictEqual(versionIdSeed.startsWith('123'), true);
+        });
+    });
+
+    describe('generateUniqueVersionId', () => {
+        it('should increase the uidCounter', () => {
+            const versionId1 = VID.generateUniqueVersionId('somestring');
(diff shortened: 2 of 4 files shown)


tests/unit/versioning/VersionID.spec.js
```diff
@@ -23,6 +23,30 @@ function generateRandomVIDs(count) {
...
 
 describe('test generating versionIds', () => {
+    describe('getVersionIdSeed', () => {
+        it('should return the correct versionIdSeed', () => {
+            const versionIdSeed = VID.getVersionIdSeed();
+            assert.strictEqual(versionIdSeed, process.pid.toString());
+        });
+
+        it('should return the correct versionIdSeed in Kubernetes', () => {
+            process.env.KUBERNETES_SERVICE_HOST = '127.0.0.1';
```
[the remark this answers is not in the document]
ghost: No impact, this is expected as only used in this test.

lib/versioning/VersionID.ts
```diff
@@ -21,6 +21,10 @@ const TEMPLATE_TS = new Array(LENGTH_TS + 1).join('0');
 const TEMPLATE_SEQ = new Array(LENGTH_SEQ + 1).join('0');
 const TEMPLATE_RG = new Array(LENGTH_RG + 1).join(' ');
 
+// Counter that is increased after each call to generateUniqueVersionId
+export let uidCounter = 0;
+export let versionIdSeed = getVersionIdSeed();
```
francoisferrand: versionIdSeed should be `const`

lib/versioning/VersionID.ts
```diff
@@ -85,6 +89,23 @@ function wait(span: number) {
...
+export function getVersionIdSeed(): string {
+    // The HOSTNAME environment variable is set by default by Kubernetes
+    // and populated with the pod name, containing a suffix with a unique id
+    // as a string. We then detect if we are running in Kubernetes and use
+    // the pod unique suffix id as the versionIdSeed.
+    // See https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/#directly-accessing-the-rest-api
+    // By default, we rely on the pid, to account for multiple workers in
+    // cluster mode.
+    // As a result, the unique id is either <pod-suffix>.<pid> or <pid>.
+    return process.env.KUBERNETES_SERVICE_HOST ?
```
francoisferrand: do we need to condition on `KUBERNETES_SERVICE_HOST` ? could we not always include `HOSTNAME` in seed, to further avoid conflicts even on non-k8s hosts, where the same PID could be used on multiple nodes, with (bad) luck?

lib/versioning/VersionID.ts
```diff
@@ -85,6 +89,23 @@ function wait(span: number) {
...
+        `${process.env.HOSTNAME?.split('-').pop() || ''}${process.pid}` : process.pid.toString();
```
francoisferrand: do we actually need to 'split' ? can't we use the full hostname anyway, for better uniqueness?

lib/versioning/VersionID.ts
```diff
@@ -85,6 +89,23 @@ function wait(span: number) {
     }
 }
 
+export function getVersionIdSeed(): string {
```
francoisferrand, line 92: is this logic used in metadata as well?
seeng how this was handled in mongo, I guess either this is done specifically with mongo backend, or the code is duplicated in metadata:
* in the first case, it means this should probably stay in MongoClientInterface;
* in that second case, we should probably create a ticket to clean this up in metadata and use the new code.

francoisferrand (approved):

lib/versioning/VersionID.ts
ghost: Do we need the split -> yes
use the full hostname -> No, because it's concatenated when used this way, which would create very long version IDs...

lib/versioning/VersionID.ts
ghost: No it is not used in metadata, but it is kind of used in the memory backend. I mean "kind of" because we actually just create a version ID without any input, but for testing purposes, being able to use unique IDs might be better.

In metadata we have this: https://github.com/scality/MetaData/blob/7fa7de94369d25935992a8bd7832db4477eeae80/lib/server/VersioningRequestProcessor.js#L107
The ID is generatedc in metadata, not by s3 client.
The algorithm is not applicable here, we cannot reuse it.

To me the versionID file is supposed to handle versionID operations, and having this primitive should not be bound to MongoDB, for instance if we want to use it with the in-mem backend (or file).

lib/versioning/VersionID.ts
ghost: I'm fine including it always yes. This will allow to have unique IDs if we deploy cloudserver in cluster mode but in multiple hosts.

Just a note on why I did this initially: to me we should only deploy s3 with cluster mode in a single host (without kube) OR multiple pods with one worker (in kube) OR multiple pods with multiple workers (in kube). All these cases are handled here.

benzekrimaha (approved):
LGTM

PR #2412, opened 2025-06-04
ARSN-499: KMIPCluster Healthcheck report healthy if 1 node is up

BourgoisMickael:
clusterHealthcheck would broadcast a healthcheck
on every node and fail if 1 node fails.

But we want to succeed if at least 1 node is up
so use the round robin method with retries to keep
it simple

### DIFF
--- a/lib/network/kmip/ClusterClient.ts
+++ b/lib/network/kmip/ClusterClient.ts
@@ -270,8 +270,24 @@ export default class ClusterClient implements KMSInterface {
     }
 
     healthcheck(...args: Parameters<Required<KMSInterface>['healthcheck']>) {
-        // for now check health of every member
-        this.clusterHealthcheck.apply(this, args);
+        // healthcheck follow round robin with retries instead of broadcast
+        const originalCallback = args.pop() as ArrayLast<typeof args>;
+        const poppedArgs = args as unknown as ArrayPopped<typeof args>;
+        const logger = args[args.length - 1] as ArrayLast<typeof poppedArgs>;
+
+        if (this.clients.length === 0) {
+            logger.warn('kmip cluster has no healthy hosts');
+            return originalCallback(errorInstances.InternalError.customizeDescription(
+                kmipMsg('Healthcheck', '', `no healthy host in the cluster`)));
+        }
+
+        const client = this.next();
+
+        client.healthcheck(
+            ...poppedArgs,
+            this.callback(client, 'healthcheck', poppedArgs, logger, originalCallback),
+        );
+        return undefined;
     }
 
     stop(cb?: Function) {
--- a/lib/network/kmip/ClusterClient.ts
+++ b/lib/network/kmip/ClusterClient.ts
@@ -188,7 +188,7 @@ export default class ClusterClient implements KMSInterface {
         if (this.clients.length === 0) {
             logger.warn('kmip cluster has no healthy hosts');
             return originalCallback(errorInstances.InternalError.customizeDescription(
-                kmipMsg('Create', args[0], `no healthy host in the cluster`)));
+                kmipMsg('Create', args[0], 'kmip cluster has no healthy hosts')));
         }
 
         const client = this.next();
@@ -208,7 +208,7 @@ export default class ClusterClient implements KMSInterface {
         if (this.clients.length === 0) {
             logger.warn('kmip cluster has no healthy hosts');
             return originalCallback(errorInstances.InternalError.customizeDescription(
-                kmipMsg('Destroy', args[0], `no healthy host in the cluster`)));
+                kmipMsg('Destroy', args[0], 'kmip cluster has no healthy hosts')));
         }
 
         const client = this.next();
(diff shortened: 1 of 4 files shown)


lib/network/kmip/ClusterClient.ts
```diff
@@ -270,7 +270,23 @@ export default class ClusterClient implements KMSInterface {
     }
 
     healthcheck(...args: Parameters<Required<KMSInterface>['healthcheck']>) {
-        // for now check health of every member
-        this.clusterHealthcheck.apply(this, args);
+        // healthcheck follow round robin with retries instead of broadcast
+        const originalCallback = args.pop() as ArrayLast<typeof args>;
+        const poppedArgs = args as unknown as ArrayPopped<typeof args>;
+        const logger = args[args.length - 1] as ArrayLast<typeof poppedArgs>;
```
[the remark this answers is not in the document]
BourgoisMickael: This is the expected behavior to mutate the array as the cb is replaced by another callback

lib/network/kmip/ClusterClient.ts
```diff
@@ -270,7 +270,23 @@ export default class ClusterClient implements KMSInterface {
...
+                kmipMsg('Healthcheck', '', `no healthy host in the cluster`)));
+        }
+
+        const client = this.next();
+
+        client.healthcheck(
+            ...poppedArgs,
+            this.callback(client, 'healthcheck', poppedArgs, logger, originalCallback),
+        );
+        return undefined;
```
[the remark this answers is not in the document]
BourgoisMickael: Because of linting in integration branches

lib/network/kmip/ClusterClient.ts
```diff
@@ -270,7 +270,23 @@ export default class ClusterClient implements KMSInterface {
...
-        this.clusterHealthcheck.apply(this, args);
+        // healthcheck follow round robin with retries instead of broadcast
+        const originalCallback = args.pop() as ArrayLast<typeof args>;
+        const poppedArgs = args as unknown as ArrayPopped<typeof args>;
+        const logger = args[args.length - 1] as ArrayLast<typeof poppedArgs>;
+
+        if (this.clients.length === 0) {
+            logger.warn('kmip cluster has no healthy hosts');
+            return originalCallback(errorInstances.InternalError.customizeDescription(
+                kmipMsg('Healthcheck', '', `no healthy host in the cluster`)));
```
jonathan-gramain: Maybe return the same logged message to avoid surprises (and you can replace backticks by simple quotes):

```suggestion
                kmipMsg('Healthcheck', '', 'kmip cluster has no healthy hosts')));
```

lib/network/kmip/ClusterClient.ts
```diff
@@ -270,7 +270,23 @@ export default class ClusterClient implements KMSInterface {
...
+        }
+
+        const client = this.next();
+
+        client.healthcheck(
```
jonathan-gramain, line 286: Are we sure the caller will retry enough times to ensure our definition of "healthy" is satisfied? Otherwise, it may be necessary to retry in this function in a round-robin fashion as you state in the comment.

BourgoisMickael: `this.callback` takes care of retrying.

Healthcheck will now act like any other action, it retries until the configured `retries` is reached, which by default will retry on every hosts of the cluster until a success or they are all marked unhealthy.

But if the client configure a smaller `retries` number then it might not go over all hosts. 

But we'll rework the healthcheck for 9.5.1 so we might provide different ways to do later.

lib/network/kmip/ClusterClient.ts
BourgoisMickael: I've reworded in every methods

lib/network/kmip/ClusterClient.ts
```diff
@@ -270,7 +270,23 @@ export default class ClusterClient implements KMSInterface {
...
 
     healthcheck(...args: Parameters<Required<KMSInterface>['healthcheck']>) {
-        // for now check health of every member
-        this.clusterHealthcheck.apply(this, args);
+        // healthcheck follow round robin with retries instead of broadcast
+        const originalCallback = args.pop() as ArrayLast<typeof args>;
+        const poppedArgs = args as unknown as ArrayPopped<typeof args>;
+        const logger = args[args.length - 1] as ArrayLast<typeof poppedArgs>;
+
+        if (this.clients.length === 0) {
```
dvasilas, line 278: Do we need this check? Isn't the check in `this.callback` enough for this case?

BourgoisMickael: Yes we need to check before we send the request the first time. 

Otherwise the round robin, (`this.next`) would return an undefined client

dvasilas: Ok I see.

dvasilas (approved):

tcarmet (approved):

jonathan-gramain (approved):

BourgoisMickael:
@bert-e reset

I've updated test

BourgoisMickael:
/reset

I've fixed test for cluster healthcheck

PR #2414, opened 2025-06-06
make crr locations invalid as location constraints

welansari:
Issue: ARSN-500

### DIFF
--- a/lib/models/BackendInfo.ts
+++ b/lib/models/BackendInfo.ts
@@ -52,7 +52,8 @@ export default class BackendInfo {
         locationConstraint: string | undefined,
         log: RequestLogger,
     ) {
-        if (!locationConstraint || !(locationConstraint in config.locationConstraints)) {
+        if (!locationConstraint || !(locationConstraint in config.locationConstraints) ||
+            config.locationConstraints[locationConstraint].isCRR) {
             log.trace('proposed locationConstraint is invalid',
                 { locationConstraint });
             return false;
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=20"
   },
-  "version": "8.2.21",
+  "version": "8.2.22",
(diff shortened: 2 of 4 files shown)


lib/models/BackendInfo.ts
```diff
@@ -52,7 +52,8 @@ export default class BackendInfo {
         locationConstraint: string | undefined,
         log: RequestLogger,
     ) {
-        if (!locationConstraint || !(locationConstraint in config.locationConstraints)) {
+        if (!locationConstraint || !(locationConstraint in config.locationConstraints) ||
+            config.locationConstraints[locationConstraint].isCrr) {
```
ghost:
```suggestion
            config.locationConstraints[locationConstraint].isCRR) {
```

as this is an acronym?

ghost (approved):

lib/models/BackendInfo.ts
francoisferrand: it is defined as `isCrr` in cloudserver and zkop, c.f. other PR, so care must be taken if we change it.

lib/models/BackendInfo.ts
```diff
@@ -52,7 +52,8 @@ export default class BackendInfo {
         locationConstraint: string | undefined,
         log: RequestLogger,
     ) {
-        if (!locationConstraint || !(locationConstraint in config.locationConstraints)) {
+        if (!locationConstraint || !(locationConstraint in config.locationConstraints) ||
```
francoisferrand, line 55: maybe easier to read this way?

```suggestion
        if (!config.locationConstraints[locationConstraint] ||
```

francoisferrand (approved):
check other PRs, may need to bump version before merging

lib/models/BackendInfo.ts
welansari: Can't really do that in typescript as `locationConstraint` can be undefined.

PR #2415, opened 2025-06-08
properly forward VaultClient errors

welansari:
VaultClient errors have the error type in the "code" property which is not properly handled in toArsenalError()

Issue: ARSN-501

### DIFF
--- a/lib/s3routes/routesUtils.ts
+++ b/lib/s3routes/routesUtils.ts
@@ -22,20 +22,29 @@ export type CallApiMethod = (
     callback: (err: ArsenalError | Error | null, ...data: any[]) => void,
 ) => void;
 
+interface VaultClientError {
+    code: string | number;
+    description: string;
+}
+
 /**
  * Convert an error to an ArsenalError instance
  * @param err - Error to convert
  * @returns ArsenalError instance
  */
-export function toArsenalError(err: ArsenalError | Error): ArsenalError {
+export function toArsenalError(err: ArsenalError | VaultClientError | Error): ArsenalError {
     if (err instanceof ArsenalError) {
         return err;
     }
-
-    // If it's a known error type, use the error instance, otherwise wrap
-    // it in an InternalError
-    return errorInstances[err.message] ||
-        errorInstances.InternalError.customizeDescription(err.message);
+    // For known error types, return the error instance as-is; otherwise, wrap it in an InternalError.
+    // - VaultClient errors identify their type via the `code` property.
+    // - ArsenalErrors received from external services (e.g., Metadata) use the `message` property for the error type.
+    return errorInstances[(err as Error).message || (err as VaultClientError).code] ||
+        errorInstances.InternalError.customizeDescription(
+            (err as Error).message ||
+            (err as VaultClientError).description ||
+            errorInstances.InternalError.description,
+        );
 }
 
 /**
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=20"
   },
-  "version": "8.2.20",
+  "version": "8.2.21",
   "description": "Common utilities for the S3 project components",
   "main": "build/index.js",
   "repository": {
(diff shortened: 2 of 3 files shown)


lib/s3routes/routesUtils.ts
```diff
@@ -22,20 +22,33 @@ export type CallApiMethod = (
...
     if (err instanceof ArsenalError) {
         return err;
     }
-
     // If it's a known error type, use the error instance, otherwise wrap
     // it in an InternalError
-    return errorInstances[err.message] ||
-        errorInstances.InternalError.customizeDescription(err.message);
+    return err.message && errorInstances[err.message] ||
+        err.code && errorInstances[err.code] ||
```
francoisferrand: do we really need to check if the field is present? indexing an object with value `undefined` just returns undefined (`{ foo:1, bar:2 }[undefined] === undefined`), so we may just do the following ?
```suggestion
    return errorInstances[err.message] ||
        errorInstances[err.code] ||
```

francoisferrand: or even simpler, maybe this:
```suggestion
    return errorInstances[err.message || err.code] ||
```

(unless we have situations where both `message` and `code` are set, but message does not match so we need to rely on code for matching arsenal error...)

lib/s3routes/routesUtils.ts
```diff
@@ -22,20 +22,33 @@ export type CallApiMethod = (
     callback: (err: ArsenalError | Error | null, ...data: any[]) => void,
 ) => void;
 
+type APIError = {
```
francoisferrand: This type should actually be defined as an "extension" of Error

lib/s3routes/routesUtils.ts
```diff
@@ -22,20 +22,33 @@ export type CallApiMethod = (
...
+    description?: string;
+}
+
 /**
  * Convert an error to an ArsenalError instance
  * @param err - Error to convert
  * @returns ArsenalError instance
  */
-export function toArsenalError(err: ArsenalError | Error): ArsenalError {
+export function toArsenalError(err: ArsenalError | APIError ): ArsenalError {
```
francoisferrand: is this `code` specific to VaultClient?
in that case, may be better to type it explicitly like this:
```suggestion
export function toArsenalError(err: ArsenalError | VaultClientError | Error): ArsenalError {
```
or (if it is called either with Arsenal or VaultClient errors)
```suggestion
export function toArsenalError(err: ArsenalError | VaultClientError): ArsenalError {
```

lib/s3routes/routesUtils.ts
welansari: Because of the different non overlapping error types, this doesn't build in TS.
The only "clutter free" way i see would be doing something like : `errorInstances[(err as any)?.code]`

lib/s3routes/routesUtils.ts
welansari: > is this `code` specific to VaultClient?

Yes it is specific to it, i made a separate type only for those errors

lib/s3routes/routesUtils.ts
welansari: I separated VaultClient errors from the normal errors. VaultClient errors are not really an extension of the Error type as they don't share the same fields

VaulClient errors look like this
```
{
  code: string,
  description: string,
  <ERROR_TYPE>: true,
}
```

lib/s3routes/routesUtils.ts
francoisferrand: would this work?
```
return errorInstances[err.message! || err.code!] ||
```

(or maybe make explicit what "kind" of inputs provides which field : `errorInstances[(err as Error).message || (err as VaultError).code] ||`, but not sure we can do it without going through a cast to any....)

package.json
```diff
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=20"
   },
-  "version": "8.2.19",
+  "version": "8.2.20",
```
francoisferrand: 8.2.20 is already released, this PR needs to be rebased

lib/s3routes/routesUtils.ts
welansari: This worked `errorInstances[(err as Error).message || (err as VaultError).code] `

francoisferrand (approved):

ghost (approved):

PR #2432, opened 2025-06-23
ARSN-503: Return the kms error without warning

BourgoisMickael:
objectCopy and objectPutCopyPart use these functions from dataWrapper and they both catch the error to return it to the client

The KMS implementation should take care of hidding sensitive internal information

Rework error message for AWS KMS to match the output from AWS and hide some non KMS related message such as Network errors to prevent leaking sensitive information like hostnames, ip, ..

### DIFF
--- a/lib/network/utils.ts
+++ b/lib/network/utils.ts
@@ -1,4 +1,6 @@
-import { errorInstances } from '../errors';
+import type { AWSError } from 'aws-sdk';
+import { ArsenalError, errorInstances } from '../errors';
+import { allowedKmsErrors } from '../errors/kmsErrors';
 
 /**
  * Normalize errors according to arsenal definitions with a custom prefix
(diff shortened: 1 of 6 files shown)


lib/network/utils.ts
```diff
@@ -28,6 +29,61 @@ export function arsenalErrorKMIP(err: string | Error) {
...
+// List of AWS KMS errors that are specific to KMS and where the error
+// can be returned to the user as is with a specific status code.
+// Avoid any error that might leak sensitive information such as hostnames
+// or ip addresses from network related errors.
+// KeyId is not considered sensitive.
+// Reference: https://docs.aws.amazon.com/kms/latest/APIReference/CommonErrors.html
+// See specific actions like CreateKey, Encrypt, Decrypt, etc. for more details.
+// Other errors not listed will return only the same error code but http status 500
+// and message will be generic and details will be in logs only.
+const allowedKmsErrors = {
```
anurag4DSB: Shouldn't we have these errors in the https://github.com/scality/Arsenal/blob/development/8.2/lib/errors/arsenalErrors.ts

BourgoisMickael: They are KMS specific errors and in the S3 context the code should be prefixed with `KMS.` 

I did not want to include many new errors and duplicate them to have 2 different codes for the same error.
We should not do anything else than return them back to client that why I only encapsulated them in arsenal error

anurag4DSB: THere is no duplication see [this](https://github.com/scality/Arsenal/blob/development/8.2/lib/errors/arsenalErrors.ts#L1076)
I think its still worth adding all errors there like the KMS one already present

anurag4DSB (approved):
I am approving the PR, your call on my comment

lib/network/utils.ts
BourgoisMickael: I've reworked kms errors in arsenal [97d9bfe](https://github.com/scality/Arsenal/pull/2432/commits/97d9bfefb6896ca839e88cdc53c53328a7fc0f99)

tmacro (approved):

PR #2439, opened 2025-06-26
block crr location for transitions

SylvainSenechal:
Issue : [ARSN-505](https://scality.atlassian.net/browse/ARSN-505)

[ARSN-505]: https://scality.atlassian.net/browse/ARSN-505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/models/LifecycleConfiguration.ts
+++ b/lib/models/LifecycleConfiguration.ts
@@ -129,14 +129,36 @@ export default class LifecycleConfiguration {
      * @param config - the CloudServer config
      * @return - LifecycleConfiguration instance
      */
-    constructor(xml: any, config: { replicationEndpoints: { site: string }[], supportedLifecycleRules: string[] }) {
+    constructor(
+        xml: any,
+        config: {
+            replicationEndpoints: { site: string }[],
+            locationConstraints?: Record<string, { isCRR: boolean }>,
+            supportedLifecycleRules: string[],
+        }
+    ) {
         this._parsedXML = xml;
-        this._storageClasses =
-            config.replicationEndpoints.map(endpoint => endpoint.site);
+        this._storageClasses = this._buildStorageClasses(config);
         this._supportedLifecycleRules = config.supportedLifecycleRules;
         this._ruleIDs = [];
         this._tagKeys = [];
         this._config = {};
+    }
+
+    /**
+     * Build the list of available storage classes for lifecycle transitions.
+     * This filters out any locations that are designated for cross-region
+     * replication (CRR), as transitions to these are not allowed.
+     * @param config - The CloudServer config
+     * @returns An array of valid storage class names.
+     */
+    _buildStorageClasses(config: {
+        replicationEndpoints: { site: string }[],
+        locationConstraints?: Record<string, { isCRR: boolean }>,
+    }) {
+        return config.replicationEndpoints
+            .map(endpoint => endpoint.site)
+            .filter(site => !config.locationConstraints?.[site]?.isCRR);
     }
 
     /**
@@ -598,7 +620,7 @@ export default class LifecycleConfiguration {
             // This differs from the AWS message. This will help the user since
             // the StorageClass does not conform to AWS specs.
             const list = `'${this._storageClasses.join("', '")}'`;
-            const msg = `'StorageClass' must be one of ${list}`;
+            const msg = `'StorageClass' must be one of ${list}, but provided: ${storageClass}`;
             return errorInstances.MalformedXML.customizeDescription(msg);
         }
         if (usedStorageClasses.includes(storageClass)) {
--- a/tests/unit/models/LifecycleConfiguration.spec.js
+++ b/tests/unit/models/LifecycleConfiguration.spec.js
@@ -13,13 +13,15 @@ const days = {
 
 const mockConfig = {
     replicationEndpoints: [
-        {
-            site: 'a',
-        },
-        {
-            site: 'b',
-        },
+        { site: 'a' },
+        { site: 'b' },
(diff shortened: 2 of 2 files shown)


lib/models/LifecycleConfiguration.ts
```diff
@@ -129,10 +130,17 @@ export default class LifecycleConfiguration {
...
+    constructor(
+        xml: any,
+        config: {
+            replicationEndpoints: { site: string }[],
+            locationConstraints?: { [key: string]: { isCRR: boolean } },
+            supportedLifecycleRules: string[],
+        }) {
         this._parsedXML = xml;
-        this._storageClasses =
+        this._storageClasses =
```
ghost:
```suggestion
        this._storageClasses =
```

lib/models/LifecycleConfiguration.ts
```diff
@@ -129,10 +130,17 @@ export default class LifecycleConfiguration {
...
      * @return - LifecycleConfiguration instance
      */
-    constructor(xml: any, config: { replicationEndpoints: { site: string }[], supportedLifecycleRules: string[] }) {
+    constructor(
+        xml: any,
+        config: {
+            replicationEndpoints: { site: string }[],
+            locationConstraints?: { [key: string]: { isCRR: boolean } },
+            supportedLifecycleRules: string[],
+        }) {
```
ghost: Nit: If we want something more readable:

```suggestion
    constructor(
        xml: any,
        config: {
            replicationEndpoints: { site: string }[],
            locationConstraints?: { [key: string]: { isCRR: boolean } },
            supportedLifecycleRules: string[],
        },
    ) {
```

lib/models/LifecycleConfiguration.ts
```diff
@@ -129,10 +130,17 @@ export default class LifecycleConfiguration {
      * @param config - the CloudServer config
      * @return - LifecycleConfiguration instance
      */
-    constructor(xml: any, config: { replicationEndpoints: { site: string }[], supportedLifecycleRules: string[] }) {
+    constructor(
+        xml: any,
+        config: {
+            replicationEndpoints: { site: string }[],
+            locationConstraints?: { [key: string]: { isCRR: boolean } },
```
ghost:
```suggestion
            locationConstraints?: Record<string, { isCRR: boolean }>,
```

tests/unit/models/LifecycleConfiguration.spec.js
```diff
@@ -13,13 +13,15 @@ const days = {
...
     replicationEndpoints: [
-        {
-            site: 'a',
-        },
-        {
-            site: 'b',
-        },
+        { site: 'a', },
+        { site: 'b', },
+        { site: 'aCrrLocation', }
```
ghost: If inline, the comma is not required by our linter

```suggestion
        { site: 'a' },
        { site: 'b' },
        { site: 'aCrrLocation' }
```

lib/models/LifecycleConfiguration.ts
```diff
@@ -594,11 +602,15 @@ export default class LifecycleConfiguration {
         rule: { Prefix?: string[]; Filter?: any };
     }) {
         const { usedStorageClasses, storageClass, ancestor, rule } = params;
+        if (this._locationConstraints[storageClass]?.isCRR) {
+            const msg = `Transition to ${storageClass} is forbidden by locationConstraints: this storageClass is a CRR`;
+            return errorInstances.InvalidRequest.customizeDescription(msg);
+        }
```
ghost: As the function is supposed to deal with storage classes, shouldn't we introduce a `_checkLocationConstraints` instead? As here, it fells strange that we are returning this error even before we confirmed the storage class exists in `_storageClasses`.

SylvainSenechal: I think it's reasonnable, and more importantly, I've been thinking that we also need to make sur this logic is only applied for Transition/NoncurrentVersionTransition. At the moment, _checkStorageClasses() is only called for transition/NoncurrentVersionTransition, but it could potentially be called for other lifecycle rules 🤔  @Kerkesni

welansari: I'm not too worried about that for now, the only lifecycle rules supported in aws are transitions and expirations (current & non current), expiration doesn't care about the storageClass.

welansari: I'm still wondering however if we should filter out the CRR location from `this._storageClasses` in the constructor, this will:
1. Avoid storing the detailed location list
2. Remove the need for any additional check as we only need to verify the existance of the storageClass like it's the case now

We could add a separate function for building the supported storageClass list and call it in the constructor, this will also make it future proof in case we decide to remove other location types, as we will just filter them out once at startup instead of having to check for each of them every time.

SylvainSenechal: @Kerkesni @williamlardier You can check current implementation, I think it's the simplest, 
The only downside is the error message will not be precise when the user provides a correct, **but crr**, storageClass in the Transition Rule : They will be told that the provided storage class doesn't exists, instead of being told that it exists but was filtered out due to CRR

welansari (approved):

lib/models/LifecycleConfiguration.ts
welansari: CRR locations not being supported in lifecycle will be documented in the product documentation, i don't think it's a big deal if the error message doesn't say it.

ghost (approved):

PR #2440, opened 2025-06-26
Use insertOne for creating new version objects

welansari:
In `putObjectVerCase1`, switch from using `updateOne` with upsert to
`insertOne` for creating new version objects.

The `putObjectVerCase1` function is designed exclusively for creating
new object versions, each with a freshly generated, unique version ID.
There is no intended scenario where a version object with the same ID
should already exist.

Using `updateOne` with `upsert:true` was therefore logically incorrect.
It could silently overwrite an existing version object in the case of a
version ID collision, leading to data corruption.

By switching to `insertOne`, the operation will now correctly fail with
a duplicate key error if a collision occurs. This fail-fast approach properly
treats a version ID collision as an error, ensuring system integrity.

**Note:** I couldn't find a reason on why it was implemented this way.
[The initial implementation of the `MongoClientInterface`](https://github.com/scality/cloudserver/pull/1064) used `UpdateOne`
in all operations without any explications.

`putObjectVerCase1` is used in the following APIs:
- PutObject -> in versioning enabled bucket to create new versions
- DeleteObject -> in versioned bucket when creating a DeleteMarker 
- DeleteObjects -> in versioned bucket when creating a DeleteMarker 

Tested this with a highly concurrent `PUT` workload that generated
collisions over multiple instances of Cloudserver , the retry logic works
fine and we end up having the desired number of versions.

Issue: ARSN-506

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -826,14 +826,11 @@ class MongoClientInterface {
         const masterKey = formatMasterKey(objName, params.vFormat);
         // initiating array of operations with version creation
         const ops: AnyBulkWriteOperation<ObjectMetastoreDocument>[] = [{
-            updateOne: {
-                filter: {
+            insertOne: {
+                document: {
                     _id: versionKey,
-                },
-                update: {
-                    $set: { _id: versionKey, value: objVal },
-                },
-                upsert: true,
+                    value: objVal,
+                } as ObjectMetastoreDocument,
             },
         }];
         // filter to get master
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=16"
   },
-  "version": "8.1.161",
+  "version": "8.1.162",
   "description": "Common utilities for the S3 project components",
   "main": "build/index.js",
   "repository": {
--- a/tests/functional/metadata/mongodb/putObject.spec.js
+++ b/tests/functional/metadata/mongodb/putObject.spec.js
@@ -1,5 +1,6 @@
 const async = require('async');
 const assert = require('assert');
+const sinon = require('sinon');
 const werelogs = require('werelogs');
 const { MongoMemoryReplSet } = require('mongodb-memory-server');
 const { errors, versioning } = require('../../../../index');
@@ -7,6 +8,7 @@ const logger = new werelogs.Logger('MongoClientInterface', 'debug', 'debug');
 const BucketInfo = require('../../../../lib/models/BucketInfo').default;
 const MetadataWrapper =
     require('../../../../lib/storage/metadata/MetadataWrapper');
+const { VersionID } = require('../../../../lib/versioning');
 const { BucketVersioningKeyFormat } = versioning.VersioningConstants;
 
 const IMPL_NAME = 'mongodb';
@@ -128,6 +130,7 @@ describe('MongoClientInterface:metadata.putObjectMD', () => {
             });
 
             afterEach(done => {
+                sinon.restore();
                 metadata.deleteBucket(BUCKET_NAME, logger, done);
             });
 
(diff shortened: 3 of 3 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -826,14 +826,11 @@ class MongoClientInterface {
         const masterKey = formatMasterKey(objName, params.vFormat);
         // initiating array of operations with version creation
         const ops: AnyBulkWriteOperation<ObjectMetastoreDocument>[] = [{
-            updateOne: {
-                filter: {
+            insertOne: {
```
welansari, line 829: **Note:** One risk i see with this is that maybe on very high workloads with lots of collisions, we might end up colliding multiple times for a same object, which will cause the requests to fail. Maybe we should return some sort of a retryable error in this case instead of an InternalError like we do now ?

francoisferrand: * Indeed, we need to consider who will retry and how: should it be the "end user" (→ need to send returnable error!), cloudserver, or even done internally in here?
* The risk of this occurring should be mitigated by the other change, related to the versionID format change to ensure uniqueness...

package.json
```diff
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=16"
   },
-  "version": "8.1.161",
+  "version": "8.1.162",
```
francoisferrand, line 6: not sure we want to bump just yet: we may want to wait for the other related changes for putObjectVerCase2 / ... (if possible, c.f. TAD) and the versionID uniqueness improvement

francoisferrand (commented):
Looks good, but retry handling (or error returned) needs to be considered ; and it may be the change to rename and/or improve docs for these 4 functions, to better explain which case they handle

lib/storage/metadata/mongoclient/MongoClientInterface.ts
welansari: > Indeed, we need to consider who will retry and how: should it be the "end user" (→ need to send returnable error!), cloudserver, or even done internally in here?

Looking at it further at this issue, `internalError` actually seems to be considered retryable so no issue there and with a quick search, it looks like aws sdk/cli by default implement some sort of retry mechanism for transient errors. So this might end up working fine actually, we perform one retry internally on the first collision, after that we return an `internalError` and the aws client (cli/sdk) performs the retry on its end.

> The risk of this occurring should be mitigated by the other change, related to the versionID format change to ensure uniqueness...

Yes this will also greatly reduce the risk of falling into that case

ghost: > Yes this will also greatly reduce the risk of falling into that case

Do we still have the risk with the second fix? To me it should be totally prevented, and this one is like a safe guard to avoid overrides in case of wrongly configured S3 that would still produce the same vID.

tests/functional/metadata/mongodb/putObject.spec.js
```diff
@@ -354,6 +357,36 @@ describe('MongoClientInterface:metadata.putObjectMD', () => {
...
+                    // We put another version of the object with the same versionId
+                    next => metadata.putObjectMD(BUCKET_NAME, OBJECT_NAME, objVal, params, logger, next),
+                ], err => {
+                    assert(err.is.InternalError,
+                        `expected InternalError, got ${err.name} with message: ${err.message}`,
+                    );
+                    // make sure the retry triggered on the first collision detection
+                    assert(genVID.calledThrice,
+                        `expected generateUniqueVersionId to be called thrice, got ${genVID.callCount} times`);
+                    done();
```
ghost, line 386: Can we add a variant, where we have a return being a new vID on the retry, so that we confirm, in that case, that no InternalError is returned, and the new vID is properly created during the retry?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
welansari: > To me it should be totally prevented, and this one is like a safe guard to avoid overrides in case of wrongly configured S3 that would still produce the same vID.

Yes, you are correct

ghost (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.ts
francoisferrand: yes, the strategy here is:
* Use `insert` to solve the issue once and for all, whatever happens with versionID or deployment;
* Improve versionID so we _avoid_ the retry and ensure we don't degrade performance

so if versionID is "unique" as expected, than this one is useless (but causes no harm) ; and if now or later there is a uniqueness issue, we will be protected by this change, and get back on our feet with a retry.

francoisferrand (approved):

PR #2445, opened 2025-07-01
support new versionId format

welansari:
see https://scality.atlassian.net/wiki/spaces/OS/pages/3225583692/VersionID+Collisions+in+Zenko

Related PR: https://github.com/scality/cloudserver/pull/5826

Issue: ARSN-502

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -37,7 +37,7 @@ import {
 import Uuid from 'uuid';
 import diskusage from 'diskusage';
 
-import { generateUniqueVersionId, getVersionIdSeed } from '../../../versioning/VersionID';
+import { generateVersionId } from '../../../versioning/VersionID';
 import * as listAlgos from '../../../algos/list/exportAlgos';
 import LRUCache from '../../../algos/cache/LRUCache';
 
@@ -84,6 +84,7 @@ export type MongoDBClientInterfaceParameters = {
     path: string,
     database: string,
     logger: werelogs.Logger,
+    instanceId: string,
     replicationGroupId: string,
     authCredentials: MongoUtils.AuthCredentials,
     isLocationTransient: Function,
@@ -237,6 +238,7 @@ class MongoClientInterface {
     private client: MongoClient | null;
     private db: Db | null;
     private path: string;
+    private instanceId: string;
     private replicationGroupId: string;
     private database: string;
     private isLocationTransient: Function;
@@ -253,7 +255,7 @@ class MongoClientInterface {
 
     constructor(params: MongoDBClientInterfaceParameters) {
         const { replicaSetHosts, writeConcern, replicaSet, readPreference, path,
-            database, logger, replicationGroupId, authCredentials,
+            database, logger, instanceId, replicationGroupId, authCredentials,
             isLocationTransient, shardCollections } = params;
         const cred = MongoUtils.credPrefix(authCredentials);
         this.mongoUrl = `mongodb://${cred}${replicaSetHosts}/` +
@@ -268,6 +270,7 @@ class MongoClientInterface {
         this.adminDb = null;
         this.logger = logger;
         this.path = path;
+        this.instanceId = instanceId;
         this.replicationGroupId = replicationGroupId;
         this.database = database;
         this.isLocationTransient = isLocationTransient;
@@ -819,7 +822,7 @@ class MongoClientInterface {
         cb: ArsenalCallback<string>,
         isRetry?: boolean,
     ) {
-        const versionId = generateUniqueVersionId(this.replicationGroupId);
+        const versionId = generateVersionId(this.instanceId, this.replicationGroupId);
         // eslint-disable-next-line
         objVal.versionId = versionId;
         const versionKey = formatVersionKey(objName, versionId, params.vFormat);
@@ -944,7 +947,7 @@ class MongoClientInterface {
         log: werelogs.Logger,
         cb: ArsenalCallback<string>,
     ) {
-        const versionId = generateUniqueVersionId(this.replicationGroupId);
+        const versionId = generateVersionId(this.instanceId, this.replicationGroupId);
         // eslint-disable-next-line
         objVal.versionId = versionId;
         const masterKey = formatMasterKey(objName, params.vFormat);
@@ -1778,7 +1781,7 @@ class MongoClientInterface {
     ) {
         const masterKey = formatMasterKey(objName, params.vFormat);
         const versionKey = formatVersionKey(objName, params.versionId, params.vFormat);
-        const _vid = generateUniqueVersionId(this.replicationGroupId);
+        const _vid = generateVersionId(this.instanceId, this.replicationGroupId);
         async.series([
             next => c.updateOne(
                 {
--- a/lib/versioning/VersionID.ts
+++ b/lib/versioning/VersionID.ts
@@ -1,13 +1,31 @@
-// VersionID format:
+// Hex VersionID format:
 //         timestamp  sequential_position  rep_group_id  other_information
 // where:
 // - timestamp              14 bytes        epoch in ms (good untill 5138)
 // - sequential_position    06 bytes        position in the ms slot (1B ops)
 // - rep_group_id           07 bytes        replication group identifier
 // - other_information      arbitrary       user input, such as a unique string
+//
+// Legacy Base62 VersionID:
+//         timestamp  sequential_position  rep_group_id
+// where:
+// - timestamp              14 bytes        epoch in ms
+// - sequential_position    06 bytes        position in the ms slot
+// - rep_group_id           07 bytes        replication group identifie
+//
+// Base62 VersionID:
+//         timestamp  sequential_position  rep_group_id  instance_id  version_id_format
+// where:
+// - timestamp              14 bytes        epoch in ms
+// - sequential_position    06 bytes        position in the ms slot
+// - rep_group_id           07 bytes        replication group identifier
+// - instance_id            06 bytes        unique instance identifier
+// - version_id_format      02 bytes        version ID format marker + version
 
 import base62Integer from 'base62';
 import baseX from 'base-x';
+import assert from 'assert';
+import { VersioningConstants } from './constants';
 const BASE62 = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ';
 const base62String = baseX(BASE62);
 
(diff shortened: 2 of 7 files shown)


lib/versioning/VersionID.ts
```diff
@@ -353,3 +353,21 @@ export function decode(str: string): string | Error {
...
+ * @param versionId - the versionId to escape
+ * @returns - the escaped versionId
+ */
+export function escapeVersionId(versionId: string): string {
+    if (!versionId) {
+        return versionId;
+    }
+
+    return versionId
+        .replace(/\x1E/g, '\\u001e');
```
ghost: Did you test if the regex is not slower than other approaches? Just to make sure we are not adding perf impact every time we need to process versionIDs (likely in all S3 apis)

lib/versioning/VersioningRequestProcessor.ts
```diff
@@ -392,8 +392,9 @@ export default class VersioningRequestProcessor {
             if (err) {
                 return callback(err);
             }
+            const escapedVid = vid? escapeVersionId(vid) : vid;
```
ghost:
```suggestion
            const escapedVid = vid ? escapeVersionId(vid) : vid;
```

lib/versioning/VersioningRequestProcessor.ts
```diff
@@ -617,8 +618,9 @@ export default class VersioningRequestProcessor {
                 if (err) {
                     return callback(err);
                 }
+                const escapedVid = vid? escapeVersionId(vid) : vid;
```
ghost:
```suggestion
                const escapedVid = vid ? escapeVersionId(vid) : vid;
```

lib/versioning/VersionID.ts
```diff
@@ -294,16 +336,38 @@ export function encode(str: string): string {
...
+ * This is useful for ensuring that the versionId can be safely used in contexts
+ * where control characters may cause issues, such as JSON serialization.
+ * 
+ * @param versionId - the versionId to escape
+ * @returns - the escaped versionId
+ */
+export function escapeVersionId(versionId: string): string {
+    if (!versionId) {
+        return versionId;
+    }
```
ghost:
```suggestion
export function escapeVersionId(versionId?: string): string {
    if (!versionId) {
        return versionId;
    }
```

lib/versioning/VersionID.ts
```diff
@@ -1,29 +1,53 @@
...
-export const versionIdSeed = getVersionIdSeed();
+/**
+ * Check if the S3_VERSION_ID_ENCODING_TYPE is set to 'hex' or not set at all.
+ * If it is, we use the legacy hex encoding for version IDs.
+ *
+ * @return - true if hex encoding type is used, false otherwise
+ */
+export function isHexEncodingType() {
+    return S3_VERSION_ID_ENCODING_TYPE === 'hex' || !S3_VERSION_ID_ENCODING_TYPE;
+}
```
ghost: I think this should just be a const?

lib/versioning/VersionID.ts
```diff
@@ -122,6 +129,21 @@ export function generateVersionId(info: string, replicationGroupId: string): str
...
+    let vidFormat = '';
+
+    if (isHexEncodingType()) {
+        // In HEX encoding, the full info data is used.
+        otherInfo = info;
+    } else {
+        // In base62, info is for the instance ID and is trimmed/padded.
+        instanceIdPadded = padRight(info, TEMPLATE_ID);
+        // versionID format, 2 bytes
+        vidFormat = VersioningConstants.VersionId.FormatMarker + versionIDFormat;
```
ghost: to be consistent with the other places

```suggestion
        vidFormat = `${VersioningConstants.VersionId.FormatMarker}${versionIDFormat}`;
```

Also that also looks like a constant, won't it be always the same? could we precompute it?

lib/versioning/VersionID.ts
```diff
@@ -122,6 +129,21 @@ export function generateVersionId(info: string, replicationGroupId: string): str
...
+        assert(vidFormat.length === LENGTH_FT, `versionID format must be ${LENGTH_FT} bytes`);
```
ghost: do we suspect cases when it's not the case? if the vidFormat is always the same string (because it's a const and another const), shouldn't it be done once during startup, so the system crashes and we know what's wrong? here, we do that for each vID generation, that seems unnecessary

lib/versioning/VersionID.ts
```diff
@@ -269,6 +286,30 @@ export function base62Decode(str: string): string | Error {
...
+ * @returns true if the versionId contains the format marker and version, false otherwise.
+ */
+function hasVersionIDFormat(versionId: string, version: string): boolean {
+    const formatMarkerIdx = versionId.length - LENGTH_FT;
+    const formatMarker = versionId.charAt(formatMarkerIdx);
+    if (formatMarker !== VersioningConstants.VersionId.FormatMarker) {
+        return false; // no format marker
+    }
+    const formatVersion = versionId.substring(formatMarkerIdx + 1);
+    return formatVersion === version;
```
ghost: if we want to avoid creating one variable

```suggestion
    const formatVersion = versionId.substring(formatMarkerIdx + 1) === version;
```

lib/versioning/VersionID.ts
```diff
@@ -269,6 +286,30 @@ export function base62Decode(str: string): string | Error {
...
+}
+
+const LEGACY_BASE62_DECODED_LENGTH = 27;
+const BASE62_DECODED_LENGTH = 35;
+const BASE62_ENCODED_LENGTH = 32;
```
ghost: Better to have them at the top of the file

lib/versioning/VersionID.ts
```diff
@@ -294,16 +336,38 @@ export function encode(str: string): string {
...
     // default format is exactly 32 characters when encoded
-    if (str.length === 32) {
+    if (str.length === BASE62_ENCODED_LENGTH) {
         const decoded: string | Error = base62Decode(str);
-        if (typeof decoded === 'string' && decoded.length !== 27) {
-            return new Error(`decoded ${str} is not length 27`);
+        // Legacy base62 version IDs (without 'info' field) are always 27 characters long.
+        // The new base62 format is always 35 characters long.
+        if (typeof decoded === 'string' &&
+            ![LEGACY_BASE62_DECODED_LENGTH, BASE62_DECODED_LENGTH].includes(decoded.length)) {
```
ghost:
```suggestion
        if (typeof decoded === 'string' && decoded.length !== LEGACY_BASE62_DECODED_LENGTH && decoded.length !== BASE62_DECODED_LENGTH) {
```

for performance

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1155,12 +1162,12 @@ class MongoClientInterface {
             ops.push(masterOp);
             return c.bulkWrite(ops, {
                 ordered: true,
-            }).then(() => cb(null, `{"versionId": "${objVal.versionId}"}`)).catch((err) => {
+            }).then(() => cb(null, `{"versionId": "${escapedVersionID}"}`)).catch((err) => {
```
jonathan-gramain: In those contexts, I'd prefer for ease of maintenance to do directly `JSON.stringify({ versionId: objVal.versionId })` and let the JSON stringifier escape accordingly.

lib/versioning/VersionID.ts
welansari: Yes it should be a constant

lib/versioning/constants.ts
```diff
@@ -10,6 +10,7 @@ export enum BucketVersioningFormat {
 export const VersioningConstants = {
     VersionId: {
         Separator: '\0',
+        FormatMarker: '?',
```
welansari, line 13: **Note for reviewers:**

I updated the versionID format marker from the record separator control character (`\x1E`) to `?` for better compatibility and performance. This way we avoid having to parse the versionID every time we read it from the DB.

The `?` character is safe to use because it never appears in the `info` section of existing hex-encoded versionIDs. In both our metadata implementations, the `info` field contains only numeric values and basic punctuation:

Metadata: `${info.term}.${info.aseq}.${this.uidCounter++}`
MongoDB: `${process.pid}.${uidCounter++}`

Also, the new versionID format is disabled by default, this will allow keeping the same branches and decide when to enabled it through feature flags or other means.

lib/versioning/VersionID.ts
welansari: This was removed

lib/storage/metadata/mongoclient/MongoClientInterface.ts
welansari: I changed to the marker character to avoid having to parse/escape on every request (see the comment below)

lib/versioning/VersionID.ts
```diff
@@ -269,6 +299,31 @@ export function base62Decode(str: string): string | Error {
...
+    if (versionId.length <= LENGTH_TS + LENGTH_SEQ + LENGTH_RG + LENGTH_FT - 1) {
+        return false; // Not enough characters for format marker
+    }
+    // For constant time lookup, we always assume that the format marker is
+    // at the end of the versionId.
+    const formatMarkerIdx = versionId.length - LENGTH_FT;
+    const formatMarker = versionId.charAt(formatMarkerIdx);
+    if (formatMarker !== VersioningConstants.VersionId.FormatMarker) {
+        return false; // no format marker
+    }
```
ghost: Nit: I'm not sure we need to create a variable for just one check

```suggestion
    if (versionId.charAt(formatMarkerIdx) !== VersioningConstants.VersionId.FormatMarker) {
        return false; // no format marker
    }
```

lib/versioning/VersionID.ts
```diff
@@ -122,6 +143,20 @@ export function generateVersionId(info: string, replicationGroupId: string): str
     // replication group ID, like PARIS; will be trimmed if exceed LENGTH_RG
     const repGroupId = padRight(replicationGroupId, TEMPLATE_RG);
 
+    let otherInfo = '';
+    let instanceIdPadded = '';
+    let formatSuffix = '';
+
+    if (S3_VERSION_ID_ENCODING_TYPE === 'hex' || !S3_VERSION_ID_ENCODING_TYPE) {
```
ghost: I would reverse the check order: if the env var is not set, no need to check for the `hex` value

ghost (approved):

lib/versioning/VersionID.ts
```diff
@@ -1,29 +1,67 @@
...
+
+// Flag to enable the new version ID (35 characters) over legacy shortID format (27 characters).
+// When enabled and S3_VERSION_ID_ENCODING_TYPE is 'base62':
+//   - Uses new format: timestamp + sequential_position + rep_group_id + instance_id + version_id_format
+//   - Includes instance_id field to differentiate version IDs across multiple instances in the same k8s cluster
+//   - Appends format marker and version identifier for format detection
+// When disabled and S3_VERSION_ID_ENCODING_TYPE is 'base62':
+//   - Uses old format: timestamp + sequential_position + rep_group_id (legacy 27-char format)
+// Falls back to hex encoding if S3_VERSION_ID_ENCODING_TYPE is 'hex' or unset
+export const ENABLE_FORMATTED_VERSION_ID = process.env.ENABLE_FORMATTED_VERSION_ID === 'true';
```
jonathan-gramain: We also use `1` to enable things via env vars in a lot of places, for least surprise you could also consider allowing `1` as truish.

lib/versioning/VersionID.ts
```diff
@@ -1,29 +1,67 @@
...
+
+// version ID format added to the end of the version ID
+const versionIDFormat = '1';
+const vidFormat = `${VersioningConstants.VersionId.FormatMarker}${versionIDFormat}`;
```
jonathan-gramain: Use all caps for consistency, and suggest renaming:

```suggestion
const VERSION_ID_FORMAT_VERSION = '1';
const VERSION_ID_FORMAT_SUFFIX = `${VersioningConstants.VersionId.FormatMarker}${VERSION_ID_FORMAT_VERSION}`;
```

lib/versioning/VersionID.ts
```diff
@@ -269,6 +299,31 @@ export function base62Decode(str: string): string | Error {
...
+ * @param versionId - The versionId string to check.
+ * @param version - The expected format version.
+ * @returns true if the versionId contains the format marker and version, false otherwise.
+ */
+function hasVersionIDFormat(versionId: string, version: string): boolean {
+    // Format marker can only exist after the required versionId sections.
+    // This check removes the risk of looking for the format marker in the
+    // replication group ID, which can technically contain any character as
+    // it's set by the end user.
+    if (versionId.length <= LENGTH_TS + LENGTH_SEQ + LENGTH_RG + LENGTH_FT - 1) {
```
jonathan-gramain:
```suggestion
    if (versionId.length < LENGTH_TS + LENGTH_SEQ + LENGTH_RG + LENGTH_FT) {
```

jonathan-gramain (approved):

PR #2468, opened 2025-07-31
ARSN-512: Fix bytesReceived being always 0

BourgoisMickael:
Introduced by ARSN-471

Needed in S3C for s3-analytics as logs are ingested by fluent-bit and aggregated into clickhouse. For PUT request we use bytesReceived instead of contentLength for ingress bytes.

### DIFF
--- a/lib/s3routes/routes.ts
+++ b/lib/s3routes/routes.ts
@@ -216,7 +216,7 @@ export default function routes(
         // @ts-ignore
         objectKey: req.objectKey,
         // @ts-ignore
-        bytesReceived: req.parsedContentLength || 0,
+        bytesReceived: req.parsedContentLength || 0, // not defined yet
         // @ts-ignore
         bodyLength: parseInt(req.headers['content-length'], 10) || 0,
     };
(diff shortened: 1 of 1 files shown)


leif-scality (approved):

lib/s3routes/routes.ts
```diff
@@ -216,7 +216,7 @@ export default function routes(
         // @ts-ignore
         objectKey: req.objectKey,
         // @ts-ignore
-        bytesReceived: req.parsedContentLength || 0,
+        bytesReceived: req.parsedContentLength || 0, // not defined yet
```
francoisferrand: not sure I understand this comment.... if it is not defined, then no point using the field here at all?

BourgoisMickael: To leave it at 0 in logs for some routes that doesn't go through normalizeRequest like some `/_/` internal routes.

But I can drop it so it won't appear

BourgoisMickael: There dropped: [141ef2a](https://github.com/scality/Arsenal/pull/2468/commits/141ef2a0aa410a089033184602d7b353ae4f3478)

jonathan-gramain (approved):

PR #2490, opened 2025-08-26
ARSN-515 ✨ Add a new condition "$exists: false" in repairMaster option to manage

DarkIsDude:
the case when the versionId does not exists

Issue: ARSN-515

https://scality.atlassian.net/browse/CLDSRV-632
https://github.com/scality/cloudserver/pull/5913

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -1143,8 +1143,9 @@ class MongoClientInterface {
             const ops: AnyBulkWriteOperation<ObjectMetastoreDocument>[] = [];
             // filter to get master
             const filter = {
-                '_id': masterKey,
-                'value.versionId': {
+                _id: masterKey,
+                $or: [
+                    { 'value.versionId': { $exists: false } },
                     // We break the semantic correctness here with
                     // $gte instead of $gt because we do not have
                     // a microVersionId to capture the micro
@@ -1155,8 +1156,11 @@ class MongoClientInterface {
                     // replication and ingestion can hopefully
                     // ensure), but this would not work e.g. in
                     // the case of an active-active replication.
-                    $gte: mstObjVal!.versionId,
-                },
+
+                    { 'value.versionId': {
+                        $gte: mstObjVal!.versionId,
+                    } },
+                ]
             };
             // values to update master
             const update = {
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -856,16 +856,9 @@ class MongoClientInterface {
         // filter to get master
         const filter = {
             _id: masterKey,
-            $or: [{
-                'value.versionId': {
-                    $exists: false,
-                },
-            },
-            {
-                'value.versionId': {
-                    $gt: objVal.versionId,
-                },
-            },
+            $or: [
+                { 'value.versionId': { $exists: false } },
+                { 'value.versionId': { $gt: objVal.versionId } },
             ],
         };
         // values to update master
@@ -1157,9 +1150,7 @@ class MongoClientInterface {
                     // ensure), but this would not work e.g. in
                     // the case of an active-active replication.
 
-                    { 'value.versionId': {
-                        $gte: mstObjVal!.versionId,
-                    } },
+                    { 'value.versionId': { $gte: mstObjVal!.versionId } },
                 ]
             };
             // values to update master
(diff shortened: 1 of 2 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1143,8 +1143,9 @@ class MongoClientInterface {
             const ops: AnyBulkWriteOperation<ObjectMetastoreDocument>[] = [];
             // filter to get master
             const filter = {
-                '_id': masterKey,
-                'value.versionId': {
+                _id: masterKey,
+                $or: [
+                    { 'value.versionId': { $exists: false } },
```
ghost, line 1141: do we really not have a versionID field?
I know we optimized recently the metadata so we don't store "default" or empty fields, so it could be a reason, but if the version is `null` it doesn't mean the field does not exist, in such case, the `$exists` will not work , as a null value is still a value. Can we verify that we are not missing this case?

DarkIsDude: It's a good question, thanks. So the use case is in this PR https://github.com/scality/cloudserver/pull/5913. So process is the following:
- Create a bucket
- Create an object
- Enable versioning (so you have a master object only)
- Put a metadata with a versionId that does not exist (so not the null one)

In this case, we must copy the null version, then create the new version and finally update the master one. When we update the master one, the versionId does not exists and that's why we added this new condition. I hope it's clear enough ? @williamlardier

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1155,8 +1156,11 @@ class MongoClientInterface {
                     // replication and ingestion can hopefully
                     // ensure), but this would not work e.g. in
                     // the case of an active-active replication.
-                    $gte: mstObjVal!.versionId,
-                },
+
+                    { 'value.versionId': {
+                        $gte: mstObjVal!.versionId,
+                    } },
```
ghost:
```suggestion
                    {
                        'value.versionId': {
                            $gte: mstObjVal!.versionId,
                        },
                    },
```

ghost (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1143,8 +1143,9 @@ class MongoClientInterface {
             const ops: AnyBulkWriteOperation<ObjectMetastoreDocument>[] = [];
             // filter to get master
             const filter = {
-                '_id': masterKey,
-                'value.versionId': {
+                _id: masterKey,
+                $or: [
```
francoisferrand, line 1140: not sure it works, but I was wondering if we could reduce duplication (and maybe make it a bit clearer?) by writing it like this:


```suggestion
                'value.versionId': $or: [
                    { $exists: false },
                    { $gte: mstObjVal!.versionId },
                ]
```

francoisferrand (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.ts
ghost: that should work if wrapped around {} yes

benzekrimaha (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: No, it's not working. Mongo does not know how to work with `$or` in nested field.

```
db.getCollection("test-or").find({
    'value.versionId': {
        $or: [
            { $exists: false },
            { $gte: 1 },
        ]
    }
})
```

But this one is working:

```
db.getCollection("test-or").find({
    $or: [
        {
            'value.versionId': { $gte: 1 }
        },
        {
            'value.versionId': { $exists: false }
        }
    ]
})
```

ghost: maybe you can use $eq for the first case?

something like `{ 'value.versionId': { $eq: { $or: [...] }}`

DarkIsDude: @williamlardier not sure to get your point. Something like that ?

```
db.getCollection("test-or").find({
    'value.versionId': {
        $eq: {
            $or: [
                { $exists: false },
                { $gte: 1 },
            ]
        }
    }
})
```

ghost: yes, but i checked and no the eq do not have any "boolean equality", the $or you have is the best approach I think.
There is the $expr that works, i confirmed it, but this would create aggregations and badly affect perf.

PR #2497, opened 2025-09-03
ARSN-518: Fix ceph test flaky race condition

BourgoisMickael:
```
  1) Object Part Copy
       With default signature
         When copy source was put by MPU
           should not corrupt object if overwriting an existing part by copying a part while the MPU is being completed:
     InternalError: We encountered an internal error. Please try again.
```

```json
{"name":"S3","clientIP":"::ffff:127.0.0.1","clientPort":36678,"httpMethod":"POST","httpURL":"/destinationbucket815502017/copycatobject?uploadId=1cb287f2705d44198930fa6b76c1069f","bucketName":"destinationbucket815502017","objectKey":"copycatobject","bodyLength":185,"bytesReceived":185,"service":"s3","action":"CompleteMultipartUpload","accountDisplayName":"Bart","accountName":"Bart","bucket":"mpuShadowBucketdestinationbucket815502017","object":"1cb287f2705d44198930fa6b76c1069f..|..00001","time":1756894576598,"req_id":"f87f0ac98e45a738d463","level":"debug","message":"internalDeleteObject: object not found or already deleted","hostname":"pkrvm7jw40e0xgp","pid":91}
{"name":"S3","clientIP":"::ffff:127.0.0.1","clientPort":36678,"httpMethod":"POST","httpURL":"/destinationbucket815502017/copycatobject?uploadId=1cb287f2705d44198930fa6b76c1069f","bucketName":"destinationbucket815502017","objectKey":"copycatobject","bodyLength":185,"bytesReceived":185,"service":"s3","action":"CompleteMultipartUpload","accountDisplayName":"Bart","accountName":"Bart","error":"DeleteConflict","bucket":"mpuShadowBucketdestinationbucket815502017","key":"1cb287f2705d44198930fa6b76c1069f..|..00001","time":1756894576598,"req_id":"f87f0ac98e45a738d463","level":"error","message":"deleteObjectNoVer: error deleting object with no version","hostname":"pkrvm7jw40e0xgp","pid":91}
{"name":"S3","clientIP":"::ffff:127.0.0.1","clientPort":36678,"httpMethod":"POST","httpURL":"/destinationbucket815502017/copycatobject?uploadId=1cb287f2705d44198930fa6b76c1069f","bucketName":"destinationbucket815502017","objectKey":"copycatobject","bodyLength":185,"bytesReceived":185,"service":"s3","action":"CompleteMultipartUpload","accountDisplayName":"Bart","accountName":"Bart","bytesSent":213,"httpCode":500,"time":1756894576599,"req_id":"f87f0ac98e45a738d463","elapsed_ms":12.21349,"level":"info","message":"responded with error XML","hostname":"pkrvm7jw40e0xgp","pid":91}
```

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -2028,6 +2028,14 @@ class MongoClientInterface {
                     if (!result || result?.deletedCount != 1) {
                         log.debug('internalDeleteObject: object not found or already deleted',
                             { bucket: bucketName, object: key });
+                        // For MPU shadow bucket objects, if the object is already deleted by another
+                        // concurrent operation (like completeMultipartUpload), treat it as success
+                        // rather than an error to avoid race condition failures
+                        if (bucketName.startsWith(constants.mpuBucketPrefix)) {
+                            log.debug('internalDeleteObject: MPU shadow object already deleted by concurrent operation',
+                                { bucket: bucketName, object: key });
+                            return cb(null, undefined);
+                        }
                         return cb(errors.NoSuchKey);
                     }
                     return cb(null, undefined);
@@ -2107,6 +2115,14 @@ class MongoClientInterface {
                 if (result.deletedCount === 0) {
                     log.debug('internalDeleteObject: object not found or already deleted', 
                         { bucket: bucketName, object: key });
+                    // For MPU shadow bucket objects, if the object is already deleted by another
+                    // concurrent operation (like completeMultipartUpload), treat it as success
+                    // rather than an error to avoid race condition failures
+                    if (bucketName.startsWith(constants.mpuBucketPrefix)) {
+                        log.debug('internalDeleteObject: MPU shadow object already deleted by concurrent operation',
+                            { bucket: bucketName, object: key });
+                        return next(null);
+                    }
                     return next(errors.DeleteConflict);
                 }
                 return next(null);
--- a/tests/functional/metadata/mongodb/raceCondition.spec.js
+++ b/tests/functional/metadata/mongodb/raceCondition.spec.js
@@ -1,0 +1,439 @@
+const async = require('async');
+const assert = require('assert');
+const werelogs = require('werelogs');
+const { MongoMemoryReplSet } = require('mongodb-memory-server');
+const { errors, versioning } = require('../../../../index');
+const logger = new werelogs.Logger('RaceConditionFunctional', 'debug', 'debug');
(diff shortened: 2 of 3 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2028,6 +2028,14 @@ class MongoClientInterface {
                     if (!result || result?.deletedCount != 1) {
                         log.debug('internalDeleteObject: object not found or already deleted',
                             { bucket: bucketName, object: key });
+                        // For MPU shadow bucket objects, if the object is already deleted by another
+                        // concurrent operation (like completeMultipartUpload), treat it as success
+                        // rather than an error to avoid race condition failures
+                        if (bucketName.startsWith(constants.mpuBucketPrefix)) {
```
francoisferrand, line 2034: If I remember correctly when this was done, we actually introduced a change in the behavior -i.e. we now return an error instead of catching the "already deleted" error-, which was needed to properly stop processing if the object was already being deleted.

Even if this is the cause of the flakiness, the proper fix should be in cloudserver: i.e. either we don't properly "catch" this `NoSuchKey` error (and maybe return it directly instead of "eating" it?), or maybe the tests themself do not properly handle this (expected) error...

@williamlardier what do you think?

ghost: I think you are discussing this commit: https://github.com/scality/Arsenal/commit/abd2d430d1af718902257d7e3981f668815008ed#diff-937f2f94dd86348c501c66fee6598a2aa8ff0a64df74808df0b28c1812954c1fR2079

Indeed, we must return a `DeleteConflict` error when the bulk operation failed completely, or succeed but the number of deleted document is 0. Meaning, the document was already deleted. We cannot ignore this error as it might lead to the silent continuation of the API, leading to more data or metadata deletion, even if the original operation failed. This was part of the data loss problems we identified.
Before, we were already returning an error if the batch was failing, but not if the deleted count was 0. In some identified cases this is one of the conditions to create data loss problems, or invalid metadata states.

So I am not sure we can say `treat it as success rather than an error to avoid race condition failures`. The change was purposely done to avoid invalid states due to such race deletions.

We can probably better handle this state at the API level, because the behavior of one API should not be in an abstraction layer such as MongoClient, which should remain independent (I know this is not perfectly the case today).

BourgoisMickael: If there is another fix to do then, I put the ticket https://scality.atlassian.net/browse/ARSN-518 to pod RR and close this PR

PR #2500, opened 2025-09-08
Reduce unnecessary computations

ghost:
- Less logs when not needed
- More usage of consts globally shared
- Better algorithm complexity
- Fail fast
- Some debt management at the same time

Issue: ARSN-521

### DIFF
--- a/lib/auth/v2/checkRequestExpiry.ts
+++ b/lib/auth/v2/checkRequestExpiry.ts
@@ -14,19 +14,19 @@ export default function checkRequestExpiry(timestamp: number, log: RequestLogger
     // timestamp is more than 15 minutes in the future, the request
     // has expired and return errors.RequestTimeTooSkewed
     const currentTime = Date.now();
-    log.trace('request timestamp', { requestTimestamp: timestamp });
-    log.trace('current timestamp', { currentTimestamp: currentTime });
+    log.trace('request and current timestamp', {
+        requestTimestamp: timestamp,
+        currentTimestamp: currentTime,
+    });
 
     const fifteenMinutes = (15 * 60 * 1000);
     if (currentTime - timestamp > fifteenMinutes) {
-        log.trace('request timestamp is not within 15 minutes of current time');
-        log.debug('request time too skewed', { timestamp });
+        log.debug('request timestamp is not within 15 minutes of current time', { timestamp });
         return errors.RequestTimeTooSkewed;
     }
 
     if (currentTime + fifteenMinutes < timestamp) {
-        log.trace('request timestamp is more than 15 minutes into future');
-        log.debug('request time too skewed', { timestamp });
+        log.debug('request timestamp is more than 15 minutes into future', { timestamp });
         return errors.RequestTimeTooSkewed;
     }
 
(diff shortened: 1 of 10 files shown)


lib/auth/v2/checkRequestExpiry.ts
```diff
@@ -14,19 +15,19 @@ export default function checkRequestExpiry(timestamp: number, log: RequestLogger
...
     const currentTime = Date.now();
-    log.trace('request timestamp', { requestTimestamp: timestamp });
-    log.trace('current timestamp', { currentTimestamp: currentTime });
-
-    const fifteenMinutes = (15 * 60 * 1000);
-    if (currentTime - timestamp > fifteenMinutes) {
-        log.trace('request timestamp is not within 15 minutes of current time');
-        log.debug('request time too skewed', { timestamp });
+    if (currentTime - timestamp > fifteenMinutesInMilliseconds) {
+        log.debug('request timestamp is not within 15 minutes of current time', {
```
francoisferrand: nit: keeping the "original" debug log maybe more friendly for support, if they are used to searching for `request time too skewed`

```suggestion
        log.debug('request time too skewed, timestamp not within 15 minutes of current time', {
```

lib/auth/v2/headerAuthCheck.ts
```diff
@@ -16,14 +18,13 @@ export function check(
...
+    if (token && !iamSecurityTokenPattern.test(token)) {
         log.debug('invalid security token', { token });
         return { err: errors.InvalidToken };
     }
 
     // Check to make sure timestamp is within 15 minutes of current time
-    let timestamp = headers['x-amz-date'] ?
-        headers['x-amz-date'] : headers.date;
+    let timestamp = headers['x-amz-date'] || headers.date;
     timestamp = Date.parse(timestamp);
```
francoisferrand: Make the variable `const` and avoid changing type:

```suggestion
    const timestamp = Date.parse(headers['x-amz-date'] || headers.date);
```

francoisferrand (approved):

lib/constants.ts
```diff
@@ -9,6 +9,12 @@ export const vaultGeneratedIamSecurityTokenSizeMax = 8192;
 // Base-64
 export const vaultGeneratedIamSecurityTokenPattern = /^[A-Za-z0-9/+=]*$/;
 
+// 15 minutes in seconds for request expiry
+export const requestExpirySeconds = 900;
+export const fifteenMinutesInMilliseconds = 900_000;
+
+export const unsignedPayload = 'UNSIGNED-PAYLOAD';
```
BourgoisMickael: If it's used only in 1 place, I'd not move it to a global constant file but keep it defined in the specific file it's being used.

Otherwise I'd prefix it to know where the variable relates to, like `authChecksumUnsignedPayload`

lib/auth/v4/queryAuthCheck.ts
```diff
@@ -86,14 +86,12 @@ export function check(
     // string 'UNSIGNED-PAYLOAD' should be
     // added to the canonicalRequest in
     // building string to sign
-    const payloadChecksum = 'UNSIGNED-PAYLOAD';
```
BourgoisMickael, line 90: move the comment above on the line `        payloadChecksum: constants.unsignedPayload,`

BourgoisMickael (approved):

PR #2501, opened 2025-09-08
Type requests in Arsenal

ghost:
The main goal is to prepare for s3 typescript migration, but also some upcoming optimizations. We use the request object today to store various info: this is convenient because we pass the request almost everywhere. However this is today not typed and not standard to modify that type. So we introduce an `ArsenalRequest` type where we properly define the scope and fields of our requests.

Issue: ARSN-523


welansari (approved):

BourgoisMickael (approved):
:clap:

PR #2502, opened 2025-09-08
Compute IP and Scheme only once

ghost:
Issue: ARSN-522

### DIFF
--- a/lib/policyEvaluator/requestUtils.ts
+++ b/lib/policyEvaluator/requestUtils.ts
@@ -1,6 +1,6 @@
 import * as ipCheck from '../ipCheck';
-import { IncomingMessage } from 'http';
 import { TLSSocket } from 'tls';
+import { type ArsenalRequest } from '../types/ArsenalRequest';
 
 export interface S3Config {
     requests: {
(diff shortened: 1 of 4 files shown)


DarkIsDude (approved):

lib/policyEvaluator/requestUtils.ts
```diff
@@ -32,27 +33,45 @@ export function getClientIp(request: IncomingMessage, s3config?: S3Config): stri
...
 
 /**
- * getHttpProtocolSecurity - Dete²object
+ * getClientIp - Gets the client IP from the request
+ * @param request - http request object
+ * @param s3config - s3 config
+ * @return - returns client IP from the request
+ */
+export function getClientIp(request: ArsenalRequest, s3config?: S3Config): string {
+    return getIpAndHttpProtocolSecurity(request, s3config).ip;
```
DarkIsDude: We still call it twice here `getIpAndHttpProtocolSecurity` ? Should we save the result in the request and, if it's in the request object, return it from there instead of reparsing ?

DarkIsDude (approved):

lib/policyEvaluator/requestUtils.ts
ghost: I renamed the function and now we call it once (we keep the two "getter" functions because we still have code using it, so we'll migrate first and then this can be removed, as we'll only use the request object, but likely this will come later when we migrate cloudserver to typescript).

ghost: This changed a bit the PR so please feel free to dismiss your approval if needed

welansari (approved):

lib/policyEvaluator/requestUtils.ts
```diff
@@ -32,27 +33,51 @@ export function getClientIp(request: IncomingMessage, s3config?: S3Config): stri
...
             // are case-insensitive.
             const ipFromHeader = request.headers[extractClientIPFromHeader.toLowerCase()]?.toString();
+            let finalIp = clientIp;
             if (ipFromHeader && ipFromHeader.trim().length) {
-                return ipFromHeader.split(',')[0].trim();
+                finalIp = ipFromHeader.split(',')[0].trim();
+            }
+            let isSecure = false;
+            if (extractProtocolFromHeader) {
+                isSecure = request.headers[extractProtocolFromHeader.toLowerCase()] === 'https';
```
BourgoisMickael: that `toLowerCase` can it be handled instead directly in cloudserver config to either validate it's lowercase or convert it to lower case (https://github.com/scality/cloudserver/blob/4738e0dd1492980e0faac8532e5b455e67c59516/lib/Config.js#L483-L511) so it's done only once at startup time ?

BourgoisMickael (approved):

lib/policyEvaluator/requestUtils.ts
ghost: Created https://github.com/scality/cloudserver/pull/5939. I'll merge the cloudserver one before this PR, to ensure we do not end up in a situation where we are not lowercasing anymore.

PR #2538, opened 2025-10-02
ARSN-528: add BucketLoggingStatus to BucketInfo model

leif-scality:
add BucketLoggingStatus to BucketInfo model


lib/models/BucketLoggingStatus.ts
```diff
@@ -0,0 +1,137 @@
...
+    static async fromXML(
+        data: string,
+    ): Promise<{ error?: { arsenalError: ArsenalError, details: any }; res?: BucketLoggingStatus; }> {
+        let parsed;
+        try {
+            parsed = await parseStringPromise(data);
+        } catch (err) {
+            return {
+                error: { arsenalError: errors.MalformedXML, details: err },
+                res: undefined,
```
jonathan-gramain: I assume we probably don't need to add an undefined `res` property (if we do, then maybe we need to revisit how we interpret the results to avoid being too javascript-centric).

jonathan-gramain (commented):
LGTM (I can review again in the final version)

lib/models/BucketInfo.ts
```diff
@@ -11,6 +11,7 @@ import { ACL as OACL } from './ObjectMD';
 import { areTagsValid, BucketTag } from '../s3middleware/tagging';
 import { VeeamCapability, VeeamSOSApiSchema, VeeamSOSApiSerializable } from './Veeam';
 import { AzureInfoMetadata } from './BucketAzureInfo';
+import BucketLoggingStatus from './BucketLoggingStatus';
 
 // WHEN UPDATING THIS NUMBER, UPDATE BucketInfoModelVersion.md CHANGELOG
```
BourgoisMickael, line 16: You should update the doc https://github.com/scality/Arsenal/blob/development/8.2/documentation/BucketInfoModelVersion.md

and bump modelVersion = 18

lib/models/BucketLoggingStatus.ts
```diff
@@ -0,0 +1,137 @@
...
+         </SimplePrefix>
+      </TargetObjectKeyFormat>
+      <TargetPrefix>string</TargetPrefix>
+   </LoggingEnabled>
+</BucketLoggingStatus>
+*/
+
+export type LoggingEnabled = {
+    TargetBucket: string;
+    TargetPrefix: string;
```
BourgoisMickael, line 44: should we include optional `TargetGrants` and `TargetObjectKeyFormat` ? https://docs.aws.amazon.com/AmazonS3/latest/API/API_LoggingEnabled.html

lib/models/BucketLoggingStatus.ts
```diff
@@ -0,0 +1,137 @@
...
+                    error: { arsenalError: errors.MalformedXML, details: 'missing TargetPrefix tag in LoggingEnabled' },
+                    res: undefined,
+                };
+            }
+
+            if (parsed.BuckerLoggingStatus.LoggingEnabled.TargetGrants) {
+                return {
+                    error: {
+                        arsenalError: errors.NotImplemented,
+                        details: 'TargetGrants tag in LoggingEnabled is not implemented',
```
BourgoisMickael:
```suggestion
                        details: 'TargetGrants field in LoggingEnabled is not implemented',
```

lib/models/BucketLoggingStatus.ts
```diff
@@ -0,0 +1,137 @@
...
+            }
+
+            if (!parsed.BucketLoggingStatus.LoggingEnabled.TargetPrefix) {
+                return {
+                    error: { arsenalError: errors.MalformedXML, details: 'missing TargetPrefix tag in LoggingEnabled' },
+                    res: undefined,
+                };
+            }
+
+            if (parsed.BuckerLoggingStatus.LoggingEnabled.TargetGrants) {
```
dvasilas:
```suggestion
            if (parsed.BucketLoggingStatus.LoggingEnabled.TargetGrants) {
```

lib/models/BucketLoggingStatus.ts
```diff
@@ -0,0 +1,137 @@
...
+                return {
+                    error: {
+                        arsenalError: errors.NotImplemented,
+                        details: 'TargetGrants tag in LoggingEnabled is not implemented',
+                    },
+                    res: undefined,
+                };
+            }
+
+            if (parsed.BuckerLoggingStatus.LoggingEnabled.TargetObjectKeyFormat) {
```
dvasilas:
```suggestion
            if (parsed.BucketLoggingStatus.LoggingEnabled.TargetObjectKeyFormat) {
```

lib/models/BucketLoggingStatus.ts
```diff
@@ -0,0 +1,137 @@
...
+                return {
+                    error: {
+                        arsenalError: errors.NotImplemented,
+                        details: 'TargetObjectKeyFormat tag in LoggingEnabled is not implemented',
```
BourgoisMickael: How is details going to be used ? Instead should we rather add a custom description directly on the error, to be returned to the client: `errorInstances.NotImplemented.customDescription('...')`

lib/models/BucketLoggingStatus.ts
BourgoisMickael: nvm I see they are not implemented

BourgoisMickael: Maybe add a comment in here to describe that those 2 fields are not implemented

lib/models/BucketLoggingStatus.ts
BourgoisMickael, line 1: You can generate a unit test with AI to covers these

lib/models/BucketLoggingStatus.ts
BourgoisMickael: is it aws that use `tag` instead of `field` in their error description ?

lib/models/BucketLoggingStatus.ts
```diff
@@ -0,0 +1,137 @@
...
+                    },
+                    res: undefined,
+                };
+            }
+
+            loggingEnabled = {
+                TargetPrefix: parsed.BucketLoggingStatus.LoggingEnabled.TargetPrefix[0],
+                TargetBucket: parsed.BucketLoggingStatus.LoggingEnabled.TargetBucket[0],
```
BourgoisMickael: Aws performs no validation on those fields ?

Because bucket name is pretty restrictive format, and prefix should probably not go beyond a certains size

leif-scality: Maybe but the docs mentions no checks https://docs.aws.amazon.com/AmazonS3/latest/API/API_LoggingEnabled.html

lib/models/BucketInfo.ts
leif-scality: Where should I do the bump?

lib/models/BucketLoggingStatus.ts
leif-scality: done

lib/models/BucketLoggingStatus.ts
leif-scality: changed, XML uses tag and element but we can use field

lib/models/BucketLoggingStatus.ts
leif-scality: fixed

lib/models/BucketLoggingStatus.ts
leif-scality: fixed

lib/models/BucketInfo.ts
BourgoisMickael: 1 or 2 line below there is the variable

documentation/BucketInfoModelVersion.md
```diff
@@ -21,7 +21,8 @@
 | v15     | -                       | Tags                     | ↪️ Forward  |
 | v16     | -                       | Capabilities(VeeamSOSApi)| ✅ New      |
 | v17     | -                       | QuotaMax                 | ✅ New      |
-| **Unified at v17**                                                         |
+| v18     | -                       | BucketLoggingStatus      | ✅ New      |
+| **Unified at v18**                                                         |
```
BourgoisMickael: This is not needed the divergence stops at v17, with unification.

But include a documentation for v18 at the end of the file

lib/models/BucketLoggingStatus.ts
BourgoisMickael: it should be changed everywhere

lib/models/BucketInfo.ts
leif-scality: done

lib/models/BucketLoggingStatus.ts
leif-scality:
```js
    const parsed = BucketLoggingStatus.fromXML(request.post);
    if (parsed.error) {
        log.trace('error processing request', { error: parsed.error, method: 'bucketPutLogging' });
        monitoring.promMetrics('PUT', bucketName, parsed.error.arsenalError, 'bucketPutLogging');
        return callback(parsed.error.arsenalError);
    }
```

lib/models/BucketLoggingStatus.ts
leif-scality: done

documentation/BucketInfoModelVersion.md
leif-scality: removed the v18 line and added description at the bottom of the file

lib/models/BucketLoggingStatus.ts
BourgoisMickael: Maybe we could check on aws what's the behavor if we pass invalid bucket name or prefix

lib/s3routes/routes/routePUT.ts
```diff
@@ -208,6 +208,13 @@ export default function routePUT(
                     return routesUtils.responseNoBody(err, resHeaders,
                         response, 200, log);
                 });
+        } else if (query.logging !== undefined) {
+            api.callApiMethod('objectPutLogging', request, response, log,
```
BourgoisMickael: This is in the object branching, shouldn't it be bucketPutLogging and in the put bucket branching ?

tests/unit/models/BucketLoggingStatus.spec.js
```diff
@@ -0,0 +1,246 @@
...
+            const loggingEnabled = {
+                TargetBucket: 'target-bucket',
+                TargetPrefix: 'app/logs/',
+            };
+            const config = new BucketLoggingStatus(loggingEnabled);
+            const result = config.getLoggingEnabled();
+            assert.strictEqual(result.TargetBucket, 'target-bucket');
+            assert.strictEqual(result.TargetPrefix, 'app/logs/');
+        });
+    });
```
BourgoisMickael: duplicate test with just above constructor describe

tests/unit/models/BucketLoggingStatus.spec.js
```diff
@@ -0,0 +1,246 @@
...
+                assert(result.res instanceof BucketLoggingStatus);
+                const loggingEnabled = result.res.getLoggingEnabled();
+                assert.strictEqual(loggingEnabled.TargetBucket, 'loggingbucket');
+                assert.strictEqual(loggingEnabled.TargetPrefix, 'my-app-logs/2025/');
+            });
+
+            it('should parse XML without LoggingEnabled (empty config)', () => {
+                const xml = '<?xml version="1.0" encoding="UTF-8"?>'+
+                '<BucketLoggingStatus xmlns="http://s3.amazonaws.com/doc/2006-03-01">'+
+                '</BucketLoggingStatus>';
```
BourgoisMickael: include an additional test with autoclosing tag to make sure we handle both cases as the autoclosing tag is documented

```xml
<BucketLoggingStatus xmlns="http://s3.amazonaws.com/doc/2006-03-01" />
```

lib/s3routes/routes/routeGET.ts
BourgoisMickael, line 1: Include a unit test for this additions in routeGET.spec.js and routePUT.spec.js

lib/s3routes/routes/routePUT.ts
leif-scality: fixed

tests/unit/models/BucketLoggingStatus.spec.js
leif-scality: removed

lib/s3routes/routes/routeGET.ts
leif-scality: done

tests/unit/models/BucketLoggingStatus.spec.js
leif-scality: done

lib/models/BucketLoggingStatus.ts
leif-scality: removed `res: undefined` lines, les is only initialized when no error occurred

lib/models/BucketLoggingStatus.ts
leif-scality: done, sorry missed some

lib/models/BucketLoggingStatus.ts
leif-scality: AWS does not check the TargetBucket format, it just returns that the bucket does not exist, cloudserver does this check so we are ok.

AWS returns `An error occurred (InvalidTargetBucketForLogging) when calling the PutBucketLogging operation: The target bucket for logging does not exist` when sending an invalid S3 TargetBucket name
```
{
    "LoggingEnabled": {
        "TargetBucket": "SCALITY_bucket-xqwertyuiopasdfghjkl",
        "TargetPrefix": "amzn-s3-demo-bucket-logs/"
    }
}
```
with `aws s3api put-bucket-logging --bucket leif-test-scality --bucket-logging-status file://../BucketLogging/logging.json  --debug`

Invalid name pulled from https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html

dvasilas (approved):
Reminder to bump Arsenal in Vault, because Vault needs to use the new actionMap.

lib/models/BucketLoggingStatus.ts
BourgoisMickael: Ok, and what about an invalid prefix, like a very large prefix beyond 1024 characters, because this might prevent any key from being written into target bucket

BourgoisMickael: Btw I tested with an bucket name below 3 chars and it returns this error:
```
An error occurred (InvalidBucketName) when calling the PutBucketLogging operation: The specified bucket is not valid.
```

So it seems the bucket name check is not very consistent it only checks for a min size of 3 char, but not invalid characters or max size.

Edit: for some reason, The InvalidBucketName is returned for bucket name above 255 bytes

BourgoisMickael: I checked a TargetPrefix of 1024 characters, there is this error:
```
An error occurred (InvalidArgument) when calling the PutBucketLogging operation: Field exceeds 800 bytes
```

So TargetPrefix should be limited to max 800 bytes

lib/models/BucketLoggingStatus.ts
```diff
@@ -0,0 +1,156 @@
...
+            if (!loggingEnabledData.TargetBucket) {
+                return {
+                    error: {
+                        arsenalError: errors.MalformedXML,
+                        details: 'missing TargetBucket field in LoggingEnabled',
+                    },
+                };
+            }
+
+            if (!loggingEnabledData.TargetPrefix) {
```
BourgoisMickael: we should authorize an empty string for TargetPrefix

leif-scality: done

lib/models/BucketLoggingStatus.ts
```diff
@@ -0,0 +1,185 @@
...
+                return {
+                    error: {
+                        arsenalError: errors.MalformedXML,
+                        details: 'missing TargetPrefix field in LoggingEnabled',
+                    },
+                };
+            } else if (loggingEnabledData.TargetPrefix[0].length > 800) {
+                return {
+                    error: {
+                        arsenalError: errors.InvalidArgument.customizeDescription("Field exceeds 800 bytes"),
```
BourgoisMickael: use `errorInstance` instead of errors with `customizeDescription`.

For InvalidArgument you can also use `.addMetadataEntry` to provide more detail about the invalid argument to match

```
<Error><Code>InvalidArgument</Code><Message>Field exceeds 800 bytes</Message><ArgumentName>TargetPrefix</ArgumentName><ArgumentValue>aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa</ArgumentValue><RequestId>JC31ANY4DBB8WDJB</RequestId><HostId>439BCfJVeOlGoi/bKUpL15wjFZs7620DBAM3SDhsH3VyU+CO4n2p+FWMsx/NzzsS5xhWE2U5P/J3P52ueLVGvKR4T+KdUsXU</HostId></Error>
```

lib/models/BucketLoggingStatus.ts
```diff
@@ -0,0 +1,185 @@
...
+                loggingEnabledData.TargetPrefix === null ||
+                loggingEnabledData.TargetPrefix === undefined
+            ) {
+                return {
+                    error: {
+                        arsenalError: errors.MalformedXML,
+                        details: 'missing TargetPrefix field in LoggingEnabled',
+                    },
+                };
+            } else if (loggingEnabledData.TargetPrefix[0].length > 800) {
```
BourgoisMickael: prefer to use constants to document those values

leif-scality: done

lib/models/BucketLoggingStatus.ts
leif-scality: done

lib/models/BucketLoggingStatus.ts
BourgoisMickael: I'd prefer to have them directly in the file `lib/models/BucketLoggingStatus.ts` because they won't be used anywhere else, they don't need to be global

leif-scality: done

lib/models/BucketLoggingStatus.ts
```diff
@@ -0,0 +1,193 @@
...
+                        details: 'missing TargetPrefix field in LoggingEnabled',
+                    },
+                };
+            } else if (loggingEnabledData.TargetPrefix[0].length > bucket_logging_status_target_prefix_max_length) {
+                return {
+                    error: {
+                        arsenalError: errorInstances.InvalidArgument.
+                            customizeDescription(
+                                `Field exceeds ${bucket_logging_status_target_prefix_max_length} bytes`).
+                            addMetadataEntry('TargetPrefix', loggingEnabledData.TargetPrefix[0]),
```
BourgoisMickael: There is the correct usage:

```suggestion
                            addMetadataEntry('invalidArguments', [{ ArgumentName: 'TargetPrefix', ArgumentValue: loggingEnabledData.TargetPrefix[0] }]),
```

cf example usage: https://github.com/scality/cloudserver/pull/4497/files

Note this won't match perferctly with aws it will return `ArgumentName1` and `ArgumentValue1`, but I guess it's ok

BourgoisMickael: Dot should be at the beginning of the line, not the end.

Weird the formatter or linter did not catch that

```suggestion
                        arsenalError: errorInstances.InvalidArgument
                            .customizeDescription(
                                `Field exceeds ${bucket_logging_status_target_prefix_max_length} bytes`)
                            .addMetadataEntry('TargetPrefix', loggingEnabledData.TargetPrefix[0]),
```

lib/models/BucketLoggingStatus.ts
```diff
@@ -0,0 +1,193 @@
+import { parseString } from 'xml2js';
+import errors, { ArsenalError, errorInstances } from '../errors';
+
+/** BucketLoggingStatus constants, not documented by AWS but found via testing */
+const bucket_logging_status_target_bucket_min_length = 3;
+const bucket_logging_status_target_bucket_max_length = 255;
+const bucket_logging_status_target_prefix_max_length = 800;
```
BourgoisMickael: because they are specific to this field you don't need to prefix with bucket logging.
And I'd say uppercase is prefered for constants

```suggestion
const TARGET_BUCKET_MIN_LENGTH = 3;
const TARGET_BUCKET_MAX_LENGTH = 255;
const TARGET_PREFIX_MAX_LENGTH = 800;
```

BourgoisMickael (approved):

lib/models/BucketLoggingStatus.ts
leif-scality: done

lib/models/BucketLoggingStatus.ts
leif-scality: done

leif-scality: done

BourgoisMickael (approved):

leif-scality:
ping

leif-scality:
ping

leif-scality:
ping

PR #2554, opened 2025-10-20
ARSN-531: record server access log fields in s3routes

leif-scality:
Extract fields to be used by cloudserver to write the server access logs

### DIFF
--- a/lib/s3routes/routesUtils.ts
+++ b/lib/s3routes/routesUtils.ts
@@ -14,6 +14,23 @@ const jsutil = require('../jsutil');
 
 const ALLOW_INVALID_META_HEADERS = !!process.env.ALLOW_INVALID_META_HEADERS;
 
+function storeServerAccessLogFields(
+    res: http.ServerResponse,
+    endTurnAroundTime: bigint,
+    errorCode?: string,
+    bytesSent?: number,
+) {
+    // @ts-expect-error
+    if (res.serverAccessLog) {
+        // @ts-expect-error
+        res.serverAccessLog.errorCode = errorCode;
+        // @ts-expect-error
+        res.serverAccessLog.endTurnAroundTime = endTurnAroundTime;
+        // @ts-expect-error
+        res.serverAccessLog.bytesSent = bytesSent;
+    }
+}
+
 export type CallApiMethod = (
     methodName: string,
     request: http.IncomingMessage,
@@ -97,6 +114,7 @@ export function okHeaderResponse(
     setCommonResponseHeaders(headers, response, log);
     log.debug('response http code', { httpCode });
     response.writeHead(httpCode);
+    storeServerAccessLogFields(response, process.hrtime.bigint());
     return response.end(() => {
         log.end().info('responded to request', {
             httpCode: response.statusCode,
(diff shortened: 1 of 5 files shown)


lib/s3routes/routesUtils.ts
```diff
@@ -1140,7 +1174,7 @@ export function normalizeRequest(
     // the x-amz-decoded-content-length
     const contentLength = request.headers['x-amz-decoded-content-length'] ?
         request.headers['x-amz-decoded-content-length'] :
-        request.headers['content-length'];
+        request.headers['content-length']; // here
```
dvasilas:
```suggestion
        request.headers['content-length'];
```

lib/s3routes/routesUtils.ts
```diff
@@ -480,6 +504,8 @@ export function retrieveData(
                 // in the callback of data.get()? The 'close' event is not
                 // called if 'end' is.
                 currentStream = readable;
+
+                storeServerAccessLogFields(response, process.hrtime.bigint());
```
dvasilas: I think this is not in the correct place.
It is inside the `eachSeries` for location, so if there are multiple locations, it will be called multiple times.

It may be better to place it inside:
```
if (!response.headersSent) {
    response.writeHead(httpCodeOnSucess);
}
`

File: lib/s3routes/routesUtils.ts
Comment (dvasilas) [inline lib/s3routes/routesUtils.ts:1]: `retrieveDataAzure` is out of context, but you could add a TODO comment saying that it does not capture log access log fields in case this feature is ever used in Artesca.

File: lib/s3routes/routesUtils.ts (hunk 2 above)
Comment (dvasilas) [reply lib/s3routes/routesUtils.ts]: Also (more important), shouldn't this include `bytesSent`? It is used by `objectGet`.

File: lib/s3routes/routesUtils.ts (hunk 1 above)
Comment (leif-scality) [reply lib/s3routes/routesUtils.ts]: removed

File: lib/s3routes/routesUtils.ts (hunk 2 above)
Comment (leif-scality) [reply lib/s3routes/routesUtils.ts]: fixed, calculated the `bytesSent` using `_computeContentLengthFromLocation`

File: lib/s3routes/routesUtils.ts
Comment (leif-scality) [reply lib/s3routes/routesUtils.ts:1]: added the comment

File: lib/s3routes/routesUtils.ts [hunk 3]
```diff
@@ -474,16 +501,24 @@ export function retrieveData(
...
                 currentStream = readable;
+
                 return readable.pipe(response, { end: false });
             }
         ), err => {
             currentStream = null;
+            storeServerAccessLogFields(
+                response,
+                first_byte_sent_timestamp,
+                err?.name || undefined,
```
dvasilas: Why do we use `err.name` instead of `err.message`  that we do in other cases?

lib/s3routes/routesUtils.ts
```diff
@@ -415,6 +440,8 @@ export function retrieveData(
         _destroyReadable(currentStream);
     });
 
+    let first_byte_sent_timestamp: bigint = process.hrtime.bigint();
```
dvasilas, line 443: Isn't here too early to set a value to `first_byte_sent_timestamp` ?

leif-scality: It get reset in line 504 if we enter the loop, else we go directly into the error case

lib/s3routes/routesUtils.ts
leif-scality: fixed

dvasilas (approved):

nicolas2bert (approved):

leif-scality:
ping

PR #2560, opened 2025-10-21
ARSN-533: Get raft session ID from metadata on getBucketAttributes and getBucketAndObjectMD

dvasilas:
Related to: https://github.com/scality/bucketclient/pull/282

### DIFF
--- a/lib/storage/metadata/MetadataWrapper.js
+++ b/lib/storage/metadata/MetadataWrapper.js
@@ -217,14 +217,14 @@ class MetadataWrapper {
 
     getBucket(bucketName, log, cb) {
         log.debug('getting bucket from metadata');
-        this.client.getBucketAttributes(bucketName, log, (err, data) => {
+        this.client.getBucketAttributes(bucketName, log, (err, data, raftSessionId) => {
             if (err) {
                 log.debug('error from metadata', { implName: this.implName,
                     error: err });
                 return cb(err);
             }
             log.trace('bucket retrieved from metadata');
-            return cb(err, BucketInfo.fromObj(data));
+            return cb(err, BucketInfo.fromObj(data), raftSessionId);
         });
     }
 
@@ -282,15 +282,15 @@ class MetadataWrapper {
         log.debug('getting bucket and object from metadata',
             { database: bucketName, object: objName });
         this.client.getBucketAndObject(bucketName, objName, params, log,
-            (err, data) => {
+            (err, data, raftSessionId) => {
                 if (err) {
                     log.debug('error from metadata', { implName: this.implName,
                         err });
                     return cb(err);
                 }
                 log.debug('bucket and object retrieved from metadata',
-                    { database: bucketName, object: objName });
-                return cb(err, data);
+                    { database: bucketName, object: objName, raftSessionId });
+                return cb(err, data, raftSessionId);
             });
     }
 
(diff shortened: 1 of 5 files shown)


lib/models/BucketInfo.ts
```diff
@@ -80,13 +80,15 @@ export type BucketMetadata = {
     capabilities?: Capabilities,
     quotaMax: bigint | number,
     bucketLoggingStatus?: BucketLoggingStatus,
+    raftSessionId?: number,
```
jonathan-gramain: Do we really have to store this info in bucket metadata? It doesn't look like a good place to me, since it's not a property of the bucket, but extra info that we want the caller to know. I think ideally we should pass it along in a side channel (can be an extra callback arg, or within the `BucketInfo` returned by `getBucketAndObject` such as `{ bucket, object, raftSessionId }`)

dvasilas: I initially chose to add it to `BucketInfo` to reduce the footprint of change (we could retrieve the raft session ID from the bucket info without having to modify `getBucketAttributes` and `getBucketAndObject` callbacks.

But indeed, this doesn't fit with bucket metadata. Fixed.

lib/storage/metadata/MetadataWrapper.js
```diff
@@ -282,15 +282,15 @@ class MetadataWrapper {
...
         this.client.getBucketAndObject(bucketName, objName, params, log,
-            (err, data) => {
+            (err, data, raftSessionId) => {
                 if (err) {
                     log.debug('error from metadata', { implName: this.implName,
                         err });
                     return cb(err);
                 }
                 log.debug('bucket and object retrieved from metadata',
                     { database: bucketName, object: objName });
```
BourgoisMickael: Could be included in the log debug line

tests/unit/storage/metadata/MetadataWrapper.spec.js
```diff
@@ -0,0 +1,160 @@
...
+            metadataWrapper.client = mockClient;
+        });
+
+        it('should propagate raftSessionId as third callback parameter when present', (done) => {
+            mockClient.getBucketAttributes.callsFake((bucketName, log, callback) => {
+                process.nextTick(() => callback(null, testBucketMD, testRaftSessionId));
+            });
+
+            metadataWrapper.getBucket(bucketName, mockLog, (err, bucketInfo, raftSessionId) => {
+                assert.strictEqual(err, null);
```
BourgoisMickael: Should rather be assert.ifError

BourgoisMickael (approved):

tests/unit/storage/metadata/MetadataWrapper.spec.js
dvasilas: Done, thanks.

lib/storage/metadata/MetadataWrapper.js
dvasilas: Good idea, done.

nicolas2bert (approved):

lib/storage/metadata/MetadataWrapper.js
```diff
@@ -282,15 +282,15 @@ class MetadataWrapper {
...
+            (err, data, raftSessionId) => {
                 if (err) {
                     log.debug('error from metadata', { implName: this.implName,
                         err });
                     return cb(err);
                 }
                 log.debug('bucket and object retrieved from metadata',
-                    { database: bucketName, object: objName });
-                return cb(err, data);
+                    { database: bucketName, object: objName, raftSessionId: raftSessionId });
```
jonathan-gramain:
```suggestion
                    { database: bucketName, object: objName, raftSessionId });
```

jonathan-gramain (approved):

PR #2566, opened 2025-10-29
Test github review / process

DarkIsDude:


francoisferrand (changes requested):
this PR is way too large, please split in multiple commits.

package.json
```diff
@@ -17,36 +17,36 @@
...
-    "JSONStream": "^1.0.0",
-    "level": "~1.6.0",
-    "level-sublevel": "~6.6.1",
-    "node-forge": "^0.7.1",
-    "simple-glob": "^0.1",
-    "socket.io": "~1.7.3",
-    "socket.io-client": "~1.7.3",
-    "utf8": "2.1.2",
-    "uuid": "^3.0.1",
+    "ajv": "6.10.0",
```
benzekrimaha: can we have a commit / package , that way we can track in case a bump has a breaking change and it's more clear

PR #2571, opened 2025-11-10
✨(ARSN-538) Accept undefined prefix for ReplicationConfiguration

DarkIsDude:
https://scality.atlassian.net/browse/ARSN-538

### DIFF
--- a/lib/models/ReplicationConfiguration.ts
+++ b/lib/models/ReplicationConfiguration.ts
@@ -45,7 +45,7 @@ export type Rule = {
 
 export type Destination = { StorageClass: string[]; Bucket: string };
 export type XMLRule = {
-    Prefix: string[];
+    Prefix?: string[];
     Status: Status[];
     ID?: string[];
     Destination: Destination[];
@@ -155,7 +155,7 @@ export default class ReplicationConfiguration {
     _buildRuleObject(rule: XMLRule) {
         const base = {
             id: '',
-            prefix: rule.Prefix[0],
+            prefix: rule.Prefix?.[0] ?? '',
             enabled: rule.Status[0] === 'Enabled',
         };
         const obj: Rule = { ...base };
@@ -238,6 +238,7 @@ export default class ReplicationConfiguration {
                 'Number of defined replication rules cannot exceed 1000'
             );
         }
+
         const err = this._parseEachRule(Rule);
         if (err) {
             return err;
@@ -257,9 +258,11 @@ export default class ReplicationConfiguration {
                 this._parsePrefix(rules[i]) ||
                 this._parseID(rules[i]) ||
                 this._parseDestination(rules[i]);
+
             if (err) {
                 return err;
             }
+
             rulesArr.push(this._buildRuleObject(rules[i]));
         }
         this._rules = rulesArr;
@@ -283,16 +286,18 @@ export default class ReplicationConfiguration {
      * @param rule - The rule object from this._parsedXML
      */
     _parsePrefix(rule: XMLRule) {
-        const prefix = rule.Prefix && rule.Prefix[0];
-        // An empty string prefix should be allowed.
-        if (!prefix && prefix !== '') {
+        const prefix = rule.Prefix?.[0] ?? '';
+
+        if (typeof prefix !== 'string') {
             return errors.MalformedXML;
         }
+
         if (prefix.length > 1024) {
             return errorInstances.InvalidArgument.customizeDescription(
                 'Rule prefix cannot be longer than maximum allowed key length of 1024'
             );
         }
+    
         // Each Prefix in a list of rules must not overlap. For example, two
         // prefixes 'TaxDocs' and 'TaxDocs/2015' are overlapping. An empty
         // string prefix is expected to overlap with any other prefix.
@@ -304,6 +309,7 @@ export default class ReplicationConfiguration {
                 );
             }
         }
+
         this._configPrefixes.push(prefix);
         return undefined;
     }
--- a/lib/models/ReplicationConfiguration.ts
+++ b/lib/models/ReplicationConfiguration.ts
@@ -286,6 +286,10 @@ export default class ReplicationConfiguration {
      * @param rule - The rule object from this._parsedXML
      */
     _parsePrefix(rule: XMLRule) {
+        if (Array.isArray(rule.Prefix) && rule.Prefix.length > 1) {
+            return errors.MalformedXML;
+        }
+
         const prefix = rule.Prefix?.[0] ?? '';
 
         if (typeof prefix !== 'string') {
--- a/tests/unit/models/ReplicationConfiguration.spec.ts
+++ b/tests/unit/models/ReplicationConfiguration.spec.ts
@@ -114,7 +114,6 @@ describe('ReplicationConfiguration.parseConfiguration()', () => {
         const rules = instance.getRules();
         expect(rules.length).toEqual(1);
         expect(rules[0].enabled).toBe(true);
-        // should have generated a new random ID
         expect(typeof rules[0].id).toBe('string');
         expect(rules[0].prefix).toEqual('');
     });
--- a/tests/unit/models/ReplicationConfiguration.spec.ts
+++ b/tests/unit/models/ReplicationConfiguration.spec.ts
@@ -95,6 +95,30 @@ describe('ReplicationConfiguration.parseConfiguration()', () => {
         expect(rules[0].prefix).toEqual('');
     });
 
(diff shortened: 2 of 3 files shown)


tests/unit/models/ReplicationConfiguration.spec.ts
```diff
@@ -95,6 +95,29 @@ describe('ReplicationConfiguration.parseConfiguration()', () => {
         expect(rules[0].prefix).toEqual('');
     });
 
+    it('should succeed for a valid configuration without a prefix', () => {
```
SylvainSenechal: Add William as a reviewer as it's touching the db, I'm not too sure I can review confidently. 
Also, is it documented on aws that replication configurations may not have a prefix anymore ? Can it impact some logic elsewhere if prefix is sometimes removed ?

DarkIsDude: I don't think so, with Typescript I hope we are safe in Arsenal 🙏

lib/models/ReplicationConfiguration.ts
```diff
@@ -283,16 +286,18 @@ export default class ReplicationConfiguration {
      * @param rule - The rule object from this._parsedXML
      */
     _parsePrefix(rule: XMLRule) {
-        const prefix = rule.Prefix && rule.Prefix[0];
-        // An empty string prefix should be allowed.
-        if (!prefix && prefix !== '') {
+        const prefix = rule.Prefix?.[0] ?? '';
+
+        if (typeof prefix !== 'string') {
```
francoisferrand, line 295: when can this happen?
`rule.Prefix` is types as optional `string[]`, then we it gets a default if null/undefined : so it may only be a string....

tests/unit/models/ReplicationConfiguration.spec.ts
```diff
@@ -95,6 +95,29 @@ describe('ReplicationConfiguration.parseConfiguration()', () => {
...
+        expect(result).toBeUndefined();
+        expect(instance.getRole()).toEqual(TEST_ROLE);
+        expect(instance.getDestination()).toEqual('arn:aws:s3:::crr-dest');
+        const rules = instance.getRules();
+        expect(rules.length).toEqual(1);
+        expect(rules[0].enabled).toBe(true);
+        expect(typeof rules[0].id).toBe('string');
+        expect(rules[0].prefix).toEqual('');
+    });
+
```
francoisferrand, line 234: Should have a test validating that we return malformed xml when the "prefix is invalid" (whatever it means) - or return that test if not relevant anymore (c.f. earlier comment).

the expression used to get prefix in `_parsePrefix` (and later malformed xml condition) is not trivial, would be better to make this plain with unit tests for `_parsePrefix` : which show exactly what "works" and what does not, and covers the 4 code-path (I think) in
```javascript
const prefix = rule.Prefix?.[0] ?? '';
if (typeof prefix !== 'string') {
```

lib/models/ReplicationConfiguration.ts
```diff
@@ -155,7 +155,7 @@ export default class ReplicationConfiguration {
     _buildRuleObject(rule: XMLRule) {
         const base = {
             id: '',
-            prefix: rule.Prefix[0],
+            prefix: rule.Prefix?.[0] ?? '',
```
francoisferrand, line 158: this is duplicated with `_parsePrefix`, can we avoid this and do the "parsing" just once?

(though `parsePrefix` name is actually incorrect, it really just validates the prefix...)

francoisferrand (approved):

lib/models/ReplicationConfiguration.ts
DarkIsDude: Maybe the typing is not the right one, but prefix can be an array, an object. Indeed, at this step we just parsed the XML without any validation. That's why we have this check. Does it make sense ?

tests/unit/models/ReplicationConfiguration.spec.ts
DarkIsDude: Make sense 🙏. I added them

lib/models/ReplicationConfiguration.ts
benzekrimaha: Indeed no need to check for the type here , can only be a string if it's not undefined/null

DarkIsDude: @benzekrimaha are you double sure ? This is accepted with the current version:

```
<?xml version="1.0" encoding="UTF-8"?>
<ReplicationConfiguration
	xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
	<Role>arn:aws:iam::003641996138:role/scality-internal/storage-manager-role</Role>
	<Role>arn:aws:iam::003641996138:role/scality-internal/storage-manager-role</Role>
	<Rule>
		<ID>testfdg</ID>
		<Prefix>
			<Test>OUI</Test>
		</Prefix>
		<Filter>
			<Prefix>fd</Prefix>
		</Filter>
		<Status>Enabled</Status>
		<Destination>
			<Bucket>arn:aws:s3:::aa-bb-2</Bucket>
		</Destination>
	</Rule>
</ReplicationConfiguration>
```

When I guess it should not. That's why we have this check (you can check new tests).

SylvainSenechal: Side question, with this request, would (and should) prefix be equal to String('<Test>OUI</Test>') ?  

Also other side question, why is XMLRule.Prefix an array ? It doesn't make much sense, on top of that we see in the function that we are getting the first element only. What am I missing 🤔

DarkIsDude: That good questions ! For the first one I tried with AWS and the request is rejected in that case, so I guess the answer is no, the request should be rejected (what I did). The second part it's how the xml parser works. In that case for example the Test will be an array of length 1 and a string OUI. Should I add one more check to make sure const prefix length is one and only one ?

lib/models/ReplicationConfiguration.ts
SylvainSenechal: The caller of _buildRuleObject is already doing all the parsing verifications. I think it would be complicated to use _parsePrefix in this function, as it would now have to return an error.

What I would do is rename the rule parameter like this : 

```suggestion
    _buildRuleObject(rule: XMLRule) {
    _buildRuleObject(validRule: XMLRule) {
```

SylvainSenechal (approved):
approved but review the few comments from the others

lib/models/ReplicationConfiguration.ts
SylvainSenechal: Test is up to you, but no need if you wanna merge now. 
Ah yeah the xml parser parse everything as an array

benzekrimaha (approved):

tests/unit/models/ReplicationConfiguration.spec.ts
```diff
@@ -69,6 +69,143 @@ function getPreferredReadXMLConfig(hasPreferredRead) {
...
+            const result = instance.parseConfiguration();
+            expect(result).toEqual(errors.InvalidArgument);
+        });
+
+        it('should return MalformedXML if a prefix is an array', () => {
+            const repConfig = {
+                Rule: [
+                    {
+                        ID: ['rule1'],
+                        Prefix: [['']],
```
francoisferrand: this works in javascript, but I don't see what XML would lead to this : prefix is an array because there could be multiple `<prefix>` tag (possible in XML, though not allowed in this case).

The content of each prefix cannot be an "array", only either a string... or maybe an object (if nested XML tags are used)

test shouldould be added for multiple prefixes though:
```suggestion
                        Prefix: ["foo", "bar"],
```
(i.e. in XML : `<prefix>foo</prefix><prefix>bar</prefix>`)

tests/unit/models/ReplicationConfiguration.spec.ts
```diff
@@ -69,6 +69,143 @@ function getPreferredReadXMLConfig(hasPreferredRead) {
...
+            const result = instance.parseConfiguration();
+            expect(result).toEqual(errors.MalformedXML);
+        });
+
+        it('should return MalformedXML if a prefix is an object', () => {
+            const repConfig = {
+                Rule: [
+                    {
+                        ID: ['rule1'],
+                        Prefix: [{}],
```
francoisferrand: same not possible in XML, the object should have keys...

```suggestion
                        Prefix: [{
                        	Foo: 'bar',
                        }],
```
(i.e. in XML: `<prefix><foo>bar</foo></prefix>`)

francoisferrand (approved):

PR #2576, opened 2025-12-02
ARSN-540: export rate limit API action map and bump `package.json` to 8.2.41

anurag4DSB:
Per requirements we should be able to call the  bucket rate limit APIs without
assuming role (like SUR admin APIs) using the normal credentials of a service user. For the same, we
need access to the rate limit action map in the cloudserver using this export.


anurag4DSB:
ping

francoisferrand (approved):

nicolas2bert (approved):

PR #2579, opened 2025-12-05
adding tests for externalclients and gcpservice

benzekrimaha:
Issue: ARSN-524


tests/unit/storage/data/DummyService.js
```diff
@@ -88,6 +88,13 @@ class AzureDummyContainerClient {
             readableStreamBody: new DummyObjectStream(offset, length || OBJECT_SIZE),
         };
     }
+
+    async deleteIfExists() {
+        if (this.key === 'externalBackendTestBucket/externalBackendMissingKey') {
+            return { succeeded: false };
+        }
+        return { succeeded: true };
+    }
```
DarkIsDude, line 96: I don't know how you imagine your futur but you can do :

`return { succeeded: this.key !== 'externalBackendTestBucket/externalBackendMissingKey' }`

tests/unit/storage/data/external/GcpService.spec.js
```diff
@@ -108,7 +108,7 @@ const invalidDnsBucketNames = [
 
 function invalidDnsBucketNameHandler() {
     return (req, res) => {
-        assert(req.headers.host, host);
+         assert(req.headers.host, host);
```
DarkIsDude: Indent issue ?

tests/unit/storage/data/external/GcpService.spec.js
```diff
@@ -329,3 +329,131 @@ describe('GcpService dnsStyle tests', () => {
...
+            assert.ifError(err);
+            assert.deepStrictEqual(res.TagSet, [
+                { Key: 'owner', Value: 'arsenal' },
+                { Key: 'color', Value: 'blue' },
+            ]);
+            done();
+        });
+    });
+
+    it('createMultipartUpload should reject missing parameters', done => {
```
DarkIsDude:
```suggestion
    it('createMultipartUpload should reject if parameters are missing', done => {
```

DarkIsDude: same for next one

DarkIsDude (approved):

lib/storage/data/external/GCP/GcpService.js
francoisferrand: does not seem to be used either

lib/storage/data/external/GCP/GcpService.js
```diff
@@ -616,6 +569,6 @@ function gcpSigningMiddleware(config) {
     };
 }
 
-Object.assign(GcpClient.prototype, gcpApis);
+Object.assign(GcpService.prototype, gcpApis);
```
francoisferrand, line 450: can we move this to the constructor of GcpService, to make it easier to notice?

(or even rewrite this without `Object.assign` : it was done to "extend" the generated class, but now the class is manually written: so we may more cleanly add the methods?)

lib/storage/data/external/GCP/GcpService.js
francoisferrand: does not seem to be used
(please double check every single "legacy" method in here, everything we drop is something we don't have to maintain anymore)

francoisferrand: does not seem to be called?

francoisferrand: don't seem to be called?

tests/unit/storage/data/external/GcpService.spec.js
```diff
@@ -347,6 +364,435 @@ describe('GcpService dnsStyle tests', () => {
...
+            Key: 'object',
+        }, err => {
+            assert(err);
+            assert(err.is.InvalidRequest);
+            done();
+        });
+    });
+});
+
+describe('GcpService mpu helper behavior', () => {
```
francoisferrand: should be split in multiple tests: for each separate file (GcpService & each file in GcpApis)

tests/unit/storage/data/external/GcpService.spec.js
```diff
@@ -300,7 +317,7 @@ describe('GcpService pathStyle tests', () => {
     });
 
     operations.forEach(test => it(`GCP::${test.op}`, done => {
-        client[test.op](test.params, err => {
+        callOperation(client, test.op, test.params, err => {
```
francoisferrand: these test execute the code (i.e. give some code coverage), but do not verify that code behaves as expected: so not so useful... should add some assertions on the result or calls it makes...

tests/unit/storage/data/external/GcpService.spec.js
```diff
@@ -181,6 +184,20 @@ const operations = [
...
+            .then(() => cb(null))
+            .catch(err => cb(err));
+    }
+    if (op === 'listVersions') {
+        return client.send(new ListObjectVersionsCommand(params))
+            .then(() => cb(null))
+            .catch(err => cb(err));
+    }
+    return client[op](params, cb);
+}
```
francoisferrand: what is the point to introduce such a wrapper?
it really just means that listObejcts and listObject should not be handled through the same `operations` table : either as extra (separate) tests, or through another table which would contain the type object to create or object to send...

tests/unit/storage/data/external/GcpService.spec.js
```diff
@@ -347,6 +364,435 @@ describe('GcpService dnsStyle tests', () => {
...
+            dataStoreName: 'unit-location',
+        });
+    });
+
+    afterEach(() => {
+        jest.restoreAllMocks();
+    });
+
+    it('putObjectTagging should merge tags into metadata', done => {
+        jest.spyOn(client, 'headObject')
```
francoisferrand, line 496: we usually use `Sinon.js` for mocks : any reason to switch to `jest` ? or is it not the case in arsenal?

(nothing against jest mocks in itself, but consistency is a good thing... especially with untyped languages like JS)

tests/unit/storage/data/external/GcpService.spec.js
```diff
@@ -347,6 +364,435 @@ describe('GcpService dnsStyle tests', () => {
...
+            bucketName: 'unit-bucket',
+            dataStoreName: 'unit-location',
+        });
+    });
+
+    afterEach(() => {
+        jest.restoreAllMocks();
+    });
+
+    it('abortMultipartUpload should reject missing parameters', done => {
```
francoisferrand: use `describe` to provide context, so that the text like looks like a proper sentence: "it should reject missing parameters"

```suggestion
describe('abortMultipartUpload', () => {
   it('should reject missing parameters', done => {
   
   [...]
})

describe('completeMultipartUpload', () => {
   [...]
```

(same for all these test)

tests/unit/storage/data/external/GcpService.spec.js
```diff
@@ -347,6 +364,435 @@ describe('GcpService dnsStyle tests', () => {
...
+                Bucket: params.Bucket,
+                MPU: params.MPU,
+                Prefix: createMpuKey(params.Key, params.UploadId),
+            }, expect.any(Function));
+            done();
+        });
+    });
+
+    describe('removeParts', () => {
+        it('should list versions via send(ListObjectVersionsCommand) and delete each version', done => {
```
francoisferrand:
```suggestion
        it('should list object versions and delete them all', done => {
```

tests/unit/storage/data/external/GcpService.spec.js
```diff
@@ -347,6 +364,435 @@ describe('GcpService dnsStyle tests', () => {
...
+            const helper = new MpuHelper(service);
+            helper.removeParts({ MPU: 'mpu-bucket', Prefix: 'pfx/' }, err => {
+                assert.ifError(err);
+                expect(service.send).toHaveBeenCalledTimes(1);
+                expect(service.deleteObject).toHaveBeenCalledTimes(2);
+                done();
+            });
+        });
+
+        it('should return error when send(ListObjectVersionsCommand) rejects', done => {
```
francoisferrand:
```suggestion
        it('should return error when list objects command fails', done => {
```

tests/unit/storage/data/external/GcpService.spec.js
```diff
@@ -347,6 +364,435 @@ describe('GcpService dnsStyle tests', () => {
...
+                    Prefix: createMpuKey(params.Key, params.UploadId, 'parts'),
+                    Marker: createMpuKey(params.Key, params.UploadId,
+                        params.PartNumberMarker, 'parts'),
+                    MaxKeys: params.MaxParts,
+                });
+                done();
+            });
+        });
+
+        it('should return error when command rejects', done => {
```
francoisferrand:
```suggestion
        it('should return error when command fails', done => {
```

lib/storage/data/external/GCP/GcpService.js
benzekrimaha: Keeping GcpApis separate avoids a noisy/huge diff and makes it easier to iterate on these highly customized APIs in isolation. I kept the Object.assign(GcpService.prototype, gcpApis) at module scope (not in the constructor) because putting it in the constructor would re-run for every new GcpService(), repeatedly overwriting prototype methods.

tests/unit/storage/data/external/GcpService.spec.js
benzekrimaha: Switched to sinon

lib/storage/data/external/GCP/GcpService.js
SylvainSenechal: Yeah not much we can do here. Maybe replace this line with something explicit
```suggestion
GcpClient.prototype.abortMultipartUpload = gcpApis.abortMultipartUpload;
GcpClient.prototype.completeMultipartUpload = gcpApis.completeMultipartUpload;
GcpClient.prototype.createMultipartUpload = gcpApis.createMultipartUpload;
...
...
...
```

This way it won't be 'hidden' into a single line, but not sure its super nice either

tests/unit/storage/data/DummyService.js
```diff
@@ -88,6 +88,12 @@ class AzureDummyContainerClient {
             readableStreamBody: new DummyObjectStream(offset, length || OBJECT_SIZE),
         };
     }
+
+    async deleteIfExists() {
```
SylvainSenechal, line 92: I don't see this function being used anywhere 🤔 except once in code that was already existing in azureClient test. How come the test didn't fail without this function

tests/unit/storage/data/external/ExternalClients.spec.js
```diff
@@ -167,6 +185,16 @@ describe('external backend clients', () => {
             assert.strictEqual(errorHandled, true);
         });
 
+        it(`${backend.name} delete() should delete the requested key without error`, async () => {
+            const key = 'externalBackendTestKey';
```
SylvainSenechal, line 189: nit: seems these key/bucket names are important as they are also hardcoded in the dummy service, but we could consider using constants dummyKey/dummyBucketName defined in dummy service and importing them

SylvainSenechal (approved):

lib/storage/data/external/GCP/GcpService.js
francoisferrand: > putting it in the constructor would re-run for every new GcpService(), repeatedly overwriting prototype methods.

```
   constructor() {
     Object.assign(this, gcpApis);
   }
```

putting it in the constructor would not overwrite prototype method: only the "current" object's methods.
Would that be an issue?

francoisferrand (approved):
with this, are we ready for 8.3.0 release?
- if we don't have anything, should bump in this PR so we can immediately release
- or align with @DarkIsDude to release once https://github.com/scality/Arsenal/pull/2581 is merged (on 8.2, with 8.3 waterflow), merge this one "as is", and do the bump #2581.

PR #2581, opened 2025-12-17
✨ ARSN-544: Add new custom permission for Optional Object Attributes

DarkIsDude:
https://scality.atlassian.net/browse/ARSN-544

https://scality.atlassian.net/browse/VAULT-668

https://scality.atlassian.net/browse/CLDSRV-812

Must be bumped in Vault2 and Cloudserver

### DIFF
--- a/lib/policyEvaluator/utils/actionMaps.ts
+++ b/lib/policyEvaluator/utils/actionMaps.ts
@@ -6,6 +6,7 @@ const sharedActionMap = {
     bucketDeleteWebsite: 's3:DeleteBucketWebsite',
     bucketDeleteTagging: 's3:PutBucketTagging',
     bucketGet: 's3:ListBucket',
+    listObjectsV2OptionalAttributes: 'scality:ListBucketOptionalObjectAttributes',
     bucketGetACL: 's3:GetBucketAcl',
     bucketGetCors: 's3:GetBucketCORS',
     bucketGetEncryption: 's3:GetEncryptionConfiguration',
--- a/lib/policyEvaluator/utils/actionMaps.ts
+++ b/lib/policyEvaluator/utils/actionMaps.ts
@@ -6,7 +6,6 @@ const sharedActionMap = {
     bucketDeleteWebsite: 's3:DeleteBucketWebsite',
     bucketDeleteTagging: 's3:PutBucketTagging',
     bucketGet: 's3:ListBucket',
-    listObjectsV2OptionalAttributes: 'scality:ListBucketOptionalObjectAttributes',
     bucketGetACL: 's3:GetBucketAcl',
     bucketGetCors: 's3:GetBucketCORS',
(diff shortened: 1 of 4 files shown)


SylvainSenechal (approved):

francoisferrand (approved):

lib/storage/metadata/MetadataWrapper.js
```diff
@@ -341,7 +359,7 @@ class MetadataWrapper {
 
     listObject(bucketName, listingParams, log, cb) {
         if (listingParams.listingType === undefined) {
-            // eslint-disable-next-line
+
```
francoisferrand, line 362: extra empty line to be removed (also below)

lib/storage/metadata/MetadataWrapper.js
```diff
@@ -25,9 +25,26 @@ let bucketNotificationHook;
 function _parseListEntries(entries) {
     return entries.map(entry => {
         const tmp = JSON.parse(entry.value);
+
+        const userMetadata = Object.keys(tmp)
```
francoisferrand, line 29: I see we "fill" the content of listing both here & in cloudserver: for better performance (avoid duplicated checks and extra work) and to reduce coupling, I wonder if "filtering" of optional attributes should not be done here, just once ?

francoisferrand (approved):

lib/policyEvaluator/utils/actionMaps.ts
```diff
@@ -6,6 +6,7 @@ const sharedActionMap = {
     bucketDeleteWebsite: 's3:DeleteBucketWebsite',
     bucketDeleteTagging: 's3:PutBucketTagging',
     bucketGet: 's3:ListBucket',
+    listObjectsV2OptionalAttributes: 'scality:ListBucketOptionalObjectAttributes',
```
SylvainSenechal: this is very nit, but it looks like this map should be used for s3:x actions only, i see other people put their scality:x action in other maps under this one

lib/storage/metadata/MetadataWrapper.js
```diff
@@ -25,6 +25,22 @@ let bucketNotificationHook;
...
```
SylvainSenechal, line 29: Adjacent to Francois's comment, I haven't read all the prs so maybe later it will make more sense, but I'm a bit surprised : aren't these new user metadata supposed to be returned conditionnally ?
I was expecting something like 
```
function _parseListEntries(entries, needMetadata) {
    if (needMetadata) {
        // get user metadata logic
    }
}
```

I haven't investigated too much, maybe it's too complicated to do because _parseListEntries is used everywhere ? Or maybe not worth it ? I wonder if there are situations where the user metadata are big and there could be a perf impac, since I guess this function is used quite often on every list objects and probably a few other apis ?

PR #2584, opened 2026-01-06
Improve streaming logic for get request with sdk v3

SylvainSenechal:
Issue: ARSN-546

### DIFF
--- a/lib/storage/data/external/AwsClient.js
+++ b/lib/storage/data/external/AwsClient.js
@@ -18,7 +18,6 @@ const { S3Client,
     NotFound } = require('@aws-sdk/client-s3');
 const { Upload } = require('@aws-sdk/lib-storage');
 const werelogs = require('werelogs');
-const { Readable } = require('stream');
 const errors = require('../../../errors').default;
 const errorInstances = require('../../../errors').errorInstances;
 const MD5Sum = require('../../../s3middleware/MD5Sum').default;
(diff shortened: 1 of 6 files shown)


lib/storage/data/external/GCP/GcpService.js
```diff
@@ -354,9 +354,11 @@ class GcpClient extends S3Client {
                 if (objectGeneration) {
                     result.VersionId = objectGeneration;
                 }
-                return callback && callback(null, result);
+                return callback(null, result);
             })
-            .catch(err => callback?.(err));
+            .catch(err => {
+                return callback(err);
```
DarkIsDude, line 360: Why did you remove the optional chaining ?

package.json
```diff
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=20"
   },
-  "version": "8.2.41",
+  "version": "8.3.0",
```
DarkIsDude, line 6: Why this bump ?

lib/storage/data/external/GCP/GcpService.js
SylvainSenechal: Well, the question is more why was it here in the first place. I think it was just too defensive, the tests passed without optional callback. I also suspect this function is not used at all, but can't confirm 100%. Will recheck with Maha

DarkIsDude: The goal was maybe in someone using it as it's a service (CloudServer?). It's make it compatible with promise and callback ? Ok to wait Maha if you prefer

SylvainSenechal: Will need to recheck with Maha, but I believe these api aren't even used
From what I understand, there was a gcp-2017-11-01.api.json file, that was defining a set of api in a similar way as what we had for backbeat client, and it had to be migrated for v3, so she migrated all the apis from this .json file, but some of them aren't used. 
But the diff is here is not really removing something from what we currently have in prod, it's removing from what was added in the migration pr. And I really do believe that anyway these are always used with a callback when they are used

package.json
SylvainSenechal: I wanna use it directly in Cloudserver after merging, so that we can also update Zenko and see rapidly if there is an issue

tests/unit/s3routes/routesUtils/responseStreamData.spec.js
```diff
@@ -177,32 +176,28 @@ describe('routesUtils.responseStreamData', () => {
...
-                    const nOpenSockets = Object.keys(awsAgent.sockets).length;
-                    assert.strictEqual(nOpenSockets, 0);
-                    doneOnce();
-                }, 1000),
-                isclosed: true,
-            }, undefined, logger.newRequestLogger());
-        } catch (err) {
-            doneOnce(err);
-        }
+        responseStreamData(undefined, {}, {}, [{
```
SylvainSenechal, line 179: try / catch was added on Maha's pr : not needed, reverting to what we have on dev branch

lib/storage/data/external/AwsClient.js
```diff
@@ -121,62 +120,6 @@ class AwsClient {
...
-    _normalizeSdkStream(body) {
-        if (!body) {
-            return Readable.from([]);
-        }
-
-        const isReadable = Readable.isReadable
-            ? Readable.isReadable(body)
-            : body instanceof Readable || typeof body?.pipe === 'function';
-
-        if (isReadable) {
```
SylvainSenechal, line 133: Did a getObject against cloudserver within a separate script : body is always readable in nodeJs, and it will just always return here. This function could be useful if the code was ran on a Web engine

lib/storage/data/external/AwsClient.js
```diff
@@ -316,36 +257,14 @@ class AwsClient {
...
-                            backendType: this.clientType,
-                            dataStoreName: this._dataStoreName,
-                            error: destroyErr,
-                        });
-                    }
-                    return stream;
+                    return originalDestroy(err);
                 };
 
-                stream.abort = abortRequest;
```
SylvainSenechal, line 345: The stream is manipulated in the function retrieveData() from routesUtils.ts.
The only function we need is destroy() and pipe(). Pipe is already available on the stream, here we just override destroy() to add the abortController logic

DarkIsDude (approved):

francoisferrand (approved):

PR #2588, opened 2026-01-26
Support the new API GetObjectAttributes

maeldonn:
Issue: ARSN-549


DarkIsDude (commented):
No a lot to add, IA spotted them

delthas (approved):
LGTM apart from AI comments

francoisferrand (commented):
see comments from Copilot

DarkIsDude (approved):

package.json
```diff
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=20"
   },
-  "version": "8.2.43",
+  "version": "8.2.44",
```
francoisferrand: please don't bump just yet: bump should be done at the "last minute" before merging, to avoid conflicts..

Also there are multiple PRs to merge (as discussed in slack), so need to align and "order" the PR so we avoid making a release for every PR (and many useless releases, accordingly)

francoisferrand (approved):

package.json
maeldonn: Ok, I dropped the bump commit.

delthas (approved):

DarkIsDude (approved):

SylvainSenechal (approved):

lib/policyEvaluator/utils/actionMaps.ts
```diff
@@ -58,6 +60,10 @@ const actionMapBucketGet = {
     listObjectsV2OptionalAttributes: 'scality:ListBucketOptionalObjectAttributes',
 };
 
+const actionMapObjectGetAttributes = {
+    objectGetAttributesWithUserMetadata: 'scality:GetObjectAttributes',
```
francoisferrand: - nit: `With` does not really add any value to the name, so best keep it short(er)
- should the string (right side) not also explicit what it is used for (ie the difference between s3:GetObjectAttributes and scality:GetObjectAttributes) ?

```suggestion
    objectGetAttributesUserMetadata: 'scality:GetObjectAttributesUserMetadata',
```
or if we wanted to have a permissions with larger scope, for "all" custom (non-standard) attributes we would add:
```suggestion
    objectGetAttributesCustom: 'scality:GetObjectAttributesCustom',
```

maeldonn: updated to scality:GetObjectAttributesCustom

delthas (approved):

PR #2590, opened 2026-02-09
ARSN-551: don't require x-amz-content-sha256 header to be signed

leif-scality:
In practice AWS does not require `x-amz-content-sha256` to be signed. So we remove the requirement


leif-scality:
ping

BourgoisMickael (approved):

dvasilas (approved):

leif-scality:
ping

PR #2591, opened 2026-02-11
ARSN-552: don't throw in case of bad Date inputs

leif-scality:
Don't throw and crash cloudserver in case of an invalid Date header

Repro code:
```
curl -v -H "Date: BAD_DATE" -H "Authorization: AWS4-HMAC-SHA256 Credential=accessKey1/20260211/us-east-1/s3/aws4_request, SignedHeaders=host, Signature=d459d5b2a2395b4c65d8f8aa2729b22c5abb04614fafbd93ab4fe203e76d21a3" -H "X-Amz-Content-Sha256: fa8d015f89da2a769d1cea7e3bd77a5670d098d7844cda148a40c1304e5b778b" localhost:8000/leif-us-east-1/test-file.txt
```

```
2026-02-11T09:33:06.467Z: uncaughtException:
RangeError: Invalid time value
    at Date.toISOString (<anonymous>)
    at convertUTCtoISO8601 (/app/node_modules/arsenal/build/lib/auth/v4/timeUtils.js:26:43)
    at Object.check [as headers] (/app/node_modules/arsenal/build/lib/auth/v4/headerAuthCheck.js:98:57)
    at extractParams (/app/node_modules/arsenal/build/lib/auth/auth.js:122:47)
    at Object.doAuth (/app/node_modules/arsenal/build/lib/auth/auth.js:140:17)
    at /app/lib/api/api.js:238:33
    at nextTask (/app/node_modules/async/dist/async.js:5327:14)
    at Object.waterfall (/app/node_modules/async/dist/async.js:5337:5)
    at processRequest (/app/lib/api/api.js:237:22)
    at Object.callApiMethod (/app/lib/api/api.js:418:16)
{"name":"S3","time":1770802386467,"error":"Invalid time value","stack":"RangeError: Invalid time value\n    at Date.toISOString (<anonymous>)\n    at convertUTCtoISO8601 (/app/node_modules/arsenal/build/lib/auth/v4/timeUtils.js:26:43)\n    at Object.check [as headers] (/app/node_modules/arsenal/build/lib/auth/v4/headerAuthCheck.js:98:57)\n    at extractParams (/app/node_modules/arsenal/build/lib/auth/auth.js:122:47)\n    at Object.doAuth (/app/node_modules/arsenal/build/lib/auth/auth.js:140:17)\n    at /app/lib/api/api.js:238:33\n    at nextTask (/app/node_modules/async/dist/async.js:5327:14)\n    at Object.waterfall (/app/node_modules/async/dist/async.js:5337:5)\n    at processRequest (/app/lib/api/api.js:237:22)\n    at Object.callApiMethod (/app/lib/api/api.js:418:16)","level":"fatal","message":"caught error","hostname":"xps","pid":886}
```

### DIFF
--- a/lib/auth/auth.ts
+++ b/lib/auth/auth.ts
@@ -216,7 +216,9 @@ function generateV4Headers(
     payload?: string,
 ) {
     Object.assign(request, { headers: {} });
-    const amzDate = convertUTCtoISO8601(Date.now());
+    // Date.now() should always return a valid date so we assert non null.
+    const amzDate = convertUTCtoISO8601(Date.now())!;
+
     // get date without time
     const scopeDate = amzDate.slice(0, amzDate.indexOf('T'));
     const region = 'us-east-1';
--- a/lib/auth/v4/headerAuthCheck.ts
+++ b/lib/auth/v4/headerAuthCheck.ts
@@ -6,6 +6,7 @@ import {
     checkTimeSkew,
     convertUTCtoISO8601,
     convertAmzTimeToMs,
+    isValidISO8601Compact,
 } from './timeUtils';
 import {
     extractAuthItems,
@@ -82,20 +83,19 @@ export function check(
     // check request timestamp
     const xAmzDate = request.headers['x-amz-date'];
     if (xAmzDate) {
-        const xAmzDateArr = xAmzDate.split('T');
-        // check that x-amz- date has the correct format and after epochTime
-        if (xAmzDateArr.length === 2 && xAmzDateArr[0].length === 8
-          && xAmzDateArr[1].length === 7
-          && Number.parseInt(xAmzDateArr[0], 10) > 19700101) {
-            // format of x-amz- date is ISO 8601: YYYYMMDDTHHMMSSZ
-            timestamp = request.headers['x-amz-date'];
+        if (isValidISO8601Compact(xAmzDate)) {
+            timestamp = xAmzDate;
         }
     } else if (request.headers.date) {
-        timestamp = convertUTCtoISO8601(request.headers.date);
+        if (isValidISO8601Compact(request.headers.date)) {
+            timestamp = request.headers.date;
+        } else {
+            timestamp = convertUTCtoISO8601(request.headers.date);
+        }
     }
     if (!timestamp) {
         log.debug('missing or invalid date header',
-            { method: 'auth/v4/headerAuthCheck.check' });
+            { 'method': 'auth/v4/headerAuthCheck.check', 'x-amz-date': xAmzDate, 'Date': request.headers.date });
         return { err: errorInstances.AccessDenied.
             customizeDescription('Authentication requires a valid Date or ' +
           'x-amz-date header') };
--- a/lib/auth/v4/timeUtils.ts
+++ b/lib/auth/v4/timeUtils.ts
@@ -15,15 +15,33 @@ export function convertAmzTimeToMs(timestamp: string) {
 }
 
 /**
- * Convert UTC timestamp to ISO 8601 timestamp
- * @param timestamp of UTC form: Fri, 10 Feb 2012 21:34:55 GMT
- * @return ISO8601 timestamp of form: YYYYMMDDTHHMMSSZ
- */
-export function convertUTCtoISO8601(timestamp: string | number) {
-    // convert to ISO string: YYYY-MM-DDTHH:mm:ss.sssZ.
-    const converted = new Date(timestamp).toISOString();
-    // Remove "-"s and "."s and milliseconds
-    return converted.split('.')[0].replace(/-|:/g, '').concat('Z');
+* Convert UTC timestamp to ISO 8601 compact format
+* @param timestamp - UTC timestamp (e.g., 'Fri, 10 Feb 2012 21:34:55 GMT') or Unix timestamp
+* @return ISO8601 timestamp of form YYYYMMDDTHHMMSSZ, or undefined if invalid
+*
+* @example
+* convertUTCtoISO8601('Fri, 10 Feb 2012 21:34:55 GMT'); // '20120210T213455Z'
+* convertUTCtoISO8601(1328910895000); // '20120210T213455Z'
+* convertUTCtoISO8601('invalid'); // undefined
+*/
+export function convertUTCtoISO8601(timestamp: string | number): string | undefined {
+    if (timestamp == null) {
+        return undefined;
+    }
+
+    const date = new Date(timestamp);
+
+    if (isNaN(date.getTime())) {
+        return undefined;
+    }
+
+    try {
+        // Can throw RangeError.
+        const converted = date.toISOString();
+        return converted.split('.')[0].replace(/-|:/g, '').concat('Z');
+    } catch {
+        return undefined;
+    }
 }
 
 /**
@@ -41,7 +59,8 @@ export function checkTimeSkew(timestamp: string, expiry: number, log: RequestLog
     if ((currentTime + fifteenMinutes) < parsedTimestamp) {
         log.debug('current time pre-dates timestamp', {
             parsedTimestamp,
-            currentTimeInMilliseconds: currentTime });
+            currentTimeInMilliseconds: currentTime
+        });
         return true;
     }
     const expiryInMilliseconds = expiry * 1000;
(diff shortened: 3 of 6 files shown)


lib/auth/v4/headerAuthCheck.ts
```diff
@@ -81,21 +82,18 @@ export function check(
...
-        // check that x-amz- date has the correct format and after epochTime
-        if (xAmzDateArr.length === 2 && xAmzDateArr[0].length === 8
-          && xAmzDateArr[1].length === 7
-          && Number.parseInt(xAmzDateArr[0], 10) > 19700101) {
-            // format of x-amz- date is ISO 8601: YYYYMMDDTHHMMSSZ
-            timestamp = request.headers['x-amz-date'];
-        }
+    if (isValidISO8601Compact(xAmzDate)) {
+        timestamp = xAmzDate;
     } else if (request.headers.date) {
```
BourgoisMickael, line 89: Now this fallsback when x-amz-date is invalid, where it should maybe only happens if it is missing ?

If x-amz-date is invalid should you rather return an error instead of falling back to date header ?

leif-scality: true. reverted to the old behavior, timestamp will be undefined if the test fails and we will return an error

lib/auth/v4/timeUtils.ts
```diff
@@ -41,16 +59,85 @@ export function checkTimeSkew(timestamp: string, expiry: number, log: RequestLog
...
+
+    // Check all other positions are digits
+    for (let i = 0; i < 16; i++) {
+        if (i === 8 || i === 15) {
+            continue; // Skip T and Z
+        }
+        if (str[i] < '0' || str[i] > '9') {
+            return false;
+        }
+    }
```
BourgoisMickael: This could be simplified and more maintainable with a regex.

Ideally we should use a date library for date manipulation or validation, but I see we don't have any in arsenal

lib/auth/v4/timeUtils.ts
```diff
@@ -15,15 +15,33 @@ export function convertAmzTimeToMs(timestamp: string) {
...
+* Convert UTC timestamp to ISO 8601 compact format
+* @param timestamp - UTC timestamp (e.g., 'Fri, 10 Feb 2012 21:34:55 GMT') or Unix timestamp
+* @return ISO8601 timestamp of form YYYYMMDDTHHMMSSZ, or undefined if invalid
+*
+* @example
+* convertUTCtoISO8601('Fri, 10 Feb 2012 21:34:55 GMT'); // '20120210T213455Z'
+* convertUTCtoISO8601(1328910895000); // '20120210T213455Z'
+* convertUTCtoISO8601('invalid'); // undefined
+*/
+export function convertUTCtoISO8601(timestamp: unknown): string | undefined {
```
BourgoisMickael: instead of `unknown` the type should be `string | number | null` (maybe even `| undefined` but there is no check for undefined.

lib/auth/auth.ts
```diff
@@ -216,7 +216,9 @@ function generateV4Headers(
     payload?: string,
 ) {
     Object.assign(request, { headers: {} });
-    const amzDate = convertUTCtoISO8601(Date.now());
+    // Date.now() should always return a valid date so we assert non null.
+    const amzDate: string = convertUTCtoISO8601(Date.now())!;
```
BourgoisMickael: if there is a non-null assertion, then you don't need to add the type `string`, it's the only type left as output of the function

tests/unit/auth/v4/timeUtils.spec.js
```diff
@@ -25,12 +27,193 @@ describe('convertAmzTimeToMs function', () => {
...
+    it('should convert Unix timestamp (number) to ISO8601 timestamp', () => {
+        // Use Date.UTC to ensure correct timestamp for 2015-02-08 20:14:05 UTC
+        const input = Date.UTC(2015, 1, 8, 20, 14, 5); // Month is 0-indexed
+        const expectedOutput = '20150208T201405Z';
+        const actualOutput = convertUTCtoISO8601(input);
+        assert.strictEqual(actualOutput, expectedOutput);
+    });
+
+    it('should convert Date.now() to ISO8601 timestamp', () => {
+        const now = Date.now();
```
BourgoisMickael: prefer a fixed date like the test above should be enough, this test doesn't provide more coverage

tests/unit/auth/v4/timeUtils.spec.js
```diff
@@ -25,12 +27,193 @@ describe('convertAmzTimeToMs function', () => {
...
+        assert.strictEqual(actualOutput, undefined);
+    });
+
+    it('should return undefined for NaN', () => {
+        const actualOutput = convertUTCtoISO8601(NaN);
+        assert.strictEqual(actualOutput, undefined);
+    });
+
+    it('should return undefined for object', () => {
+        const actualOutput = convertUTCtoISO8601({});
```
BourgoisMickael: this should be a typescript error if the function is typed correctly

tests/unit/auth/v4/timeUtils.spec.js
```diff
@@ -25,12 +27,193 @@ describe('convertAmzTimeToMs function', () => {
...
+        const actualOutput = convertUTCtoISO8601(input);
+        assert.strictEqual(actualOutput, undefined);
+    });
+});
+
+describe('isValidISO8601Compact function', () => {
+    describe('valid timestamps', () => {
+        it('should return true for valid ISO8601 compact format', () => {
+            const input = '20160208T201405Z';
+            assert.strictEqual(isValidISO8601Compact(input), true);
```
BourgoisMickael:
```suggestion
            assert.strictEqual(isValidISO8601Compact('20160208T201405Z'), true);
```

you can go one line like the describe a little below here

tests/unit/auth/v4/timeUtils.spec.js
```diff
@@ -25,12 +27,193 @@ describe('convertAmzTimeToMs function', () => {
...
+            assert.strictEqual(isValidISO8601Compact(input), false);
+        });
+
+        it('should return false for day 00', () => {
+            const input = '20160200T201405Z';
+            assert.strictEqual(isValidISO8601Compact(input), false);
+        });
+    });
+
+    describe('null/undefined/non-string inputs', () => {
```
BourgoisMickael: anything other than string should raise a typescript error here, you should not be able to use another type

tests/unit/auth/v4/timeUtils.spec.js
```diff
@@ -25,12 +27,193 @@ describe('convertAmzTimeToMs function', () => {
...
+        });
+
+        it('should return true for leap year Feb 29', () => {
+            const input = '20200229T120000Z'; // 2020 is leap year
+            assert.strictEqual(isValidISO8601Compact(input), true);
+        });
+    });
+
+    describe('invalid format', () => {
+        it('should return false for string with wrong length', () => {
```
BourgoisMickael: you could use an array of (test name, input) and loop over it

```js
[
  { name: 'should return false for string with wrong length', input: '2016020T201405Z' },
  // ...
].forEach(t => it(t.name, () => assert.strictEqual(isValidISO8601Compact(t.input), false))
```

tests/unit/auth/v4/timeUtils.spec.js
BourgoisMickael: Oh the file is javascript, that's why you have no typescript error in there

tests/unit/auth/v4/timeUtils.spec.js
leif-scality: the function accepts `any` even in typescript so we need to test the case. Even with the type `string | number`

tests/unit/auth/v4/timeUtils.spec.js
leif-scality: the function accepts `any` even in typescript so we need to test the case. Even with the type `string | number`

lib/auth/v4/timeUtils.ts
leif-scality: Do we want to compile and run a regex for each request? I agree the regex is simpler, but the code is not that complicated

lib/auth/v4/timeUtils.ts
leif-scality: replaced with `string | number`

lib/auth/auth.ts
leif-scality: removed the type

tests/unit/auth/v4/timeUtils.spec.js
leif-scality: removed the test we already have a test with `new Date`

tests/unit/auth/v4/timeUtils.spec.js
leif-scality: refactored to array

tests/unit/auth/v4/timeUtils.spec.js
leif-scality: done

tmacro (approved):

lib/auth/v4/timeUtils.ts
jonathan-gramain: I tend to agree with @BourgoisMickael , I'd prefer a regexp, it's cleaner and less error-prone (and maybe more efficient too, but that would have to be tested). Normally if you use a static regexp (`/.../`) it's pre-compiled automatically.

Also you can use the result of the regexp parsing to extract components, instead of using `substring`.

lib/auth/v4/timeUtils.ts
```diff
@@ -41,16 +59,85 @@ export function checkTimeSkew(timestamp: string, expiry: number, log: RequestLog
...
+    const second = str.substring(13, 15);
+
+    // Construct standard ISO format and validate
+    const isoString = `${year}-${month}-${day}T${hour}:${minute}:${second}.000Z`;
+    const date = new Date(isoString);
+
+    try {
+        // date.toISOString() can throw.
+        // date.toISOString() === isoString check prevents silent date corrections (30 February to 1 March)
+        return !isNaN(date.getTime()) && date.toISOString() === isoString;
```
jonathan-gramain: nitpick: you can use `Number.isNaN` to be strict about checking exclusively a `NaN` value (rather than anything that is not parsable to a valid number)

tests/unit/auth/v4/timeUtils.spec.js
```diff
@@ -9,6 +9,8 @@ const convertAmzTimeToMs =
     require('../../../../lib/auth/v4/timeUtils').convertAmzTimeToMs;
 const convertUTCtoISO8601 =
     require('../../../../lib/auth/v4/timeUtils').convertUTCtoISO8601;
+const isValidISO8601Compact =
+    require('../../../../lib/auth/v4/timeUtils').isValidISO8601Compact;
```
jonathan-gramain:
```suggestion
const {
    checkTimeSkew,
    convertAmzTimeToMs,
    convertUTCtoISO8601,
    isValidISO8601Compact,
} = require('../../../../lib/auth/v4/timeUtils');
```

lib/auth/v4/timeUtils.ts
leif-scality: changed to regex

lib/auth/v4/timeUtils.ts
leif-scality: changed to Number.isNaN

tests/unit/auth/v4/timeUtils.spec.js
leif-scality: done

lib/auth/v4/timeUtils.ts
```diff
@@ -41,16 +59,65 @@ export function checkTimeSkew(timestamp: string, expiry: number, log: RequestLog
...
+* ```typescript
+* isValidISO8601Compact('20160208T201405Z');  // true
+* isValidISO8601Compact('20160230T201405Z');  // false (Feb 30 invalid)
+* isValidISO8601Compact('20160208T251405Z');  // false (25 hours invalid)
+* isValidISO8601Compact('2016-02-08T20:14:05Z'); // false (wrong format)
+* isValidISO8601Compact('abcd0208T201405Z');  // false (contains letters)
+* ```
+*/
+export function isValidISO8601Compact(str: string): boolean {
+    if (str == null || typeof str !== 'string') {
```
jonathan-gramain: nitpick: `typeof str !== 'string'` would suffice

jonathan-gramain (approved):

leif-scality:
ping

leif-scality:
ping

PR #2599, opened 2026-03-23
ARSN-560: return AccessDenied error if x-amz-timestamp values before epoch

leif-scality:
# Issue                                                                                                                                               

The ceph s3-tests test_object_create_bad_amz_date_before_epoch_aws4 and test_bucket_create_bad_amz_date_before_epoch_aws4 are failing in CI. These tests send an AWS4-signed request with X-Amz-Date: 19500707T215304Z (pre-Unix epoch, 1950) and expect a 403 with error code AccessDenied or SignatureDoesNotMatch.                                                                                                                              

Cloudserver returns RequestTimeTooSkewed instead, causing an AssertionError.                                                                        

# Root cause                                                                                                                                          

The regression was introduced in arsenal commit e3223a7 (ARSN-552, 2026-02-11). The old headerAuthCheck.ts had an inline check that explicitly rejected pre-1970 dates: Number.parseInt(xAmzDateArr[0], 10) > 19700101

ARSN-552 replaced this with a call to the new isValidISO8601Compact function, which no longer rejects pre-1970 dates — JavaScript's Date handles them fine as negative Unix timestamps, so the round-trip validation passes for 1950-07-07.                                                          

With a valid timestamp, execution reaches validateCredentials, which compares the credential scope date (20260323) against the x-amz-date timestamp date (19500707), finds a mismatch, and returns RequestTimeTooSkewed — the wrong error.

# Fix                                                       

Add a pre-1970 guard in isValidISO8601Compact in arsenal/lib/auth/v4/timeUtils.ts, restoring the original behavior:                                 
```
  const [, year, month, day, hour, minute, second] = match;
  if (parseInt(year, 10) < 1970) {
      return false;
  }
```
  With this fix, isValidISO8601Compact('19500707T215304Z') returns false, timestamp stays undefined, and headerAuthCheck returns AccessDenied — matching the test expectation and the pre-ARSN-552 behavior.

### DIFF
--- a/lib/auth/v4/headerAuthCheck.ts
+++ b/lib/auth/v4/headerAuthCheck.ts
@@ -5,8 +5,7 @@ import constructStringToSign from './constructStringToSign';
 import {
     checkTimeSkew,
     convertUTCtoISO8601,
-    convertAmzTimeToMs,
-    isValidISO8601Compact,
+    parseISO8601Compact,
 } from './timeUtils';
 import {
     extractAuthItems,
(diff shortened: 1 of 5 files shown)


lib/auth/v4/timeUtils.ts
```diff
@@ -94,6 +94,7 @@ export function checkTimeSkew(timestamp: string, expiry: number, log: RequestLog
 * isValidISO8601Compact('20160208T251405Z');  // false (25 hours invalid)
 * isValidISO8601Compact('2016-02-08T20:14:05Z'); // false (wrong format)
 * isValidISO8601Compact('abcd0208T201405Z');  // false (contains letters)
+* isValidISO8601Compact('19500707T215304Z');  // false (pre-Unix epoch)
```
fredmnl: Major. Semantically checking for `< 1970` in `isValidISO8601Compact` isn't great because we are now denying valid ISO8601 inputs. `19500707T215304Z` is actually a valid ISO8601 compact time

lib/auth/v4/timeUtils.ts
```diff
@@ -109,6 +110,10 @@ export function isValidISO8601Compact(str: string): boolean {
 
     const [, year, month, day, hour, minute, second] = match;
 
+    if (parseInt(year, 10) < 1970) {
```
fredmnl: Nit. If it's easy to do, I would prefer `unix < 0` rather than `year < 1970` which would be a bit more meaningful. If it's adding more than a single line of code, I'm fine with keeping it written that way.

fredmnl (commented):
Two comments come to mind but I'm not coming with a solution, so apologies for that. Other than that, I'm all good with the analysis and the intent of the fix.

lib/auth/v4/timeUtils.ts
leif-scality: replaced `isValidISO8601Compact` with `ParseISO8601Compact` that returns a Date object if it managed to parse the string, the epoch check is done outside the function using `date.getTime() < 0`

lib/auth/v4/timeUtils.ts
leif-scality: replaced isValidISO8601Compact with ParseISO8601Compact that returns a Date object if it managed to parse the string, the epoch check is done outside the function using date.getTime() < 0

lib/auth/v4/timeUtils.ts
```diff
@@ -116,8 +110,17 @@ export function isValidISO8601Compact(str: string): boolean {
     try {
         // date.toISOString() can throw.
         // date.toISOString() === isoString check prevents silent date corrections (30 February to 1 March)
-        return !Number.isNaN(date.getTime()) && date.toISOString() === isoString;
+        if (date.toISOString() !== isoString) {
+            return undefined;
+        }
+
+        if (Number.isNaN(date.getTime())) {
```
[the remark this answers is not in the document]
leif-scality: reverted order

lib/auth/v4/headerAuthCheck.ts
```diff
@@ -79,33 +78,46 @@ export function check(
...
+                const convertedParsed = parseISO8601Compact(converted);
+                if (convertedParsed) {
+                    timestampCompact = converted;
+                    timestampParsed = convertedParsed;
+                }
+            }
         }
     }
-    if (!timestamp) {
+    const beforeEpoch = timestampParsed && timestampParsed.getTime() < 0;
```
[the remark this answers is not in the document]
leif-scality: https://scality.atlassian.net/browse/ARSN-562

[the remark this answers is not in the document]
leif-scality: https://scality.atlassian.net/browse/ARSN-562

lib/auth/v4/headerAuthCheck.ts
```diff
@@ -79,33 +78,46 @@ export function check(
...
+    if (!timestampCompact || !timestampParsed || beforeEpoch) {
         log.debug('missing or invalid date header',
             { 'method': 'auth/v4/headerAuthCheck.check', 'x-amz-date': xAmzDate, 'Date': request.headers.date });
-        return { err: errorInstances.AccessDenied.
-            customizeDescription('Authentication requires a valid Date or ' +
-          'x-amz-date header') };
+        return {
+            err: errorInstances.AccessDenied.
+                customizeDescription('Authentication requires a valid Date or ' + 'x-amz-date header')
```
BourgoisMickael:
```suggestion
                customizeDescription('Authentication requires a valid Date or x-amz-date header')
```

BourgoisMickael (approved):

fredmnl (approved):

leif-scality:
ping

leif-scality:
ping

PR #2603, opened 2026-04-01
Bound MongoDB sort memory in lifecycle current listing

delthas:
## Summary

When lifecycle v2 indexes are used for current listings, MongoDB must
re-sort results by `_id` in memory because the indexes are ordered by
`value.last-modified`. Without a cursor `.limit()`, this SORT stage
collects **all** matching documents before returning any — which can
exceed the 100MB `memLimit` and spill to disk.

On systems with low disk space (the scenario described in BB-753),
this disk spill fails, breaking lifecycle processing entirely. The
problem is not index *creation* failing, but index *usage* triggering
an unbounded sort that spills to a full disk.

## Fix

Add a cursor limit based on `maxScannedLifecycleListingEntries`
(default 10,000) in `DelimiterCurrent.genMDParamsV0()`. This flows
through to `MongoReadStream` which already supports `options.limit`
on the cursor.

With the limit, MongoDB uses a bounded top-k heap sort — keeping only
~10,000 documents in memory regardless of total matches. We use
`maxScannedLifecycleListingEntries` rather than `maxKeys` because it
counts documents scanned (not results), which maps directly to cursor
documents regardless of bucket key format (v0 or v1).

## Impact

Tested on a 100k-object bucket with `explain("executionStats")`:

| | Without `.limit()` | With `.limit(1001)` |
|---|---|---|
| **nReturned** | 448,497 | 1,001 |
| **totalDataSizeSorted** | **478 MB** | **1 MB** |
| **usedDisk** | **true** (spilled!) | false |

Issue: ARSN-565

### DIFF
--- a/lib/algos/list/delimiterCurrent.ts
+++ b/lib/algos/list/delimiterCurrent.ts
@@ -48,6 +48,19 @@ class DelimiterCurrent extends DelimiterMaster {
             };
         }
 
+        // Limit the number of documents MongoDB returns to bound the
+        // in-memory sort when lifecycle indexes are used (these indexes
+        // order by value.last-modified, forcing a re-sort on _id). Without
+        // a limit, MongoDB must sort all matching documents — which can
+        // exceed the 100MB memory cap and spill to disk.
+        // We use maxScannedLifecycleListingEntries (default 10,000) as the
+        // bound because it counts documents scanned, which maps directly
+        // to cursor documents regardless of bucket format (v0 or v1).
+        // The +1 allows the listing algorithm to detect truncation.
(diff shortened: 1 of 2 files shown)


francoisferrand (approved):

francoisferrand:
is this on the right branch?
ideally we may want this in zenko 2.13, which is still on backbeat 9.1, which is using artesca 8.2.26.
→ this is a long way from we should assess if there is a risk to go from 8.2.26 to 8.2.49...
→ how confident are we in this change? can we just put it in prod ASAP, or should we be slightly more careful?

francoisferrand (approved):

benzekrimaha (approved):

PR #2605, opened 2026-04-01
Fix getDiskUsage and add getCollectionStats

delthas:
## Summary

- **Fix `getDiskUsage`**: Was reading the nonexistent `fsFreeSize` field from MongoDB's `dbStats` command, always returning 0 for available/free space. Now correctly computes free space as `fsTotalSize - fsUsedSize`.
- **Add `getCollectionStats`**: New method on `MongoClientInterface` that exposes per-collection stats via the `collStats` command. Returns the raw stats object including `indexSizes` (per-index size breakdown) and `totalIndexSize`.

## Context

The lifecycle conductor (BB-753) needs to check available disk space before creating v2 indexes on MongoDB. When the MongoDB PV is nearly full, attempting index creation can fail (ENOSPC) or further pressure the disk. The conductor uses `getDiskUsage` for filesystem free space and `getCollectionStats` to get the `_id_` index size as an estimate for the cost of creating lifecycle indexes.

## Design decisions

- **`getDiskUsage` keeps its existing return shape** (`{ available, free, total }`) to avoid breaking the interface contract, just fixes the computation.
- **`getCollectionStats` passes stats through directly** from MongoDB rather than reshaping — the caller extracts what it needs.
- **Uses `db.command({ collStats })` instead of the deprecated `collection.stats()`** which was removed from the MongoDB Node.js driver type definitions.
- Both calls are metadata-only (~3ms and ~8ms respectively on a 1M-document collection), no collection scans or locks.

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -2524,17 +2524,32 @@ class MongoClientInterface {
 
         return this.db.command({ dbStats: 1, scale: 1 })
             .then(stats => {
-                const result = {
-                    available: stats.fsFreeSize || 0,
-                    // Same as available in MongoDB context
-                    free: stats.fsFreeSize || 0,
-                    total: stats.fsTotalSize || 0
-                };
-                return cb(null, result);
+                if (stats.fsTotalSize === undefined || stats.fsUsedSize === undefined) {
+                    this.logger.error('unexpected dbStats response: missing fsTotalSize or fsUsedSize',
+                        { stats });
+                    return cb(errors.InternalError);
+                }
+                const free = stats.fsTotalSize - stats.fsUsedSize;
+                return cb(null, { available: free, free, total: stats.fsTotalSize });
             })
             .catch(err => {
-                this.logger.error('Error getting MongoDB disk stats', 
+                this.logger.error('Error getting MongoDB disk stats',
                     { error: err.message });
+                return cb(errors.InternalError);
+            });
+    }
+
+    getCollectionStats(bucketName: string, log: werelogs.Logger, cb: ArsenalCallback<any>) {
+        if (!this.db || !this.client) {
+            return cb(errors.InternalError.customizeDescription(
+                'Cannot get collection stats: database not connected'));
+        }
+        const c = this.getCollection(bucketName);
+        return this.db.command({ collStats: c.collectionName })
+            .then(stats => cb(null, stats))
+            .catch(err => {
+                log.error('error getting collection stats',
+                    { error: err.message, bucketName });
                 return cb(errors.InternalError);
             });
     }
--- a/tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
+++ b/tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
@@ -307,21 +307,21 @@ describe('MongoClientInterface::getDiskUsage', () => {
         
         // Mock MongoDB stats response
         const mockStats = {
-            fsFreeSize: 1000000,
+            fsUsedSize: 4000000,
             fsTotalSize: 5000000
         };
-        
+
         // Mock the database and command
         testClient.db = {
             command: sinon.stub().resolves(mockStats)
         };
         testClient.client = {}; // Just to pass the initial check
-        
+
         testClient.getDiskUsage((err, result) => {
             assert.strictEqual(err, null);
             assert.deepStrictEqual(result, {
-                available: mockStats.fsFreeSize,
-                free: mockStats.fsFreeSize,
+                available: 1000000,
+                free: 1000000,
                 total: mockStats.fsTotalSize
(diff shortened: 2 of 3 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2524,10 +2524,10 @@ class MongoClientInterface {
 
         return this.db.command({ dbStats: 1, scale: 1 })
```
francoisferrand, line 2525: does this depend on specific MongoDB version(s) ?

tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
```diff
@@ -307,21 +307,21 @@ describe('MongoClientInterface::getDiskUsage', () => {
         
         // Mock MongoDB stats response
         const mockStats = {
-            fsFreeSize: 1000000,
+            fsUsedSize: 4000000,
             fsTotalSize: 5000000
         };
-        
+
```
francoisferrand, line 313: aren't we missing some (additional) tests targeting a _real_ mongodb here ? i.e. kind of integration tests, so that we would identify more easily issues like the one you fixed.

francoisferrand (approved):

francoisferrand:
note: this is an "independent" fix, could be fixed to Arsenal 8.3 if it makes integration easier/simpler...

lib/storage/metadata/mongoclient/MongoClientInterface.ts
delthas: dbStats and collStats are there since the very beginning.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2524,10 +2524,10 @@ class MongoClientInterface {
...
             .then(stats => {
+                const free = (stats.fsTotalSize || 0) - (stats.fsUsedSize || 0);
```
SylvainSenechal: kinda weird to have this || 0 as it means we could end up with a negative number 🤔 I guess its just js things, but would be better to have errors returned if needed

tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
```diff
@@ -307,21 +307,21 @@ describe('MongoClientInterface::getDiskUsage', () => {
...
         };
         testClient.client = {}; // Just to pass the initial check
-        
+
         testClient.getDiskUsage((err, result) => {
             assert.strictEqual(err, null);
             assert.deepStrictEqual(result, {
-                available: mockStats.fsFreeSize,
-                free: mockStats.fsFreeSize,
+                available: 1000000,
```
SylvainSenechal, line 323: nit : for this its nicer to work on named constants 
const usedSize = 400
const totalSize = 500

available = totalSize - usedSized

Anyways, I much much prefer the functional tests, as as we can see here, this mock tests previously failed to detect that getDiskUsage was reading from a wrong mongodb field -_-

SylvainSenechal (approved):
Next time try to re add in the PR summary the Issue: ARSN-566, so i can quickly click on it to read the jira ticket, it's also missing in your backbeat pr

delthas:
> Next time try to re add in the PR summary the Issue: [ARSN-566](https://scality.atlassian.net/browse/ARSN-566), so i can quickly click on it to read the jira ticket, it's also missing in your backbeat pr

Indeed, thanks.

[ARSN-566]: https://scality.atlassian.net/browse/ARSN-566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
delthas: Agree. Dropped those since we have better func tests now.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
delthas: OK, sending an error now.

tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
```diff
@@ -242,136 +242,6 @@ describe('MongoClientInterface::_isReplicationEntryStalled', () => {
...
-                total: 0
-            });
-            assert(testClient.db.command.calledOnce);
-            assert(testClient.db.command.calledWith({ dbStats: 1, scale: 1 }));
-            done();
-        });
-    });
-});
-
 function createBucket(client, bucketName, isVersioned, callback) {
```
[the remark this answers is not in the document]
delthas: These error paths are trivial, not worth it.

PR #2611, opened 2026-04-13
feat(ARSN-572): Propagate W3C trace context to MongoDB metadata writes

delthas:
## Context

Cloudserver recently gained OpenTelemetry tracing ([CLDSRV-884], cloudserver PR scality/cloudserver#6140). That gives us a single end-to-end trace per S3 request spanning cloudserver → vault → hdproxy → hyperiod.

But many state mutations in the S3 stack are written to MongoDB and **consumed asynchronously** by worker services via the oplog:
- **Backbeat** — replication, lifecycle expiration/transition, garbage collection
- **Sorbet** — cold-storage restore completion
- Other oplog consumers

Today the oplog entry carries no trace context, so async work performed hours or days after the S3 request cannot be tied back to the request that caused it. A user's `POST …?restore` produces a trace that ends at the metadata write; the actual cold-tier retrieval by Sorbet is invisible from the original trace.

## Solution

Stamp a W3C trace context (`traceparent`/`tracestate`) on every mutating metadata write, auto-derived from the currently-active OTEL context.

The field lives on `ObjectMDData` next to `originOp`:

```ts
traceContext?: {
    traceparent?: string;
    tracestate?: string;
};
```

A new helper `stampActiveTraceContext` in `lib/storage/metadata/captureTraceContext.ts`:
- reads the active OTEL context via `@opentelemetry/api`
- serializes it via `propagation.inject()`
- writes the result to `data.traceContext`, or clears the field if no span is active (so a stale value loaded from storage cannot leak forward into a new oplog entry)

It is exposed with TS **overloads** so callers pass whichever shape they have — an `ObjectMD` instance or a raw metadata object:

```ts
export function stampActiveTraceContext(objMd: ObjectMD): void;
export function stampActiveTraceContext(data: Record<string, any>): void;
export function stampActiveTraceContext(data: ObjectMD | Record<string, any>): void {
    const target = data instanceof ObjectMD ? data.getValue() : data;
    // ...
}
```

It is called from inside `MongoUtils.serialize()`, so **every metadata write that goes through serialize automatically gets the stamp** (the `putObject` / `putObjectVerCase4` / `repair` paths in `MongoClientInterface`). For the 2 paths that bypass `serialize` because they rewrite a doc previously read from the DB with already-escaped tags (`internalPutObject`, `internalDeleteObject`), `stampActiveTraceContext()` is called explicitly on the `ObjectMD` instance (the overload resolves `.getValue()` internally).

Consumers (backbeat, sorbet — **out of scope for this PR**) will later:
1. Read `value.traceContext.traceparent` from the oplog entry
2. `propagation.extract()` it to reconstruct a parent context
3. Wrap per-entry processing in `context.with(extractedCtx, ...)`

This closes the loop on end-to-end tracing across the async boundary.

## Design decisions

**1. Auto-inject at the arsenal write boundary (vs. param plumbing).**
The alternative was to thread a new `traceContext` parameter through the ~15 cloudserver call sites that already set `originOp`. Rejected because:
- `@opentelemetry/context-async-hooks` (loaded by the OTEL SDK at cloudserver startup) already propagates the request context through arsenal's async call chain for free. Re-reading it at the mongo write is idiomatic OTEL.
- Zero cloudserver surface change — we can iterate on arsenal independently.
- Parallels existing behavior where arsenal reads `originOp` out of params and attaches it to the `ObjectMD`.
- Avoids replicating the maintenance cost that makes `originOp` easy to forget on new API endpoints today.

**2. Only `@opentelemetry/api` as a runtime dep (~10 KB, no SDK).**
The api package is specifically designed to be safe for libraries to depend on. When no SDK is registered (today's state for anyone not opting into cloudserver OTEL), `context.active()` returns a no-op context, `trace.getSpan()` returns undefined, and `stampActiveTraceContext()` clears the field and returns. **Zero runtime cost when OTEL is off** — covered by a dedicated unit test (see Verification).

**3. Optional field, no schema migration.**
`traceContext` is absent by default — consistent with other optional ObjectMD fields like `legalHold`, `retentionDate`. No default-value change in the ObjectMDData construction. Existing consumers that don't understand the field simply ignore it; fully backward compatible.

**4. Always reflects the current write — never carries a stale context forward.**
If no span is active when `stampActiveTraceContext()` runs, it clears `traceContext` (rather than leaving whatever was loaded from storage). Same when `propagation.inject()` produces no `traceparent`. This prevents both edge-case OTEL output and the more practical hazard where a doc read for a delete/repair carries a `traceContext` from its original write that does not belong on the new oplog entry.

**5. Single helper, single contract.**
The trace-stamping API on arsenal is one function — `stampActiveTraceContext`, with overloads for the `ObjectMD` and raw-data call shapes but a single implementation. There is intentionally no public setter on `ObjectMD`: the contract is "trace context is stamped by the metadata backend at write time," and a model-level setter would invite callers in cloudserver/backbeat to bypass that contract (issue raised in review). Internally this also keeps the stamp/clear semantics in exactly one place. The overload pattern matches existing arsenal usage (`convertToEpochTime`, `uuidv4`).

## Interaction with async flows (cold restore as example)

Cold storage restore exercises both a sync and an async metadata write:

1. `ObjectRestore:Post` (sync, user-initiated via `POST /bucket/key?restore`) — cloudserver sets `originOp`; arsenal now also stamps `traceContext` pointing to the user's request trace. ✅
2. `ObjectRestore:Completed` (async, Sorbet-triggered via cloudserver's `/_/backbeat/*` route) — behavior depends on Sorbet:
   - If Sorbet forwards `traceparent` on its HTTP call (ideally extracted from step 1's stored `traceContext`): arsenal picks it up via `context.with(remoteContext, ...)` in cloudserver's `server.js`, and the Completed write ties back to the original user restore. ✅
   - If Sorbet calls without `traceparent` (today's state): safely no-ops — Completed write has no `traceContext`, no regression. ✅

Same pattern applies to lifecycle expiration/transition, CRR replication, and GC.

### Unit tests

- `tests/unit/storage/metadata/captureTraceContext.spec.js` — `stampActiveTraceContext` against a mocked `@opentelemetry/api`: no-active-span clears (and is a no-op when there was nothing to clear), traceparent-only write, traceparent + tracestate write, defensive handling of a propagator that injects no `traceparent`, overwrite of a stale value, and the `ObjectMD`-instance overload stamping onto `getValue()`.
- `tests/unit/storage/metadata/captureTraceContext.noOtel.spec.js` — drives the **real** `@opentelemetry/api` with no SDK registered (the "arsenal bumped into a consumer that never initialized OTEL" case): confirms a safe no-op that clears stale context and never throws.
- `tests/unit/storage/metadata/mongoclient/{putObject,delObject,repair}.spec.js` — sinon-stubbed tests verify the field lands on the document handed to `findOneAndReplace` / `bulkWrite` for each write path the cloudserver hot path exercises, and that it is cleared rather than carried forward when no span is active.

All unit tests in `tests/unit/models/` and `tests/unit/storage/metadata/` pass; `yarn build` is clean.

### End-to-end (follow-up on a test cluster)

1. Bump cloudserver's arsenal pin to a build of this branch; roll the cloudserver deployment on an Artesca cluster with `ENABLE_OTEL=true`
2. Make a PutObject; note the trace ID in Jaeger
3. Inspect the MongoDB oplog: `o.value.traceContext.traceparent` should be present, with characters 3-35 matching the Jaeger trace ID
4. Negative test with `ENABLE_OTEL=false` → field should be absent

Issue: ARSN-572

[CLDSRV-884]: https://scality.atlassian.net/browse/CLDSRV-884

### DIFF
--- a/lib/models/ObjectMD.ts
+++ b/lib/models/ObjectMD.ts
@@ -2,10 +2,7 @@ import * as crypto from 'crypto';
 import * as constants from '../constants';
 import * as VersionIDUtils from '../versioning/VersionID';
 import { VersioningConstants } from '../versioning/constants';
-import ObjectMDLocation, {
-    ObjectMDLocationData,
-    Location,
-} from './ObjectMDLocation';
+import ObjectMDLocation, { ObjectMDLocationData, Location } from './ObjectMDLocation';
 import ObjectMDAmzRestore from './ObjectMDAmzRestore';
 import ObjectMDArchive from './ObjectMDArchive';
 import { ObjectMDAzureInfoMetadata } from './ObjectMDAzureInfo';
@@ -193,7 +190,7 @@ export default class ObjectMD {
             'cache-control': '',
             'content-disposition': '',
             'content-encoding': '',
-            'expires': '',
+            expires: '',
             'content-length': 0,
             'content-type': '',
             'content-md5': '',
@@ -212,27 +209,27 @@ export default class ObjectMD {
             'x-amz-server-side-encryption-customer-algorithm': '',
             'x-amz-website-redirect-location': '',
             'x-amz-scal-transition-in-progress': undefined,
-            'acl': {
+            acl: {
                 Canned: 'private',
                 FULL_CONTROL: [],
                 WRITE_ACP: [],
                 READ: [],
                 READ_ACP: [],
             },
-            'key': '',
-            'location': null,
-            'azureInfo': undefined,
+            key: '',
+            location: null,
+            azureInfo: undefined,
             // versionId, isNull, nullVersionId and isDeleteMarker
             // should be undefined when not set explicitly
-            'isNull': undefined,
-            'isNull2': undefined,
-            'nullVersionId': undefined,
-            'nullUploadId': undefined,
-            'isDeleteMarker': undefined,
-            'versionId': undefined,
-            'uploadId': undefined,
-            'tags': {},
-            'replicationInfo': {
+            isNull: undefined,
+            isNull2: undefined,
+            nullVersionId: undefined,
+            nullUploadId: undefined,
+            isDeleteMarker: undefined,
+            versionId: undefined,
+            uploadId: undefined,
+            tags: {},
+            replicationInfo: {
                 status: '',
                 backends: [],
                 content: [],
@@ -243,10 +240,10 @@ export default class ObjectMD {
                 dataStoreVersionId: '',
                 isNFS: undefined,
             },
-            'dataStoreName': '',
-            'originOp': '',
-            'deleted': undefined,
-            'isPHD': undefined,
+            dataStoreName: '',
+            originOp: '',
+            deleted: undefined,
+            isPHD: undefined,
         };
     }
 
@@ -489,11 +486,15 @@ export default class ObjectMD {
      * @param checksum - ObjectMDChecksum instance
      * @return itself
      */
-    setChecksum(checksum: ObjectMDChecksum | {
-        checksumAlgorithm: ChecksumAlgorithm;
-        checksumValue: string;
-        checksumType: ChecksumType;
-    }) {
+    setChecksum(
+        checksum:
+            | ObjectMDChecksum
+            | {
+                  checksumAlgorithm: ChecksumAlgorithm;
+                  checksumValue: string;
+                  checksumType: ChecksumType;
+              },
+    ) {
         if (checksum instanceof ObjectMDChecksum) {
             this._data.checksum = checksum;
         } else if (ObjectMDChecksum.isValid(checksum) === null) {
@@ -706,9 +707,9 @@ export default class ObjectMD {
      * @param transitionTime - Date when the transition started
      * @return itself
      */
-    setTransitionInProgress(inProgress: false): this
-    setTransitionInProgress(inProgress: true, transitionTime: Date|string|number): this
-    setTransitionInProgress(inProgress: boolean, transitionTime?: Date|string|number) {
+    setTransitionInProgress(inProgress: false): this;
+    setTransitionInProgress(inProgress: true, transitionTime: Date | string | number): this;
+    setTransitionInProgress(inProgress: boolean, transitionTime?: Date | string | number) {
         this._data['x-amz-scal-transition-in-progress'] = inProgress;
         if (!inProgress || !transitionTime) {
             delete this._data['x-amz-scal-transition-time'];
@@ -1098,17 +1099,8 @@ export default class ObjectMD {
         dataStoreVersionId?: string;
         isNFS?: boolean;
     }) {
-        const {
-            status,
-            backends,
-            content,
-            destination,
-            storageClass,
-            role,
-            storageType,
-            dataStoreVersionId,
-            isNFS,
-        } = replicationInfo;
+        const { status, backends, content, destination, storageClass, role, storageType, dataStoreVersionId, isNFS } =
+            replicationInfo;
         this._data.replicationInfo = {
             status,
             backends,
@@ -1156,9 +1148,7 @@ export default class ObjectMD {
     }
 
     setReplicationSiteStatus(site: string, status: string) {
-        const backend = this._data.replicationInfo.backends.find(
-            o => o.site === site
-        );
+        const backend = this._data.replicationInfo.backends.find(o => o.site === site);
         if (backend) {
             backend.status = status;
         }
@@ -1166,9 +1156,7 @@ export default class ObjectMD {
     }
 
     getReplicationSiteStatus(site: string) {
-        const backend = this._data.replicationInfo.backends.find(
-            o => o.site === site
-        );
+        const backend = this._data.replicationInfo.backends.find(o => o.site === site);
         if (backend) {
             return backend.status;
         }
@@ -1181,9 +1169,7 @@ export default class ObjectMD {
     }
 
     setReplicationSiteDataStoreVersionId(site: string, versionId: string) {
-        const backend = this._data.replicationInfo.backends.find(
-            o => o.site === site
-        );
+        const backend = this._data.replicationInfo.backends.find(o => o.site === site);
         if (backend) {
             backend.dataStoreVersionId = versionId;
         }
@@ -1191,9 +1177,7 @@ export default class ObjectMD {
     }
 
     getReplicationSiteDataStoreVersionId(site: string) {
-        const backend = this._data.replicationInfo.backends.find(
-            o => o.site === site
-        );
+        const backend = this._data.replicationInfo.backends.find(o => o.site === site);
         if (backend) {
             return backend.dataStoreVersionId;
         }
@@ -1510,37 +1494,37 @@ export default class ObjectMD {
     }
 
     /**
-    * Set deleted flag
-    * @param {Boolean} value deleted object
-    * @return {ObjectMD}
-    */
+     * Set deleted flag
+     * @param {Boolean} value deleted object
+     * @return {ObjectMD}
+     */
     setDeleted(value) {
         this._data.deleted = value;
         return this;
     }
 
     /**
-    * Get deleted flag
-    * @return {Boolean}
-    */
+     * Get deleted flag
+     * @return {Boolean}
+     */
     getDeleted() {
         return !!this._data.deleted;
     }
 
     /**
-    * Set isPHD flag
-    * @param {Boolean} value isPHD value
-    * @return {ObjectMD}
-    */
+     * Set isPHD flag
+     * @param {Boolean} value isPHD value
+     * @return {ObjectMD}
+     */
     setIsPHD(value) {
         this._data.isPHD = value;
         return this;
     }
 
     /**
-    * Get isPHD flag
-    * @return {Boolean}
-    */
+     * Get isPHD flag
+     * @return {Boolean}
+     */
     getIsPHD() {
         return !!this._data.isPHD;
     }
--- a/lib/models/ObjectMD.ts
+++ b/lib/models/ObjectMD.ts
@@ -86,6 +86,10 @@ export type ObjectMDData = {
     replicationInfo: ReplicationInfo;
     dataStoreName: string;
     originOp: string;
+    traceContext?: {
+        traceparent: string;
+        tracestate?: string;
+    };
     microVersionId?: string;
     // Deletion flag
     // Used for keeping object metadata in the oplog event
--- a/lib/storage/metadata/captureTraceContext.ts
+++ b/lib/storage/metadata/captureTraceContext.ts
@@ -1,0 +1,51 @@
+import { context, propagation, trace } from '@opentelemetry/api';
+import ObjectMD from '../../models/ObjectMD';
+
+/**
+ * Stamp the currently-active OTEL W3C trace context onto a
+ * metadata-shaped object's `traceContext` field, in place. Always
+ * reflects the current write: if no span is active (OTEL disabled, or
+ * called outside a traced request), clears any previously-set
+ * traceContext so stale context loaded from storage is not carried
+ * forward into a new oplog entry.
+ *
+ * Mongo writes go through `MongoUtils.serialize`, which calls this — so
+ * callers there don't invoke this directly. Use this helper directly
+ * only at write sites that bypass serialize (e.g. paths that hand off
+ * a pre-built `ObjectMDData` to bulkWrite without going through tag
+ * escape).
+ *
+ * Accepts either an `ObjectMD` instance (stamps onto its underlying
+ * `getValue()` object) or a raw metadata object.
+ */
+export function stampActiveTraceContext(objMd: ObjectMD): void;
+export function stampActiveTraceContext(data: Record<string, any>): void;
+export function stampActiveTraceContext(data: ObjectMD | Record<string, any>): void {
+    const target: Record<string, any> = data instanceof ObjectMD ? data.getValue() : data;
+
+    const ctx = context.active();
+    const span = trace.getSpan(ctx);
+    if (!span) {
+        delete target.traceContext;
+        return;
+    }
+
+    const carrier: Record<string, string> = {};
+    propagation.inject(ctx, carrier, {
+        set: (c, k, v) => {
+            c[k] = v;
+        },
+    });
+    if (!carrier.traceparent) {
+        delete target.traceContext;
+        return;
+    }
+
+    const out: { traceparent: string; tracestate?: string } = {
+        traceparent: carrier.traceparent,
+    };
+    if (carrier.tracestate) {
+        out.tracestate = carrier.tracestate;
+    }
+    target.traceContext = out;
+}
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -33,7 +33,7 @@ import {
     UpdateFilter,
     MongoServerError,
     MongoBulkWriteError,
-    WriteError
+    WriteError,
 } from 'mongodb';
 import { v4 as uuidv4 } from 'uuid';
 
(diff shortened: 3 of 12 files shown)


lib/models/ObjectMD.ts
```diff
@@ -89,6 +89,10 @@ export type ObjectMDData = {
     replicationInfo: ReplicationInfo;
     dataStoreName: string;
     originOp: string;
+    traceContext?: {
+        traceparent?: string;
+        tracestate?: string;
```
DarkIsDude:
```suggestion
        parentId?: string;
        state?: string;
```

State is a string or a list of possible value ?

lib/models/ObjectMD.ts
```diff
@@ -1442,6 +1446,33 @@ export default class ObjectMD {
...
+     * a stale context does not get carried forward from an existing
+     * ObjectMD loaded from storage.
+     * @param tc - W3C trace context carrier
+     * @return itself
+     */
+    setTraceContext(tc?: { traceparent?: string; tracestate?: string }) {
+        if (tc && tc.traceparent) {
+            this._data.traceContext = tc;
+        } else {
+            delete this._data.traceContext;
```
DarkIsDude: Should we have a dedicated method to delete the context ? It can be error prone right now

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2073,6 +2089,7 @@ class MongoClientInterface {
                 const obj = doc.value;
                 const objMetadata = new ObjectMD(obj.value);
                 objMetadata.setOriginOp(originOp);
+                objMetadata.setTraceContext(captureCurrentTraceContext());
```
DarkIsDude: The goal is here to delete the object, does it make sense to save data?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1635,6 +1642,15 @@ class MongoClientInterface {
         const masterKey = formatMasterKey(objName, vFormat);
         MongoUtils.serialize(objVal);
         objVal.originOp = 's3:ObjectRemoved:Delete';
+        const tc = captureCurrentTraceContext();
```
DarkIsDude: We can't use the models function ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude, line 1: We should have test for this one. It's the most important one

DarkIsDude (changes requested):

lib/models/ObjectMD.ts
delthas: Those names are standard, come from a W3C spec: https://www.w3.org/TR/trace-context/

lib/storage/metadata/mongoclient/MongoClientInterface.ts
delthas: I'd say yes: we likely currently set the originOp today to trigger a downstream processing by smth that will consume the oplog. Whatever work is being started due to this should be so in the context of this trace.

lib/models/ObjectMD.ts
DarkIsDude: @delthas indeed, I saw that we use them in several place in the code but when stored in mongo, under the traceContext structure, it don't make sense anymore to keep `trace`

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: still no tests here ?

lib/models/ObjectMD.ts
delthas: I'd heavily lean towards keeping the standard names, as they have specific formats, with meanings specified in the specification. Specifically, `traceparent` is not a plain `parentId` at all, it is a carefull formatted string, such as `traceparent: 00-0af7651916cd43dd8448eb211c80319c-b9c7c989f97918e1-01`, which carries a version, a trace ID, a span ID, and flags, including the standard "is sampled" flag. `parentId` would be outright wrong, but even if we gave it a better name, it would be less precise and more confusing than `traceparent`, which is exactly the format the consumer would use. For example, see my [backbeat code that recreates an OTEL span from saved trace context](https://github.com/scality/backbeat/pull/2733/changes#diff-e73e540de402c8b7e8cd55a8bae58b98df0f76858c34865214e7e02bb037b6c3R27): tracecontext and traceparent are the field names used by the library.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
delthas: They were added here https://github.com/scality/Arsenal/pull/2611/changes#diff-d1ea7d7e5de4caac528eacf83f3d202b6a4842a027751f922700826668d13256R40

lib/models/ObjectMD.ts
DarkIsDude: If you prefer to keep it, then let's go. It still don't make sense for me but I'm fine to move forward

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: Thanks a lot 🙏, I missed them 😬. Why you don't add them in MongoClientInterface.*.spec.ts ? It's the standard way that we try to respect

delthas: Hmm, will look into this. I think initially this made sense but got refactored and now not so much.

delthas: Hmm yeah done mainly to refactor OTEL setup. Now split.

DarkIsDude (approved):

lib/models/ObjectMD.ts
```diff
@@ -1442,6 +1450,29 @@ export default class ObjectMD {
...
+     * Attach a W3C trace context to the metadata so it ends up in
+     * the MongoDB oplog and downstream consumers can continue the trace.
+     * Always reflects the current write: passing undefined (or a value
+     * without traceparent) clears any previously-set traceContext so that
+     * a stale context does not get carried forward from an existing
+     * ObjectMD loaded from storage.
+     * @param tc - W3C trace context carrier
+     * @return itself
+     */
+    setTraceContext(tc?: TraceContextCarrier) {
```
francoisferrand: I wonder if this is correct.
* By relying on a call to `setTraceContext`, some (bad) code may unexpectedly update the metadata without updating the trace context : and we would end up with incorrect trace... (This has happened before with changes being made without calling `setOriginOp()`)
* Especially, updating metadata (as in "manipulating the ObjectMD") may be done by backbeat, but the actual write request is made by Cloudserver : so the traceContext should be the one at the bottom (in cloudserver) - and will however "derive" from the traceContext in backbeat
* It may thus be safer to handle this at lower level, like in MetadataWrapper (or even below, in specific backends? would allow correlating more details, like metadata or mongo inner details, but requires to duplicate the code and may break layering/abstraction...)

In addition (and maybe this becomes irrelevant with the above point, or maybe not): I know ObjectMD is mostly written with simple accessors, but this is causing issues. Would be better to allow passing the trace context at the proper point in the lifecycle of the object, and introduce the parameter at that specific point. For exemple, if trace should be passed whenever we make a new MD op, this should be a second parameter of `setOriginOp()`

francoisferrand: ok, looking below it is applied in mongo client : so correct indeed, but then the problem is more related to abstraction. Setting the trace context SHOULD NOT be done by "users" of ObjectMD (cloudserver and backbeat mostly), but only by metadata backend
→ so we should not have this method.

also it will not be present for other backends -esp. metadata- : should we consider handling this in `MetadataWrapper` instead of the actual backend? If we stick with current approach (in mongoclientinterface) there should be a ticket to do the same in metadata (and thus probably export `applyTraceContext`, as it will be used there)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1361,6 +1363,7 @@ class MongoClientInterface {
         cb: ArsenalCallback<string | void>,
     ): void {
         MongoUtils.serialize(objVal);
+        applyTraceContext(objVal, captureCurrentTraceContext());
```
francoisferrand: since we always serialize objet before writing it, should we not apply traceContext in `serialize` ?

lib/storage/metadata/captureTraceContext.ts
```diff
@@ -0,0 +1,60 @@
...
+ * ObjectMDData manipulation and the ObjectMD model setter — so the
+ * set/clear semantics live in exactly one place. Always reflects the
+ * current write: passing undefined (or a value without `traceparent`)
+ * clears any previously-set context so a stale value is not carried
+ * forward from a loaded ObjectMD.
+ *
+ * Generic over `data` so callers can use either an ObjectMD's `_data`
+ * or a raw `ObjectMDData` without type gymnastics.
+ */
+export function applyTraceContext(
```
francoisferrand: this function is always called with `captureCurrentTraceContext()` in argument: should we not merge them?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1284,6 +1285,7 @@ class MongoClientInterface {
                 const obj = doc.value;
                 const objMetadata = new ObjectMD(obj);
                 objMetadata.setOriginOp(params.originOp);
+                objMetadata.setTraceContext(captureCurrentTraceContext());
                 objMetadata.setDeleted(true);
                 return next(null, objMetadata.getValue());
```
francoisferrand: - serialize() not called here, contrary to the other cases : not sure if this can create issue
- the whole parsing/serialization for adding 3 fields is probably overkill, when we could simply set the 3 fields manually (`obj.originOp = params.originOp ; obj.deleted = true ; ...`)
- we may even be able to use an aggregation pipeline to do this service side and possibly aomatically

not related to this PR, let's not fix it yet ; but please create a ticket to look at this

lib/models/ObjectMD.ts
delthas: Good call — addressed in the latest push. `setTraceContext` / `getTraceContext` are removed from `ObjectMD`, leaving the metadata backend as the only site that stamps trace context. The two `MongoClientInterface` sites that previously called `objMetadata.setTraceContext(...)` (`internalPutObject`, `internalDeleteObject`) now call `stampActiveTraceContext(objMetadata.getValue())` directly. The other 3 write sites get the stamp implicitly via `MongoUtils.serialize()` (see the related thread on line 1366). No public ObjectMD trace-context API remains, so cloudserver/backbeat can't accidentally bypass the contract.

Per the second part of your comment — the broader "apply the same in MetadataWrapper / other backends" cleanup is intentionally tracked separately rather than expanded in this PR.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
delthas: Done — `MongoUtils.serialize()` now also calls `stampActiveTraceContext(objMD)` after the tag escape, so this site and the other two paths that go through `serialize` (`putObjectVerCase4` at line 1136, `repair` at line 1639) get the stamp automatically. Two paths bypass `serialize` because they re-write a doc previously read from the DB with already-escaped tags (`internalPutObject`, `internalDeleteObject`) — those keep an explicit `stampActiveTraceContext` call on the value.

lib/storage/metadata/captureTraceContext.ts
delthas: Done — merged. The new `stampActiveTraceContext(data)` does both the capture and the apply in one step. The old two-function API (`applyTraceContext` + `captureCurrentTraceContext`) is gone.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
delthas: Acknowledged — agree this is worth a separate cleanup pass (the round-trip parse/serialize for 3 fields, and possibly an aggregation-pipeline replacement). I'll keep this scoped out of the current PR; we'll file a follow-up ARSN ticket separately to track it.

lib/models/ObjectMD.ts
```diff
@@ -89,6 +89,10 @@ export type ObjectMDData = {
     replicationInfo: ReplicationInfo;
     dataStoreName: string;
     originOp: string;
+    traceContext?: {
+        traceparent?: string;
+        tracestate?: string;
+    };
```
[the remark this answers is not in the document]
delthas: Applied — `traceparent` is now required inside `traceContext` in commit `d9198ea8`. Matches the invariant `stampActiveTraceContext` enforces (clears the field if no `traceparent` is available rather than writing a partial value), so downstream consumers can rely on `.traceparent` being a string whenever `traceContext` is set.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
delthas: Ticket filed: **ARSN-581** — *Clean up putObjectNoVerWithOplogUpdate write path*. Captures all three points (missing serialize, parse/serialize overkill, possible aggregation-pipeline replacement). https://scality.atlassian.net/browse/ARSN-581

lib/models/ObjectMD.ts
delthas: Follow-up on the second part of your comment — the broader bucketd/MetadataWrapper question — now filed as **ARSN-582** *Propagate W3C trace context to bucketd metadata writes*. The ticket description captures the open architectural choice (per-backend like MongoClient, vs MetadataWrapper-level for one-place stamping) along with the trade-offs, to be decided when it's picked up. https://scality.atlassian.net/browse/ARSN-582

lib/storage/metadata/captureTraceContext.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Mongo writes go through `MongoUtils.serialize`, which calls this — so
+ * callers there don't invoke this directly. Use this helper directly
+ * only at write sites that bypass serialize (e.g. paths that hand off
+ * a pre-built `ObjectMDData` to bulkWrite without going through tag
+ * escape).
+ *
+ * Generic over `data` so callers can pass either an ObjectMD's `_data`
+ * (via `objMd.getValue()`) or a raw `ObjectMDData`.
+ */
+export function stampActiveTraceContext(data: Record<string, any>): void {
```
francoisferrand: nit: `stampActiveTraceContext(objMD.getValue())` looks very weird, it reads like "modify the value I get from objMD" when it really modifies the metadata...

may be better instead to define multiple overloads?

```suggestion
export function stampActiveTraceContext(data: ObjectMD): void {
   return stampActiveTraceContext(data.getValue());
}
export function stampActiveTraceContext(data: ObjectMDData): void {
```

tests/unit/storage/metadata/captureTraceContext.spec.js
```diff
@@ -0,0 +1,109 @@
...
+    it('clears traceContext when no span is active', () => {
+        mockGetSpan.mockReturnValue(undefined);
+        const data = { traceContext: { traceparent: 'stale-prior-trace' } };
+        stampActiveTraceContext(data);
+        assert.strictEqual(data.traceContext, undefined);
+        // No injection should be attempted when there is no active span.
+        assert.strictEqual(mockInject.mock.calls.length, 0);
+    });
+
+    it('is a no-op (no traceContext field added) when no span is active and data had none', () => {
```
francoisferrand, line 37: can we (unit) test the pathologic case, when Otel is not enabled or initialized _AT ALL_ ? (i.e. the case where arsenal would be bumped, without any additional change, in cloudserver or s3util)

francoisferrand (approved):
LGTM, with just a small fear (probably not an issue, but I'd like to be sure) about behavior if/when Otel is not configured/imported/... at all in the component (c.f. comment).

Which brings a related question: shoud this be merged in 8.3, or in a new branch?

lib/storage/metadata/captureTraceContext.ts
delthas: Done — `stampActiveTraceContext` now has overloads so callers pass the `ObjectMD` instance directly (no more `.getValue()` at the call site):

```ts
export function stampActiveTraceContext(objMd: ObjectMD): void;
export function stampActiveTraceContext(data: Record<string, any>): void;
export function stampActiveTraceContext(data: ObjectMD | Record<string, any>): void {
    const target = data instanceof ObjectMD ? data.getValue() : data;
    // ...
}
```

The two `MongoClientInterface` call sites now pass `objMetadata` instead of `objMetadata.getValue()`. (Same overload pattern already used in arsenal: `convertToEpochTime` in `lib/policyEvaluator/utils/conditions.ts` and `uuidv4` in `lib/network/kmip/index.ts`.)

One deviation from your snippet: the non-`ObjectMD` overload stays `Record<string, any>` rather than `ObjectMDData`. `MongoUtils.serialize()` passes its loosely-typed local `ObjectMetadata` (`{ tags?; [key: string]: any }`), which isn't assignable to the strict `ObjectMDData` (lots of required fields). Tightening that path is out of scope here, so the data overload stays permissive.

tests/unit/storage/metadata/captureTraceContext.spec.js
delthas: Added — new mock-free spec `tests/unit/storage/metadata/captureTraceContext.noOtel.spec.js` that drives the **real** `@opentelemetry/api` with no SDK / propagator registered (exactly the "arsenal bumped without OTEL set up in cloudserver/s3utils" case). With no SDK, `trace.getSpan(context.active())` returns `undefined`, so the function is a safe no-op:

```js
it('is a safe no-op and clears a stale traceContext', () => {
    const data = { foo: 'bar', traceContext: { traceparent: 'stale' } };
    assert.doesNotThrow(() => stampActiveTraceContext(data));
    assert.strictEqual('traceContext' in data, false);
});
it('adds no traceContext field when none existed', () => { /* ... */ });
```

Kept it in a separate file (rather than splitting the existing spec into mocked/unmocked describes) so each file has one consistent module-mock state — no `jest.isolateModules` juggling. The existing mocked spec also gained one case covering the new `ObjectMD`-instance overload.

delthas:
> LGTM, with just a small fear (probably not an issue, but I'd like to be sure) about behavior if/when Otel is not configured/imported/... at all in the component (c.f. comment).
> 
> Which brings a related question: shoud this be merged in 8.3, or in a new branch?

Added a test for this; no issues.

But moving to 8.4 since the other repos for tracing will target their latest branch, too.

tests/unit/storage/metadata/mongoclient/delObject.spec.js
```diff
@@ -283,3 +319,72 @@ describe('MongoClientInterface:delObject', () => {
...
+        };
+    });
+
+    afterEach(() => {
+        sinon.restore();
+    });
+
+    it('stamps the active traceContext on the delete tombstone', done => {
+        otel.activateSpan();
+        client.internalDeleteObject(collection, 'bucket', 'example', {}, 's3:ObjectRemoved:Delete', logger, err => {
```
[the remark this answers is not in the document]
delthas: Good catch — fixed in commit `2d779a3c`. The string was in the `params` slot; all three calls now pass `null` for `params` and let `originOp` default to `'s3:ObjectRemoved:Delete'` (same value, so behavior is unchanged):

```js
client.internalDeleteObject(collection, 'bucket', 'example', {}, null, logger, err => {
```

Tests still pass (22/22 in delObject.spec.js).

delthas:
Rebased. No signficiant changes, so approving now.

delthas:
> LGTM — well-designed, thorough test coverage, correct handling of all write paths.One minor note: `yarn.lock` now has two resolutions for `@opentelemetry/api` (1.9.0 from an existing transitive dep and 1.9.1 from the new direct dep). Not a functional issue since the OTEL API package handles multi-version scenarios via its global registration design, but a `yarn dedupe @opentelemetry/api` would clean it up.Review by Claude Code

Ack, we don't want an explicit resolutions block and dedupe doesnt exist on yarn v1 which we use. Leaving as is.

PR #2622, opened 2026-04-24
Fix ETag fallback to dataStoreETag for ring-s3-rdma-http objects

borisfaure:
When content-md5 is absent (e.g. parts written by ring-s3-rdma-http which stores ETag only as "1:<hex>" in dataStoreETag), fall back to extracting the hex portion from dataStoreETag for use as the ETag in list entries.

Issue: ARSN-579

### DIFF
--- a/lib/storage/metadata/MetadataWrapper.js
+++ b/lib/storage/metadata/MetadataWrapper.js
@@ -41,11 +41,27 @@ function _parseListEntries(entries) {
             };
         }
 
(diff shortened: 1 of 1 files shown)


francoisferrand:
Falling back to datastore ETag seems very risky, as ETag may thus change: for exemple in case of transition (or other obscure features: transient locations...) to another storage location.
→ ETag should be written by Cloudserver when creating the object.

borisfaure:
I'll do it in a better way one day

borisfaure:
This is no longer needed

PR #2627, opened 2026-05-13
Support multi-destination CRR with V2 replication configuration

maeldonn:
Add V2 replication configuration support (Filter, Priority, per-rule Account) with multi-destination CRR. V1 remains fully supported and round-trips via a persisted format hint. Per-rule destination/role move into the Backend; legacy top-level ReplicationInfo fields kept as optional for backward-compatible reads.

Issue: ARSN-571


lib/models/ReplicationConfiguration.ts
```diff
@@ -265,10 +369,61 @@ export default class ReplicationConfiguration {
...
+     */
+    _parsePriority(rule: XMLRule) {
+        if (this._format !== 'v2') {
+            return undefined;
+        }
+        if (!rule.Priority || !rule.Priority[0]) {
+            return undefined;
+        }
+        const priority = +rule.Priority[0];
+        if (isNaN(priority) || priority < 0) {
```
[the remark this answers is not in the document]
maeldonn: Updated.

lib/models/ReplicationConfiguration.ts
```diff
@@ -435,13 +590,36 @@ export default class ReplicationConfiguration {
...
+    /**
+     * Check that the `Account` property is valid (V2 only)
+     * @param destination - The destination object from this._parsedXML
+     */
+    _parseAccount(destination: Destination) {
+        if (this._format !== 'v2' || !destination.Account) {
+            return undefined;
+        }
+
+        const account = destination.Account[0];
```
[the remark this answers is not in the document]
maeldonn: Updated.

lib/models/ObjectMD.ts
```diff
@@ -23,17 +23,19 @@ export type Backend = {
...
-    destination: string;
-    storageClass: string;
-    role: string;
-    storageType: string;
-    dataStoreVersionId: string;
+    destination?: string;
+    storageClass?: string;
+    role?: string;
+    storageType?: string;
+    dataStoreVersionId?: string;
```
francoisferrand, line 50: these fields are actually "deprecated" (as in, should not be used anymore ; and will not be set by Backbeat anymore), right?
→ can we remove them already (I don't think it creates a parsing issue, so no issue with existing metadata in prod: it will just make TS "know" we should not access them), or do we still use them for some fallback?
→ if not, should be indicated here: it's not just "optional" but really deprecated and only used for backward compatibility

lib/models/ObjectMD.ts
```diff
@@ -23,17 +23,19 @@ export type Backend = {
     site: string;
     status: string;
     dataStoreVersionId: string;
+    destination?: string;
+    role?: string;
```
francoisferrand, line 24: mandatory for CRR, right?

lib/models/ObjectMD.ts
```diff
@@ -1254,7 +1256,7 @@ export default class ObjectMD {
     }
 
     getReplicationTargetBucket() {
```
francoisferrand: should this function still be used?
(if it is just for backwards compatibility with existing backbeat, maybe time to create a new arsenal branch, that we'll bump in the next backbeat branch)

lib/models/ReplicationConfiguration.ts
```diff
@@ -282,10 +437,10 @@ export default class ReplicationConfiguration {
     }
 
     /**
-     * Check that the `Prefix` property is valid
+     * Check that the `Prefix` property is valid (V1 format)
      * @param rule - The rule object from this._parsedXML
      */
-    _parsePrefix(rule: XMLRule) {
+    _parsePrefixV1(rule: XMLRule) {
```
francoisferrand: instead of the whole v1 vs v2 "mode", and parsing twice, can't we do this locally here?

only (significant) difference is `Prefix` vs `Rules.Prefix` : we could be permissive about priority and account, and handle here in a few lines:

```js
if (Array.isArray(rule.Prefix) && Array.isArray(rule.Filter)) {
    return errors.MalformedXML; // both v0 and v1
}
if (Array.isArray(rule.Prefix)) {
    prefix = rule.Prefix?.[0];
    format = v1
} else {
    prefix = rule.Filter[0].Prefix?.[0];
    format = v2
}
```

lib/models/ReplicationConfiguration.ts
```diff
@@ -144,34 +174,95 @@ export default class ReplicationConfiguration {
...
+     * Detect whether the configuration uses V2 format.
+     * V2 is identified by the presence of any V2-only element:
+     * Filter, Priority, or Account (per-rule destination account).
+     */
+    _detectFormat(rules: XMLRule[]) {
+        const isV2 = rules.some(rule =>
+            rule.Filter ||
+            rule.Priority ||
+            rule.Destination?.[0]?.Account);
+        this._format = isV2 ? 'v2' : 'v1';
```
francoisferrand: are account and priority really specific to v2 ?

it seems to me v2 was more about the (advanced) filters, but no issue having account or priority in v1...

lib/models/ReplicationConfiguration.ts
```diff
@@ -144,34 +174,95 @@ export default class ReplicationConfiguration {
...
         };
     }
 
+    /**
+     * Detect whether the configuration uses V2 format.
+     * V2 is identified by the presence of any V2-only element:
+     * Filter, Priority, or Account (per-rule destination account).
+     */
+    _detectFormat(rules: XMLRule[]) {
+        const isV2 = rules.some(rule =>
```
francoisferrand: this is kind of doing the parsing twice : first scan all rules to idenfiy format, then parse again...
a few cycles lost, and most importantly maybe it creates coupling by doing the same thing twice (e.g. the knowledge of fitler/prefix format is in 2 places)

lib/models/ReplicationConfiguration.ts
```diff
@@ -144,34 +174,95 @@ export default class ReplicationConfiguration {
             destination: this.getDestination(),
             rules: this.getRules(),
             preferredReadLocation: this.getPreferredReadLocation(),
+            format: this.getFormat(),
```
francoisferrand, line 178: already computed in `_parseRules()`, and stored in `this._format` ?

lib/models/ReplicationConfiguration.ts
```diff
@@ -265,10 +369,61 @@ export default class ReplicationConfiguration {
...
+
+        return this._parsePrefixV1(rule);
+    }
+
+    /**
+     * Parse and validate the V2 Filter element. Filter is optional in
+     * V2; when absent or empty, the rule matches all objects.
+     */
+    _parseFilter(rule: XMLRule) {
+        const prefix = this._extractPrefix(rule);
```
francoisferrand, line 527: this function handles both v1 and v2 ; so do we really needed _parseFilter / _parsePrefixOfFilter (instead of just using _parseFilter, and litting it call _extraPrefix for v1/v2 compat' ?)

lib/models/ReplicationConfiguration.ts
```diff
@@ -435,13 +590,36 @@ export default class ReplicationConfiguration {
...
+        if (this._format === 'v1') {
+            if (this._destination && this._destination !== bucketARN) {
+                return errorInstances.InvalidRequest.customizeDescription(
+                    'The destination bucket must be the same for all rules'
+                );
+            }
+            this._destination = bucketARN;
+        }
+
+        // In V2, the destination lives on each rule; no top-level value.
```
francoisferrand: the destination is inside the rule, both v1 and v2 (i.e. AWS api concept, not our own internal storage format)

lib/models/ReplicationConfiguration.ts
```diff
@@ -474,44 +655,86 @@ export default class ReplicationConfiguration {
...
         }
-        return this._parseRole();
+        const roleErr = this._parseRole();
+        if (roleErr) {
+            return roleErr;
+        }
+        // Derive per-rule roles from Account after the top-level Role
+        // has been parsed, since role derivation depends on this._role.
+        if (this._format === 'v2' && this._rules) {
+            this._derivePerRuleRoles();
```
francoisferrand: should we do this at "parse" time (as you did), and internally store the role in each rule? or do this dynamically whenever the role is used?

```
- takes (some) space in metadata
- dupicates the information
+ may be slightly more efficient
+ maybe needed for better abstraction in other layers (i.e. work with "indepdendent" rules, not the whole replicationConfiguration)
```

lib/models/ReplicationConfiguration.ts
```diff
@@ -474,44 +655,86 @@ export default class ReplicationConfiguration {
...
+        }
+        return undefined;
+    }
+
+    /**
+     * For V2 configurations, derive per-rule destination roles from
+     * the Account field by replacing the account ID in the destination
+     * role from the top-level Role field.
+     */
+    _derivePerRuleRoles() {
```
francoisferrand: should this not be done directly in parseAccount? it already goes through the "account" field, if any

lib/models/ReplicationConfiguration.ts
```diff
@@ -474,44 +655,86 @@ export default class ReplicationConfiguration {
...
+        const { Rule } = this._parsedXML.ReplicationConfiguration;
+        this._rules!.forEach((rule, i) => {
+            const account = Rule[i].Destination[0].Account?.[0];
+            if (account) {
+                const destRole = this._deriveDestinationRole(account);
+                if (destRole) {
+                    rule.role = destRole;
```
francoisferrand: not stored everytime (e.g. if no account) : so code using role still needs to be aware, cannot be simplified that much...

lib/models/ReplicationConfiguration.ts
```diff
@@ -474,44 +655,86 @@ export default class ReplicationConfiguration {
...
-            })
-            .join('');
+        const { role, destination, rules, format } = config;
+
+        // Prefer the explicit format hint. Fall back to inferring from
+        // per-rule `destination`, which the parser always sets for V2
+        // rules (and never for V1).
+        const isV2 = format
+            ? format === 'v2'
+            : rules.some(r => r.destination !== undefined);
```
francoisferrand: seems weird to couple the output format to our internal representation.
in both AWS v1/v2 there is a `Destination`, so should not be used to distinguish...

at the same time, the difference between v1 and v2 are really about prefix.

while we could argue about being strict on parsing, there is no such problem here:
- we can generate `priority` if set (it won't be set on v1 if not allowed, no point depending on format)
- we can generate `account` if set (it won't be set on v1 if not allowed, no point depending on format)
- we need to know the format for `prefix`, but only if a prefix was actually set: if it was not don't generate anything (no prefix or filter), and if it was we know the format

francoisferrand (commented):
can be greatly simplified by not being strict in parsing:
- always parse `account`
- always parse `priority`
- when parsing prefix, remember if it was in a <filter> or directly in the <rule>

This will always lots of back-and-forth and coupling with the `_v2` variable in parsing, and simplify generation.

(And if we really want or need to be strict in parsing, we can add an extra validation at the end of parsing : after we have parsed everything, fail if `format == v1 && account && priority`. But IMHO not needed...)

lib/models/ReplicationConfiguration.ts
```diff
@@ -28,37 +28,61 @@ const validStorageClasses = ['STANDARD', 'STANDARD_IA', 'REDUCED_REDUNDANCY'];
...
+    Priority?: string[];
 };
 
 export type ReplicationConfigurationMetadata = {
     role: string,
-    destination: string,
+    destination?: string,
     rules: Rule[],
     preferredReadLocation?: string | null,
+    format?: 'v1' | 'v2',
```
DarkIsDude: that should be a new type reused everywhere?

lib/models/ReplicationConfiguration.ts
```diff
@@ -87,15 +112,13 @@ export default class ReplicationConfiguration {
...
         this._role = null;
         this._destination = null;
         this._rules = null;
         this._prevStorageClass = null;
         this._hasScalityDestination = null;
         this._preferredReadLocation = null;
+        this._format = 'v1';
     }
 
     /**
```
DarkIsDude, line 127: Why default to v1 ?

DarkIsDude: I mean if the v2 is future, we should use v2 as default ?

DarkIsDude (changes requested):
Not much more than Francois 🙏

lib/models/ObjectMD.ts
maeldonn: Added a `@deprecated` jsdoc comment.

lib/models/ObjectMD.ts
maeldonn: Yes, but optional for backward compatibility.

lib/models/ObjectMD.ts
maeldonn: This function is now deprecated.

lib/models/ReplicationConfiguration.ts
maeldonn: Yes they are, but we are going to include them in v1

lib/models/ReplicationConfiguration.ts
maeldonn: Done, simplified.

lib/models/ReplicationConfiguration.ts
maeldonn: Updated.

lib/models/ReplicationConfiguration.ts
maeldonn: Updated.

lib/models/ReplicationConfiguration.ts
maeldonn: Collapsed into a single _parsePrefix.

lib/models/ReplicationConfiguration.ts
maeldonn: Updated.

lib/models/ReplicationConfiguration.ts
maeldonn: Done.

lib/models/ReplicationConfiguration.ts
maeldonn: Default is now null. If no filter specified in the rule, format surfaces 'v2' as the default when no rule provides a prefix.

lib/models/ReplicationConfiguration.ts
maeldonn: Can't be done in parseAccount, at that point the role hasn't been parsed yet. The derivation needs the top-level Role, so it's deferred once both are known.

lib/models/ReplicationConfiguration.ts
maeldonn: Switched to storing the per-rule `account` instead of the derived role. Top-level `Role` stays the single source of truth, and `resolveDestinationRole()` is exposed as a static helper for consumers.

lib/models/ReplicationConfiguration.ts
maeldonn: resolveDestinationRole returns the right ARN whether or not the rule has an `account`.

DarkIsDude (approved):
LGTM 🙏

lib/models/ObjectMD.ts
francoisferrand: should we not remove them?

if they are used, some features will not work: so they MUST NOT be used in newer cloudserver/backbeat/s3utils which support multi-destination/... ?
→ if this is correct, we can remove them UNLESS we want to use this arsenal branch in older branches?

lib/models/ReplicationConfiguration.ts
```diff
@@ -153,25 +185,78 @@ export default class ReplicationConfiguration {
...
         // StorageClass is an optional property.
-        if (rule.Destination[0].StorageClass) {
-            obj.storageClass = rule.Destination[0].StorageClass[0];
+        const storageClass = rule.Destination[0].StorageClass?.[0];
+        if (storageClass) {
+            obj.storageClass = storageClass;
+        }
+
+        if (rule.Priority?.[0] !== undefined) {
+            obj.priority = +rule.Priority[0];
```
francoisferrand: would be more explicit with `Number` or `parseInt`
```suggestion
            obj.priority = Number(rule.Priority[0]);
```

tests/unit/models/ReplicationConfiguration.spec.ts
```diff
@@ -507,4 +560,664 @@ describe('ReplicationConfiguration.parseConfiguration()', () => {
...
+                }],
+            };
+            const instance = new ReplicationConfiguration(
+                { ReplicationConfiguration: repConfig }, null, mockS3ServerConfig);
+            const result = instance.parseConfiguration();
+            expect(result).toBeUndefined();
+            const rules = instance.getRules();
+            expect(rules.length).toEqual(1);
+            expect(rules[0].prefix).toEqual('docs/');
+            expect(rules[0].destination).toEqual('arn:aws:s3:::crr-dest');
```
francoisferrand, line 806: expect(instance.getFormat()).toEqual('v1');

lib/models/ReplicationConfiguration.ts
```diff
@@ -265,32 +352,54 @@ export default class ReplicationConfiguration {
 
             rulesArr.push(this._buildRuleObject(rules[i]));
         }
+
+        const hasAnyPriority = rulesArr.some(r => r.priority !== undefined);
```
francoisferrand: the rule applies in any case : when 2 rules have overlapping prefix, they MUST have either a different destination (i.e. region for AWS, storageClass for us) or different priority

i.e. algorithm may be to create one "bucket" for each destination/priority, and compare prefixes in the bucket (or do a NxN comparison: check each rule either does not have the same destination, priority or "startsWith" all other rules)

tests/unit/models/ReplicationConfiguration.spec.ts
```diff
@@ -507,4 +552,596 @@ describe('ReplicationConfiguration.parseConfiguration()', () => {
...
+                    priority: 1,
+                    storageClass: 'ring',
+                    destination: 'arn:aws:s3:::bucket-a',
+                    role: 'arn:aws:iam::222222222222:role/crr-trust-role',
+                }],
+                format: 'v2',
+            })).not.toThrow();
+        });
+    });
 });
```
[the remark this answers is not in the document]
francoisferrand: round-trip test for both v1 and v2 would be nice... (including the "unknown" case)

lib/models/ReplicationConfiguration.ts
```diff
@@ -474,44 +619,52 @@ export default class ReplicationConfiguration {
...
+        const isV2 = (format ?? 'v1') === 'v2';
+
+        const rulesXML = rules.map(rule => {
+            const { prefix, enabled, storageClass, id, priority, destination: ruleDest, account } = rule;
+
+            const ID = `<ID>${escapeForXml(id)}</ID>`;
+            const Status = `<Status>${enabled ? 'Enabled' : 'Disabled'}</Status>`;
+
+            const prefixContent = prefix === '' ? '<Prefix/>' : `<Prefix>${escapeForXml(prefix)}</Prefix>`;
+            const prefixXML = isV2 ? `<Filter>${prefixContent}</Filter>` : prefixContent;
```
francoisferrand: do we HAVE TO specify a prefix (i.e. empty <Prefix/>) ?

if the user sends no prefix at all (neither v1 nor v2), this XML serialization would end up adding an empty filter/prefix : `<Filer><Prefix/></Filter>` instead of "nothing" ?

lib/models/ReplicationConfiguration.ts
maeldonn: Updated.

lib/models/ReplicationConfiguration.ts
maeldonn: Tested against AWS (PUT then GET, raw XML on the wire):

| Input | AWS returns |
|---|---|
| v2 `<Filter/>` | `<Filter/>` |
| v2 `<Filter><Prefix></Prefix></Filter>` | `<Filter><Prefix></Prefix></Filter>` |
| v1 `<Prefix></Prefix>` | `<Prefix></Prefix>` |

AWS preserves "no Prefix" vs "empty Prefix" on the wire even though they're semantically equivalent. We were collapsing both to `prefix: ''` and always emitting `<Prefix/>`.

Fixed in the latest push: `Rule.prefix` is now optional. Parser returns `undefined` when no `<Prefix>` is present; `getConfigXML` emits `<Filter/>` for v2 / nothing-but-`<Prefix/>` for v1 when undefined. Added a round-trip test covering both shapes.

lib/models/ReplicationConfiguration.ts
```diff
@@ -153,25 +186,165 @@ export default class ReplicationConfiguration {
...
+            const newP = item.rule.priority ?? -Infinity;
+            const curP = cur?.rule.priority ?? -Infinity;
+            if (!cur || newP > curP) {
+                byKey.set(dedupKey, item);
+            }
+        }
+
+        const backends: Backend[] = [];
+        for (const item of byKey.values()) {
+            const existing = existingBackends?.find(b => b.site === item.site);
```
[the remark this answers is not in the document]
maeldonn: Updated — `resolveBackends` lookup now matches existing CRR backends on the `(site, destination, role)` tuple. Regression test added.

lib/models/ReplicationConfiguration.ts
```diff
@@ -153,25 +186,158 @@ export default class ReplicationConfiguration {
...
```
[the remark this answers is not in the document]
maeldonn: Updated — `resolveBackends` lookup now matches existing CRR backends on the `(site, destination, role)` tuple. Regression test added.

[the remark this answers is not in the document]
maeldonn: Updated — `resolveBackends` lookup now matches existing CRR backends on the `(site, destination, role)` tuple. Regression test added.

lib/models/ReplicationConfiguration.ts
```diff
@@ -153,25 +186,158 @@ export default class ReplicationConfiguration {
...
```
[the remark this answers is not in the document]
maeldonn: Updated — `resolveBackends` lookup now matches existing CRR backends on the `(site, destination, role)` tuple. Regression test added.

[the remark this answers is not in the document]
maeldonn: Updated — `resolveBackends` lookup now matches existing CRR backends on the `(site, destination, role)` tuple. Regression test added.

lib/models/ReplicationConfiguration.ts
```diff
@@ -265,56 +432,132 @@ export default class ReplicationConfiguration {
...
+                }
+            }
+        }
+
+        for (const group of groups.values()) {
+            if (group.length < 2) {
+                continue;
+            }
+            group.sort((a, b) => (a.prefix < b.prefix ? -1 : a.prefix > b.prefix ? 1 : 0));
+            for (let i = 1; i < group.length; i++) {
```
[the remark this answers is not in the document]
maeldonn: Updated — `_findRuleOverlap` now compares all pairs within each site group (no longer adjacent-only). Regression test added for the non-adjacent case.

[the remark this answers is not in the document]
maeldonn: Updated — `_findRuleOverlap` now compares all pairs within each site group (no longer adjacent-only). Regression test added for the non-adjacent case.

lib/models/ReplicationConfiguration.ts
```diff
@@ -153,25 +185,78 @@ export default class ReplicationConfiguration {
...
+
         // StorageClass is an optional property.
-        if (rule.Destination[0].StorageClass) {
-            obj.storageClass = rule.Destination[0].StorageClass[0];
+        const storageClass = rule.Destination[0].StorageClass?.[0];
+        if (storageClass) {
+            obj.storageClass = storageClass;
+        }
+
+        if (rule.Priority?.[0] !== undefined) {
```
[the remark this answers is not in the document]
maeldonn: Updated — `_parsePriority` now rejects empty `<Priority>` explicitly; `_buildRuleObject` gates the `Number()` conversion on a typeof check. Test added.

lib/models/ReplicationConfiguration.ts
```diff
@@ -144,6 +175,7 @@ export default class ReplicationConfiguration {
             destination: this.getDestination(),
```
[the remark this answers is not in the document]
maeldonn: Updated — `_destination` is now `string | undefined`, matching `ReplicationConfigurationMetadata.destination`.

lib/models/ObjectMD.ts
```diff
@@ -1151,72 +1156,60 @@ export default class ObjectMD {
...
     }
 
-    setReplicationSiteStatus(site: string, status: string) {
-        const backend = this._data.replicationInfo.backends.find(o => o.site === site);
+    /**
+     * With just `{ site }`, returns the first match. With
+     * `{ site, destination, role }`, uniquely identifies one backend
+     * among multiple sharing a site.
+     */
+     _findBackend(key: BackendKey): Backend | undefined {
```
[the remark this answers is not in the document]
maeldonn: Fixed.

tests/unit/models/ReplicationConfiguration.spec.ts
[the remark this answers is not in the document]
maeldonn: Round-trip tests are in place — v1, v2 multi-destination, no-prefix, empty-prefix, and legacy (no `format` field) cases are covered in the `round-trip parse / getConfigXML` describe block (spec.ts:1559).

lib/models/ObjectMD.ts
```diff
@@ -1229,20 +1213,47 @@ export default class ObjectMD {
...
+     */
+    getReplicationTargetBucket(key?: BackendKey): string {
+        if (key !== undefined) {
+            const backend = this._findBackend(key);
+            if (backend?.destination) {
+                return backend.destination.split(':').slice(-1)[0];
+            }
+        }
+        const destBucketArn = this._data.replicationInfo.destination ?? '';
         return destBucketArn.split(':').slice(-1)[0];
```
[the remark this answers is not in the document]
maeldonn: Pre-existing behavior — `''` was already the no-destination sentinel before this PR.

lib/models/ObjectMD.ts
```diff
@@ -1151,72 +1146,61 @@ export default class ObjectMD {
...
+    _findBackend(key: BackendKey): Backend | undefined {
+        return this._data.replicationInfo.backends.find(
+            b =>
+                b.site === key.site &&
+                (key.destination === undefined || b.destination === key.destination) &&
+                (key.role === undefined || b.role === key.role),
+        );
+    }
+
+    setReplicationSiteStatus(key: BackendKey, status: string) {
```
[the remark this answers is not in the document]
maeldonn: Callers in CloudServer and Backbeat are being updated in lockstep in their companion PRs.

lib/models/ObjectMD.ts
```diff
@@ -1151,72 +1146,61 @@ export default class ObjectMD {
...
-        this._data.replicationInfo.storageClass = storageClass;
-        return this;
-    }
-
-    setReplicationTargetBucket(destination: string) {
-        this._data.replicationInfo.destination = destination;
-        return this;
-    }
-
+    /** @deprecated in favor of per-backend role for multi-destination CRR. */
```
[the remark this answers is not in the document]
maeldonn: Callers in CloudServer and Backbeat are being updated in lockstep in their companion PRs.

lib/models/ObjectMD.ts
```diff
@@ -237,11 +254,11 @@ export default class ObjectMD {
...
                 backends: [],
                 content: [],
-                destination: '',
-                storageClass: '',
-                role: '',
-                storageType: '',
-                dataStoreVersionId: '',
+                destination: undefined,
+                storageClass: undefined,
+                role: undefined,
```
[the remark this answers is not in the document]
maeldonn: These fields are deprecated — downstream code is being migrated to read per-backend values from `backends[]`. The `''` vs `undefined` divergence is acceptable on a deprecated field.

lib/models/ReplicationConfiguration.ts
```diff
@@ -265,56 +436,124 @@ export default class ReplicationConfiguration {
...
 
-        const prefix = rule.Prefix?.[0] ?? '';
+        if (hasFilter || hasTopPrefix) {
+            const ruleFormat: ReplicationFormat = hasFilter ? 'v2' : 'v1';
+            if (this._format === null) {
+                this._format = ruleFormat;
+            } else if (this._format !== ruleFormat) {
+                return errors.MalformedXML;
+            }
+        }
```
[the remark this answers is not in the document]
maeldonn: `<Account>` is supported in both v1 and v2 — it is not a v2 marker, so pinning the format on its presence would be incorrect.

lib/models/ReplicationConfiguration.ts
```diff
@@ -414,34 +645,36 @@ export default class ReplicationConfiguration {
...
-            );
+            return errorInstances.InvalidArgument.customizeDescription('The specified bucket is not valid');
         }
-        // We can replicate objects only to one destination bucket.
-        if (this._destination && this._destination !== bucketARN) {
-            return errorInstances.InvalidRequest.customizeDescription(
-                'The destination bucket must be the same for all rules'
-            );
+
+        return undefined;
```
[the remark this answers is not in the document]
maeldonn: No downstream callers rely on `getDestination()` for legacy records — destinations are read per-rule now.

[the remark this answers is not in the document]
maeldonn: Intentional — multi-destination is supported in v1 as well, per the design.

lib/models/ObjectMD.ts
```diff
@@ -1092,30 +1109,8 @@ export default class ObjectMD {
...
-            content,
-            destination,
-            storageClass: storageClass || '',
-            role,
-            storageType: storageType || '',
-            dataStoreVersionId: dataStoreVersionId || '',
-            isNFS,
-        };
+    setReplicationInfo(replicationInfo: ReplicationInfo) {
+        this._data.replicationInfo = { ...replicationInfo };
```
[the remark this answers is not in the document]
maeldonn: Pre-existing behavior — the previous destructure-and-rebuild was also a shallow copy of `backends`. Not introduced by this PR.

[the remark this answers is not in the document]
maeldonn: These fields are deprecated — downstream code is being migrated to per-backend reads, so the `''` vs `undefined` divergence is acceptable.

lib/models/ObjectMD.ts
```diff
@@ -20,17 +20,34 @@ export type Backend = {
...
+
+/**
+ * Identifies one backend in `replicationInfo.backends`. Matches
+ * `ReplicationConfiguration.resolveBackends` dedup key.
+ */
+export type BackendKey = {
+    site: string;
+    destination?: string;
+    role?: string;
 };
```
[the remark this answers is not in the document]
maeldonn: Added re-exports for `BackendKey` and `ReplicationFormat` in `lib/models/index.ts`.

lib/models/ReplicationConfiguration.ts
```diff
@@ -153,25 +185,163 @@ export default class ReplicationConfiguration {
...
+     * role)`. Highest-priority rule wins on collision; missing
+     * `priority` is treated as the lowest.
+     *
+     * @param isCloud - returns true if the named site is a cloud
+     *   backend; supplied by the caller because the location-config
+     *   model lives outside this class.
+     * @param existingBackends - prior backends used to carry forward
+     *   `dataStoreVersionId` for sites that already had a backend.
+     */
+    static resolveBackends(
```
SylvainSenechal, line 269: I think I'm the only one in this team who likes this, so you may or may not do it, it's very verbose and it's gonna add 50 or 100 lines of doc : 

For these kind of function, I like having one, or even more examples (as this function handle multiple cases including legacy entries) of input this function would take, and what it would respond from that input, it helps understanding what its gonna do. This is kinda what you already documented in citadel, but I wouldn't expect people to check citadel that often (and even then they would have to go through the whole design when they just wanna understand that function)

lib/models/ObjectMD.ts
SylvainSenechal, line 1113: wonder if this will clash with : 
https://github.com/scality/Arsenal/pull/2628/changes#diff-a7c36ae2e6efb606b627a70a905847bc33a781f53eca944ec00fcbcd40583d4bR1130

lib/models/ReplicationConfiguration.ts
maeldonn: Skipping, the parameter docs and the dedup-key description in the function comment already cover the non-obvious behavior, and adding worked examples here would significantly expand the JSDoc without much additional clarity. Happy to revisit if it comes up again.

lib/models/ObjectMD.ts
maeldonn: Yes, it will textually conflict with #2628 on this method, both PRs touch the same lines. Resolvable by hand though, the resolver just needs to keep this PR's spread (which intentionally drops the `|| ''` defaults on deprecated fields) and layer #2628's `isReplica` derivation on top.

lib/models/ObjectMD.ts
```diff
@@ -1151,72 +1146,61 @@ export default class ObjectMD {
...
 
+    /** @deprecated in favor of per-backend dataStoreVersionId for multi-destination CRR. */
     setReplicationDataStoreVersionId(versionId: string) {
         this._data.replicationInfo.dataStoreVersionId = versionId;
         return this;
     }
 
-    setReplicationSiteDataStoreVersionId(site: string, versionId: string) {
-        const backend = this._data.replicationInfo.backends.find(o => o.site === site);
+    setReplicationSiteDataStoreVersionId(key: BackendKey, versionId: string) {
```
SylvainSenechal, line 1181: I know this was already the existing behavior, but reading this : I feel like if backend is not found, the function should return an error no ? 
I'm pretty sure the called of setReplicationSiteDataStoreVersionId doesn't want to continue if it was not able to set versionId on the given backendKey

lib/models/ObjectMD.ts
```diff
@@ -1229,20 +1213,47 @@ export default class ObjectMD {
         return this._data.replicationInfo.content;
     }
 
-    getReplicationRoles() {
-        return this._data.replicationInfo.role;
+    /**
```
SylvainSenechal, line 1216: nit : I think i would document this one with an example objectMd (only the fields we are interested in) of what we can have in 
replicationInfo.role and backends[], to see what i would return more concretely

maybe break down more the if statement too, that deals with 2 concerns at the same time (key undefined, and top being defined). Also I think we can expect top to always be defined since it's the historical value (unless we don't set it anymore with the new code).

SylvainSenechal: edit : I see some v1/v2 example of replication info in the next file, but still, nice to be able to see everything on the same file with a specific example for the function

lib/models/ObjectMD.ts
maeldonn: Fair point, but this is pre-existing behavior. Keeping it out of this PR to stay scoped to multi-destination CRR. Happy to do it in a follow-up ticket if you'd like.

lib/models/ObjectMD.ts
maeldonn: Added a worked example in the JSdoc.

SylvainSenechal (approved):
I think this is good, now we need thorough functional testing

PR #2628, opened 2026-05-22
Upgrade microVersionId to timestamp-ordered format and add isReplica

maeldonn:
Replace the random 8-byte hex microVersionId with a 20-character timestamp-ordered identifier matching the versionId scheme, and add encode/decode/compare helpers in a new MicroVersionID module. Add an optional isReplica flag to ReplicationInfo to distinguish replica writes from user writes, enabling cascaded CRR.

Issue: ARSN-578

Crr cascaded design : https://github.com/scality/citadel/pull/349
Related PRs :
Cloudserver : https://github.com/scality/cloudserver/pull/6179
CloudserverClient : https://github.com/scality/cloudserverclient/pull/24
Backbeat : https://github.com/scality/backbeat/pull/2747
S3utils : https://github.com/scality/s3utils/pull/395

### DIFF
--- a/lib/models/ObjectMD.ts
+++ b/lib/models/ObjectMD.ts
@@ -1,4 +1,3 @@
-import * as crypto from 'crypto';
 import * as constants from '../constants';
 import * as VersionIDUtils from '../versioning/VersionID';
 import { VersioningConstants } from '../versioning/constants';
@@ -49,6 +48,7 @@ export type ReplicationInfo = {
     /** @deprecated in favor of per-backend dataStoreVersionId for multi-destination. */
     dataStoreVersionId?: string;
     isNFS?: boolean;
+    isReplica?: boolean;
 };
 
 export type ObjectMDTag = {
@@ -260,6 +260,7 @@ export default class ObjectMD {
                 storageType: undefined,
                 dataStoreVersionId: undefined,
                 isNFS: undefined,
+                isReplica: undefined,
             },
             dataStoreName: '',
             originOp: '',
@@ -1168,6 +1169,26 @@ export default class ObjectMD {
         return this;
     }
 
+    /**
+     * Mark this object as the result of a replication write (replica),
+     * as opposed to a write originating from a user request.
+     *
+     * @param isReplica - true if this object was written by replication
+     * @return itself
+     */
+    setReplicationIsReplica(isReplica: boolean) {
+        this._data.replicationInfo.isReplica = isReplica;
+        return this;
+    }
+
+    /**
+     * Get whether this object was written by replication (replica).
+     * @return true if this object is a replica
+     */
+    getReplicationIsReplica(): boolean {
+        return this._data.replicationInfo.isReplica === true || this._data.replicationInfo.status === 'REPLICA';
+    }
+
     getReplicationSiteStatus(key: BackendKey): string | undefined {
         return this._findBackend(key)?.status;
     }
@@ -1348,35 +1369,28 @@ export default class ObjectMD {
     }
 
     /**
-     * Create or update the microVersionId field
+     * Update the microVersionId
      *
-     * This field can be used to force an update in MongoDB. This can
-     * be needed in the following cases:
+     * This field can be used to force an update in MongoDB when no other
+     * metadata field changes, to detect a change for CRR,
+     * and to manage conflicts during concurrent updates using conditions on this field.
      *
-     * - in case no other metadata field changes
-     *
-     * - to detect a change when fields change but object version does
-     *   not change e.g. when ingesting a putObjectTagging coming from
-     *   S3C to Zenko
-     *
-     * - to manage conflicts during concurrent updates, using
-     *   conditions on the microVersionId field.
-     *
-     * It's a field of 16 hexadecimal characters randomly generated
-     *
+     * @param instanceId - instance identifier (e.g. config.instanceId)
+     * @param replicationGroupId - replication group ID (e.g. config.replicationGroupId)
      * @return itself
      */
-    updateMicroVersionId() {
-        this._data.microVersionId = crypto.randomBytes(8).toString('hex');
+    updateMicroVersionId(instanceId: string, replicationGroupId: string) {
+        this._data.microVersionId = VersionIDUtils.generateVersionId(instanceId, replicationGroupId);
+        return this;
     }
 
     /**
-     * Get the microVersionId field, or null if not set
+     * Get the microVersionId field, or undefined if not set
      *
-     * @return the microVersionId field if exists, or {null} if it does not exist
+     * @return the microVersionId field if set, or undefined if not
      */
     getMicroVersionId() {
-        return this._data.microVersionId || null;
+        return this._data.microVersionId || undefined;
     }
 
     /**
--- a/lib/versioning/VersionID.ts
+++ b/lib/versioning/VersionID.ts
@@ -53,8 +53,7 @@ export const S3_VERSION_ID_ENCODING_TYPE = process.env.S3_VERSION_ID_ENCODING_TY
 //   - Uses old format: timestamp + sequential_position + rep_group_id (legacy 27-char format)
 // Falls back to hex encoding if S3_VERSION_ID_ENCODING_TYPE is 'hex' or unset
 export const ENABLE_FORMATTED_VERSION_ID =
-    process.env.ENABLE_FORMATTED_VERSION_ID === 'true' ||
-    process.env.ENABLE_FORMATTED_VERSION_ID === '1';
+    process.env.ENABLE_FORMATTED_VERSION_ID === 'true' || process.env.ENABLE_FORMATTED_VERSION_ID === '1';
 
 // version ID format added to the end of the version ID
 const VERSION_ID_FORMAT_VERSION = '1';
@@ -101,11 +100,7 @@ const MAX_SEQ = Math.pow(10, LENGTH_SEQ) - 1; // good for 1 billion ops
  */
 export function getInfVid(replicationGroupId: string) {
     const repGroupId = padRight(replicationGroupId, TEMPLATE_RG);
-    return (
-        padLeft(MAX_TS, TEMPLATE_TS) +
-        padLeft(MAX_SEQ, TEMPLATE_SEQ) +
-        repGroupId
-    );
+    return padLeft(MAX_TS, TEMPLATE_TS) + padLeft(MAX_SEQ, TEMPLATE_SEQ) + repGroupId;
 }
 
 // internal state of the module
@@ -232,9 +227,7 @@ export function hexDecode(str: string): string | Error {
  */
 const B62V_TOTAL = LENGTH_TS + LENGTH_SEQ;
 const B62V_HALF = B62V_TOTAL / 2;
-const B62V_EPAD = '0'.repeat(
-    Math.ceil(B62V_HALF * (Math.log(10) / Math.log(62)))
-);
+const B62V_EPAD = '0'.repeat(Math.ceil(B62V_HALF * (Math.log(10) / Math.log(62))));
 const B62V_DPAD = '0'.repeat(B62V_HALF);
 const B62V_STRING_EPAD = '0'.repeat(32 - 2 * B62V_EPAD.length);
 
@@ -355,11 +348,14 @@ export function decode(str: string): string | Error {
         const decoded: string | Error = base62Decode(str);
         // Legacy base62 version IDs (without 'info' field) are always 27 characters long.
         // The new base62 format is always 35 characters long.
-        if (typeof decoded === 'string' &&
+        if (
+            typeof decoded === 'string' &&
             decoded.length !== LEGACY_BASE62_DECODED_LENGTH &&
-            decoded.length !== BASE62_DECODED_LENGTH) {
-            return new Error(`decoded ${str} is not length ` +
-                `${LEGACY_BASE62_DECODED_LENGTH} or ${BASE62_DECODED_LENGTH}`);
+            decoded.length !== BASE62_DECODED_LENGTH
+        ) {
+            return new Error(
+                `decoded ${str} is not length ` + `${LEGACY_BASE62_DECODED_LENGTH} or ${BASE62_DECODED_LENGTH}`,
+            );
         }
         return decoded;
     }
--- a/lib/versioning/VersionID.ts
+++ b/lib/versioning/VersionID.ts
@@ -365,3 +365,33 @@ export function decode(str: string): string | Error {
     }
     return new Error(`cannot decode str ${str.length}`);
 }
+
+export enum Ordering {
+    OLDER = 'older',
+    YOUNGER = 'younger',
+    EQUAL = 'equal',
+    NOT_COMPARABLE = 'notComparable',
+}
+
+/**
+ * Compare two raw (decoded) version IDs.
+ *
+ * Returns NOT_COMPARABLE when either value is absent or shorter than
+ * LEGACY_BASE62_DECODED_LENGTH (rejects the old 16-char random-hex format).
+ * Callers are responsible for deciding how to handle NOT_COMPARABLE.
+ */
+export function compare(v1: string | null | undefined, v2: string | null | undefined): Ordering {
+    if (
+        typeof v1 !== 'string' ||
+        v1.length < LEGACY_BASE62_DECODED_LENGTH ||
+        typeof v2 !== 'string' ||
+        v2.length < LEGACY_BASE62_DECODED_LENGTH
+    ) {
+        return Ordering.NOT_COMPARABLE;
+    }
+    if (v1 === v2) {
+        return Ordering.EQUAL;
+    }
+
+    return v1 > v2 ? Ordering.OLDER : Ordering.YOUNGER;
+}
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
     "engines": {
         "node": ">=20"
     },
-    "version": "8.4.8",
+    "version": "8.4.9",
     "config": {
         "mongodbMemoryServer": {
             "version": "8.0.23"
--- a/tests/unit/models/ObjectMD.spec.js
+++ b/tests/unit/models/ObjectMD.spec.js
@@ -112,6 +112,7 @@ describe('ObjectMD class setters/getters', () => {
                 storageType: undefined,
                 dataStoreVersionId: undefined,
                 isNFS: undefined,
+                isReplica: undefined,
             },
         ],
         [
@@ -132,6 +133,7 @@ describe('ObjectMD class setters/getters', () => {
                 storageType: 'aws_s3',
                 dataStoreVersionId: '',
                 isNFS: undefined,
+                isReplica: undefined,
             },
         ],
         ['DataStoreName', null, ''],
@@ -181,6 +183,28 @@ describe('ObjectMD class setters/getters', () => {
                 assert.strictEqual(value, defaultValue);
             }
         });
+    });
+
+    it('getReplicationIsReplica: returns true when status is REPLICA', () => {
+        md.setReplicationStatus('REPLICA');
+        assert.strictEqual(
+            md.getReplicationIsReplica(),
+            true,
+            'status REPLICA alone can be used to determine isReplica',
+        );
+    });
+
+    it('getReplicationIsReplica: survives status transition to PENDING (cascade)', () => {
+        md.setReplicationInfo({
+            status: 'REPLICA',
+            isReplica: true,
+            backends: [],
+            content: [],
+            destination: '',
+            role: '',
+        });
+        md.setReplicationStatus('PENDING');
+        assert.strictEqual(md.getReplicationIsReplica(), true, 'isReplica must survive status changes');
     });
 
     it('ObjectMD::setReplicationSiteStatus', () => {
@@ -386,22 +410,22 @@ describe('ObjectMD class setters/getters', () => {
     });
 
     it('ObjectMD::microVersionId unset', () => {
-        assert.strictEqual(md.getMicroVersionId(), null);
+        assert.strictEqual(md.getMicroVersionId(), undefined);
     });
 
     it('ObjectMD::microVersionId set', () => {
-        const generatedIds = new Set();
+        const generatedIds = [];
         for (let i = 0; i < 100; ++i) {
-            md.updateMicroVersionId();
-            generatedIds.add(md.getMicroVersionId());
+            md.updateMicroVersionId('instance', 'RG001');
+            generatedIds.push(md.getMicroVersionId());
         }
         // all generated IDs should be different
-        assert.strictEqual(generatedIds.size, 100);
-        generatedIds.forEach(key => {
-            // length is always 16 in hex because leading 0s are
-            // also encoded in the 8-byte random buffer.
-            assert.strictEqual(key.length, 16);
-        });
+        assert.strictEqual(new Set(generatedIds).size, 100);
+        // microVersionIds use the versionId format (reversed time ordered):
+        // newer values sort before older ones lexicographically.
+        for (let i = 1; i < generatedIds.length; ++i) {
+            assert(generatedIds[i] < generatedIds[i - 1]);
+        }
     });
 
     it('ObjectMD::set/getRetentionMode', () => {
--- a/tests/unit/versioning/VersionID.spec.js
+++ b/tests/unit/versioning/VersionID.spec.js
@@ -1,3 +1,4 @@
+const crypto = require('crypto');
 const VID = require('../../../lib/versioning/VersionID');
 const VersioningConstants = require('../../../lib/versioning/constants').VersioningConstants;
 const assert = require('assert');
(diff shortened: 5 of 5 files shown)


lib/models/ObjectMD.ts
```diff
@@ -35,6 +35,7 @@ export type ReplicationInfo = {
     storageType: string;
     dataStoreVersionId: string;
     isNFS?: boolean;
+    isReplica?: boolean;
```
SylvainSenechal, line 51: Just to double check, before that, we were using status == 'REPLICA', right ?

lib/models/ObjectMD.ts
```diff
@@ -242,6 +243,7 @@ export default class ObjectMD {
                 storageType: '',
                 dataStoreVersionId: '',
                 isNFS: undefined,
+                isReplica: undefined,
```
SylvainSenechal, line 263: nit and not a javascript pro but can discuss undefined vs '', since i see many other fields with ''
(i have ZERO idea if its important or not, just opening discussing)

lib/models/ObjectMD.ts
```diff
@@ -1353,12 +1378,15 @@ export default class ObjectMD {
...
+     * The value is a timestamp-ordered identifier matching the
+     * versionId scheme, so two microVersionIds can be compared with
+     * {@link MicroVersionIdUtils.compare} to determine which one is
+     * more recent.
      *
      * @return itself
      */
     updateMicroVersionId() {
-        this._data.microVersionId = crypto.randomBytes(8).toString('hex');
+        this._data.microVersionId = MicroVersionIdUtils.generate();
```
SylvainSenechal: So we are gonna have the newer objects with microVersionId that has a different format from older code version. 
I think we discussed a bit the impact but just want to double check here too, is that problematic (I think we said the currently microVersionId is not used at all ?)

lib/versioning/MicroVersionId.ts
SylvainSenechal, line 1: is it possible to add for each of these function a example value of what it creates : 
an example of microversionId
an example of its encoded value
I think it will a bit to understand the structure 
Maybe something similar to the first lines of VersionID.ts file

lib/models/ObjectMD.ts
maeldonn: We do it again. We are just adding the field for now.

lib/models/ObjectMD.ts
maeldonn: Matches the isNFS pattern right above. Booleans typed as optional default to undefined so they are omitted from the serialized JSON

lib/models/ObjectMD.ts
maeldonn: No impact, this id is generated to create a new event in the mongo oplog

lib/versioning/MicroVersionId.ts
```diff
@@ -0,0 +1,98 @@
...
+ * Generate a timestamp-ordered microVersionId.
+ *
+ * The returned value is a 20-character decimal string where the
+ * timestamp and sequence are stored in reversed form (MAX - value),
+ * so that string comparison orders microVersionIds from newest to
+ * oldest, matching the versionId scheme.
+ *
+ * @return - the generated microVersionId
+ */
+export function generate(): string {
```
SylvainSenechal: I didnt expect the generation to be this complex, I guess the code in VersionID partially explains it, but it looks a bit over engineered.
Feels weird to build an id based on data from another id, also not even sure the + 1 thing because of the millisecond precision is enough since we have multiple cloudserver instance running. I guess thats why VersionID also uses instance_id

lib/versioning/MicroVersionId.ts
```diff
@@ -0,0 +1,115 @@
...
+ * @param replicationGroupId - replication group id. Normally a
+ *   configured, stable value identifying the writer's cluster (same
+ *   one passed to {@link generateVersionId}). Callers bypassing
+ *   CloudServer that have no such stable identity (e.g. the CRR
+ *   resync tool) should supply a random token to avoid colliding with
+ *   concurrent writers.
+ *
+ * @return - the generated microVersionId
+ */
+export function generate(replicationGroupId: string): string {
```
SylvainSenechal: discussed briefly in person with Maël already but, i had no idea id generation was this complicated
I think here its mostly inspired from the VersionID file to have something similar, but yeah this whole thing looks kinda smelly.
A good ID generation should be independent from basically anything,
- we shouldn't have to call wait,
- we shouldn't have to rely on a previous stored lastTimetamp
- we shouldn't have to do weird conditionnal + 1 increment on current timestamp

I mean at the end of the day, why don't we use uuidv7 ? It's globally unique, time ordered and sortable, and particularly recommended for distributed system

SylvainSenechal: Maybe we want uuids to be as small as possible, because I remember in other context we talked about how important it is to have metadata that doesn't occupy too much storage (especially important when users have many multiple small objects, we end up having metadata that is larger than the object itself).

Claude also tells me for mongodb calls ordering, it might be easier to have reversed ordered Version Id but not sure its relevant here (and not sure its true anyways)

lib/versioning/MicroVersionId.ts
maeldonn: Done

maeldonn, line 1: @SylvainSenechal We should check the pros/cons of using uuid v7

lib/versioning/MicroVersionId.ts
```diff
@@ -0,0 +1,140 @@
...
+let lastSeq = 0;
+
+/**
+ * Generate a timestamp-ordered microVersionId.
+ *
+ * The returned value is a 27-character string where the timestamp and
+ * sequence are stored in reversed form (MAX - value), followed by the
+ * replication group id normalized to {@link LENGTH_RG} characters
+ * (space-padded if shorter, truncated if longer). String comparison
+ * orders microVersionIds from newest to oldest, matching the
```
SylvainSenechal: discussed briefly in person with Maël already but, i had no idea id generation was this complicated
I think here its mostly inspired from the VersionID file to have something similar, but yeah this whole looks rather smelly.
A good ID generation should be independent from basically anything,
- we shouldn't have to call wait,
- we should have to rely on a previous stored lastTimetamp
- we should have to do weird conditionnal + 1 increment on current timestamp

I mean at the end of the day, why don't we use uuidv7 ? It's globally unique, time ordered and sortable

SylvainSenechal: Re thinking some more 🌚 
We want two things from this microVersionID : 

- Avoid infinite loop cascading : uuidv7 is good for this : when we receive the request, we compare the uuidv7 received with whats currently stored, if its the same, it means we have a loop. uuidv7 does not suffer from collision (2 ids generated in the same millisecond have 1/2^74 collision chance)
- Deal with stale object : If we receive an object that is older than the currently stored object, we don't want to store it. uuidv7 can be compared and ordered with a millisecond precision (which is largely enough, imo even ~1 second would be enough)


Also, uuidv7 storage space is 128 bits, which I think is the same order of magnitude with this 27 chars microVersionId

lib/versioning/MicroVersionId.ts
maeldonn: Checked and not possible for this use case.

lib/versioning/MicroVersionId.ts
maeldonn: Uuid v7 doesn't actually solve the same-ms case across sites, so it's not a real win over what we have. Snowflake-style (ms timestamp + per-generator counter + node bits) gives us:

- Strict monotonicity per generator (the `wait` / `lastTimestamp` / `+1` logic is the counter, not smell)
- Deterministic cross-site tiebreak via the node bits
- Same equality check for loop detection

We need to keep the versionId format. But i'm sure we can do a little refactor to make the file clearer.

lib/versioning/MicroVersionId.ts
maeldonn: A few clarifications on the "smells":

- wait is called once per process lifetime, not per id
- lastTimestamp + 1 is the per-ms counter giving strict monotonicity per generator
- with uuid v7, two writers generating in the same ms are unordered

So we are keeping the version id format. We can plan a refactoring as follow-up if you want.

lib/versioning/MicroVersionId.ts
francoisferrand, line 1: why duplicate this module : can't we directly use the versionID primitives to generate and compare microVersionIDs ?

- The whole point of using the VersionID format was to reuse a proven solution... This is rendered moot by duplication...
- Also need to avoid the confusion : `versionID` is a field -like `microVersionID`- in Object metadata; but also a _format_, and the VersionID.ts module is about the format only (can be used for multiple fields)

→ if we cannot reuse the `VersionID` _format_ for the `microVersionID` _field_, there is something fishy to investigate...

maeldonn: You're right that the duplication is a code smell. Looking more carefully, there are two ways to fix it:

Cheap path, keep `MicroVersionId.ts` but make it a thin wrapper:
```ts
const raw = generateVersionId('', rg).slice(0, LENGTH_TS + LENGTH_SEQ + LENGTH_RG);
```
The ts+seq+rg fields are always the first 27 chars of `generateVersionId`'s output regardless of env vars (`ENABLE_FORMATTED_VERSION_ID`, base62, etc.), so slicing gives a stable format. Pair with `hexEncode`/`hexDecode` directly for encode/decode. No refactor of `VersionID.ts`. Downsides: if versionId's prefix layout ever changes, microVersionId silently breaks.

Clean path, extract a `TimestampId.ts` module that owns the format primitive (ts+seq+rg, padding, hex encode/decode, state machine), and make `VersionID.ts` and `MicroVersionId.ts` thin field-specific wrappers on top of it. Matches your framing (format vs field) cleanly: `TimestampId` = format-only, `VersionID`/`MicroVersionId` = fields. I've prototyped this on `improvement/ARSN-583/timestamp-id`.

I'd lean toward the clean path since it directly resolves the layering ambiguity you flagged, but the cheap path is a valid answer too. What do you think ?

francoisferrand: why do we "cut" the end of versionID ?

the benefit of reusing versionID is that it garantees really unique, but this stands only if we keep the whole string (including instance id, to prevent multiple pods in the same cluster from generating the same id at the same time)

maeldonn: I'd missed what the instance-id segment was actually doing. Should we just use `VersionID` as-is then? I can drop `MicroVersionId.ts`, call `generateVersionId(instanceId, rg)` from `ObjectMD.updateMicroVersionId` and reuse `encode`/`decode`. What do you think ?

francoisferrand: AFAIK, reusing the whole versionID "as is" should be good & safe, so I thought that's what we would do indeed...

lib/models/ObjectMD.ts
```diff
@@ -1353,12 +1377,20 @@ export default class ObjectMD {
...
+     *   microVersionIds unique across concurrent writers in the same
+     *   replication group
+     * @param replicationGroupId - replication group id
      *
      * @return itself
      */
-    updateMicroVersionId() {
-        this._data.microVersionId = crypto.randomBytes(8).toString('hex');
+    updateMicroVersionId(instanceId: string, replicationGroupId: string) {
+        this._data.microVersionId = VersionIDUtils.generateVersionId(instanceId, replicationGroupId);
```
francoisferrand, line 1383: are the instanceID and replicaitionGroupId not usually pick'ed up (in arsenal!) by env variables?
(i.e. do we need to have params, or can we just pick the right values automatically)

lib/models/ObjectMD.ts
```diff
@@ -1151,6 +1160,26 @@ export default class ObjectMD {
...
+    setReplicationIsReplica(isReplica: boolean) {
+        this._data.replicationInfo.isReplica = isReplica;
+        return this;
+    }
+
+    /**
+     * Get whether this object was written by replication (replica).
+     * @return true if this object is a replica
+     */
+    getReplicationIsReplica() {
```
SylvainSenechal: i think we should still also have an || data.replicationInfo.status === 'REPLICA' for cross compatiblity with older code ?

lib/versioning/VersionID.ts
```diff
@@ -369,3 +369,36 @@ export function decode(str: string): string | Error {
...
+ */
+export function isMicroVersionIdComparable(id: string | null | undefined): boolean {
+    return typeof id === 'string' && id.length >= LEGACY_BASE62_DECODED_LENGTH;
+}
+
+/**
+ * Compare an incoming (raw) microVersionId against the one stored in object
+ * metadata for CRR cascade loop/stale detection.
+ */
+export function checkCrrCascadeEvent(
```
SylvainSenechal: should this function be moved to cloudserver ? 
not fully sure, also not sure yet if backbeat will use it or not

lib/models/ObjectMD.ts
SylvainSenechal: call it "isReplica" ? and the setter "setReplica" ?

lib/models/ObjectMD.ts
```diff
@@ -1151,6 +1160,26 @@ export default class ObjectMD {
...
     }
 
+    /**
+     * Mark this object as the result of a replication write (replica),
+     * as opposed to a write originating from a user request.
+     *
+     * @param isReplica - true if this object was written by replication
+     * @return itself
+     */
+    setReplicationIsReplica(isReplica: boolean) {
```
SylvainSenechal, line 1179: wonder if we should also set 
replicationInfo.status = 'REPLICA'

lib/models/ObjectMD.ts
```diff
@@ -1115,6 +1123,7 @@ export default class ObjectMD {
             storageType: storageType || '',
             dataStoreVersionId: dataStoreVersionId || '',
             isNFS,
+            isReplica: isReplica !== undefined ? isReplica : (status === 'REPLICA' ? true : undefined),
```
SylvainSenechal:
```suggestion
            isReplica: isReplica !== undefined ? isReplica : (status === 'REPLICA' ? true : false),
```

not sure if undefined or false when status not REPLICA

lib/models/ObjectMD.ts
SylvainSenechal: For instance id, we have this in cloudserver so maybe we could : 

const instanceId = process.env.CLOUDSERVER_INSTANCE_ID || config.instanceId;

But for replicationGroupId, it's not an env variable.

And this function will also be used by S3UTILS so I think we have to leave it like this

lib/models/ObjectMD.ts
```diff
@@ -1327,35 +1364,28 @@ export default class ObjectMD {
...
     /**
-     * Get the microVersionId field, or null if not set
+     * Get the microVersionId field, or undefined if not set
      *
-     * @return the microVersionId field if exists, or {null} if it does not exist
+     * @return the microVersionId field if set, or undefined if not
      */
     getMicroVersionId() {
-        return this._data.microVersionId || null;
+        return this._data.microVersionId || undefined;
```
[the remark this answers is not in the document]
SylvainSenechal: no worries this function wasn't even called before https://github.com/search?q=org%3Ascality%20getMicroVersionId&type=code

i switched to undefined to align with getEncodedVersionId, and because it fits better the client that can use string | undefined

lib/versioning/VersionID.ts
SylvainSenechal: nope, also used by backbeat

lib/models/ObjectMD.ts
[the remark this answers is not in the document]
SylvainSenechal: second time you're saying it

lib/models/ObjectMD.ts
```diff
@@ -1353,12 +1371,18 @@ export default class ObjectMD {
...
+     * more recent.
+     *
+     * @param replicationGroupId - replication group id, used to keep
+     *   microVersionIds unique across concurrent writers
      *
      * @return itself
      */
-    updateMicroVersionId() {
-        this._data.microVersionId = crypto.randomBytes(8).toString('hex');
+    updateMicroVersionId(replicationGroupId: string) {
```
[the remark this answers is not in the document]
maeldonn: updateMicroVersionId() is useless

lib/models/ObjectMD.ts
maeldonn: For now it's just a simple getter.

maeldonn: It's to keep the same names as other getters/setters

lib/models/ObjectMD.ts
maeldonn: No that's legacy, we don't want anymore to use status to track replicas

lib/models/ObjectMD.ts
maeldonn: Since it's a boolean i would say `false`. But if we can save some storage in database, we should use `undefined`.

lib/versioning/VersionID.ts
```diff
@@ -369,3 +369,40 @@ export function decode(str: string): string | Error {
...
+
+/**
+ * Returns true if a stored (raw) microVersionId is time-ordered and comparable.
+ *
+ * Returns false for:
+ *   - null / undefined (field absent)
+ *   - the legacy 16-char random-hex value written by old ObjectMD code
+ *   - any other string shorter than the minimum valid raw length (27 chars)
+ */
+export function isMicroVersionIdComparable(id: string | null | undefined): boolean {
```
maeldonn:
```suggestion
export function isComparable(id?: string | null): boolean {
```

maeldonn: Can the field really be null ? 🤔

lib/versioning/VersionID.ts
```diff
@@ -369,3 +369,40 @@ export function decode(str: string): string | Error {
...
+
+/**
+ * Compare an incoming (raw) microVersionId against the one stored in object
+ * metadata for CRR cascade loop/stale detection.
+ *
+ * Returns 'proceed' whenever either ID is non-comparable (absent, legacy
+ * 16-char hex, or otherwise too short) : a legacy incoming ID must never
+ * produce a false 'stale' or 'loop' event
+ */
+export function checkCrrCascadeEvent(
```
maeldonn:
```suggestion
export function compare(
```

maeldonn: 'loop' | 'stale' | 'proceed' are specific to CRR not to versionID format. We should just expose a simple compare function in this file

lib/models/ObjectMD.ts
```diff
@@ -1130,6 +1143,9 @@ export default class ObjectMD {
 
     setReplicationStatus(status: string) {
         this._data.replicationInfo.status = status;
+        if (status === 'REPLICA') {
```
maeldonn: I think this is useless since we are not going anymore to set status to 'REPLICA'

lib/models/ObjectMD.ts
```diff
@@ -1151,6 +1167,27 @@ export default class ObjectMD {
...
+        this._data.replicationInfo.isReplica = isReplica;
+        return this;
+    }
+
+    /**
+     * Get whether this object was written by replication (replica).
+     * @return true if this object is a replica
+     */
+    getReplicationIsReplica(): boolean {
+        return this._data.replicationInfo.isReplica === true
```
maeldonn: Not sure if we need logic in this getter. Is it used somewhere ?

lib/models/ObjectMD.ts
SylvainSenechal: I would say false too, avoid as much undefined variable as possible 🤔 What do other think ?

lib/models/ObjectMD.ts
SylvainSenechal: Shouldn't we at least keep it to avoid weird state ?
Really not sure about this, maybe we can get thoughts from somebody else.

Also are you sure we aren't gonna use setReplicationStatus ? In backbeat, in ObjectQueueEntry.js, I see some setReplicationStatus('REPLICA'), I haven't touched them for crrCascade, maybe you did for multi destination, but not sure.

lib/models/ObjectMD.ts
SylvainSenechal: I don't understand the comment 🤔 

What logic in this getter is not needed, the || this._data.replicationInfo.status === 'REPLICA'; ?

Do you mean we should only return isReplica ?

lib/versioning/VersionID.ts
SylvainSenechal: ah yeah something that would return

'equal' | 'older' | 'younger' 🤔 

It's a bit annoying though, because in the first if, it's just not comparable and we return proceed.
Feels like we would need another utility function on top of this utility function if we want something nice usable for crr cascade 🤔 Or need to return a fourth  'not comparable' value, then it can work  

Like in backbeat I would prefer to call 
if checkCrrCascadeEvent(id1, id2) === 'proceed' {do proceed logic}
than 
if compare(id1,id1) === ('older' || 'not comparable') {do proceed logic} 🤔 



I want some more THOUGHTS before making a change

lib/versioning/VersionID.ts
SylvainSenechal: yeah 
But I think the real mess here is : We are calling typescript code with javascript code. This would be easier to deal with if it was all typescript, here I did something defensive 
Image

SylvainSenechal: Re-using this comment to ask :
you want some name changes 
isMicroVersionIdComparable -> isComparable
checkCrrCascadeEvent -> compare

I say why not but its a bit annoying that the repo using this code 
has to do : 

const { versioning } = require('arsenal');
versioning.VersionID.checkCrrCascadeEvent()

Or 
const { decode, encode, checkCrrCascadeEvent } = versioning.VersionID;
checkCrrCascadeEvent()


Like I mean the namespace isnt microversionId exactly 🤔

lib/models/ObjectMD.ts
benzekrimaha: I'd also say lets keep it to avoid breaking any behavior

lib/models/ObjectMD.ts
benzekrimaha: Even if it's legacy , we can still keep it for the moment I think this avoid any breaking behavior

lib/versioning/VersionID.ts
```diff
@@ -369,3 +369,40 @@ export function decode(str: string): string | Error {
...
+    existingRaw: string | null | undefined,
+): 'loop' | 'stale' | 'proceed' {
+    if (!isMicroVersionIdComparable(incomingRaw) || !isMicroVersionIdComparable(existingRaw)) {
+        return 'proceed';
+    }
+    const existing = existingRaw as string;
+    if (incomingRaw === existing) {
+        return 'loop';
+    }
+    if (incomingRaw > existing) {
```
benzekrimaha: is comparing the full id intentional?
the ordering after timestamp/seq includes rg/instance data, so same-slot writes can be classified as stale based on ids, not time

lib/versioning/VersionID.ts
```diff
@@ -369,3 +369,40 @@ export function decode(str: string): string | Error {
...
+/**
+ * Returns true if a stored (raw) microVersionId is time-ordered and comparable.
+ *
+ * Returns false for:
+ *   - null / undefined (field absent)
+ *   - the legacy 16-char random-hex value written by old ObjectMD code
+ *   - any other string shorter than the minimum valid raw length (27 chars)
+ */
+export function isMicroVersionIdComparable(id: string | null | undefined): boolean {
+    return typeof id === 'string' && id.length >= LEGACY_BASE62_DECODED_LENGTH;
```
benzekrimaha: is length really enough to decide this is comparable?
if this can come from metadata or headers, a malformed 27+ chars string may produce a stale decision..

lib/versioning/VersionID.ts
```diff
@@ -369,3 +369,40 @@ export function decode(str: string): string | Error {
     }
     return new Error(`cannot decode str ${str.length}`);
 }
+
+/**
+ * Returns true if a stored (raw) microVersionId is time-ordered and comparable.
+ *
+ * Returns false for:
+ *   - null / undefined (field absent)
```
francoisferrand: isn't this the case for "lack" of version id, i.e. the very first version (before versioning was enabled) or the very first microVersionId (we don't allocate one when creating object) ?
→ a null or missing versionId is actually comparable...

(maybe not an issue in the bigger picture, as that case if handled separately by caller; but the name is not really consistent...)

lib/versioning/VersionID.ts
```diff
@@ -369,3 +369,40 @@ export function decode(str: string): string | Error {
...
+ * Returns 'proceed' whenever either ID is non-comparable (absent, legacy
+ * 16-char hex, or otherwise too short) : a legacy incoming ID must never
+ * produce a false 'stale' or 'loop' event
+ */
+export function checkCrrCascadeEvent(
+    incomingRaw: string,
+    existingRaw: string | null | undefined,
+): 'loop' | 'stale' | 'proceed' {
+    if (!isMicroVersionIdComparable(incomingRaw) || !isMicroVersionIdComparable(existingRaw)) {
+        return 'proceed';
```
francoisferrand: should not return _strings_, it is hard to track/refactor (hidden coupling) and dangerous...
Introduce constant instead, which are more easily manageable code entities

francoisferrand: not sure what we should do if versionID are not comparable (except for the `null` case):
- in any this is an error, should not happen
- not sure if we should replicate (solves the problem, but may lose data...) or just stop there (e.g. fail replication)
- can this really happen (in design) ?

lib/versioning/VersionID.ts
francoisferrand: is this function really needed?

it has 4 (simple) conditions, and returns 4 outcomes : so may need a switch at every place it is called, with as many lines as here...
maybe a simpler `compareMicroVersionId()` (return <0, ==0 or >0, like common `compare` functions) would be more appropriate? (i.e. remove the business logic element, stick to "math" operation on versoinID)

lib/versioning/VersionID.ts
SylvainSenechal: Yeah its fine, last part is indeed random but all we want when doing comparison is : 
- Timestamp ordered
- Stable, non random comparison when equal timestamp

lib/versioning/VersionID.ts
SylvainSenechal: I don't like this one that much. I was also thinking about checking length == 27 || 35, which correspond to the possible length of the id, but this is all just very annoying :
there are multiple formats + some environment variables in the generateVersionId function that create varying length version ID depending on their values. I think this isComparable function being dirty is just the result of VersionID not being super clean itself, but here I think it's enough, what I really wanna dodge is the old microVersionID format that was much smaller than that

lib/versioning/VersionID.ts
SylvainSenechal: I wanna have this function in arsenal, as we need to do comparison in both cloudserver and backbeat, but yeah I changed it to make it more generic

lib/versioning/VersionID.ts
SylvainSenechal: Here I was being cautious, in reality most of the time we won't have an issue.
The situation where we will have an issue is :
For a first, new cascade replication, the incomingMicroVersionID will be set, but currently in the, the microVersionID is never set (except if s3utils was used), so existing microVersionID could be unset. In that case, the caller will consider that the replciation can go, as microVersionId not set means the metadata is older than metadata with a microVersion ID

lib/versioning/VersionID.ts
SylvainSenechal: "isn't this the case for "lack" of version id, i.e. the very first version (before versioning was enabled) or the very first microVersionId (we don't allocate one when creating object) ?" => both 🤔 

Why do you say null or missing versionID is comparable ? Because you think an object with a microVersionId set is always younger than one with no microVersionID ? anyways yeah we will have some handling at the caller site

lib/models/ObjectMD.ts
```diff
@@ -1348,35 +1376,28 @@ export default class ObjectMD {
...
+     * metadata field changes, to detect a change for CRR,
+     * and to manage conflicts during concurrent updates using conditions on this field. 
+     * 
+     * @param instanceId - instance identifier (e.g. config.instanceId)
+     * @param replicationGroupId - replication group ID (e.g. config.replicationGroupId)
      * @return itself
      */
-    updateMicroVersionId() {
-        this._data.microVersionId = crypto.randomBytes(8).toString('hex');
+    updateMicroVersionId(instanceId: string, replicationGroupId: string) {
```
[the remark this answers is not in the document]
maeldonn: Yes will be updated.

lib/models/ObjectMD.ts
francoisferrand: Long time ago, some fields were created with default '' value (so they would always be present).
However this ends up taking quite a bit of space (in metadata), and mostly useless: so we changed to `undefined` in many places _where it seemed safe enough_.
Some remain either because the value `''` itself is really relevant, or before there is a high risk it would break something to change it (i.e. fixing the places where it is used would be a large endeavor)

francoisferrand (approved):

lib/models/ObjectMD.ts
```diff
@@ -1348,35 +1370,28 @@ export default class ObjectMD {
...
     /**
-     * Get the microVersionId field, or null if not set
+     * Get the microVersionId field, or undefined if not set
      *
-     * @return the microVersionId field if exists, or {null} if it does not exist
+     * @return the microVersionId field if set, or undefined if not
      */
     getMicroVersionId() {
-        return this._data.microVersionId || null;
+        return this._data.microVersionId || undefined;
```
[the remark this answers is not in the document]
maeldonn: Callers will be updated.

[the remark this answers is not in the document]
maeldonn: Callers will be updated.

benzekrimaha (approved):

PR #2629, opened 2026-05-26
Extract shared timestamp-id primitives into TimestampId module

maeldonn:
Move the constants, padding/hex helpers, and the stateful ts+seq+rg
generator out of VersionID into a new TimestampId module that both
VersionID and MicroVersionId consume. Add
a dedicated test suite covering constants, helpers, and generator
behaviour.

Issue: ARSN-583


francoisferrand:
@maeldonn not relevant anymore since we reuse VersionID format, should be closed?

PR #2631, opened 2026-06-02
Probe with HeadBucketCommand in AwsClient healthcheck

maeldonn:
HeadObjectCommand requires both Bucket and Key. Sending it with only Bucket is rejected by the v3 SDK validator, making every aws_s3-typed location (real AWS, RING-S3) report an error on every healthcheck. HeadBucketCommand only needs Bucket and is the correct probe.

Issue: ARSN-585

### DIFF
--- a/lib/storage/data/external/AwsClient.js
+++ b/lib/storage/data/external/AwsClient.js
@@ -1,6 +1,7 @@
 const { S3Client,
     PutObjectCommand,
     HeadObjectCommand,
+    HeadBucketCommand,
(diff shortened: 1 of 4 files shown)


SylvainSenechal (approved):

delthas (approved):

francoisferrand (approved):
need to bump Arsenal version in package.json (on both branches) for immediate release...

benzekrimaha (approved):

PR #2632, opened 2026-06-03
feat: add reusable OpenTelemetry tracing module

delthas:
## What

Extracts the OpenTelemetry tracing code duplicated across backbeat, cloudserver and vault into a single shared module, `lib/tracing/`, consumed via `require('arsenal/build/lib/tracing')`.

It provides:

- `init(options)` / `close()` / `isEnabled()` — SDK bootstrap (ParentBased(TraceIdRatio) sampler, span limits, traces-only, fail-fast asserts, bounded-flush shutdown race). Config assembly is split into a testable `buildSdkConfig()`.
- `loadTrustedHosts()` / `makeTrustedHostsHook()` — outbound trust boundary that strips `traceparent`/`tracestate` on calls to untrusted hosts. `OTEL_TRUSTED_HOSTS` supports NO_PROXY-style `.suffix` entries (e.g. `.svc.cluster.local` trusts every in-cluster service, subdomains and apex). We do not bother doing a trim() on each host, the value is expected to be well-formed, avoiding swallowing formatting mistakes.
- `isHealthPath(url, pathSet)` — probe/scrape span filter.
- `instrumentApiMethod(handler, name)` — per-API-method span wrapper (scope `scality.api`, span `api.<method>`, `error.type` attribute; service disambiguated by `service.name` on the resource).
- `kafka.*` — W3C trace-context propagation over node-rdkafka headers.

## Design notes

- **Consumers own their instrumentations.** `init()` takes a lazy `instrumentations: () => Instrumentation[]` thunk, invoked only when OTEL is enabled. So arsenal carries no `instrumentation-*` packages — only the OTEL SDK core, as `optionalDependencies`; `@opentelemetry/api` stays a hard (inert) dependency.
- **Deep-require, not `require('arsenal').tracing`.** The arsenal barrel eagerly loads ioredis/mongodb; reaching `init()` through it would load them before OTEL can patch them. The `lib/tracing` subpath loads nothing instrumentable at import time.
- Off by default (`ENABLE_OTEL=true` to enable); see `lib/tracing/README.md` for env vars and usage.

Consumer adoption (backbeat, cloudserver, vault) lands in separate PRs once this ships.

Issue: ARSN-586

### DIFF
--- a/lib/tracing/README.md
+++ b/lib/tracing/README.md
@@ -1,0 +1,133 @@
+# OpenTelemetry tracing
+
+Shared OTEL tracing bootstrap and helpers for the S3 platform services
+(backbeat, cloudserver, vault). Arsenal owns the parts that are hard to get
+right and identical across services — SDK lifecycle, the outbound trust
+boundary, span conventions — while each consumer supplies its own
+instrumentation packages.
+
+## Import
+
+Deep-require the built module; do **not** use `require('arsenal').tracing`:
+
+```js
+const tracing = require('arsenal/build/lib/tracing');
+```
+
+The arsenal barrel (`require('arsenal')`) eagerly loads `ioredis` (via
+`lib/metrics`) and `mongodb` (via `lib/storage`); reaching `init()` through it
+would load those before OpenTelemetry can patch them, silently disabling the
+instrumentation. The deep-require pulls only the tracing module, which loads
+nothing instrumentable at import time.
+
+## Enabling
+
+Tracing is **off** unless `ENABLE_OTEL=true`. When enabled, `init()` requires
+`OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` and fails fast if it is unset.
+
+- `ENABLE_OTEL` — master switch (`isEnabled()`); must be exactly `true`.
+  Default: off.
+- `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` — OTLP/HTTP traces collector URL;
+  required when enabled.
+- `OTEL_SAMPLING_RATIO` — root sampling ratio in `[0, 1]`. Default `0.01`.
+- `OTEL_SERVICE_NAME` — overrides the `serviceName` option.
+- `OTEL_SERVICE_VERSION` — overrides the `serviceVersion` option.
+- `OTEL_SERVICE_NAMESPACE` — resource `service.namespace`. Default `scality`.
+- `OTEL_TRUSTED_HOSTS` — comma-separated hostnames trusted for outbound trace
+  propagation; a `.suffix` entry matches any subdomain. Default: loopback only.
+
+## Lifecycle
+
+```js
+const tracing = require('arsenal/build/lib/tracing');
+
+tracing.init({
+    serviceName: 'cloudserver',
+    serviceVersion: require('./package.json').version,
+    // Lazy thunk — invoked once inside init(), only when OTEL is enabled, so
+    // the instrumentation packages (and their patch hooks) never load when
+    // OTEL is off. The consumer owns these packages and their options.
+    instrumentations: () => {
+        const { HttpInstrumentation } = require('@opentelemetry/instrumentation-http');
+        const { IORedisInstrumentation } = require('@opentelemetry/instrumentation-ioredis');
+        const { MongoDBInstrumentation } = require('@opentelemetry/instrumentation-mongodb');
+        return [
+            // makeHttpInstrumentationConfig wires the trust boundary + health
+            // filter. Outbound-only services (backbeat) spread it and add
+            // disableIncomingRequestInstrumentation: true.
+            new HttpInstrumentation(
+                tracing.makeHttpInstrumentationConfig({
+                    healthPaths: ['/live', '/_/healthcheck', '/metrics'],
+                }),
+            ),
+            new IORedisInstrumentation({ requireParentSpan: true }),
+            new MongoDBInstrumentation({ enhancedDatabaseReporting: false }),
+        ];
+    },
+});
+
+// On shutdown — bounded flush (~5s), safe to call concurrently / when never inited:
+await tracing.close();
+```
+
+`init()` is idempotent and a no-op when disabled. Build the instrumentation
+array entirely inside the thunk so packages resolve from the consumer's
+`node_modules` and load lazily.
+
+## API
+
+- `init(options)` — boot the SDK. Options: `serviceName` (required),
+  `serviceVersion?`, `instrumentations?: () => Instrumentation[]`.
+- `close()` — bounded-flush shutdown; idempotent.
+- `isEnabled()` — `process.env.ENABLE_OTEL === 'true'`.
+- `makeHttpInstrumentationConfig({ healthPaths? })` → params for the consumer's
+  `HttpInstrumentation`: the outbound trust-boundary `requestHook` plus an
+  `ignoreIncomingRequestHook` that drops OPTIONS and the given health/probe
+  paths (none by default). arsenal never disables inbound spans — an
+  outbound-only service spreads the result and adds
+  `disableIncomingRequestInstrumentation: true`.
+- `instrumentApiMethod(handler, methodName)` — wrap a callback/async/sync
+  handler in an `api.<methodName>` span (scope `arsenal.api`, `err.code` →
+  `error.type`); returns the handler unchanged when OTEL is off.
+- `kafka.*` — trace-context propagation over node-rdkafka headers (scope
+  `arsenal.kafka`): `traceHeadersFromEntry`, `traceHeadersFromCurrentContext`,
+  `contextFromKafkaHeaders`, `startLinkedSpanFromKafkaEntry`.
+
+## Examples
+
+Instrument API handlers (cloudserver / vault) — wrap each handler so calls
+produce `api.<name>` spans:
+
+```js
+const { instrumentApiMethod } = require('arsenal/build/lib/tracing');
+for (const [name, handler] of Object.entries(api)) {
+    api[name] = instrumentApiMethod(handler, name);
+}
+```
+
+Kafka propagation (backbeat) — producer stamps headers from the active span or
+an oplog entry; consumer starts a new trace linked to the upstream span:
+
+```js
+const { kafka } = require('arsenal/build/lib/tracing');
+
+// producer side
+producer.send(payload, kafka.traceHeadersFromCurrentContext());
+producer.send(payload, kafka.traceHeadersFromEntry(objectMd));
+
+// consumer side
+const { ctx, span } = kafka.startLinkedSpanFromKafkaEntry(entry, 'replicate');
+try {
+    /* ...work within ctx... */
+} finally {
+    span.end();
+}
+```
+
+## Dependencies
+
+`@opentelemetry/api` is a hard dependency (inert until an SDK is registered).
+The SDK-core packages (`sdk-node`, `sdk-trace-base`, `resources`,
+`exporter-trace-otlp-http`) are **optional** dependencies, required lazily in
+`init()`. The `instrumentation-*` packages are **not** arsenal dependencies —
+consumers bring their own and pass them via the `instrumentations` thunk.
--- a/lib/tracing/bootstrap.ts
+++ b/lib/tracing/bootstrap.ts
@@ -1,0 +1,125 @@
+import assert from 'assert';
+
+import { DEFAULT_SAMPLING_RATIO, SPAN_LIMITS, SHUTDOWN_DEADLINE_MS } from './constants';
+
+export interface InitOptions {
+    // service.name (OTEL_SERVICE_NAME env overrides).
+    serviceName: string;
+    // service.version (OTEL_SERVICE_VERSION env overrides).
+    serviceVersion?: string;
+    // Lazy invoked once inside init() only when OTEL is enabled, so the
+    // consumer's require()s and instrumentation patch hooks never load when off.
+    instrumentations?: () => any[];
+}
+
+let sdk: any = null;
+
+let diagLog: any = null;
+function getDiagLog(): any {
+    if (!diagLog) {
+        const werelogs = require('werelogs');
+        diagLog = new werelogs.Logger('tracing');
+    }
+    return diagLog;
+}
+
+function makeDiagLogger(): any {
+    const log = getDiagLog();
+    const fwd =
+        (level: string) =>
+        (m: string, ...a: unknown[]) =>
+            log[level](m, a.length ? { args: a } : undefined);
+    return {
+        error: fwd('error'),
+        warn: fwd('warn'),
+        info: fwd('info'),
+        debug: fwd('debug'),
+        verbose: fwd('trace'),
+    };
+}
+
+export function isEnabled(): boolean {
+    return process.env.ENABLE_OTEL === 'true';
+}
+
+export function init(options: InitOptions): void {
+    if (!isEnabled() || sdk) {
+        return;
+    }
+
+    const endpoint = process.env.OTEL_EXPORTER_OTLP_TRACES_ENDPOINT;
+    assert(endpoint, 'ENABLE_OTEL=true but OTEL_EXPORTER_OTLP_TRACES_ENDPOINT is unset');
+    assert(options.serviceName, 'tracing.init: options.serviceName is required');
+
+    let samplingRatio = DEFAULT_SAMPLING_RATIO;
+    if (process.env.OTEL_SAMPLING_RATIO !== undefined) {
+        const parsed = Number(process.env.OTEL_SAMPLING_RATIO);
+        assert(
+            Number.isFinite(parsed) && parsed >= 0 && parsed <= 1,
+            `OTEL_SAMPLING_RATIO must be a finite number in [0, 1], got: ${process.env.OTEL_SAMPLING_RATIO}`,
+        );
+        samplingRatio = parsed;
+    }
+
+    // diag + NodeSDK are required lazily here (never at module top) so OTEL-off
+    // processes load nothing beyond @opentelemetry/api.
+    const { diag, DiagLogLevel } = require('@opentelemetry/api');
+    const { NodeSDK } = require('@opentelemetry/sdk-node');
+    diag.setLogger(makeDiagLogger(), DiagLogLevel.WARN);
+
+    sdk = new NodeSDK(_buildSdkConfig(options, endpoint, samplingRatio));
+    sdk.start();
+}
+
+export function _buildSdkConfig(options: InitOptions, endpoint: string, samplingRatio: number): any {
+    const { resourceFromAttributes } = require('@opentelemetry/resources');
+    const { OTLPTraceExporter } = require('@opentelemetry/exporter-trace-otlp-http');
+    const { ParentBasedSampler, TraceIdRatioBasedSampler } = require('@opentelemetry/sdk-trace-base');
+
+    return {
+        resource: resourceFromAttributes({
+            'service.name': process.env.OTEL_SERVICE_NAME || options.serviceName,
+            'service.version': process.env.OTEL_SERVICE_VERSION || options.serviceVersion,
+            // Operators set OTEL_SERVICE_NAME per-pod but not the namespace,
+            // so the 'scality' default marks Scality-owned traces.
+            'service.namespace': process.env.OTEL_SERVICE_NAMESPACE || 'scality',
+        }),
+        traceExporter: new OTLPTraceExporter({ url: endpoint }),
+        // Disable OTLP log + metric exporters with empties
+        logRecordProcessors: [],
+        metricReaders: [],
+        spanLimits: SPAN_LIMITS,
+        // ParentBased so a service honors the upstream sampled flag; without
+        // it, it re-samples at `ratio` and the pipeline rate collapses to
+        // ratio × ratio.
+        sampler: new ParentBasedSampler({
+            root: new TraceIdRatioBasedSampler(samplingRatio),
+        }),
+        instrumentations: options.instrumentations ? options.instrumentations() : [],
+    };
+}
+
+export async function close(): Promise<void> {
+    if (!sdk) {
+        return;
+    }
+    // Capture the sdk via the IIFE param and null the module ref synchronously
+    // (before any await) so concurrent callers don't both shutdown() — the SDK
+    // isn't idempotent. The IIFE owns the rejection so a late failure after the
+    // timeout wins the race can't crash the process.
+    const shutdown = (async (running: any) => {
+        try {
+            await running.shutdown();
+        } catch (err) {
+            getDiagLog().error('tracing close failed', { err });
+        }
+    })(sdk);
+    sdk = null;
+    await Promise.race([
+        shutdown,
+        // .unref() so the timer doesn't keep the event loop alive.
+        new Promise<void>(resolve => {
+            setTimeout(resolve, SHUTDOWN_DEADLINE_MS).unref();
+        }),
+    ]);
+}
--- a/lib/tracing/httpHooks.ts
+++ b/lib/tracing/httpHooks.ts
@@ -1,0 +1,92 @@
+import type { ClientRequest } from 'http';
+import type { Span } from '@opentelemetry/api';
+
+// makeHttpInstrumentationConfig builds the params a consumer passes to its
+// HttpInstrumentation: the outbound trust boundary and the inbound health-path
+// span filter.
+
+const LOOPBACK_HOSTS = ['localhost', '127.0.0.1', '::1'];
+
+const HOST_BRACKET_RE = /^\[([^\]]+)\](?::\d+)?$/;
+const HOST_PORT_RE = /:\d+$/;
+const TRAILING_DOT_RE = /\.$/;
+
+// Operator-supplied bare hostnames (+ always-trusted loopback). A '.'-prefixed
+// entry is a NO_PROXY-style suffix (see makeTrustedHostsHook). Unset → only
+// loopback, so every other outbound call gets traceparent stripped.
+export function loadTrustedHosts(): Set<string> {
+    const hosts = new Set(LOOPBACK_HOSTS);
+    const raw = process.env.OTEL_TRUSTED_HOSTS;
+    if (typeof raw === 'string' && raw.length > 0) {
+        raw.split(',').forEach(entry => {
+            if (entry) {
+                hosts.add(entry.toLowerCase());
+            }
+        });
+    }
+    return hosts;
+}
+
+// Bare hostname for matching. IPv6 is bracket-handled separately so the :port
+// strip can't eat into the address (naive strip of `[::1]` → `:`). A trailing
+// dot (FQDN form) is stripped so it matches the same patterns.
+function normalizeHostHeader(s: string): string {
+    const bracket = HOST_BRACKET_RE.exec(s);
+    if (bracket) {
+        return bracket[1].toLowerCase();
+    }
+    return s.replace(HOST_PORT_RE, '').replace(TRAILING_DOT_RE, '').toLowerCase();
+}
+
+// HTTP requestHook that strips traceparent/tracestate (and tags the span
+// suppressed) on outbound calls to untrusted hosts, so trace IDs don't leak off-platform.
+export function makeTrustedHostsHook(trustedHosts: Set<string>) {
+    // '.'-prefixed entries are suffix patterns, partitioned once (not per request).
+    const suffixes = [...trustedHosts].filter(h => h.startsWith('.'));
+    const isTrusted = (host: string): boolean =>
+        trustedHosts.has(host) ||
+        // endsWith keeps the dot boundary (`xsvc.cluster.local` not matched);
+        // the slice(1) check also trusts the bare apex.
+        suffixes.some(sfx => host.endsWith(sfx) || host === sfx.slice(1));
+
+    return function requestHook(span: Span, request: ClientRequest): void {
+        // Only ClientRequest exposes getHeader/removeHeader, not inbound IncomingMessage.
+        if (!request || typeof request.getHeader !== 'function') {
+            return;
+        }
+        const host = normalizeHostHeader((request.getHeader('host') || '').toString());
+        if (isTrusted(host)) {
+            return;
+        }
+        if (typeof request.removeHeader === 'function') {
+            request.removeHeader('traceparent');
+            request.removeHeader('tracestate');
+        }
+        if (span && typeof span.setAttribute === 'function') {
+            span.setAttribute('scality.trace.suppressed', true);
+        }
+    };
+}
+
+// True when url's path (query string stripped) exactly matches an entry in
+// pathSet — used to skip spans for health/probe endpoints. The set differs per
+// service, so the caller passes it in.
+export function isHealthPath(url: string | undefined, pathSet: Set<string>): boolean {
+    if (typeof url !== 'string' || url.length === 0) {
+        return false;
+    }
+    const qIdx = url.indexOf('?');
+    const path = qIdx === -1 ? url : url.slice(0, qIdx);
+    return pathSet.has(path);
+}
+
+// HttpInstrumentation params (the consumer constructs it): trust-boundary
+// requestHook + an ignore hook for OPTIONS and the given health paths.
+export function makeHttpInstrumentationConfig(options: { healthPaths?: string[] } = {}): Record<string, any> {
+    const healthPaths = new Set(options.healthPaths ?? []);
+    return {
+        ignoreIncomingRequestHook: (req: { method?: string; url?: string }) =>
+            req.method === 'OPTIONS' || isHealthPath(req.url, healthPaths),
+        requestHook: makeTrustedHostsHook(loadTrustedHosts()),
+    };
+}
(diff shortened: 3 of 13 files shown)


index.ts
```diff
@@ -35,6 +35,7 @@ export * as constants from './lib/constants';
 export * as https from './lib/https';
 export * as metrics from './lib/metrics';
 export * as network from './lib/network';
+export * as tracing from './lib/tracing';
```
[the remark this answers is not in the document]
delthas: Removed — the `export * as tracing` barrel export is gone as of the latest push, so the deep-require (`require('arsenal/build/lib/tracing')`) is now the only entry point and the footgun no longer exists. Thanks!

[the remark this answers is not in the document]
delthas: Removed — the `export * as tracing` barrel export is gone as of the latest push, so the deep-require (`require('arsenal/build/lib/tracing')`) is now the only entry point and the footgun no longer exists. Thanks!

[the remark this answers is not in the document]
delthas: Removed — the `export * as tracing` barrel export is gone as of the latest push, so the deep-require (`require('arsenal/build/lib/tracing')`) is now the only entry point and the footgun no longer exists. Thanks!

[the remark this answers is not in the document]
delthas: Removed — the `export * as tracing` barrel export is gone as of the latest push, so the deep-require (`require('arsenal/build/lib/tracing')`) is now the only entry point and the footgun no longer exists. Thanks!

lib/tracing/kafkaTraceContext.ts
```diff
@@ -0,0 +1,75 @@
...
+    return propagation.extract(ROOT_CONTEXT, carrier);
+}
+
+// Start a new root span that LINKS to (not a child of) the upstream span in the
+// Kafka entry. Kafka hops can fire hours after the original request, so links —
+// not parent/child — keep traces small. Caller must span.end(). New trace each read.
+export function startLinkedSpanFromKafkaEntry(
+    kafkaEntry: { headers?: KafkaHeader[] }, operationName: string,
+): { ctx: Context; span: Span } {
+    const tracer = trace.getTracer('backbeat');
```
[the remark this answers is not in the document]
delthas: Fixed — the hardcoded `'backbeat'` scope is now `'scality.kafka'`; per-service attribution comes from `service.name` on the resource, not the tracer scope. Thanks!

lib/tracing/httpHooks.ts
```diff
@@ -0,0 +1,61 @@
...
+
+// Operator-supplied bare hostnames (+ always-trusted loopback). A '.'-prefixed
+// entry is a NO_PROXY-style suffix (see makeTrustedHostsHook). Unset → only
+// loopback, so every other outbound call gets traceparent stripped.
+export function loadTrustedHosts(): Set<string> {
+    const hosts = new Set(LOOPBACK_HOSTS);
+    const raw = process.env.OTEL_TRUSTED_HOSTS;
+    if (typeof raw === 'string' && raw.length > 0) {
+        raw.split(',').forEach(entry => {
+            if (entry) {
```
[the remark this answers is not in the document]
delthas: Intentional — `OTEL_TRUSTED_HOSTS` is operator-set with bare entries (no spaces); we deliberately don't normalize the values, consistent with the existing 'operator supplies bare hosts' behavior. Leaving as-is.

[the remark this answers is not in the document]
delthas: Intentional — `OTEL_TRUSTED_HOSTS` is operator-set with bare entries (no spaces); we deliberately don't normalize the values, consistent with the existing 'operator supplies bare hosts' behavior. Leaving as-is.

lib/tracing/httpHooks.ts
```diff
@@ -0,0 +1,61 @@
...
+                hosts.add(entry.toLowerCase());
+            }
+        });
```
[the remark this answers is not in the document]
delthas: Intentional — `OTEL_TRUSTED_HOSTS` is operator-set with bare entries (no spaces); we deliberately don't normalize the values, consistent with the existing 'operator supplies bare hosts' behavior. Leaving as-is.

lib/tracing/httpHooks.ts
```diff
@@ -0,0 +1,76 @@
...
```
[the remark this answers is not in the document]
delthas: Intentional — `OTEL_TRUSTED_HOSTS` is operator-set with bare entries (no spaces); we deliberately don't normalize the values, consistent with the existing 'operator supplies bare hosts' behavior. Leaving as-is.

lib/tracing/instrumentation.ts
```diff
@@ -0,0 +1,106 @@
...
+        return originalCallback.call(this, err, ...results);
+    };
+
+    const ctx = trace.setSpan(context.active(), span);
+    try {
+        return context.with(ctx, () => apiMethod.apply(self, wrappedArgs));
+    } catch (err) {
+        endSpan(span, err);
+        throw err;
+    }
```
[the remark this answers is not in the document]
delthas: A handler that invokes its callback and then throws synchronously is itself buggy, and OTEL ignores the second `end()`. Keeping the original instrumentation's verbatim behavior rather than adding a guard — same handler-contract class we're treating as out of scope. Thanks for the catch.

lib/tracing/httpHooks.ts
```diff
@@ -0,0 +1,76 @@
...
+// Operator-supplied bare hostnames (+ always-trusted loopback). A '.'-prefixed
+// entry is a NO_PROXY-style suffix (see makeTrustedHostsHook). Unset → only
+// loopback, so every other outbound call gets traceparent stripped.
+export function loadTrustedHosts(): Set<string> {
+    const hosts = new Set(LOOPBACK_HOSTS);
+    const raw = process.env.OTEL_TRUSTED_HOSTS;
+    if (typeof raw === 'string' && raw.length > 0) {
+        raw.split(',').forEach(entry => {
+            if (entry) {
+                hosts.add(entry.toLowerCase());
```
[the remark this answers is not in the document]
delthas: Intentional — `OTEL_TRUSTED_HOSTS` is operator-set with bare entries (no spaces); we deliberately don't normalize the values, consistent with the existing 'operator supplies bare hosts' behavior. Leaving as-is.

lib/tracing/instrumentation.ts
```diff
@@ -0,0 +1,110 @@
...
+function getTracer(): Tracer {
+    if (tracer) {
+        return tracer;
+    }
+    const { version } = require('../../package.json');
+    tracer = trace.getTracer(TRACER_NAME, version);
+    return tracer;
+}
+
+// Test-only: drop the cached tracer so a freshly-installed provider is picked up.
```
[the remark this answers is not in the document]
delthas: Verified the compiled build actually resolves this — `build/package.json` is present in the `tsc` output, so the built `getTracer()` returns the version without a MODULE_NOT_FOUND. That said, relying on a relative require across the source/build boundary is fragile, so I've removed the optional tracer scope-version read entirely: `getTracer()` no longer reads `package.json`. Thanks for the catch!

delthas:
@francoisferrand @DarkIsDude intentionally tagging you here as this is moving [cloudserver code you already reviewed](https://github.com/scality/cloudserver/pull/6140) into this PR.

lib/tracing/httpHooks.ts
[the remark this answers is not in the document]
delthas: Intentional — `OTEL_TRUSTED_HOSTS` is operator-set with bare entries (no spaces); we deliberately don't normalize the values, consistent with the existing 'operator supplies bare hosts' behavior. Leaving as-is. (Re-flagged on each push; same decision as the earlier resolved threads.)

lib/tracing/bootstrap.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// Bound the flush so an unreachable collector can't block past Kubernetes'
+// 30s grace period (BatchSpanProcessor's own export timeout is also 30s).
+const SHUTDOWN_DEADLINE_MS = 5000;
+
+export async function close(): Promise<void> {
+    // Null sdk before awaiting so concurrent callers don't both shutdown()
+    // (the SDK isn't idempotent on shutdown).
+    const local = sdk;
```
francoisferrand: "local" is not very descriptive : we know it's "local" -which is both useless (the declaration is here) and false (it is is used in shutdown, where it is definitely not local)- but now what it is...

to avoid sync, a trick may be to pass a parameter to shutdown (`shutdown = (async sdk => {})(sdk)`) or to use `once()` ?

lib/tracing/httpHooks.ts
```diff
@@ -0,0 +1,76 @@
...
+        });
+    }
+    return hosts;
+}
+
+// Bare hostname for matching. IPv6 is bracket-handled separately so the :port
+// strip can't eat into the address (naive strip of `[::1]` → `:`). A trailing
+// dot (FQDN form) is stripped so it matches the same patterns.
+function normalizeHostHeader(s: string): string {
+    const bracket = /^\[([^\]]+)\](?::\d+)?$/.exec(s);
```
francoisferrand: should we not compile the regex just once?

lib/tracing/instrumentation.ts
```diff
@@ -0,0 +1,109 @@
+import { trace, context, SpanKind, SpanStatusCode, Tracer, Span } from '@opentelemetry/api';
+import { isEnabled } from './bootstrap';
+
+// Shared by every service — service identity comes from service.name on the
+// resource, not re-encoded here.
+const TRACER_NAME = 'scality.api';
```
francoisferrand: should it convey "how" this was collected: 'arsenal.api' ?

lib/tracing/kafkaTraceContext.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// Start a new root span that LINKS to (not a child of) the upstream span in the
+// Kafka entry. Kafka hops can fire hours after the original request, so links —
+// not parent/child — keep traces small. Caller must span.end(). New trace each read.
+export function startLinkedSpanFromKafkaEntry(
+    kafkaEntry: { headers?: KafkaHeader[] },
+    operationName: string,
+): { ctx: Context; span: Span } {
+    const tracer = trace.getTracer('scality.kafka');
```
francoisferrand: - should have a constant for the trace name
- same as above: scality.kafka or arsenal.kafka ?

lib/tracing/README.md
```diff
@@ -0,0 +1,101 @@
...
+        const { IORedisInstrumentation } = require('@opentelemetry/instrumentation-ioredis');
+        const { MongoDBInstrumentation } = require('@opentelemetry/instrumentation-mongodb');
+        const healthPaths = new Set(['/live', '/_/healthcheck', '/metrics']);
+        return [
+            new HttpInstrumentation({
+                ignoreIncomingRequestHook: req =>
+                    req.method === 'OPTIONS' || tracing.isHealthPath(req.url, healthPaths),
+                // Trust boundary: strips traceparent/tracestate on outbound
+                // calls to hosts outside OTEL_TRUSTED_HOSTS.
+                requestHook: tracing.makeTrustedHostsHook(tracing.loadTrustedHosts()),
```
francoisferrand: `tracing.loadTrustedHosts()` should be called just once and stored in a `const`

lib/tracing/README.md
```diff
@@ -0,0 +1,101 @@
...
+- `makeTrustedHostsHook(trustedHosts)` → an HTTP `requestHook` that strips
+  trace headers on outbound calls to untrusted hosts.
+- `isHealthPath(url, pathSet)` — path-in-set test (query stripped) for an
+  `ignoreIncomingRequestHook`.
+- `instrumentApiMethod(handler, methodName)` — wrap a callback/async/sync
+  handler in an `api.<methodName>` span (scope `scality.api`, `err.code` →
+  `error.type`); returns the handler unchanged when OTEL is off.
+- `kafka.*` — trace-context propagation over node-rdkafka headers:
+  `traceHeadersFromEntry`, `traceHeadersFromCurrentContext`,
+  `contextFromKafkaHeaders`, `startLinkedSpanFromKafkaEntry`.
```
francoisferrand, line 94: would help to have examples how to use these...
(esp. kafka stuff ; but instrumentApiMethod would be nice as well, if only to clarify that this is needed for http instrumentation to work)

lib/tracing/README.md
```diff
@@ -0,0 +1,101 @@
...
+        const { MongoDBInstrumentation } = require('@opentelemetry/instrumentation-mongodb');
+        const healthPaths = new Set(['/live', '/_/healthcheck', '/metrics']);
+        return [
+            new HttpInstrumentation({
+                ignoreIncomingRequestHook: req =>
+                    req.method === 'OPTIONS' || tracing.isHealthPath(req.url, healthPaths),
+                // Trust boundary: strips traceparent/tracestate on outbound
+                // calls to hosts outside OTEL_TRUSTED_HOSTS.
+                requestHook: tracing.makeTrustedHostsHook(tracing.loadTrustedHosts()),
+            }),
```
francoisferrand: instead of exporting 4 functions, and making the same call in each module: should we create a similar function in arsenal ?

```js
function newHttpInstrumention(healthPath...) {
	const trustedHosts = tracing.loadTrustedHosts();
    const healthPaths = new Set(healthPath...);
	return new HttpInstrumentation{
	      ignoreIncomingRequestHook: req =>
	          req.method === 'OPTIONS' || tracing.isHealthPath(req.url, healthPaths),
	      // Trust boundary: strips traceparent/tracestate on outbound
	      // calls to hosts outside OTEL_TRUSTED_HOSTS.
	      requestHook: tracing.makeTrustedHostsHook(trustedHosts),
	}
}
```

or maybe more flexible to have a function returning just the http instrumentation params : so the params may be further customized if needed before creating `HttpInstrumentation` object

package.json
```diff
@@ -57,13 +57,19 @@
         "xml2js": "^0.6.2"
```
francoisferrand, line 57: missing `@opentelemetry/api` in the "regular" deps

francoisferrand (commented):
also missing some tests to cover corner cases

lib/tracing/README.md
delthas: I was wondering about this. But for example `backbeat` intentionally drops incoming requests since it's only health checks, so then we'd have to have a parameter for that or something.

Returning the instrumentation params would be a flexible way to add this, I'll look into it.

lib/tracing/bootstrap.ts
delthas: Adopted the param-passing style: close() now captures the sdk via the IIFE param (`running`) and nulls the module ref synchronously before any await — no more `local`. (The synchronous null-before-await is still the concurrency guard against double-shutdown.)

lib/tracing/httpHooks.ts
delthas: Done — hoisted the host-matching regexes to module constants (`HOST_BRACKET_RE`/`HOST_PORT_RE`/`TRAILING_DOT_RE`), compiled once.

lib/tracing/instrumentation.ts
delthas: Renamed the scope to `arsenal.api` (conveys it's arsenal's instrumentation; service identity stays on the resource via service.name).

lib/tracing/kafkaTraceContext.ts
delthas: Done — extracted a `TRACER_NAME` constant and renamed it to `arsenal.kafka` (matching `arsenal.api`).

lib/tracing/README.md
delthas: Subsumed by the new `makeHttpInstrumentationConfig` helper, which calls `loadTrustedHosts()` once internally.

lib/tracing/README.md
delthas: Added an Examples section to the README with instrumentApiMethod (handler wrapping) and kafka (producer/consumer propagation) usage.

lib/tracing/README.md
delthas: Done — added `makeHttpInstrumentationConfig({ healthPaths })`: it returns the HttpInstrumentation *params* (trust-boundary requestHook + the OPTIONS/health ignore hook), so the consumer constructs the HttpInstrumentation and arsenal still needs no instrumentation-http dep. Consumers now make a single call; the 3 building-block functions are off the public barrel. (Spread the result to customize further.)

package.json
delthas: `@opentelemetry/api` is already a regular `dependencies` entry (`^1.9.0`); it predates this PR so it's not in the diff. Only the heavy SDK-core packages are optional.

lib/tracing/kafkaTraceContext.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// Serialize the active OTEL context into node-rdkafka traceparent/tracestate headers.
+export function traceHeadersFromCurrentContext(): KafkaHeader[] | undefined {
+    const span = trace.getSpan(context.active());
+    if (!span) {
+        return undefined;
+    }
+    const spanCtx = span.spanContext();
+    if (!spanCtx || !spanCtx.traceId || !spanCtx.spanId) {
```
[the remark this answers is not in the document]
delthas: Fixed — `traceHeadersFromCurrentContext` now guards with `!spanCtx || !trace.isSpanContextValid(spanCtx)`, matching `startLinkedSpanFromKafkaEntry`. Added a test for the all-zero invalid context.

lib/tracing/httpHooks.ts
```diff
@@ -0,0 +1,98 @@
...
+export function loadTrustedHosts(): Set<string> {
+    const hosts = new Set(LOOPBACK_HOSTS);
+    const raw = process.env.OTEL_TRUSTED_HOSTS;
+    if (typeof raw === 'string' && raw.length > 0) {
+        raw.split(',').forEach(entry => {
+            if (entry) {
+                hosts.add(entry.toLowerCase());
+            }
+        });
+    }
```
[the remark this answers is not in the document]
delthas: Intentional — `OTEL_TRUSTED_HOSTS` is operator-set with bare, space-free entries; arsenal doesn't normalize the values (consistent with the existing 'operator supplies bare hosts' behavior). The space-before-dot case fails safe (over-strips traceparent, no leak). Leaving as-is.

lib/tracing/httpHooks.ts
```diff
@@ -0,0 +1,99 @@
...
+// Operator-supplied bare hostnames (+ always-trusted loopback). A '.'-prefixed
+// entry is a NO_PROXY-style suffix (see makeTrustedHostsHook). Unset → only
+// loopback, so every other outbound call gets traceparent stripped.
+export function loadTrustedHosts(): Set<string> {
+    const hosts = new Set(LOOPBACK_HOSTS);
+    const raw = process.env.OTEL_TRUSTED_HOSTS;
+    if (typeof raw === 'string' && raw.length > 0) {
+        raw.split(',').forEach(entry => {
+            if (entry) {
+                hosts.add(entry.toLowerCase());
```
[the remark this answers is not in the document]
delthas: Same as the other trust-hosts thread: declining intentionally — entries are operator-set and bare, and arsenal does no value normalization. Leaving as-is.

lib/tracing/instrumentation.ts
```diff
@@ -0,0 +1,109 @@
...
+        return context.with(ctx, () => apiMethod.apply(self, wrappedArgs));
+    } catch (err) {
+        endSpan(span, err);
+        throw err;
+    }
+}
+
+function instrumentAsyncHandler(self: unknown, apiMethod: ApiMethod, spanName: string, args: any[]): any {
+    const span = getTracer().startSpan(spanName, { kind: SpanKind.INTERNAL });
+    const ctx = trace.setSpan(context.active(), span);
```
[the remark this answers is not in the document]
delthas: Degenerate case: a callback that throws after being invoked is itself buggy, and OTEL ignores the second `end()`. Matches the original instrumentation's behavior; declining a guard (same handler-contract class as the earlier resolved thread).

lib/tracing/bootstrap.ts
```diff
@@ -0,0 +1,113 @@
...
+export function init(options: InitOptions): void {
+    if (!isEnabled() || sdk) {
+        return;
+    }
+
+    const endpoint = process.env.OTEL_EXPORTER_OTLP_TRACES_ENDPOINT;
+    assert(endpoint, 'ENABLE_OTEL=true but OTEL_EXPORTER_OTLP_TRACES_ENDPOINT is unset');
+    assert(options.serviceName, 'tracing.init: options.serviceName is required');
+
+    let samplingRatio = 0.01;
```
francoisferrand: should this be a constant somewhere *default sample ratio" ?

lib/tracing/bootstrap.ts
```diff
@@ -0,0 +1,113 @@
...
+        traceExporter: new OTLPTraceExporter({ url: endpoint }),
+        // Explicit empties: NodeSDK silently starts OTLP log + metric
+        // exporters otherwise. We only emit traces.
+        logRecordProcessors: [],
+        metricReaders: [],
+        spanLimits: {
+            attributeValueLengthLimit: 4096,
+            attributeCountLimit: 128,
+            eventCountLimit: 128,
+            linkCountLimit: 128,
```
francoisferrand: should be constants as well?
do we need to make them configurable?

lib/tracing/bootstrap.ts
```diff
@@ -0,0 +1,113 @@
...
+        sampler: new ParentBasedSampler({
+            root: new TraceIdRatioBasedSampler(samplingRatio),
+        }),
+        instrumentations: options.instrumentations ? options.instrumentations() : [],
+    };
+}
+
+// Bound the flush so an unreachable collector can't block past Kubernetes'
+// 30s grace period (BatchSpanProcessor's own export timeout is also 30s).
+const SHUTDOWN_DEADLINE_MS = 5000;
```
francoisferrand: quite a few constant it seems, may be better to introduce `lib/tracing/constants.ts` for these?

lib/tracing/instrumentation.ts
```diff
@@ -0,0 +1,126 @@
...
+// instrumentApiMethod gates on isEnabled() before any code path here that
+// calls getApi(), so it's safe to assume the require succeeds when reached.
+let api: typeof import('@opentelemetry/api') | null = null;
+function getApi(): typeof import('@opentelemetry/api') {
+    if (api) {
+        return api;
+    }
+    api = require('@opentelemetry/api');
+    return api!;
+}
```
francoisferrand, line 16: Can't we make it simpler?

```suggestion
const api = isEnabled() ? require('@opentelemetry/api') : null;
```

lib/tracing/kafkaTraceContext.ts
```diff
@@ -0,0 +1,107 @@
+import type { Context, Span, Link } from '@opentelemetry/api';
+
+type KafkaHeader = Record<string, Buffer | string>;
+
+const TRACER_NAME = 'arsenal.kafka';
+
+// Lazy-loaded so OTEL-off processes (and arsenal consumers that don't use the
+// kafka helpers at all, e.g. cloudserver) never pull in @opentelemetry/api.
+let api: typeof import('@opentelemetry/api') | null = null;
```
francoisferrand, line 8: same here, can't we just do this?

```suggestion
const api: isEnabled() ? import('@opentelemetry/api') : null;
```

lib/tracing/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+Shared OTEL tracing bootstrap and helpers for the S3 platform services
+(backbeat, cloudserver, vault). Arsenal owns the parts that are hard to get
+right and identical across services — SDK lifecycle, the outbound trust
+boundary, span conventions — while each consumer supplies its own
+instrumentation packages.
+
+## Import
+
+Deep-require the built module; do **not** use `require('arsenal').tracing`:
```
francoisferrand, line 11: nit: is there a some annotation/... we could set on the module to warn or fail when required via arsenal ? (e.g. in case somebody is using it without deep-requiring?)

package.json
francoisferrand: should we not align the version though? (`^2.7.1`)

francoisferrand (approved):

lib/tracing/bootstrap.ts
```diff
@@ -0,0 +1,113 @@
+import assert from 'assert';
+
+export interface InitOptions {
+    // service.name (OTEL_SERVICE_NAME env overrides).
+    serviceName: string;
+    // service.version (OTEL_SERVICE_VERSION env overrides).
+    serviceVersion?: string;
+    // Lazy thunk invoked once inside init() only when OTEL is enabled, so the
```
DarkIsDude:
```suggestion
    // Lazy invoked once inside init() only when OTEL is enabled, so the
```

lib/tracing/bootstrap.ts
```diff
@@ -0,0 +1,113 @@
...
+            `OTEL_SAMPLING_RATIO must be a finite number in [0, 1], got: ${process.env.OTEL_SAMPLING_RATIO}`,
+        );
+        samplingRatio = parsed;
+    }
+
+    // diag + NodeSDK are required lazily here (never at module top) so OTEL-off
+    // processes load nothing beyond @opentelemetry/api.
+    const { diag, DiagConsoleLogger, DiagLogLevel } = require('@opentelemetry/api');
+    const { NodeSDK } = require('@opentelemetry/sdk-node');
+    diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.WARN);
```
DarkIsDude: Or should we use werelogs ? I think your is the good one but did you check ?

lib/tracing/bootstrap.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    sdk = new NodeSDK(buildSdkConfig(options, endpoint, samplingRatio));
+    sdk.start();
+}
+
+// Split from init() so it's unit-testable without starting the SDK. Not on the
+// public barrel.
+export function buildSdkConfig(options: InitOptions, endpoint: string, samplingRatio: number): any {
```
DarkIsDude:
```suggestion
export function _buildSdkConfig(options: InitOptions, endpoint: string, samplingRatio: number): any {
```

A convention is to add _ to private method that need to be exported ? The comment don't help us to understand anything, splitting is always good

lib/tracing/bootstrap.ts
```diff
@@ -0,0 +1,113 @@
...
+        resource: resourceFromAttributes({
+            'service.name': process.env.OTEL_SERVICE_NAME || options.serviceName,
+            'service.version': process.env.OTEL_SERVICE_VERSION || options.serviceVersion,
+            // Operators set OTEL_SERVICE_NAME per-pod but not the namespace,
+            // so the 'scality' default marks Scality-owned traces.
+            'service.namespace': process.env.OTEL_SERVICE_NAMESPACE || 'scality',
+        }),
+        traceExporter: new OTLPTraceExporter({ url: endpoint }),
+        // Explicit empties: NodeSDK silently starts OTLP log + metric
+        // exporters otherwise. We only emit traces.
```
DarkIsDude:
```suggestion
        // Disable OTLP log + metric exporters with empties
```

lib/tracing/index.ts
```diff
@@ -0,0 +1,12 @@
+import * as bootstrap from './bootstrap';
+
+export { makeHttpInstrumentationConfig } from './httpHooks';
+export { instrumentApiMethod, endSpan } from './instrumentation';
+export type { InitOptions } from './bootstrap';
+export * as kafka from './kafkaTraceContext';
+
+// Value bindings, not `export { ... } from` — the latter compiles to a
+// non-configurable getter that consumer tests can't stub (e.g. tracing.close).
```
DarkIsDude:
```suggestion
```

Our tests will broke and it's pretty classic stuff from stub

lib/tracing/README.md
DarkIsDude, line 1: There is a `documentation` folder. Not sure where we should put that

lib/tracing/httpHooks.ts
```diff
@@ -0,0 +1,91 @@
...
+        }
+        if (span && typeof span.setAttribute === 'function') {
+            span.setAttribute('scality.trace.suppressed', true);
+        }
+    };
+}
+
+// True if url's path (query stripped) is in pathSet — for filtering probe/scrape
+// spans at ingest. The path set differs per service, so the caller passes it in.
+export function isHealthPath(url: unknown, pathSet: Set<string>): boolean {
```
DarkIsDude: This is not what the function does. This is how you use it, but it's more contains or something like that ? Why we must use unknown ? We can use string | null | undefined | ... ?

DarkIsDude (changes requested):

lib/tracing/instrumentation.ts
DarkIsDude: Not sure because if I remember well we want to be able to disable it on the fly (for example start with disable in test and then enable them).

package.json
delthas: There's no `api@2.x` to align to — the `api` package deliberately stays on 1.x as the stable, inert facade while the SDK packages (sdk-node, sdk-trace-base, resources) are on 2.x. So `^1.9.0` is correct.

lib/tracing/bootstrap.ts
delthas: Done — extracted to `lib/tracing/constants.ts` as `DEFAULT_SAMPLING_RATIO`.

lib/tracing/bootstrap.ts
delthas: Done — moved to `SPAN_LIMITS` in `constants.ts`. Kept fixed rather than env-configurable: they're conservative protective caps, not per-deployment tunables. Can revisit if a need comes up.

lib/tracing/bootstrap.ts
delthas: Done — added `lib/tracing/constants.ts` (DEFAULT_SAMPLING_RATIO, SPAN_LIMITS, SHUTDOWN_DEADLINE_MS, and the tracer scopes + span-name prefix).

lib/tracing/instrumentation.ts
delthas: Keeping the lazy `getApi()` getter. Evaluating `isEnabled()` once at module load freezes the decision, which breaks (a) enabling OTEL after this module is first required and (b) tests that toggle it via stubbing — as noted on the thread. The getter still loads nothing when OTEL is off (require deferred to first use).

lib/tracing/kafkaTraceContext.ts
delthas: Same as the instrumentation thread: a load-time `isEnabled()` check freezes the toggle, and `import()` is async — it would force these helpers to become async and change their call signatures. Keeping the lazy runtime require.

lib/tracing/README.md
delthas: There's no clean runtime hook to tell "required via the arsenal barrel" from "deep-required" — the module can't see how it was reached. The deep-require requirement is documented at the top of this README and enforced by the import path consumers use. Leaving as-is.

lib/tracing/bootstrap.ts
delthas: Applied.

lib/tracing/bootstrap.ts
delthas: Done — switched to a werelogs adapter: `new werelogs.Logger('tracing')` wired into `diag.setLogger`, so OTEL SDK diagnostics flow through the structured logging stack. This matches arsenal's per-module Logger pattern (ClusterRPC, HealthProbe/Probe/REST servers, and several storage backends all do `new werelogs.Logger(...)`).

lib/tracing/bootstrap.ts
delthas: Done — renamed to `_buildSdkConfig` and dropped the comment.

lib/tracing/bootstrap.ts
delthas: Done — applied your suggestion verbatim.

lib/tracing/index.ts
delthas: Done — dropped the comment; the value-binding-for-stubbing pattern is classic enough to stand on its own.

lib/tracing/README.md
delthas: Keeping it co-located. arsenal already has 5 module-level READMEs under `lib/` (network/kmip, network/kmsAWS, network/kmip/codec, executables/pensieveCreds, storage/metadata/proxy); the root `documentation/` folder is for cross-cutting topics (model versions, listing algos). So module docs living with the module is the established convention here.

lib/tracing/httpHooks.ts
delthas: Done — reworded to "True when url's path (query string stripped) exactly matches an entry in pathSet — used to skip spans for health/probe endpoints", and tightened `url: unknown` -> `url: string | undefined` to match the caller (`req.url`).

DarkIsDude (approved):
LGTM

lib/tracing/instrumentation.ts
```diff
@@ -0,0 +1,121 @@
...
+    };
+
+    const ctx = trace.setSpan(context.active(), span);
+    try {
+        return context.with(ctx, () => apiMethod.apply(self, wrappedArgs));
+    } catch (err) {
+        endSpan(span, err);
+        throw err;
+    }
+}
```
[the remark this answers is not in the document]
delthas: This only double-ends in a degenerate path: the consumer's own callback throwing synchronously after we've already ended the span. OTEL treats `end()` on a finished span as a no-op, and the extra `recordException`/`setStatus` on a dead span is harmless debug noise in that edge case only. The shape mirrors the original per-repo instrumentation; not adding a `spanEnded` flag for a buggy-handler path. Resolving.

PR #2634, opened 2026-06-05
test: functional test for putBucketIndexes error handling + bump test mongo to 8.0.23

delthas:
## Summary

Adds a functional test for `MongoClientInterface.putBucketIndexes` error handling — including the case where MongoDB 7.1+ refuses an index build because available disk falls below `indexBuildMinAvailableDiskSpaceMB`.

Companion to [BB-783](https://scality.atlassian.net/browse/BB-783) (Backbeat conductor metric for index-creation failures, scality/backbeat#2751) and [BB-784](https://scality.atlassian.net/browse/BB-784) (Prometheus alert on the metric, scality/backbeat#2753), under the [ARTESCA-16740](https://scality.atlassian.net/browse/ARTESCA-16740) "Empty bucket feature" epic.

## Commits

1. **`test: pin mongodb-memory-server to MongoDB 8.0.23`** — version pin via `package.json` `config.mongodbMemoryServer.version`. The library default (7.0.14) doesn't enforce `indexBuildMinAvailableDiskSpaceMB`; the test needs 7.1+ behavior. 8.0.23 also aligns with what Artesca 4.1+ runs in production.

2. **`test: add functional test for putBucketIndexes error handling`** — new spec covering three cases:
   - **Happy path**: valid spec → success.
   - **`indexBuildMinAvailableDiskSpaceMB` refusal**: set the parameter to 100 TiB at runtime, attempt an index build, restore the previous value in `finally`. MongoDB returns `OutOfDiskSpace` (code 14031) → Arsenal maps to `errors.InternalError`.
   - **Spec conflict**: existing index name + different keys → `IndexKeySpecsConflict` → `errors.InternalError`.

## Empirical observation

The disk-kill test confirms MongoDB returns:

```
{"errmsg":"available disk space of N bytes is less than required minimum of M","code":14031,"codeName":"OutOfDiskSpace"}
```

Arsenal's existing catch handler at `MongoClientInterface.putBucketIndexes` correctly maps this to `errors.InternalError` — same path used by all non-`NamespaceNotFound` MongoDB errors. Backbeat's `putBucketIndexesFailed` metric on BB-783 is cause-agnostic (fires on any `putBucketIndexes` error), so this gives us the signal regardless of the specific MongoDB error code.

## Notes

- The disk-kill test reads the current `indexBuildMinAvailableDiskSpaceMB` value via `getParameter`, sets it to 100 TiB during the test, and restores the original in `finally` — does not leak state to other tests.
- Async style follows @francoisferrand's review: `util.promisify` on `putBucketIndexes` in `beforeAll`, then `try { await ... } catch { assert; return } assert.fail(...)` for the two negative-path tests.

Issue: ARSN-588

[BB-783]: https://scality.atlassian.net/browse/BB-783?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BB-784]: https://scality.atlassian.net/browse/BB-784?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ARTESCA-16740]: https://scality.atlassian.net/browse/ARTESCA-16740?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/functional/metadata/mongodb/putBucketIndexes.spec.js
+++ b/tests/functional/metadata/mongodb/putBucketIndexes.spec.js
@@ -1,0 +1,147 @@
+const assert = require('assert');
+const util = require('util');
+const werelogs = require('werelogs');
+const { MongoMemoryReplSet } = require('mongodb-memory-server');
+
+const logger = new werelogs.Logger('MongoClientInterface', 'debug', 'debug');
+const BucketInfo = require('../../../../lib/models/BucketInfo').default;
+const MetadataWrapper = require('../../../../lib/storage/metadata/MetadataWrapper');
+
+const IMPL_NAME = 'mongodb';
+const DB_NAME = 'metadata';
+const BUCKET_NAME = 'test-put-bucket-indexes';
+
+const BUCKET_MD = {
+    _owner: 'testowner',
+    _ownerDisplayName: 'testdisplayname',
+    _creationDate: new Date().toJSON(),
+    _acl: {
+        Canned: 'private',
+        FULL_CONTROL: [],
+        WRITE: [],
+        WRITE_ACP: [],
+        READ: [],
+        READ_ACP: [],
+    },
+    _mdBucketModelVersion: 10,
+    _transient: false,
+    _deleted: false,
+    _serverSideEncryption: null,
+    _versioningConfiguration: null,
+    _locationConstraint: 'us-east-1',
+    _readLocationConstraint: null,
+    _cors: null,
+    _replicationConfiguration: null,
+    _lifecycleConfiguration: null,
+    _uid: '',
+    _isNFS: null,
+    ingestion: null,
+};
+
+const mongoserver = new MongoMemoryReplSet({
+    debug: false,
+    instanceOpts: [{ port: 27024 }],
+    replSet: {
+        name: 'rs0',
+        count: 1,
+        DB_NAME,
+        storageEngine: 'wiredTiger',
+    },
+});
+
+describe('MongoClientInterface:putBucketIndexes error handling', () => {
+    let metadata;
+    let putBucketIndexes;
+
+    beforeAll(async () => {
+        await mongoserver.start();
+        await mongoserver.waitUntilRunning();
+
+        const opts = {
+            mongodb: {
+                replicaSetHosts: 'localhost:27024',
+                writeConcern: 'majority',
+                replicaSet: 'rs0',
+                readPreference: 'primary',
+                database: DB_NAME,
+            },
+        };
+        metadata = new MetadataWrapper(IMPL_NAME, opts, null, logger);
+        metadata.setup = util.promisify(metadata.setup.bind(metadata));
+        metadata.createBucket = util.promisify(metadata.createBucket.bind(metadata));
+        metadata.close = util.promisify(metadata.close.bind(metadata));
+
+        await metadata.setup();
+        putBucketIndexes = util.promisify(metadata.client.putBucketIndexes.bind(metadata.client));
+
+        const bucketMD = BucketInfo.fromObj({
+            _name: BUCKET_NAME,
+            ...BUCKET_MD,
+        });
+        await metadata.createBucket(BUCKET_NAME, bucketMD, logger);
+    });
+
+    afterAll(async () => {
+        await metadata.close();
+        await mongoserver.stop();
+    });
+
+    const okSpec = [
+        {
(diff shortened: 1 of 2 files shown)


tests/functional/metadata/mongodb/putBucketIndexes.spec.js
```diff
@@ -0,0 +1,155 @@
...
+    ];
+
+    it('should succeed on a valid index spec', done => {
+        metadata.client.putBucketIndexes(BUCKET_NAME, okSpec, logger, err => {
+            assert.ifError(err);
+            done();
+        });
+    });
+
+    it('should return InternalError when indexBuildMinAvailableDiskSpaceMB exceeds available disk', async () => {
```
SylvainSenechal, line 100: nit, since you mentioned it returns this error : 
{"errmsg":"available disk space of N bytes is less than required minimum of M","code":14031,"codeName":"OutOfDiskSpace"}

would be better to assert on codeName=outOfDiskSpace instead internal error ?

SylvainSenechal: ah no cause it comes from putBucketIndexes, it would require modifying this function. ok np

SylvainSenechal (approved):

tests/functional/metadata/mongodb/putBucketIndexes.spec.js
```diff
@@ -0,0 +1,155 @@
...
+                    if (!err) {
+                        reject(new Error('expected an error from putBucketIndexes'));
+                        return;
+                    }
+                    try {
+                        assert(err.is.InternalError, `expected InternalError, got ${err.name} (${err.message})`);
+                        resolve();
+                    } catch (assertErr) {
+                        reject(assertErr);
+                    }
```
francoisferrand: would be simpler to just promisify `metadata.client.putBucketIndexes` (or a small lambda around it), and process error in catch:

```
try {
   await promisify(cb => {
      [...]
      metadata.client.putBucketIndexes(BUCKET_NAME, spec, logger, cb);
   });
} catch(err) {
   assert(err.is.InternalError, `expected InternalError, got ${err.name} (${err.message})`);
   return
} finally { ...}
assert(case, 'expected an error from putBucketIndexes');
```

(and I think there is some `mocha` to simplify this kind of "negative" tests)

tests/functional/metadata/mongodb/putBucketIndexes.spec.js
```diff
@@ -0,0 +1,155 @@
...
+                writeConcern: 'majority',
+                replicaSet: 'rs0',
+                readPreference: 'primary',
+                database: DB_NAME,
+            },
+        };
+        metadata = new MetadataWrapper(IMPL_NAME, opts, null, logger);
+        metadata.setup = util.promisify(metadata.setup.bind(metadata));
+        metadata.createBucket = util.promisify(metadata.createBucket.bind(metadata));
+        metadata.close = util.promisify(metadata.close.bind(metadata));
```
francoisferrand, line 72: not related to this PR : since we are somewhat in a process of migrating to async, we should probably take the time to async'ify MetadataWrapper itself (in a way which allows the API to be used both with callback and as asynchronous methods) -> will make it easier to migrate the test of the code

francoisferrand (approved):

tests/functional/metadata/mongodb/putBucketIndexes.spec.js
delthas: Applied in 4d0d5c1f. Pulled `putBucketIndexes` into a top-level `util.promisify` handle in `beforeAll`, then used `try { await ... } catch (err) { assert ...; return; } assert.fail(...)` for both negative-path cases. Tightens the diff and removes the `new Promise` wrapper.

tests/functional/metadata/mongodb/putBucketIndexes.spec.js
delthas: Agreed — would be a nice cleanup. Out of scope here, and not tracking it as a follow-up ticket for now.

PR #2637, opened 2026-06-10
optimize checkArnMatch: equality fast-path and memoized matchers

delthas:
`checkArnMatch` built six `RegExp` objects per call (regex-escape + compile for the relative-id and each of the five ARN segments), for every (request context × policy statement × Resource entry). A V8 profile of Vault under warm authorization load showed this as the worker's dominant CPU cost (`StringPrototypeReplace` + `RegExpConstructor` + `StringPrototypeSplit` under `evaluatePolicy`, plus the resulting GC churn).

Three commits:
1. **Equality fast-path** — a wildcard-free portion translates to a fully-escaped, anchored regExp, so testing it is equivalent to string equality: compare those portions directly and only build a `RegExp` for portions containing `*`, `?` or `${`.
2. **Matcher memoization** — compiled matchers are pure functions of `(caseSensitive, policyArn)`; cache them in a 10000-entry `LRUCache`. No invalidation needed; the LRU cap bounds the cardinality introduced by policy-variable substitution; a miss costs what every call cost before.
3. **Dead-code removal** — `checkArnMatch` was the only caller of `handleWildcardInResource`, which now calls `handleWildcards` per-portion. The helper is unused across arsenal and its consumers (Vault, CloudServer), so it is removed rather than left as a dead public export.

Behavior-preserving: verified by differential testing against the previous implementation over a matrix of literal/wildcard patterns, case-insensitive mode, policy-variable literals (`${*}`), empty-account service ARNs, and malformed ARNs — zero differences. Measured on a 1000-account Vault authz benchmark (warm, drift-controlled A→B→A): **+18–19% throughput, p95 −16%**.

Issue: ARSN-590

### DIFF
--- a/lib/policyEvaluator/utils/checkArnMatch.ts
+++ b/lib/policyEvaluator/utils/checkArnMatch.ts
@@ -21,27 +21,25 @@ export default function checkArnMatch(
     // object name in S3)
     // Join on ":" in case there were ":" in the relativeID at the end
     // of the arn
-    const policyRelativeId = caseSensitive ? regExofArn.slice(5).join(':') :
-        regExofArn.slice(5).join(':').toLowerCase();
+    const policyRelativeId = caseSensitive
+        ? regExofArn.slice(5).join(':')
+        : regExofArn.slice(5).join(':').toLowerCase();
     const policyRelativeIdRegEx = new RegExp(policyRelativeId);
     // Check to see if the relative-id matches first since most likely
     // to diverge.  If not a match, the resource is not applicable so return
     // false
-    if (!policyRelativeIdRegEx.test(caseSensitive ?
-        requestRelativeId : requestRelativeId.toLowerCase())) {
+    if (!policyRelativeIdRegEx.test(caseSensitive ? requestRelativeId : requestRelativeId.toLowerCase())) {
         return false;
     }
     // Check the other parts of the ARN to make sure they match.  If not,
     // return false.
     for (let j = 0; j < 5; j++) {
         const segmentRegEx = new RegExp(regExofArn[j]);
-        const requestSegment = caseSensitive ? requestArnArr[j] :
-            requestArnArr[j].toLowerCase();
+        const requestSegment = caseSensitive ? requestArnArr[j] : requestArnArr[j].toLowerCase();
         const policyArnArr = policyArn.split(':');
         // We want to allow an empty account ID for utapi and SUR service ARNs to not
         // break compatibility.
-        if (j === 4 && policyArnAllowedEmptyAccountId.includes(policyArnArr[2])
-            && policyArnArr[4] === '') {
+        if (j === 4 && policyArnAllowedEmptyAccountId.includes(policyArnArr[2]) && policyArnArr[4] === '') {
             continue;
         } else if (!segmentRegEx.test(requestSegment)) {
             return false;
--- a/lib/policyEvaluator/utils/checkArnMatch.ts
+++ b/lib/policyEvaluator/utils/checkArnMatch.ts
@@ -1,5 +1,12 @@
+import LRUCache from '../../algos/cache/LRUCache';
 import { handleWildcards } from './wildcards';
 import { policyArnAllowedEmptyAccountId } from '../../constants';
+
+// Compiled matchers are pure functions of (policyArn, caseSensitive), so the
+// cache needs no invalidation; the LRU cap bounds the cardinality introduced
+// by policy-variable substitution (per-request values in the policy arn).
+const ARN_MATCHER_CACHE_ENTRIES = 10000;
+const arnMatcherCache = new LRUCache(ARN_MATCHER_CACHE_ENTRIES);
 
 // A wildcard-free portion translates to a fully-escaped, anchored regExp,
 // so testing it is equivalent to comparing for string equality: keep the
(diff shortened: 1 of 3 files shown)


SylvainSenechal (approved):
Looks solid

lib/policyEvaluator/utils/checkArnMatch.ts
```diff
@@ -1,5 +1,66 @@
...
+type PortionMatcher = { literal: string } | { regExp: RegExp };
+
+type ArnMatcher = {
+    relativeId: PortionMatcher;
+    segments: PortionMatcher[];
+    skipAccountIdCheck: boolean;
+};
+
+function hasWildcards(portion: string): boolean {
+    return portion.includes('*') || portion.includes('?') || portion.includes('${');
```
francoisferrand: nit: since we are talking efficiency, calling `includes()` 3 times is intrinsically not very efficient (in C, this would be easy to fix, not sure in JS...)

lib/policyEvaluator/utils/checkArnMatch.ts
```diff
@@ -1,5 +1,66 @@
...
+    const segments: PortionMatcher[] = [];
+    for (let j = 0; j < 5; j++) {
+        const portion = policyArnArr[j];
+        if (portion !== undefined && !hasWildcards(portion)) {
+            segments.push({ literal: portion });
+        } else {
+            segments.push({
+                regExp: new RegExp(portion && handleWildcards(portion)),
+            });
+        }
```
francoisferrand: nit: keep a single segments.push, either with extra variable or ternary operator?

(tempting to do the same both `relativeId` above, but the "nested" ternary operator may make it harder to swallow...)

lib/policyEvaluator/utils/checkArnMatch.ts
```diff
@@ -1,5 +1,66 @@
-import { handleWildcardInResource } from './wildcards';
+import LRUCache from '../../algos/cache/LRUCache';
+import { handleWildcards } from './wildcards';
 import { policyArnAllowedEmptyAccountId } from '../../constants';
+
+// Compiled matchers are pure functions of (policyArn, caseSensitive), so the
+// cache needs no invalidation; the LRU cap bounds the cardinality introduced
+// by policy-variable substitution (per-request values in the policy arn).
+const ARN_MATCHER_CACHE_ENTRIES = 10000;
+const arnMatcherCache = new LRUCache(ARN_MATCHER_CACHE_ENTRIES);
```
francoisferrand, line 9: how much memory are we talking here?

lib/policyEvaluator/utils/checkArnMatch.ts
```diff
@@ -1,5 +1,66 @@
...
+    // of the arn
+    let relativeId: PortionMatcher;
+    if (policyArnArr.length === 6 && !hasWildcards(policyArnArr[5])) {
+        relativeId = {
+            literal: caseSensitive ? policyArnArr[5] : policyArnArr[5].toLowerCase(),
+        };
+    } else {
+        const source = policyArnArr.slice(5).map(handleWildcards).join(':');
+        relativeId = {
+            regExp: new RegExp(caseSensitive ? source : source.toLowerCase()),
```
francoisferrand, line 45: nit: wildcard match is relatively simple -much more than "general" regex. So I wonder if there would not be a deeper optimization where we use a simpler/faster engine....

(also, may be better to build a more "complex" expression to match the whole arn, instead of doing all the splitting/part-by-part matching/... in JS : like use `[^:]*` for wildcard instead of `.*`)

francoisferrand (approved):

lib/policyEvaluator/utils/checkArnMatch.ts
delthas: Good call — replaced the three `includes()` with a single module-level `RegExp` (`/[*?]|\$\{/`) compiled once at load (no `g` flag, so `.test()` is stateless and safe to reuse), one scan instead of three. Done in 224e9bf4.

Worth noting it only runs on a cache *miss* (the compiled matcher is memoized), so it's off the warm path either way — but the single-regex form reads better regardless.

lib/policyEvaluator/utils/checkArnMatch.ts
delthas: Done in 224e9bf4 — collapsed into a single `segments.push()` with a ternary.

Agreed on leaving `relativeId` as the explicit `if/else`: its regExp branch joins `slice(5)` and applies the case-folding only on that side, so a ternary there would be exactly the nested/hard-to-read kind you flagged.

lib/policyEvaluator/utils/checkArnMatch.ts
delthas: Measured the steady-state ceiling at the 10000-entry cap (filling the cache with distinct policy ARNs, `heapUsed` before/after, `--expose-gc`):

- **literal ARNs** (the common case, all-`{literal}` matchers, 0 RegExp): **~8 MB** (~843 B/entry)
- **worst case** (every portion a wildcard → up to 6 compiled RegExps/entry): **~21 MB** (~2.2 KB/entry)

So it's a bounded one-time ceiling (LRU evicts beyond 10k), realistically single-digit MB in production where ARNs are mostly literal. I left the cap at 10000 and kept it out of a code comment per the latest review; happy to lower it (e.g. 2000 → ~1.7–4 MB) or expose it as configurable if the worst case is a concern for a given deployment.

lib/policyEvaluator/utils/checkArnMatch.ts
delthas: I explored this — the short version is there's no easy win left, because the cost you're intuiting (regExp *compilation*) is already eliminated:

- the matcher is memoized, so each pattern compiles **once**, not per call;
- wildcard-free portions skip regExp entirely (string `===`), ~15× faster than `.test()`.

For the remaining wildcard portions, a *compiled* `RegExp.test` is already as fast as hand-rolled string matching on the warm path (micro-bench, ns/op):

| op | ns/op |
|---|---|
| `RegExp.test` `prefix*` (compiled once) | 50.4 |
| `startsWith` `prefix*` | 49.5 |
| `===` literal | 3.2 |

On the single whole-ARN regex with `[^:]*`: it would replace ~6 cheap ops with one (~15 ns saved, on wildcard ARNs only), but it needs *different* wildcard semantics per part — `[^:]*` for segments 0–4 but `.*` for the relative-id, where `*` legitimately spans `:` today, likely a marginal gain for a larger change. Preferring to keep this aside for now.

PR #2638, opened 2026-06-10
fix matching of resource ARNs whose relative-id contains a colon

delthas:
A policy `Resource` ARN whose relative-id contains `':'` (legal in S3 object keys, e.g. `bucket/data:2024/report.csv`) never matched any request — not even a byte-identical ARN. `checkArnMatch` anchors each `':'`-split portion individually (`^…$`) and then re-joins the relative-id portions, producing an unsatisfiable pattern like `^bucket/data$:^2024/….$` (`$` = end-of-input, so a required `':'` after it can never match). The request side already re-joins split portions correctly; only the policy side was broken.

Fix: wildcard-translate the **joined** relative-id as one unit, yielding a single anchored pattern. Six-portion ARNs compile to byte-identical patterns — no behavior change for them (all pre-existing tests pass unchanged); ARNs with fewer than six portions keep their previous behavior.

⚠ **Behavior change**: policy statements on colon-containing resources that were silently dead start applying after upgrade — `Allow` statements start granting (as their authors intended) and `Deny` statements start denying (they previously failed open).

Issue: ARSN-591

### DIFF
--- a/lib/policyEvaluator/utils/checkArnMatch.ts
+++ b/lib/policyEvaluator/utils/checkArnMatch.ts
@@ -40,7 +40,9 @@ function compileArnMatcher(policyArn: string, caseSensitive: boolean): ArnMatche
             literal: caseSensitive ? policyArnArr[5] : policyArnArr[5].toLowerCase(),
         };
     } else {
-        const source = policyArnArr.slice(5).map(handleWildcards).join(':');
+        // Translate the joined relative-id as one anchored unit so an embedded
+        // ':' (legal in S3 keys) matches; <6-portion ARNs have none, match any.
+        const source = policyArnArr.length >= 6 ? handleWildcards(policyArnArr.slice(5).join(':')) : '';
         relativeId = {
             regExp: new RegExp(caseSensitive ? source : source.toLowerCase()),
         };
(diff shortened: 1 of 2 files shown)


lib/policyEvaluator/utils/wildcards.ts
```diff
@@ -43,13 +41,20 @@ export const handleWildcards = (string: string) => {
...
+    // Wildcards do NOT span segments of the ARN (separated by ":")
 
     // Example: all elements in specific bucket:
     // "Resource": "arn:aws:s3:::my_corporate_bucket/*"
     // ARN format:
     // arn:partition:service:region:namespace:relative-id
     const arnArr = arn.split(':');
-    return arnArr.map(portion => handleWildcards(portion));
+    // The first five portions are the fixed ARN segments. The relative-id
+    // (everything past the fifth ':') may itself contain ':' and is
```
SylvainSenechal: nit: as you wrote a comment, could indicate why it may contain ":"

SylvainSenechal (approved):
nice catch, hopefully we don't get RDs with client suddenly having issues with their policies 🌚

lib/policyEvaluator/utils/wildcards.ts
delthas: Done — noted that S3 object keys may legally contain ':' (53bafdca)

benzekrimaha (approved):

PR #2639, opened 2026-06-11
feat(ARSN-592): expose manual instrumentation API in lib/tracing

delthas:
## Summary

Adds `startApiSpan(action)` to `lib/tracing` alongside the existing `instrumentApiMethod` — a manual-lifecycle helper for consumers whose dispatch site owns the span start + end points directly.

```ts
export interface ApiSpan {
    end(err?: any): void;
    withContext<T>(fn: () => T): T;
}

export function startApiSpan(action: string): ApiSpan;
```

- Span name `api.<action>` (same `SPAN_PREFIX` as `instrumentApiMethod`).
- `end(err?)` reuses the existing exported `endSpan(span, err?)` so the `recordException` + status + `error.type` ceremony stays in one place. Same err-as-optional convention.
- `withContext(fn)` runs `fn` with the span set on the active context so child auto-spans (mongo, ioredis, http) nest underneath.
- OTEL-off: returns a no-op object — no `@opentelemetry/api` load.

## Motivation

[VAULT-708](https://scality.atlassian.net/browse/VAULT-708) consumes the tracing module via vault PR [#203](https://github.com/scality/vault2/pull/203). Reviewer feedback ([thread](https://github.com/scality/vault2/pull/203#discussion_r3387477976)) pointed out that `instrumentApiMethod`'s wrap-once-at-module-load shape — natural for cloudserver's flat \`api[name]=handler\` table — forces vault to add a per-Route wrap cache + a static-this \`.bind\` dance + a new lazy method on Route just to amortize a wrap on a single centralized dispatch site (\`Router._startRequest\`). \`startApiSpan\` lets vault call the helper inline at the dispatch site instead, no caching layer needed.

Both APIs produce the same span output and share the underlying \`endSpan\` ceremony; the choice is stylistic at the consumer.

## Tests

5 new unit tests in \`tests/unit/tracing/instrumentation.spec.js\` (13 total in the file):

- \`end()\` on success → span ends with status OK
- \`end(err)\` → span ends with status ERROR + \`error.type\` from \`err.code\`
- \`withContext(fn)\` sets the started span as the active context inside \`fn\` (asserted by running inside an outer span and checking the active span flips correctly before / inside / after)
- \`withContext(fn)\` returns the value \`fn\` returns
- OTEL-off returns a no-op object — \`end()\`, \`end(err)\`, \`withContext\` all safe

## Related

- Parent: ARSN-586 — the original tracing module ([#2632](https://github.com/scality/Arsenal/pull/2632))
- Consumer: VAULT-708 / scality/vault2#203

Issue: ARSN-592

[VAULT-708]: https://scality.atlassian.net/browse/VAULT-708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/tracing/instrumentation.ts
+++ b/lib/tracing/instrumentation.ts
@@ -119,3 +119,31 @@ export function instrumentApiMethod<T extends (...args: any[]) => any>(apiMethod
         return instrumentAsyncHandler(this, apiMethod, spanName, args);
     } as T;
 }
+
+// Manual span lifecycle for consumers whose dispatch owns the start + end sites
+// directly (e.g. a centralized Router method). Prefer `instrumentApiMethod` for
+// flat dispatch tables where wrap-once-at-module-load fits naturally.
+export interface ApiSpan {
+    // `end()` marks the span OK; `end(err)` marks it ERROR + records the
+    // exception (same err-as-optional shape as `endSpan` above).
+    end(err?: any): void;
+    // Runs `fn` with the span set as the active context so child auto-spans
+    // (mongo, ioredis, http) nest underneath.
+    withContext<T>(fn: () => T): T;
+}
+
+export function startApiSpan(action: string): ApiSpan {
+    if (!isEnabled()) {
+        return {
+            end: () => {},
+            withContext: fn => fn(),
+        };
+    }
+    const { trace, context, SpanKind } = getApi();
+    const span = getTracer().startSpan(`${SPAN_PREFIX}${action}`, { kind: SpanKind.INTERNAL });
+    const ctx = trace.setSpan(context.active(), span);
+    return {
+        end: err => endSpan(span, err),
+        withContext: fn => context.with(ctx, fn),
+    };
+}
--- a/lib/tracing/index.ts
+++ b/lib/tracing/index.ts
@@ -1,7 +1,8 @@
 import * as bootstrap from './bootstrap';
 
 export { makeHttpInstrumentationConfig } from './httpHooks';
-export { instrumentApiMethod, endSpan } from './instrumentation';
+export { instrumentApiMethod, endSpan, startApiSpan } from './instrumentation';
+export type { ApiSpan } from './instrumentation';
 export type { InitOptions } from './bootstrap';
 export * as kafka from './kafkaTraceContext';
 
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
     "engines": {
         "node": ">=20"
     },
-    "version": "8.4.7",
+    "version": "8.4.8",
     "config": {
         "mongodbMemoryServer": {
             "version": "8.0.23"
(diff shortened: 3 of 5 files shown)


lib/tracing/instrumentation.ts
```diff
@@ -119,3 +119,48 @@ export function instrumentApiMethod<T extends (...args: any[]) => any>(apiMethod
...
+//     it surfaces, rather than calling `cb` a second time.
+export function withApiSpan(
+    action: string,
+    work: (done: (err: any, ...results: any[]) => void) => void,
+    cb: (err: any, ...results: any[]) => void,
+): void {
+    if (!isEnabled()) {
+        work(cb);
+        return;
+    }
```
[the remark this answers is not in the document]
delthas: Resolved by removing the catch from both paths instead. After reviewing francois's feedback on vault PR #203 in aggregate, the right move was to revert `withApiSpan` (which added more abstraction) back to `startApiSpan` (minimal lifecycle helper that consumers wrap explicitly). Both OTEL-on and OTEL-off return a span-like object; neither catches sync throws. The asymmetry you flagged is now gone because there's no catch anywhere — any sync throw from a consumer's work bubbles up consistently regardless of OTEL state. Consumers whose handlers are strictly callback-based (the vault convention) write `span.end(err)` inline in their callback.

lib/tracing/instrumentation.ts
```diff
@@ -119,3 +119,33 @@ export function instrumentApiMethod<T extends (...args: any[]) => any>(apiMethod
...
+    end(err?: any): void;
+    // Runs `fn` with the span set as the active context so child auto-spans
+    // (mongo, ioredis, http) nest underneath.
+    withContext<T>(fn: () => T): T;
+}
+
+const NO_OP_API_SPAN: ApiSpan = {
+    end: () => {},
+    withContext: fn => fn(),
+};
```
francoisferrand: nit: may be more readable to inline this "const" inside `startApiSpan`? Otherwise please find a better name (sequence of short capital abreviations is really hard to read...)

francoisferrand (approved):

lib/tracing/instrumentation.ts
delthas: Done in `15cd5805` — inlined the no-op object directly into the `!isEnabled()` branch of `startApiSpan` and dropped the `NO_OP_API_SPAN` const, so the awkward name is gone entirely.

DarkIsDude (approved):

DarkIsDude:
PS: codecov is not happy when it should. The code is very simple to test 🙏

delthas:
> PS: codecov is not happy when it should. The code is very simple to test 🙏

codecov patch is happy. It's just that the overall project score went below 75% earlier so all PRs have this. :stuck_out_tongue:  Still, added an additional test.

PR #2640, opened 2026-06-11
ARSN-594: guarantee a single callback in RESTClient

tcarmet:
RESTClient's request methods have two independent completion paths - the response handler and the request 'error' handler - with no guard between them, so a socket error arriving after the response was received invokes the callback a second time. When the file data backend (DataFileInterface) is consumed by retrieveData, that double invocation reaches the async.eachSeries iteration callback and throws "Callback was already called", surfacing as an uncaughtException that crashes the server process on a mid-stream client disconnect. This was reproduced in cloudserver's file-ft-tests.

This guarantees each RESTClient method's callback fires at most once, mirroring the guarantee sproxydclient already provides at its request layer. A unit test reproduces the response-then-socket-error double call.

A broader data-layer backstop - enforcing the single-callback contract once at the DataWrapper.get boundary so every consumer and any future backend is covered by a single guard - is tracked separately in ARSN-595.

### DIFF
--- a/lib/network/rest/RESTClient.ts
+++ b/lib/network/rest/RESTClient.ts
@@ -6,6 +6,7 @@ import * as utils from './utils';
 import errors, { ArsenalError } from '../../errors';
 import { http as HttpAgent } from 'httpagent';
 import * as stream from 'stream';
+import { once } from '../../jsutil';
 
 function setRequestUids(reqHeaders: http.IncomingHttpHeaders, reqUids: string) {
     // inhibit 'assignment to property of function parameter' -
@@ -160,6 +161,9 @@ export default class RESTClient {
         reqUids: string,
         callback: (error: Error | null, key?: string) => void,
     ) {
+        // Guarantee a single callback invocation across the response and
+        // request 'error' paths (see get()).
+        const cbOnce = once(callback);
         const log = this.createLogger(reqUids);
         const headers = {};
         setRequestUids(headers, reqUids);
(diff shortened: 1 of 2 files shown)


dvasilas (approved):
You also need a version bump?

BourgoisMickael (approved):

lib/s3routes/routesUtils.ts
```diff
@@ -429,61 +429,68 @@ export function retrieveData(
...
-                    log.error('failed to get object', {
-                        error: err,
-                        method: 'retrieveData',
+        (current, next) => {
+            // create the once() wrapper at iteration scope: the data layer
+            // may invoke its callback more than once on connection teardown
+            // races, and a per-invocation wrapper would let `next` be called
+            // twice, crashing the process through async's double-callback
+            // detection
+            const cbOnce = jsutil.once(next);
```
jonathan-gramain: Shouldn't we fix the data layer itself to guarantee exactly one callback invocation there? Looking at sproxydclient `_createRequest` method, it seems it was updated to avoid double-callback invocation already in https://github.com/scality/sproxydclient/commit/1c21c1a991f8b362c1b645aac9ccd01590d47b59, so in principle it shouldn't be necessary to add higher-level handling of this unless the fix is incomplete.

tcarmet: you're right sproxydclient is properly fixed and will not face this issue, the problem lies with the file backend on the `file-ft-tests` in cloudserver which uses DataFileInterface in arsenal, so I'm working on applying a similar fix there as well and reviewing if I should remove the guard from here or if there's a more global location to add it.

tcarmet: Ended up fixing it in RESTClient only (the file backend's client, mirroring sproxydclient's guard), and dropped the higher-level retrieveData guard. Tracked the broader DataWrapper.get backstop as ARSN-595 in case we decide to go for it later.

lib/network/rest/RESTClient.ts
```diff
@@ -160,6 +161,9 @@ export default class RESTClient {
         reqUids: string,
         callback: (error: Error | null, key?: string) => void,
     ) {
+        // Guarantee a single callback invocation across the response and
+        // request 'error' paths (see get()).
+        const cb = once(callback);
```
jonathan-gramain: Typically named `cbOnce` to mark the fact that it's wrapped hence may be called multiple times.

jonathan-gramain (approved):

lib/network/rest/RESTClient.ts
tcarmet: Done - renamed to `cbOnce` across all four methods.

PR #2643, opened 2026-06-16
ARSN-596: treat NamespaceExists as idempotent success in createBucket

tcarmet:
MongoDB returns error code 48 (`NamespaceExists`) when `createCollection` is called on a collection that already exists - a benign outcome under concurrent create/drop/create sequences on MPU shadow buckets. The existing code let this error fall through to the outer catch, which converted every `createCollection` error into `InternalError`. That 500 response was the root cause of `mongo-v0-ft-tests` and `mongo-v1-ft-tests` failing very frequently in cloudserver.

A unit test was added that stubs `createCollection` to reject with a `NamespaceExists` error and asserts the callback receives no error. We ran the cloudserver mongo functional test suite 20 times in parallel. 4 runs failed on pre-existing unrelated flakes; the target failure did not appear in any run. Full results: https://github.com/scality/cloudserver/actions/runs/27648800616

Issue: ARSN-596

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -466,21 +466,28 @@ class MongoClientInterface {
                 // "constants.usersBucket" and "PENSIEVE" since it has already
                 // been created
                 if (bucketName !== constants.usersBucket && bucketName !== PENSIEVE) {
-                    return this.db!.createCollection(bucketName).then(() => {
-                        if (this.shardCollections) {
-                            const cmd = {
-                                shardCollection: `${this.database}.${bucketName}`,
-                                key: { _id: 1 },
-                            };
-                            return this.adminDb!.command(cmd, {})
-                                .then(() => cb(null))
-                                .catch(err => {
-                                    log.error('createBucket: enabling sharding', { error: err });
-                                    return cb(errors.InternalError);
-                                });
-                        }
-                        return cb(null);
-                    });
+                    return this.db!.createCollection(bucketName)
+                        .catch(err => {
+                            if (err.codeName === 'NamespaceExists') {
+                                return;
+                            }
+                            throw err;
+                        })
+                        .then(() => {
+                            if (this.shardCollections) {
+                                const cmd = {
+                                    shardCollection: `${this.database}.${bucketName}`,
+                                    key: { _id: 1 },
+                                };
+                                return this.adminDb!.command(cmd, {})
+                                    .then(() => cb(null))
+                                    .catch(err => {
+                                        log.error('createBucket: enabling sharding', { error: err });
+                                        return cb(errors.InternalError);
+                                    });
+                            }
+                            return cb(null);
+                        });
                 }
                 return cb(null);
             })
--- a/tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
+++ b/tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
@@ -1900,6 +1900,26 @@ describe('MongoClientInterface, createBucket', () => {
         });
     });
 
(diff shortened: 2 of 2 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -466,21 +466,28 @@ class MongoClientInterface {
...
-                                });
-                        }
-                        return cb(null);
-                    });
+                    return this.db!.createCollection(bucketName)
+                        .catch(err => {
+                            if (err.codeName === 'NamespaceExists') {
+                                return;
+                            }
+                            throw err;
```
jonathan-gramain, line 474: I think we could call `cb` here to satisfy the contract of `createBucket` and call the callback in any case, as the `then` clause is not followed by a `catch` clause. It seems like the error case was already missing in the original code, causing the exception to be thrown and caught in the outermost layer. Calling the callback in any case and not throwing an exception is probably a better design choice. We could otherwise add a final `.catch(err => cb(errors.InternalError))` to catch any thrown error and turn it into an error callback call there instead of relying on the outermost layer.

francoisferrand:
Not really a review, but just providing some history/context as there is a slightly deeper question here. `MongoClientInterface` used to have this kind of idempotent behavior (not sure about this specific API), which ended up creating race conditions in Cloudserver (and ghost/orphan objects at customers), and it was changed ~last year ([1], [2] maybe missing some) to report errors more strictly: pushing the handling of errors up to cloudserver.

If I remember correctly (@benzekrimaha may remember as well?), it -the earlier, idempotent, approach, not sure about this specific error in this code path- created issues when multiple operations were needed (MPU or master+version), which could not be handled properly in cloudserver without knowing if the call "really" succeeded...

i.e. there are 2 kinds of use cases:
- making the methods idempotent and masking the error is fine for simple CRUD api (i.e. `s3:createBucket` api)
- but we are doing a sequence "create X then do Y", we may end up having a race condition where we do twice `Y`

Also when creating a bucket, we may be able to set some parameters (quota, activating Veeam...) : for the API to be idempotent, these would need to be set even with this error

So we should tread carefully here:
- Reviewing all places where this specific api is used, to confirm that idempotency is acceptable in every case
- Confirm what is the behavior of the other implementation of this interface? All implementations should have the same idempotency caracteristic
- Having some functions idempotent and some not will make it harder to correctly deal with the interface, would be best to stick to one rule for the whole metadata interface: some functions cannot be idempotent, so do we really want to mix-and-match?

[1] https://github.com/scality/Arsenal/pull/2363
[2] https://github.com/scality/Arsenal/pull/2387

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -466,21 +466,28 @@ class MongoClientInterface {
...
-                                    log.error('createBucket: enabling sharding', { error: err });
-                                    return cb(errors.InternalError);
-                                });
-                        }
-                        return cb(null);
-                    });
+                    return this.db!.createCollection(bucketName)
+                        .catch(err => {
+                            if (err.codeName === 'NamespaceExists') {
+                                return;
```
francoisferrand, line 472: will this not make us go through the `shardCollections` below?
could this cause a race condition if 2 `shardCollection` commands are sent to mongo in parallel, and possibly let the call fail (and still return an error)?

tcarmet:
thank you @jonathan-gramain and @francoisferrand for the extra context and review, I'll report back, need a bit of time to investigate and experiment with the tips and info you gave me.

DarkIsDude (commented):
Let me know when it's reviewable again 🙏

tcarmet:
Closing since it has been replaced by #2665 

Thanks to everyone involved on pod RR for taking over this work, much appreciated :heart:

PR #2648, opened 2026-06-18
Allow Days=0 in lifecycle rules (relax day-based actions to nonnegative)

delthas:
## What

Allow `Days = 0` in S3 lifecycle rules. Previously the day-based actions required a value `>= 1`; this accepts `0` (nonnegative) for:

- `Expiration` → `Days`
- `NoncurrentVersionExpiration` → `NoncurrentDays`
- `AbortIncompleteMultipartUpload` → `DaysAfterInitiation`

All three now go through the existing `_checkDays()` helper (already used by `Transition` / `NoncurrentVersionTransition`), which accepts `0..MAX_DAYS` and returns `must be nonnegative` / `must not exceed` errors. As a side effect this adds the `MAX_DAYS` (2³¹−1) upper bound that these three actions previously lacked, consistent with the transition actions.

## Why

`Days = 0` is meant as an explicit *"empty this bucket"* intent. Backbeat (BB-779) keys on it to skip building the v2 lifecycle index, which is expensive (~5–10× its size in transient disk during the merge-sort) and pointless when a bucket is being drained.

`Days = 1` can't serve this purpose because legitimate short-retention buckets (logs, etc.) use it. Since `Days = 0` was rejected everywhere until now, it is an unambiguous marker that cannot collide with any previously-valid configuration.

This intentionally diverges from AWS S3 (which requires `Days >= 1`) — that divergence is exactly what makes `0` a safe internal signal.

## Falsy-zero fixes

Accepting `0` exposed several spots that treated the day value as falsy and would have silently dropped a `Days = 0` rule. Fixed so the rule is preserved end to end:

- `_parseAction` copied action times with `if (action[t])` → now `!== undefined`
- `getConfigXml` serialized `Expiration` `Days` with `days ? …` → now `days !== undefined` (otherwise GetBucketLifecycle returned an `Expiration` with no action time)
- `_validateRules` skipped the type assertion when `days === 0`
- `LifecycleUtils.getApplicableRules` guards/comparisons dropped `0` before it could be selected as the earliest applicable action — this is the path that feeds expiration decisions

## Scope

- `NewerNoncurrentVersions` keeps its `>= 1` rule (it is a count, not a number of days).
- Consumers are handled separately: the cloudserver dependency bump + functional tests, and the backbeat skip-index heuristic (BB-779).

Issue: ARSN-597


lib/models/LifecycleConfiguration.ts
francoisferrand, line 1: do we also need to check in pensieve-api if some fix/cleanup is needed?

francoisferrand, line 1: to be confirmed by product team : can this be allowed everywhere, or do we need to make this opt-in so it can be selectively allowed?

lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
```diff
@@ -267,12 +267,14 @@ export default class LifecycleUtils {
                 const ncd = 'NoncurrentDays';
                 const nncv = 'NewerNoncurrentVersions';
 
-                if (ncvExpiration[ncd]) {
-                    if (!store[ncve]?.[ncd] || ncvExpiration[ncd] < store[ncve][ncd]) {
+                const ruleNcd = ncvExpiration[ncd];
+                if (typeof ruleNcd === 'number') {
+                    const storedNcd = store[ncve]?.[ncd];
+                    if (typeof storedNcd !== 'number' || ruleNcd < storedNcd) {
```
francoisferrand: should we not compare to `undefined` instead of comparing to number? (like "days" in LifecycleConfiguration). This would make consistent between validation and usage...

lib/models/LifecycleConfiguration.ts
```diff
@@ -1011,10 +1011,13 @@ export default class LifecycleConfiguration {
             return abortObj;
         }
         const daysInt = parseInt(subAbort.DaysAfterInitiation[0], 10);
-        if (daysInt < 1) {
-            abortObj.error = errorInstances.InvalidArgument.customizeDescription(
-                'DaysAfterInitiation is not a positive integer',
-            );
+        const error = this._checkDays({
```
francoisferrand, line 1039: should we add a log when days is 0 ?

- may help to identify case when customer (mistakenly) deleted everything with a malformed rule
- but maybe this would be lost in logs (lots of activity in cloudserver), maybe a metrics would help better?
- or even too complex to generate the log from here?
...or maybe this is overkill, and enough to just allow this "extended" setting?

lib/models/LifecycleConfiguration.ts
```diff
@@ -969,10 +963,10 @@ export default class LifecycleConfiguration {
                     'deleteMarker',
                     'transition',
                     'nonCurrentVersionTransition',
-                    'newerNoncurrentVersions'
+                    'newerNoncurrentVersions',
                 ];
                 actionTimes.forEach(t => {
-                    if (action[t]) {
+                    if (action[t] !== undefined) {
```
DarkIsDude: Why this one ?

lib/models/LifecycleConfiguration.ts
DarkIsDude: I think a log is a good idea. No cost and can save us many times

lib/models/LifecycleConfiguration.ts
```diff
@@ -1131,13 +1137,15 @@ export default class LifecycleConfiguration {
         };
 
         const daysInt = parseInt(subNVExp.NoncurrentDays[0], 10);
-        if (daysInt < 1) {
-            const msg = 'NoncurrentDays is not a positive integer';
-            const error = errorInstances.InvalidArgument.customizeDescription(msg);
+        const error = this._checkDays({
+            days: daysInt,
+            field: 'NoncurrentDays',
+            ancestor: 'NoncurrentVersionExpiration',
```
DarkIsDude, line 1163: I don't remember exactly the design but should we allow everything to be 0 days ? My point is : if we want to empty a bucket, maybe we should only allow to expire "current version" and first clean everything else after one day ? It's a matter of safety but allowing 0 everywhere give more flexibility to our customer.

lib/models/LifecycleConfiguration.ts
DarkIsDude: Go it, it's to allow 0. But are we sure we want to check only undefined ? Not null ?

lib/models/LifecycleConfiguration.ts
delthas: I'd say more flexibility, and safety can be done with a UI warning etc. Currently users can mess up their install badly if they do weird stuff with the API anyway.

lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
delthas: Good consistency point, but the asymmetry is type-driven. `Expiration.Days` is an optional key (`Days?: number`), so it's `undefined` when unset and `=== undefined` is the right check. `NoncurrentVersionExpiration.NoncurrentDays` is typed `number | null`: the rule builder (`addNCVExpiration` in `LifecycleRule`) always sets the key and uses `null` to mean "unset", so `=== undefined` would never match — and would misread a freshly-built `null` as a real value.

I've refactored these guards to `!= null` / `== null` (a64961e3): keeps `0` valid, treats a `null` stored value as "not yet set", and reads cleaner than the `typeof` checks.

lib/models/LifecycleConfiguration.ts
delthas: Done in 4f0ac7f6. The constructor now takes a werelogs logger and `_checkDays` emits `log.info('lifecycle rule configured with a 0-day action time', { field, ancestor })` whenever a day-based action resolves to `0`. Centralizing in `_checkDays` covers Expiration / NoncurrentVersionExpiration / AbortIncompleteMultipartUpload / Transition from a single place.

CloudServer will pass its request logger, so the line is correlated to the request (wired in the CLDSRV-928 PR).

lib/models/LifecycleConfiguration.ts
delthas: Right — it's to allow `0` (the previous `if (action[t])` dropped a legitimate `days: 0`).

On null: the sub-parsers (`_parseExpiration`, `_parseNoncurrentVersionExpiration`, ...) only ever return `number | undefined`, never `null`, so `!== undefined` was sufficient. But I've switched it to `!= null` (a64961e3) so it covers both and matches the `LifecycleUtils` guards — no separate null handling needed.

lib/models/LifecycleConfiguration.ts
delthas: Ack, tracked in [PSVAPI-128](https://scality.atlassian.net/browse/PSVAPI-128)

lib/models/LifecycleConfiguration.ts
delthas: Product said OK for merge as is, no need for separate warning / feature-flag.

lib/models/LifecycleConfiguration.ts
delthas: Product said OK for merge as is, no need for separate warning / feature-flag.

lib/models/LifecycleConfiguration.ts
```diff
@@ -1005,22 +999,22 @@ export default class LifecycleConfiguration {
         }
         if (filter && filter.Tag) {
             abortObj.error = errorInstances.InvalidRequest.customizeDescription(
-                'Tag-based filter cannot be used with ' +
-                'AbortIncompleteMultipartUpload action',
+                'Tag-based filter cannot be used with ' + 'AbortIncompleteMultipartUpload action',
```
francoisferrand:
```suggestion
'Tag-based filter cannot be used with AbortIncompleteMultipartUpload action',
```

(a few like this below)

lib/models/LifecycleConfiguration.ts
```diff
@@ -593,6 +598,12 @@ export default class LifecycleConfiguration {
                 `'${field}' in ${ancestor} action must not exceed ${MAX_DAYS}`,
             );
         }
+        if (days === 0) {
+            this._log.info('lifecycle rule configured with a 0-day action time', {
```
francoisferrand: nit: should it be a warning?
(not sure really: it is not standard and dangerous.... but at the same time it is now valid and supported...)

lib/models/LifecycleConfiguration.ts
```diff
@@ -1011,10 +1011,13 @@ export default class LifecycleConfiguration {
             return abortObj;
         }
         const daysInt = parseInt(subAbort.DaysAfterInitiation[0], 10);
```
francoisferrand, line 1038: this will fail with an empty string, right?
(e.g. one significant risk that 0 may introduce is the case where serialization/convertion/... ends up turning some default/unset field into a 0 : and a user which did not configure anything would get his data wiped)

seems to me we should be fine (we check the `subAbort.DaysAfterInitiation` is not empty, and then I expect parseInt() to throw on empty string?), but worth double checking...

lib/models/LifecycleConfiguration.ts
```diff
@@ -1064,10 +1067,13 @@ export default class LifecycleConfiguration {
...
             const daysInt = parseInt(subExp.Days[0], 10);
-            if (daysInt < 1) {
-                expObj.error = errorInstances.InvalidArgument.customizeDescription(
-                    'Expiration days is not a positive integer',
-                );
+            const error = this._checkDays({
+                days: daysInt,
+                field: 'Days',
+                ancestor: 'Expiration',
+            });
```
francoisferrand: nit: fits on one line...

```suggestion
            const error = this._checkDays({ days: daysInt, field: 'Days', ancestor: 'Expiration' });
```

lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
```diff
@@ -267,12 +267,14 @@ export default class LifecycleUtils {
...
-                    if (!store[ncve]?.[ncd] || ncvExpiration[ncd] < store[ncve][ncd]) {
+                const ruleNcd = ncvExpiration[ncd];
+                if (ruleNcd != null) {
+                    const storedNcd = store[ncve]?.[ncd];
+                    if (storedNcd == null || ruleNcd < storedNcd) {
                         store[ncve] = {
                             ...store[ncve],
                             ID: rule.ID,
-                            [ncd]: ncvExpiration[ncd],
+                            [ncd]: ruleNcd,
```
francoisferrand: nit: IMHO the extra variables does not actually help readability here, I would rather have kept the diff short as well as the symmetry with other cases...

(same for storedDai below)

tests/unit/models/LifecycleConfiguration.spec.js
```diff
@@ -85,7 +85,13 @@ const invalidActions = [
         tag: 'AbortIncompleteMultipartUpload',
         label: 'invalid-days',
         error: 'InvalidArgument',
-        errMessage: 'DaysAfterInitiation is not a positive integer',
+        errMessage: "'DaysAfterInitiation' in AbortIncompleteMultipartUpload " + 'action must be nonnegative',
```
francoisferrand: to be confirmed, but I think the error message match what AWS generates: so we should not change them....

tests/unit/models/LifecycleConfiguration.spec.js
```diff
@@ -546,6 +567,97 @@ describe('LifecycleConfiguration class getLifecycleConfiguration', () => {
             done();
         });
     });
+
```
francoisferrand, line 573: should we add a check to verify that "empty" expiration (`<Expiration></Expiration>`) or days (`<Expiration><Days></Days></Expiration>`) do not generate an rule with actions.days 0 ?

or is it covered already by existing tests? (or maybe we still want to test it explicitly, for documentation & be reassured?)

tests/unit/models/LifecycleConfiguration.spec.js
delthas: Good prompt to double-check — I reproduced the actual AWS errors (`aws s3api put-bucket-lifecycle-configuration` with negative and zero day values). How the three day-validator messages compare:

| Field | Pre-PR arsenal | This PR (`_checkDays`) | AWS (reproduced) |
|---|---|---|---|
| Expiration `Days` | `Expiration days is not a positive integer` | `'Days' in Expiration action must be nonnegative` | `'Days' for Expiration action must be a positive integer` |
| `NoncurrentDays` | `NoncurrentDays is not a positive integer` | `'NoncurrentDays' in NoncurrentVersionExpiration action must be nonnegative` | `'NoncurrentDays' for NoncurrentVersionExpiration action must be a positive integer` |
| `DaysAfterInitiation` | `DaysAfterInitiation is not a positive integer` | `'DaysAfterInitiation' in AbortIncompleteMultipartUpload action must be nonnegative` | `'DaysAfterInitiation' for AbortIncompleteMultipartUpload action must be a positive integer` |

Two things stand out:

- **Neither the pre-PR nor this PR's message actually matches AWS.** AWS uses `'<Field>' for <Action> action must be a positive integer`; the pre-PR strings were `<Field> is not a positive integer` (no quotes, no "for … action"). So this PR didn't regress *away* from an AWS-matching string — the pre-PR wording differed from AWS too.
- **AWS rejects both `0` and negative** with the same "must be a positive integer" (verified). Since this PR now *accepts* `Days=0`, our error only fires for negatives — which is why I used "nonnegative" rather than "positive integer" (saying "must be a positive integer" while we accept `0` would contradict itself).

Options if we want to get closer to AWS:

1. **AWS phrasing + accurate bound:** `'<Field>' for <Action> action must be nonnegative` — adopts AWS's `for <Action> action` structure but keeps "nonnegative" (correct, since `0` is valid here).
2. **Exact AWS string:** `'<Field>' for <Action> action must be a positive integer` — byte-identical to AWS for the negative case; downside: it says "positive integer" although we accept `0`.
3. Keep as-is.

WDYT?

lib/models/LifecycleConfiguration.ts
delthas: Good catch — and it does *not* throw: `parseInt('', 10)` returns `NaN`, and `_checkDays` let `NaN` slip through (all of `<0`, `>MAX`, `===0` are false), so an empty `<Days/>` produced a `days: NaN` rule with **no** error. Fixed in b9f803c3 with a `Number.isNaN` guard at the top of `_checkDays` (returns `MalformedXML`), which covers all three day fields centrally. Explicit tests added (see the empty-case thread).

lib/models/LifecycleConfiguration.ts
delthas: Done in b9f803c3.

lib/models/LifecycleConfiguration.ts
delthas: Collapsed this and the other split error-message literals to single strings in b9f803c3.

lib/models/LifecycleConfiguration.ts
delthas: Agreed — switched to `log.warn` in b9f803c3. It's valid/supported, but "0 days = empty the bucket" is dangerous enough to warrant a warning in the audit trail.

lib/s3middleware/lifecycleHelpers/LifecycleUtils.ts
delthas: Inlined both in b9f803c3 — dropped the `storedNcd`/`storedDai` locals so it mirrors the Expiration case's style. (tsc still narrows the `const`-key element access fine.)

tests/unit/models/LifecycleConfiguration.spec.js
delthas: Not covered before — added explicit tests in b9f803c3 that empty `<Days>` / `<NoncurrentDays>` / `<DaysAfterInitiation>` now error with `'<field>' in <action> action must be an integer` instead of silently producing a `days: NaN` rule. (This is the test side of the empty-string/NaN fix above.)

tests/unit/models/LifecycleConfiguration.spec.js
francoisferrand: > "must be a positive integer" while we accept 0 would contradict itself

not really, we can play on "strictly" positive ambiguity ;-)

> Options if we want to get closer to AWS: [...] WDYT?

since we don't already match, not sure the argument really matters... maybe product team has more precise requirement there.

lib/models/LifecycleConfiguration.ts
```diff
@@ -587,13 +589,25 @@ export default class LifecycleConfiguration {
...
             const msg = `'${field}' in ${ancestor} action must be nonnegative`;
             return errorInstances.InvalidArgument.customizeDescription(msg);
         }
         if (days > MAX_DAYS) {
             return errorInstances.MalformedXML.customizeDescription(
-                `'${field}' in ${ancestor} action must not exceed ${MAX_DAYS}`);
+                `'${field}' in ${ancestor} action must not exceed ${MAX_DAYS}`,
+            );
+        }
+        if (days === 0) {
```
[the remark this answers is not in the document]
delthas: Good catch, thanks. Fixed in 397b89e2 — I scoped it a touch narrower than the suggestion: the warn now fires only for the object-deleting expirations (`Expiration` + `NoncurrentVersionExpiration`), i.e. the genuinely destructive "empty the bucket" 0-day case. `Transition`/`NoncurrentVersionTransition` (which already accepted `Days=0` = "transition immediately" before this PR) and `AbortIncompleteMultipartUpload` (aborts in-progress uploads, not stored objects) no longer warn. Added tests: NCV `NoncurrentDays=0` warns; AbortMPU `DaysAfterInitiation=0` parses cleanly without warning.

tests/unit/models/LifecycleConfiguration.spec.js
delthas: OK, asked Florent

delthas: Suggest we keep this as a "nit" thread and approve if all else is OK and I'll address when Florent replies

francoisferrand (approved):
LGTM, pending Florent inptu on error messages

DarkIsDude (approved):

benzekrimaha (approved):

tests/unit/models/LifecycleConfiguration.spec.js
delthas: Florent said let's match AWS exactly. Changing to mimick AWS exactly.

PR #2655, opened 2026-06-25
Bumps werelogs from 8.2.2 to 8.2.4 and bumps arsenal to 8.3.13

maeldonn:
Issue: ARSN-603

### DIFF
--- a/lib/storage/data/external/AwsClient.js
+++ b/lib/storage/data/external/AwsClient.js
@@ -102,10 +102,8 @@ class AwsClient {
                     applyRegion(err.region);
                     return;
                 }
-                this._logger.error('error during setup', {
-                    error: err,
-                    method: 'AwsClient.setup',
-                });
+                logHelper(this._logger, 'error', 'error during setup',
+                    err, this._dataStoreName, this.clientType);
                 throw err;
             }
         });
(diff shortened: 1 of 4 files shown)


lib/storage/data/external/AwsClient.js
```diff
@@ -102,10 +102,8 @@ class AwsClient {
                     applyRegion(err.region);
                     return;
                 }
-                this._logger.error('error during setup', {
-                    error: err,
-                    method: 'AwsClient.setup',
-                });
+                logHelper(this._logger, 'error', 'error during setup',
```
francoisferrand, line 105: nit: what is this `logHelper` ? what is the benefit vs direct call to log API (except reducing readability) ?
(it is used in this file, but not everywhere; the name does not indicate what it does; and the parameter seems to give lot of flexibility...)

francoisferrand (approved):

delthas (approved):

lib/storage/data/external/AwsClient.js
maeldonn: `logHelper` (in `external/utils.js`) is already used by all the other error logs here. It normalizes the backend error fields (`dataStoreName`, `backendType`, extended request IDs). This change adds the location name to the setup error log to align it with the rest, per CLDSRV-927.

PR #2656, opened 2026-06-25
Add manual release CI workflow (release.yaml)

delthas:
## Summary

Arsenal has no release workflow, unlike `cloudserver`, `vault2`, and `backbeat`, which each ship a `workflow_dispatch` release pipeline. Today arsenal releases are cut by hand via `gh release create <version> --generate-notes`.

This adds `.github/workflows/release.yaml`, triggered manually (`workflow_dispatch`), that:

- Reads the version from `package.json` (no manual tag input), so the released tag can never drift from what consumers pin (`git+https://github.com/scality/arsenal#<tag>`).
- Fails loudly if a release for that version already exists, guarding against accidental re-runs (`softprops/action-gh-release` otherwise updates an existing release silently).
- Creates the GitHub release + tag with auto-generated notes (`generate_release_notes: true`), targeting the dispatched commit.

Arsenal is a library (`private: true`, no Docker image), so the workflow reduces to the GitHub-release step only — no Docker/oras steps. Build artifacts continue to be published by the existing per-push `compile` job in `tests.yaml`. The release name is kept as the bare version (e.g. `8.5.1`) to match arsenal's existing release history.

Issue: ARSN-604


.github/workflows/release.yaml
```diff
@@ -0,0 +1,41 @@
...
+run-name: release ${{ github.ref_name }}
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+jobs:
+  release:
```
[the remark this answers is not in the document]
delthas: Good catch. Addressed in 53291ddc with a fail-loud guard step: the workflow now exits with an error if dispatched from anything other than a `development/*` or `hotfix/*` branch. I used an explicit step (rather than a job-level `if`) so a wrong-branch dispatch fails visibly instead of skipping silently with a green check. Also allowed `hotfix/*` so a release can still be cut from a hotfix branch when needed.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,50 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Reject disallowed branch
+        if: >-
+          ${{ !startsWith(github.ref, 'refs/heads/development/')
+          && !startsWith(github.ref, 'refs/heads/hotfix/') }}
+        run: |
+          echo "::error::Releases must run from a development/* or" \
+            "hotfix/* branch (got ${{ github.ref_name }})"
```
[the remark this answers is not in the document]
delthas: Addressed: `github.ref_name` is now passed through a `REF_NAME` env var and referenced as `$REF_NAME` in the shell, so a branch name with shell metacharacters can no longer be interpolated into the `run` block. (Note: the branch was squashed to a single commit, so the fix is in the current HEAD rather than a follow-up commit.)

.github/workflows/release.yaml
```diff
@@ -0,0 +1,52 @@
...
+          echo "::error::Releases must run from a development/* or" \
+            "hotfix/* branch (got $REF_NAME)"
+          exit 1
+
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Read version from package.json
+        id: version
+        run: echo "version=$(jq -r .version package.json)" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
delthas: Added the guard: the step now fails with a clear error if `package.json` has no `.version` (or it is `null`) instead of cutting a release tagged `null`.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,52 @@
...
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
+          if gh release view "$VERSION" --repo "$GITHUB_REPOSITORY" \
+              >/dev/null 2>&1; then
+            echo "::error::Release $VERSION already exists"
+            exit 1
+          fi
+
+      - name: Create Release
+        uses: softprops/action-gh-release@v2
```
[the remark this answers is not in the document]
delthas: Keeping a floating major tag rather than SHA-pinning, to match the convention across arsenal's other workflows (`actions/checkout@v4`, `codecov/codecov-action@v4`, `scality/action-artifacts@v4`) and the sibling repos. Bumped from `@v2` to `@v3` so we track the current major (matches vault2); the only change in v3 is the runner moving from Node 20 to Node 24, which `ubuntu-latest` supports.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,58 @@
...
+            exit 1
+          fi
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
+      - name: Fail if release already exists
+        env:
+          GH_TOKEN: ${{ github.token }}
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
+          if gh release view "$VERSION" --repo "$GITHUB_REPOSITORY" \
```
francoisferrand, line 47: this only checks the _release_, not the _tag_.
if the release was previously aborted or removed, but the tag is still there, we could end up in a weird state: the tag was created -possibly on another commit- and we would now release a different commit (and possibly move the tag)

francoisferrand (approved):
As logic is added to release workflow, bugs are also added : which is esp. critical for releases, which are only used seldom and especially when we need (urgently) a release... Thus (non-trivial) release workflow should be tested, as we do for exemple in [Zenko](https://github.com/scality/Zenko/blob/development/2.15/tests/workflows/release.spec.ts)

maeldonn (commented):
Need a test

benzekrimaha (approved):

.github/workflows/release.yaml
delthas: Addressed in 43f7f537. Added a `Fail if tag already exists` guard (after a `git fetch --tags`) that aborts the release if `refs/tags/$VERSION` already exists, in addition to the existing release check — so a dangling tag left by an aborted/removed release can no longer be moved onto a different commit. Also added workflow tests under `tests/workflows` (act-js + mock-github) that cover this tag guard along with the branch and version guards, run in CI via the new `check-workflows` job.

delthas:
Added some tests, inspired by Zenko.

maeldonn (commented):
Do we need a separate package? Root already has jest/ts-jest/typescript, so
@kie/act-js and @kie/mock-github could just be root devDeps. Avoids a second
yarn.lock to maintain. What's the reasoning?

.github/workflows/check-workflows.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  test-workflows:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install NodeJS
+        uses: actions/setup-node@v4
```
francoisferrand:
```suggestion
        uses: actions/setup-node@v6
```

please check all actions, you are using outdated ones...

.github/workflows/check-workflows.yaml
```diff
@@ -0,0 +1,30 @@
+---
+name: check-workflows
+
+on:
+  push:
+    branches-ignore:
+      - 'development/**'
```
francoisferrand: there is also `q/*/**` to ignore : otherwise it will run twice on Bert-e's queue, since there are 2 branches : q/<version> and q/<version>/<branch>

.github/workflows/check-workflows.yaml
```diff
@@ -0,0 +1,30 @@
...
+    paths:
+      - '.github/workflows/**'
+      - 'tests/workflows/**'
```
francoisferrand: best to run on everything IMHO
- you never know if the workflow may depend on what's in the repo
- cheap anyway
- best to run periodically to find out before it breaks

.github/workflows/check-workflows.yaml
francoisferrand, line 1: do we need a separate workflow file?
best to keep in same workflow as other test/premerge, like in other repos

tests/workflows/release.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+import { exec as execCb } from "node:child_process";
+import { promisify } from "node:util";
+
+const exec = promisify(execCb);
+const IMAGE = "ghcr.io/catthehacker/ubuntu:act-latest";
+
+let github: MockGithub;
+let act: Act;
+
+async function setupRepo(versionFixture: string, branch: string) {
```
francoisferrand: seems to do more than setup "repo" :  it also setups `act`

tests/workflows/release.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+            },
+        },
+    });
+    await github.setup();
+    act = new Act(github.repo.getPath("arsenal"));
+    act.setWorkflowFile(".github/workflows/release.yaml");
+    act.setPlatforms("ubuntu-latest", IMAGE);
+    act.setEnv("GITHUB_REPOSITORY", "scality/Arsenal");
+    act.setEnv("GITHUB_REF", `refs/heads/${branch}`);
+}
```
francoisferrand, line 35: we need the GitHub variable for asynchronous cleanup in `afterEach` (unless we can register a cleanup step dynamically) ; but not the case for act : could we return it instead of relying on global variable?

.github/workflows/check-workflows.yaml
delthas: Agreed — merged into `tests.yaml` as a `test-workflows` job and removed `check-workflows.yaml`, matching the other repos (10087e38).

.github/workflows/check-workflows.yaml
delthas: Done — folded into `tests.yaml`, so it runs on every push like the other test jobs (dropped the `paths:` filter).

.github/workflows/check-workflows.yaml
delthas: Moot now — moved into `tests.yaml` as a job, so it follows the same trigger policy as the other test jobs rather than a separate `branches-ignore`. If we want `q/*/**` excluded, that'd be a repo-wide change to `tests.yaml` affecting all its jobs — happy to do it separately if you'd like.

.github/workflows/check-workflows.yaml
delthas: Bumped in 10087e38 — the `test-workflows` steps use `actions/checkout@v6` + `actions/setup-node@v6`, and `release.yaml`'s checkout is now `@v6` too.

tests/workflows/release.spec.ts
delthas: Renamed to `setup()` in 10087e38 — it sets up the mock repo and returns the configured `act`, so the name no longer implies repo-only.

tests/workflows/release.spec.ts
delthas: Done — `setup()` now returns `act`; only `github` stays module-scoped, since the async `afterEach` teardown needs it.

delthas:
On whether the workflow tests need a separate `tests/workflows` package vs. adding `@kie/act-js`/`@kie/mock-github` to the root devDeps — it's deliberate:

1. **Node version.** `@kie/mock-github@2.0.2` declares `engines: node >=24`, but arsenal's root builds/tests on node 22 — a root `yarn install` on node 22 fails its engine check. The separate package runs its own CI job on node 24 while the rest of arsenal stays on 22, without forcing the whole library onto node 24.

2. **Toolchain divergence.** The harness rides the latest tooling to match act-js/mock-github (TypeScript 6, jest 30, `@types/node` 24); arsenal's root is on TypeScript 5.6 / jest 29 / node-22 types. Sharing would pull arsenal's own `tsc` build and unit/coverage suite onto TS 6 + jest 30 — a big toolchain bump driven by a test-only harness.

3. **Isolation.** act-js brings a large, CI-only tree (the `act` binary download, `tar`, express, nock, …). Keeping it in `tests/workflows` keeps arsenal's own manifest/lockfile focused on what arsenal actually depends on, and out of arsenal's dependency-review surface. arsenal's `coverage` also runs `jest tests` (all of `tests/`), so a shared jest would sweep the act-js *container* suite into the unit/coverage run — a separate jest is a clean boundary.

4. **Consistency.** Zenko and vault2 use this exact layout — a `tests/workflows` package with its own lockfile, run via `yarn --cwd tests/workflows` on node 24.

The extra `yarn.lock` is a fair cost, but #1 in particular is effectively a blocker for the root approach. Happy to revisit if arsenal's root moves to node 24 + matching tooling.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,71 @@
...
+        run: |
+          echo "::error::Releases must run from a development/* or" \
+            "hotfix/* branch (got $REF_NAME)"
+          exit 1
+
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Fetch tags
+        run: git fetch --tags --force --quiet
```
francoisferrand:
```suggestion
      with:
        fetch-depth: 0
        fetch-tags: true
```

francoisferrand (approved):

.github/workflows/release.yaml
delthas: Applied in 33223405 — `checkout` now uses `fetch-depth: 0` + `fetch-tags: true` and the separate `Fetch tags` step is gone.

One thing worth flagging: with `fetch-depth: 0`, act's real `checkout` mutates the mock-github repo, which intermittently collided with the next test's `setup()` (`remote origin already exists`, ~50% flaky) because mock-github shares a single repo path per run. Fixed by giving each test a unique `setupPath` (MockGithub's 2nd constructor arg) — the suite is now stable across repeated runs.

PR #2657, opened 2026-06-29
Publish arsenal to GitHub Packages and npm on release

delthas:
## Summary

Builds on the manual release workflow from ARSN-604 (#2656): on release, **build and dual-publish `@scality/arsenal` to GitHub Packages and npmjs**. arsenal is a public, Apache-2.0 library, so per the DevEx TAD it goes to both registries. Non-breaking — existing git-dependency consumers (cloudserver, backbeat, vault2) are untouched; migrating them to the registry is a later, organic step.

`release.yaml` is split into jobs so the build is produced once and each publish can be retried independently of it and of the other:

- **build** — after the branch / version / tag guards: install, `yarn build`, `npm pack`, generate a signed build-provenance attestation for the tarball (`actions/attest-build-provenance`), and upload the `.tgz` as the artifact.
- **publish-github** — download the tarball and publish it to GitHub Packages with the built-in `GITHUB_TOKEN`.
- **publish-npm** — publish the same tarball to npmjs via **OIDC trusted publishing** (`id-token: write`, no stored token) with `--provenance`.
- **release** — create the GitHub release + tag, linking to both packages.

Both publish jobs use the committed `package.json` version (no auto-bump), check the registry with `npm view` first and **skip if that version is already published** — so a partial dual-publish can be re-run cleanly — and publish under a `v<major.minor>` dist-tag so an older release line never moves `latest`. Building and packing once, then attesting and publishing the same tarball to both registries, gives a single GitHub attestation (`gh attestation verify <tarball> --repo scality/Arsenal`) covering whatever lands in either registry, plus npm-native provenance on npmjs.

### Name: committed, not stamped

`package.json` is renamed to **`@scality/arsenal`** directly — no publish-time name stamp / `jq`. Git-dependency consumers resolve arsenal by their dependency **key**, not the package's `name`, so committing the scoped name keeps `require('arsenal')` working for cloudserver/backbeat/vault2 (verified end-to-end: real arsenal committed as `@scality/arsenal`, consumed under the unchanged bare `arsenal` git key, installs to `node_modules/arsenal` and resolves both bare and deep `arsenal/build/...` imports). `package.json` also drops `private` and adds `types`/`files` + `publishConfig.access: public`. The version stays committed / manually bumped — auto-stamping is deferred to a later step.

The workflow tests under `tests/workflows` stub the publish path, so the guard coverage is unchanged.

### Prerequisite (npmjs trusted publishing)

The npm publish uses **OIDC trusted publishing** — no `NPM_TOKEN`. Before the first npm release, a **trusted publisher** must be registered for `@scality/arsenal`: GitHub org `scality`, repo `Arsenal`, workflow `release.yaml` (the same setup Sylvain did for `@scality/cloudserverclient`). Since `@scality/arsenal` isn't on npm yet, bootstrap it one of two ways:

- **One-time manual publish** by anyone with `@scality` publish rights (`yarn build && npm publish --access public`) to create the package, then register the trusted publisher. That first version has no provenance (local publish); the workflow provenance-signs every release after it and skips the bootstrapped version via its `npm view` check.
- **Org-level trusted publishing**, if the `@scality` npm org allows creating packages via OIDC — then the first workflow run creates it, no manual step.

GitHub Packages needs no bootstrap (built-in `GITHUB_TOKEN`).

Aligned with the DevEx initiative (S3C-9725); mirrors the dual-publish model of `hdclient` / `cloudserverclient`.

Issue: ARSN-605

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -4,6 +4,8 @@ run-name: release ${{ github.ref_name }}
 
 on:
   workflow_dispatch:
+  release:
+    types: [published]
 
 permissions: {}
 
@@ -19,7 +21,8 @@ jobs:
     steps:
       - name: Reject disallowed branch
         if: >-
-          ${{ !startsWith(github.ref, 'refs/heads/development/')
+          ${{ github.event_name == 'workflow_dispatch'
+          && !startsWith(github.ref, 'refs/heads/development/')
           && !startsWith(github.ref, 'refs/heads/hotfix/') }}
         env:
           REF_NAME: ${{ github.ref_name }}
@@ -44,7 +47,21 @@ jobs:
           fi
           echo "version=$version" >> "$GITHUB_OUTPUT"
 
+      - name: Verify tag matches package.json version
+        if: ${{ github.event_name == 'release' }}
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
+          TAG: ${{ github.event.release.tag_name }}
+        run: |
+          tag=${TAG#v}
+          if [ "$tag" != "$VERSION" ]; then
+            echo "::error::Release tag $TAG does not match package.json" \
+              "version $VERSION"
+            exit 1
+          fi
+
       - name: Fail if release already exists
+        if: ${{ github.event_name == 'workflow_dispatch' }}
         env:
           GH_TOKEN: ${{ github.token }}
           VERSION: ${{ steps.version.outputs.version }}
@@ -56,6 +73,7 @@ jobs:
           fi
 
       - name: Fail if tag already exists
+        if: ${{ github.event_name == 'workflow_dispatch' }}
         env:
           VERSION: ${{ steps.version.outputs.version }}
         run: |
@@ -154,6 +172,7 @@ jobs:
   release:
     needs: [build, publish-github, publish-npm]
     runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'workflow_dispatch' }}
     permissions:
       contents: write
     steps:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -5,12 +5,17 @@ run-name: release ${{ github.ref_name }}
 on:
   workflow_dispatch:
 
-permissions:
-  contents: write
+permissions: {}
 
 jobs:
-  release:
+  build:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+      attestations: write
+    outputs:
+      version: ${{ steps.version.outputs.version }}
     steps:
       - name: Reject disallowed branch
         if: >-
@@ -60,12 +65,108 @@ jobs:
             exit 1
           fi
 
+      - name: Install NodeJS
+        uses: actions/setup-node@v6
+        with:
+          node-version: '24'
+          cache: yarn
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile --prefer-offline
+
+      - name: Build
+        run: yarn build
+
+      - name: Pack tarball
+        run: npm pack --ignore-scripts
+
+      - name: Attest build provenance
+        uses: actions/attest-build-provenance@v4
+        with:
+          subject-path: '*.tgz'
+
+      - name: Upload package artifact
+        uses: actions/upload-artifact@v7
+        with:
+          name: package
+          path: '*.tgz'
+          if-no-files-found: error
+          retention-days: 7
+
+  publish-github:
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+    steps:
+      - name: Download package artifact
+        uses: actions/download-artifact@v8
+        with:
+          name: package
+
+      - name: Install NodeJS
+        uses: actions/setup-node@v6
+        with:
+          node-version: '24'
+          registry-url: 'https://npm.pkg.github.com'
+          scope: '@scality'
+
+      - name: Publish to GitHub Packages
+        env:
+          NODE_AUTH_TOKEN: ${{ github.token }}
+          VERSION: ${{ needs.build.outputs.version }}
+        run: |
+          if [ -n "$(npm view "@scality/arsenal@$VERSION" version 2>/dev/null)" ]; then
+            echo "::notice::@scality/arsenal@$VERSION already on GitHub Packages; skipping"
+          else
+            npm publish *.tgz --tag "v${VERSION%.*}"
+          fi
+
+  publish-npm:
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - name: Download package artifact
+        uses: actions/download-artifact@v8
+        with:
+          name: package
+
+      - name: Install NodeJS
+        uses: actions/setup-node@v6
+        with:
+          node-version: '24'
+          registry-url: 'https://registry.npmjs.org'
+          scope: '@scality'
+
+      - name: Publish to npm
+        env:
+          VERSION: ${{ needs.build.outputs.version }}
+        run: |
+          if [ -n "$(npm view "@scality/arsenal@$VERSION" version 2>/dev/null)" ]; then
+            echo "::notice::@scality/arsenal@$VERSION already on npm; skipping"
+          else
+            npm publish *.tgz --provenance --tag "v${VERSION%.*}"
+          fi
+
+  release:
+    needs: [build, publish-github, publish-npm]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
       - name: Create Release
         uses: softprops/action-gh-release@v3
         env:
           GITHUB_TOKEN: ${{ github.token }}
         with:
-          name: ${{ steps.version.outputs.version }}
-          tag_name: ${{ steps.version.outputs.version }}
+          name: ${{ needs.build.outputs.version }}
+          tag_name: ${{ needs.build.outputs.version }}
           generate_release_notes: true
+          append_body: true
+          body: |
+            GitHub Packages: https://github.com/${{ github.repository }}/pkgs/npm/arsenal
+            npm: https://www.npmjs.com/package/@scality/arsenal
           target_commitish: ${{ github.sha }}
--- a/tests/workflows/release.spec.ts
+++ b/tests/workflows/release.spec.ts
@@ -1,52 +1,55 @@
-import { Act } from "@kie/act-js";
-import { MockGithub } from "@kie/mock-github";
-import path from "path";
-import { exec as execCb } from "node:child_process";
-import { promisify } from "node:util";
-import { mkdtempSync } from "node:fs";
-import os from "node:os";
+import { Act } from '@kie/act-js';
+import { MockGithub } from '@kie/mock-github';
+import path from 'path';
+import { exec as execCb } from 'node:child_process';
+import { promisify } from 'node:util';
+import { mkdtempSync } from 'node:fs';
+import os from 'node:os';
 
 const exec = promisify(execCb);
-const IMAGE = "ghcr.io/catthehacker/ubuntu:act-latest";
+const IMAGE = 'ghcr.io/catthehacker/ubuntu:act-latest';
 
 let github: MockGithub;
 
 async function setup(versionFixture: string, branch: string): Promise<Act> {
     // unique setup path per test so act's checkout residue can't collide with another test's repo
-    const setupPath = mkdtempSync(path.join(os.tmpdir(), "arsenal-workflows-"));
-    github = new MockGithub({
-        repo: {
-            arsenal: {
-                currentBranch: branch,
-                files: [
-                    { src: path.resolve(__dirname, "../..", ".github"), dest: ".github" },
-                    { src: path.resolve(__dirname, "fixtures", versionFixture), dest: "package.json" },
-                ],
+    const setupPath = mkdtempSync(path.join(os.tmpdir(), 'arsenal-workflows-'));
+    github = new MockGithub(
+        {
+            repo: {
+                arsenal: {
+                    currentBranch: branch,
+                    files: [
+                        { src: path.resolve(__dirname, '../..', '.github'), dest: '.github' },
+                        { src: path.resolve(__dirname, 'fixtures', versionFixture), dest: 'package.json' },
+                    ],
+                },
             },
         },
-    }, setupPath);
+        setupPath,
+    );
     await github.setup();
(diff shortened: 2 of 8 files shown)


.github/workflows/release.yaml
```diff
@@ -60,6 +61,38 @@ jobs:
...
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile --prefer-offline
+
+      - name: Stamp package name and files
+        run: |
+          tmp=$(mktemp)
+          jq '.name = "@scality/arsenal" | .files = ["build"] | del(.private)' \
+            package.json > "$tmp"
+          mv "$tmp" package.json
```
francoisferrand: why do we need `jq`? Isn't there a way to write/commit the package.json "correctly" so we don't need this?

.github/workflows/release.yaml
```diff
@@ -60,6 +61,38 @@ jobs:
...
+
+      - name: Build
+        run: yarn build
+
+      - name: Publish to GitHub Packages
```
francoisferrand, line 132: c.f. earlier discussions (TAD and pr in guidelines) : the conclusion was that we should always push to Github registry (public & private packages) + publish public packages to NPM.
→ should this be implemented already?

(to do that, should upload artifact after build and upload in separate jobs, so each one uploads the exact same build AND can be retried independently in case of infra failure)

francoisferrand, line 132: should be in separate Job, to allow retrying upload on infra failure

francoisferrand: esp. I think GitHub npm registry may require some auth and/or configuration to be used, so slightly less seamless...

also AFAIU we can't "create" package on npm, we need platform-eng to do it, setup creds, ... as Sylvain did for cloudserverclient.

.github/workflows/release.yaml
```diff
@@ -60,6 +61,38 @@ jobs:
...
+        env:
+          NODE_AUTH_TOKEN: ${{ github.token }}
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
```
francoisferrand, line 136: (putting this comment in a random place)

there has been quite some noise around package signature/..., to ensure packages are not forged and used to injects backdoors/... ; and GitHub has added signing infrastructure to ease this: is it used automatically? can you verify what is means/requires?

francoisferrand: maybe this is the missing `--provenance` flag?

.github/workflows/release.yaml
```diff
@@ -60,6 +61,38 @@ jobs:
...
+          if [ -n "$(npm view "@scality/arsenal@$VERSION" version 2>/dev/null)" ]; then
+            echo "::notice::@scality/arsenal@$VERSION already published; skipping"
+          else
+            npm publish --ignore-scripts
```
francoisferrand: why do we ignore scripts? is this the standard way, or should we "fix" package.json so we don't need such trick?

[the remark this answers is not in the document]
delthas: Added `--tag "v${VERSION%.*}"` on both publishes, so a release from an older line (e.g. `development/8.3` after `8.4`) publishes under its own `v8.3` dist-tag rather than moving `latest` backward.

.github/workflows/release.yaml
delthas: The only field that can't be committed is `name`: consumers use arsenal as a git dependency via `require('arsenal')` (resolving `arsenal/build/...`), so committing `@scality/arsenal` would relocate it to `node_modules/@scality/arsenal` and break them until they migrate to the scoped package (a later step). With `files`/`types` committed and `private` dropped (pointless for a public package — cloudserverclient doesn't set it either), the stamp is now a single line: the `name`. Once consumers migrate we commit the scoped name and drop the `jq` entirely.

.github/workflows/release.yaml
delthas: You're right — arsenal is a public (Apache-2.0) library, so I've implemented the full dual-publish per the TAD: it now publishes to **both** GitHub Packages and npmjs. Restructured to build-once → artifact → separate publish jobs (`publish-github`, `publish-npm`), each consuming the same artifact and independently retriable, plus a `release` job.

`NPM_TOKEN` is already an org secret; the npm job uses an `npmjs` environment — repo admins should add protection rules to it (mirroring hdclient) before the first npm release. No npm package pre-creation needed: the org token creates `@scality/arsenal` on first publish with `access: public`.

Will ask plaform-eng to add it before merging.

delthas: Done — split into `publish-github` and `publish-npm` jobs (both `needs: build`), each consuming the uploaded artifact, so an infra failure on either can be retried without rebuilding, independently of the other. The `npm view` idempotency check makes a partial dual-publish safe to re-run (the registry that already has the version is skipped) — which also covers the E409 atomicity gotcha you raised on the Guidelines pilot.

.github/workflows/release.yaml
delthas: Good call — now that we publish to npmjs, npm's `--provenance` applies, so I've added it (`npm publish --provenance` + `id-token: write` on the npm job) for a signed Sigstore/transparency-log attestation. GitHub Packages doesn't consume npm provenance; if we want GH-side attestation too, `actions/attest-build-provenance` is the native equivalent — easy to add on the same artifact later.

.github/workflows/release.yaml
delthas: The publish jobs ship the artifact built in `build` and install no dependencies, so npm's lifecycle scripts must not run — notably `prepare: yarn build`, which would need the absent dev deps and rebuild instead of shipping the exact artifact. `--ignore-scripts` keeps publish decoupled from build, not a `package.json` workaround.

.github/workflows/release.yaml
delthas: Update: implemented the GitHub-side attestation now rather than deferring. `build` packs the tarball and runs `actions/attest-build-provenance` on it (`id-token: write` + `attestations: write`); both publish jobs then publish that *same* tarball, so one signed SLSA build-provenance attestation covers the exact bytes in both registries — verifiable with `gh attestation verify <tarball> --repo scality/Arsenal` — on top of npm-native `--provenance` on the npmjs side.

.github/workflows/release.yaml
delthas: Small update after the restructure: `build` now packs the tarball (`npm pack --ignore-scripts`, where `prepare: yarn build` would otherwise fire) and both jobs publish that prebuilt `*.tgz`, so `--ignore-scripts` moved onto `npm pack` — the `npm publish *.tgz` calls no longer need it, since publishing a prebuilt tarball doesn't run the pack lifecycle.

.github/workflows/release.yaml
delthas: Follow-up — turns out we don't need `jq` at all, so I've dropped the stamp entirely: `package.json` now commits `name: @scality/arsenal` directly.

The premise I gave above (that committing the scoped name would break `require('arsenal')`) is actually wrong for git deps: **yarn and npm resolve a git dependency by the dependency _key_ the consumer declares, not by the package's own `name` field.** So a consumer keeping `"arsenal": "git+…"` still gets `node_modules/arsenal` and `require('arsenal')` regardless of what we commit as the name.

Verified end-to-end: real arsenal committed as `@scality/arsenal`, consumed under the unchanged bare `arsenal` git key (yarn v1 **and** npm, with a lockfile + the `prepare` build hook) → installs to `node_modules/arsenal`, no name-mismatch error, and both `require('arsenal')` and deep `arsenal/build/lib/...` imports resolve. Checked cloudserver/backbeat/vault2 — all their arsenal imports are `arsenal` or `arsenal/build/*`, covered by `files: ["build"]`.

So: scoped name committed, no publish-time stamp, no `jq`. Consumers can rename their dependency key to `@scality/arsenal` whenever (cosmetic until the eventual registry move); nothing forces a flag-day.

.github/workflows/release.yaml
delthas: Follow-up on auth: switched the npm publish to **OIDC trusted publishing** — matching what Sylvain set up for cloudserverclient — so there's now **no `NPM_TOKEN` and no `npmjs` environment**. The `publish-npm` job just carries `id-token: write`; `npm publish` exchanges the GitHub OIDC token for a short-lived, package-scoped credential (and provenance comes for free). Bumped the release jobs to node 24 so npm is new enough for trusted publishing.

Updated prerequisite: an `@scality` npm-org admin registers a trusted publisher for `@scality/arsenal` — GitHub org `scality`, repo `Arsenal`, workflow `release.yaml`. Since the package isn't on npm yet, that's either org-level trusted-publishing config or a one-time bootstrap publish to create it first — same situation as cloudserverclient.

.github/workflows/release.yaml
```diff
@@ -60,12 +65,108 @@ jobs:
...
+        run: |
+          if [ -n "$(npm view "@scality/arsenal@$VERSION" version 2>/dev/null)" ]; then
+            echo "::notice::@scality/arsenal@$VERSION already on GitHub Packages; skipping"
+          else
+            npm publish *.tgz --tag "v${VERSION%.*}"
+          fi
+
+  publish-npm:
+    needs: build
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
delthas: Intentional — this job uses **OIDC trusted publishing** (`id-token: write`), not a token, so there's no `environment: npmjs` and no `NPM_TOKEN`. `npm publish` exchanges the GitHub OIDC token for a short-lived, package-scoped npm credential; there's no environment secret to gate. This matches cloudserverclient's `publish-npm`. (See the auth follow-up on François's thread.)

.github/workflows/release.yaml
```diff
@@ -60,12 +65,108 @@ jobs:
...
+      - name: Install NodeJS
+        uses: actions/setup-node@v6
+        with:
+          node-version: '24'
+          registry-url: 'https://registry.npmjs.org'
+          scope: '@scality'
+
+      - name: Publish to npm
+        env:
+          VERSION: ${{ needs.build.outputs.version }}
```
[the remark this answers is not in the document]
delthas: Intentional — OIDC trusted publishing, so no `NODE_AUTH_TOKEN`. `npm publish` authenticates via the OIDC token exchange (not the `.npmrc` `_authToken`), so it won't 401. The `npm view` idempotency check is an **anonymous read of the public registry** — verified with an empty token it returns the version for an existing package and a 404 (not 401) for a missing one, which is exactly what the guard expects (404 → empty → publish; exists → skip). cloudserverclient publishes to npm the same tokenless way.

tests/workflows/release.spec.ts
francoisferrand, line 1: these tests should be executed in CI (pre-merge)

delthas: They already do — the **Test GitHub Actions workflows** job (added in ARSN-604, in `tests.yaml`) runs `tests/workflows` on every push to the branch, so it gates pre-merge. It ran on this PR and passed 5/5: https://github.com/scality/Arsenal/actions/runs/28672537753/job/85038858308

maeldonn (commented):
LGTM but we need to get team alignment on this first

.github/workflows/release.yaml
```diff
@@ -5,12 +5,17 @@ run-name: release ${{ github.ref_name }}
...
 
-permissions:
-  contents: write
+permissions: {}
 
 jobs:
-  release:
-    runs-on: ubuntu-latest
+  build:
+    runs-on: ubuntu-22.04
```
francoisferrand: ubuntu-22 is quite old, should we not use ubuntu-24.04 instead (or even ubuntu-26.04) ?

francoisferrand: (btw, for runners maybe latest could be appropriate, esp. in jobs where we control all deps otherwise: i.e. either actions or tools manually installed)

francoisferrand (approved):

.github/workflows/release.yaml
delthas: Good call — switched all four `release.yaml` jobs to `ubuntu-latest`. Agreed the runner OS needn't be pinned here since these jobs control their own toolchain (node via `setup-node`, actions pinned).

I also normalized the rest of the repo's workflows (`codeql`, `dependency-review`, `lint`, `tests`, `tests-kmip`) from `ubuntu-22.04` → `ubuntu-latest` in a separate commit (**Use ubuntu-latest for CI runners**), preceded by a prettier commit for the two files that were prettier-dirty. Left `ubuntu-26.04` for now since I haven't confirmed it's an available runner label yet.

delthas:
(Unrelated kmip failure due to https://github.com/scality/Arsenal/pull/2668 ; will rebase once merged)

.github/workflows/release.yaml
```diff
@@ -5,12 +5,17 @@ run-name: release ${{ github.ref_name }}
 on:
   workflow_dispatch:
```
tcarmet, line 6: In my experience when working with making release in our repos I believe we tend to use GitHub Releases for libraries and github actions workflow dispatch for components. If you look at the latest arsenal releases itself, it has been inconsistent: https://github.com/scality/Arsenal/releases

I see two options to suggest, my only request is that if we use GitHub Releases we should be good. Overall I want to ensure we don't have to be like, "hey be careful now let's not forget to trigger the workflow"

So option 1 we keep both available, I believe the step that makes the github release will not be affected if a release has already been made and should work as expected.
```suggestion
  workflow_dispatch: {}
  release:
    types: [published]
```

Or we remove the dispatch, might be easier to handle:

```suggestion
  release:
    types: [published]
```

tcarmet (commented):
Looking good! Happy to see this change. I have one request on it which I made it clear, otherwise LGTM.

.github/workflows/release.yaml
delthas: I decided to go with option 1. In the future I'd ideally like to drop bump commits, and generate releases one-click with a release workflow dispatch. (But this is for later and I suppose will be up for larger discussion.) So I'd like to support manual dispatches from today already. (Also this would be more consistent for a new developer to have every release be done through a manual workflow instead of an exception for Arsenal where you have to create a GitHub release manually.)

Adapted suggestion 1 (to prevent failing on an existing tag on... a tag-triggered workflow) and added tests.

tcarmet: Sounds good! Yeah I'd love to drop the bump commits, I have some thoughts in the area so if you'd like to chat with a human about it :joy: feel free to do so :)

delthas:
@maeldonn Rerequesting review (after François + Thomas ACK)

maeldonn (approved):
LGTM

PR #2662, opened 2026-07-02
Skip head-check for scality backends in MultipleBackendGateway

delthas:
## Problem

`MultipleBackendGateway.head()` feature-detects the location client's `head()` method and then passes it the whole `objectGetInfo` **object**. sproxydclient 8.2.0+ added a `head(key)` method that `assert`s a 40-char string key. Once arsenal bumped to sproxydclient 8.2.1 (ARSN-606), the head-check performed during object GET began crashing the CloudServer worker with an uncaught `AssertionError: 'object' !== 'string'` on multiple-backend sproxyd locations (`S3DATA=scality`), taking down object reads (reproduced on MPU / partNumber GETs).

Before 8.2.0 sproxydclient had no `head()`, so the feature-detect skipped it for sproxyd and this path stayed dormant. The buggy caller dates back to ARSN-128.

Affected released versions: `8.4.14` (sproxydclient 8.2.0), `8.4.15` (8.2.1), and the unreleased `development/8.5` tip. Targeting `development/8.4` so the fix waterfalls up to 8.5.

## Approaches considered (discussion)

The head-check exists only to verify **external cloud backend** data state — arsenal's own comment: `// head is used during get just to check external backend data state`. sproxyd was never head-checked before 8.2.0, and object reads worked fine; the crash comes from that check being *accidentally* activated for sproxyd the moment sproxydclient grew a `head()`.

Three ways to fix it:

| | Change | Effect | Trade-off |
|---|---|---|---|
| **(A) — this PR** | Skip `scality` clients in `MultipleBackendGateway.head()` | Restores pre-8.2.0 behavior: sproxyd is not head-checked | Minimal, behavior-preserving, no new round-trip; mirrors the `clientType === 'scality'` handling already in `get()`/`delete()` |
| (B) | Extract `objectGetInfo.key` in `head()` and call sproxyd's `head` with the string | Sends a real HEAD to sproxyd on **every** GET | Introduces a per-read round-trip that never existed; sproxyd doesn't need this existence check |
| (C) | Make `sproxydclient.head()` accept an `objectGetInfo` object | Same as (B) | Same extra round-trip, and pushes caller-shaped logic into a low-level, string-keyed API (its `get`/`put`/`delete` are all string-keyed) |

**This PR implements (A)** — the minimal regression fix. (B) and (C) make the accidental call *succeed* rather than not making it, at the cost of a HEAD-per-GET to sproxyd that never existed. Flagging for discussion: happy to switch to (B)/(C) if there's a reason sproxyd *should* be HEAD-checked before reads, or if reviewers prefer consistency with `get()`/`delete()`.

Note: sproxydclient's `head()` (SPRXCLT-21) stays available for a deliberate future use; this PR only stops the GET path from invoking it by accident.

## Fix

Skip `scality` clients in `MultipleBackendGateway.head()` (option A above). Adds `head()` to the sproxyd/azure mock clients and two regression tests: sproxyd `head` is skipped (guards the crash), external azure `head` is still invoked.

Issue: ARSN-607

### DIFF
--- a/lib/storage/data/MultipleBackendGateway.js
+++ b/lib/storage/data/MultipleBackendGateway.js
@@ -92,8 +92,11 @@ class MultipleBackendGateway {
         }
         const objectGetInfo = objectGetInfoArr[0];
         const client = this.clients[objectGetInfo.dataStoreName];
-        if (client.head === undefined) {
-            // no-op if unsupported client method
+        if (client.head === undefined || client.clientType === 'scality') {
+            // no-op if the client does not support head, or for scality
+            // (sproxyd) backends: the head-check only verifies external cloud
+            // backend data state, and sproxyd's head expects a string key
+            // rather than the objectGetInfo object (see get()/delete()).
             return process.nextTick(callback);
         }
         return client.head(objectGetInfo, reqUids, callback);
--- a/lib/storage/data/MultipleBackendGateway.js
+++ b/lib/storage/data/MultipleBackendGateway.js
@@ -24,13 +24,11 @@ class MultipleBackendGateway {
     }
 
     put(hashedStream, size, keyContext, backendInfo, reqUids, callback) {
-        const controllingLocationConstraint =
-            backendInfo.getControllingLocationConstraint();
+        const controllingLocationConstraint = backendInfo.getControllingLocationConstraint();
         const client = this.clients[controllingLocationConstraint];
         if (!client) {
             const log = createLogger(reqUids);
-            log.error('no data backend matching controlling locationConstraint',
-                { controllingLocationConstraint });
+            log.error('no data backend matching controlling locationConstraint', { controllingLocationConstraint });
             return process.nextTick(() => {
                 callback(errors.InternalError);
             });
@@ -53,13 +51,16 @@ class MultipleBackendGateway {
                 return callback(errors.InternalError);
             }
         }
-        return client.put(writeStream, size, keyContext, reqUids,
+        return client.put(
+            writeStream,
+            size,
+            keyContext,
+            reqUids,
             (err, key, dataStoreVersionId, dataStoreSize, dataStoreMD5) => {
                 const log = createLogger(reqUids);
                 log.debug('put to location', { controllingLocationConstraint });
                 if (err) {
-                    log.error('error from datastore',
-                        { error: err, dataStoreType: client.clientType });
+                    log.error('error from datastore', { error: err, dataStoreType: client.clientType });
                     return callback(errors.ServiceUnavailable);
                 }
                 const dataRetrievalInfo = {
@@ -71,14 +72,21 @@ class MultipleBackendGateway {
                     dataStoreMD5,
                 };
                 return callback(null, dataRetrievalInfo);
-            // sproxyd accepts keyschema, send as null so sproxyd generates key
-            // send metadata as param for AzureClient in Arsenal
-            }, null, this.metadata);
+                // sproxyd accepts keyschema, send as null so sproxyd generates key
+                // send metadata as param for AzureClient in Arsenal
+            },
+            null,
+            this.metadata,
+        );
     }
 
     head(objectGetInfoArr, reqUids, callback) {
-        if (!objectGetInfoArr || !Array.isArray(objectGetInfoArr)
-        || !objectGetInfoArr[0] || !objectGetInfoArr[0].dataStoreName) {
+        if (
+            !objectGetInfoArr ||
+            !Array.isArray(objectGetInfoArr) ||
+            !objectGetInfoArr[0] ||
+            !objectGetInfoArr[0].dataStoreName
+        ) {
             // no-op if no stored data store name
             return process.nextTick(callback);
         }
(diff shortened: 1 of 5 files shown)


lib/storage/data/MultipleBackendGateway.js
```diff
@@ -208,41 +262,58 @@ class MultipleBackendGateway {
...
-                                        log.error('Error decrementing location ' +
-                                    'metric following object PUT failure',
-                                        { error: error.message });
-                                    }
-                                    return cbOnce(err);
-                                });
+                            // if error putting part, counter should be decremented
+                            return this.locStorageCheckFn(location, -size, log, error => {
+                                if (error) {
+                                    log.error('Error decrementing location ' + 'metric following object PUT failure', {
```
BourgoisMickael:
```suggestion
                                    log.error('Error decrementing location metric following object PUT failure', {
```

lib/storage/data/MultipleBackendGateway.js
```diff
@@ -285,73 +356,89 @@ class MultipleBackendGateway {
...
+                        if (err) {
+                            // if error copying obj, counter should be decremented
+                            return this.locStorageCheckFn(
+                                destLocationConstraintName,
+                                -storeMetadataParams.size,
+                                log,
+                                error => {
+                                    if (error) {
+                                        log.error(
+                                            'Error decrementing location ' + 'metric following object PUT failure',
```
BourgoisMickael:
```suggestion
                                            'Error decrementing location metric following object PUT failure',
```

fredmnl (approved):

lib/storage/data/MultipleBackendGateway.js
```diff
@@ -285,73 +356,89 @@ class MultipleBackendGateway {
...
+                    },
+                );
+            });
         }
-        return cb(errorInstances.NotImplemented
-            .customizeDescription('Can not copy object from ' +
-            `${client.clientType} to ${client.clientType}`));
+        return cb(
+            errorInstances.NotImplemented.customizeDescription(
+                'Can not copy object from ' + `${client.clientType} to ${client.clientType}`,
```
BourgoisMickael:
```suggestion
                `Can not copy object from '${client.clientType} to ${client.clientType}`,
```

lib/storage/data/MultipleBackendGateway.js
```diff
@@ -285,73 +356,89 @@ class MultipleBackendGateway {
...
-                sourceLocationConstraintName, config,
-                log, cb);
+            return client.uploadPartCopy(request, awsSourceKey, sourceLocationConstraintName, config, log, cb);
         }
-        return cb(errorInstances.NotImplemented.customizeDescription(
-            'Can not copy object from ' +
-          `${client.clientType} to ${client.clientType}`));
+        return cb(
+            errorInstances.NotImplemented.customizeDescription(
+                'Can not copy object from ' + `${client.clientType} to ${client.clientType}`,
```
BourgoisMickael:
```suggestion
                `Can not copy object from ${client.clientType} to ${client.clientType}`,
```

tests/unit/storage/data/wrapperClientRoutes.spec.js
```diff
@@ -107,29 +102,38 @@ describe('Routes from DataWrapper to backend client', () => {
...
+    });
+
+    // ARSN-607: the head-check only verifies external cloud backend data
+    // state. sproxydclient's head() expects a 40-char string key, but the
+    // gateway holds the objectGetInfo object; it must skip scality backends
+    // rather than pass them the object (which crashed object GET).
+    it('should not send a head request to a sproxyd (scality) backend', done => {
+        const objectGetInfo = genObjGetInfo('sproxyd');
+        dw.head([objectGetInfo], log, err => {
+            assert.ifError(err);
```
BourgoisMickael, line 140: this should add a spy on the mock sproxyd head to verify it's actually not called (instead of relying on the fact that it fails when passing an object)

BourgoisMickael: Btw I think the other unit tests are not good as well, they should all spy and assert the client method has been called

BourgoisMickael (approved):

lib/storage/data/MultipleBackendGateway.js
delthas: Done — collapsed into the prettier commit (600f7dd6).

lib/storage/data/MultipleBackendGateway.js
delthas: Done — collapsed into the prettier commit (600f7dd6).

lib/storage/data/MultipleBackendGateway.js
delthas: Done — collapsed into the prettier commit (600f7dd6). Applied the clean form (dropped the stray `'`), matching your other suggestion.

lib/storage/data/MultipleBackendGateway.js
delthas: Done — collapsed into the prettier commit (600f7dd6).

tests/unit/storage/data/wrapperClientRoutes.spec.js
delthas: Done in d4b27764 — added a `jest.spyOn` on the sproxyd `head` asserting it is **not** called (and on the azure `head` asserting it **is**). Also took your broader point: all the routing tests in this file now spy on the client method and assert it was called.

lib/storage/data/MultipleBackendGateway.js
```diff
@@ -285,73 +356,88 @@ class MultipleBackendGateway {
...
-                                        }
-                                        return cb(err);
-                                    });
-                            }
-                            return cb(null, dataRetrievalInfo);
-                        });
-                });
+            return this.locStorageCheckFn(destLocationConstraintName, storeMetadataParams.size, log, err => {
+                if (err) {
+                    cb(err);
```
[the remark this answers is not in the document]
delthas: Unrelated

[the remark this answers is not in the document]
delthas: Unrelated

PR #2665, opened 2026-07-06
Improvement/arsn 596 namespace exists createbucket

benzekrimaha:
### **Context** 
`MongoClientInterface.createBucket()` lazily creates the MongoDB collection backing a bucket via `db.createCollection(bucketName)`. When concurrent create/drop/create sequences occur on the same bucket (typically MPU shadow buckets), MongoDB rejects the call with error code 48 (`codeName: 'NamespaceExists'`) because the collection already exists. The generic .catch at the end of the chain converts every error — including this benign collision — into `errors.InternalError`, which propagates as HTTP 500 to callers.

This causes intermittent `We encountered an internal error failures` in the afterEach cleanup hooks of cloudserver's mongo-v0-ft-tests / mongo-v1-ft-tests, and has blocked merges on q/9.3 and q/9.4 (tracked in CLDSRV-919).

### **Fix**

- `NamespaceExists` is now treated as idempotent success: the collection existing is exactly the desired postcondition of this step, so the error is intercepted and the flow continues. This does not change the createBucket API contract: the metastore write above is already an idempotent `updateOne(..., { upsert: true })`, and collection existence is an internal implementation detail of the Mongo backend (other metadata backends have no equivalent concept). It also mirrors the existing behavior of deleteBucket, which ignores `NamespaceNotFound` when dropping the collection.

- **Sharded deployments**: the flow now proceeds to shardCollection even when the collection pre-exists. Two concurrent createBucket calls can therefore race on the shardCollection admin command, so AlreadyInitialized (code 23) is also treated as success — the collection is always sharded with the same { _id: 1 } key, so both outcomes are equivalent.

- **Error contrac**t: collection-creation errors are now handled locally with an explicit `cb(errors.InternalError)` instead of relying on the outer catch of the metastore update; the callback is invoked exactly once on every path, and any error other than the two benign ones above still returns InternalError as before.

This addresses the three review concerns raised on the earlier attempt (#2643): local error handling with the callback always called, the shardCollection race, and whether idempotency here is safe given the interface contract and other backends.

Issue: ARSN-596

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -466,21 +466,54 @@ class MongoClientInterface {
                 // "constants.usersBucket" and "PENSIEVE" since it has already
                 // been created
                 if (bucketName !== constants.usersBucket && bucketName !== PENSIEVE) {
-                    return this.db!.createCollection(bucketName).then(() => {
-                        if (this.shardCollections) {
-                            const cmd = {
-                                shardCollection: `${this.database}.${bucketName}`,
-                                key: { _id: 1 },
-                            };
-                            return this.adminDb!.command(cmd, {})
-                                .then(() => cb(null))
-                                .catch(err => {
-                                    log.error('createBucket: enabling sharding', { error: err });
-                                    return cb(errors.InternalError);
-                                });
-                        }
-                        return cb(null);
-                    });
+                    return this.db!.createCollection(bucketName)
+                        .catch(err => {
+                            // MongoDB returns NamespaceExists (code 48) when
+                            // the collection already exists, e.g. on
+                            // concurrent create/drop/create sequences on MPU
+                            // shadow buckets. The collection being there is
+                            // the desired outcome, so treat it as success:
+                            // this mirrors deleteBucket, which ignores
+                            // NamespaceNotFound when dropping the collection.
+                            if (err.codeName !== 'NamespaceExists') {
+                                throw err;
+                            }
+                            log.debug('createBucket: collection already exists', { bucketName });
+                        })
+                        .then(() => {
+                            if (this.shardCollections) {
+                                const cmd = {
+                                    shardCollection: `${this.database}.${bucketName}`,
+                                    key: { _id: 1 },
+                                };
+                                return this.adminDb!.command(cmd, {})
+                                    .catch(err => {
+                                        // Concurrent createBucket calls may
+                                        // race on shardCollection: sharding
+                                        // an already-sharded collection fails
+                                        // with AlreadyInitialized. The
+                                        // collection is sharded (with the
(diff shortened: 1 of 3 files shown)


delthas (approved):

DarkIsDude (approved):

tcarmet (approved):
Thank you for taking over @benzekrimaha, it's much appreciated and I look forward to see the results in cloudserver's CI :)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -466,21 +466,54 @@ class MongoClientInterface {
...
+                    return this.db!.createCollection(bucketName)
+                        .catch(err => {
+                            // MongoDB returns NamespaceExists (code 48) when
+                            // the collection already exists, e.g. on
+                            // concurrent create/drop/create sequences on MPU
+                            // shadow buckets. The collection being there is
+                            // the desired outcome, so treat it as success:
+                            // this mirrors deleteBucket, which ignores
+                            // NamespaceNotFound when dropping the collection.
+                            if (err.codeName !== 'NamespaceExists') {
```
francoisferrand, line 478: in case of conflict, CreateBucket must return `BucketAlreadyExists` (this is used in cloudserver for _conditional_ followups) → eating the error does not provide the expected semantics

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -466,21 +466,54 @@ class MongoClientInterface {
...
+                                throw err;
+                            }
+                            log.debug('createBucket: collection already exists', { bucketName });
+                        })
+                        .then(() => {
+                            if (this.shardCollections) {
```
francoisferrand, line 484: shouldn't this be done only on _real_ success : in case of conflict, another call will succeed and perform the `shardCollection` call?

francoisferrand (commented):
the current fix will create issues in cloudserver, c.f. for exemple https://github.com/scality/cloudserver/blob/780e103da378d77cd2a30d121dabbb1c81add24c/lib/api/apiUtils/bucket/bucketCreation.js#L56

```
                    return metadata.createBucket(usersBucket,
                        freshBucket, log, err => {
                            // Note: In the event that two
                            // users' requests try to create the
                            // usersBucket at the same time,
                            // this will prevent one of the users
                            // from getting a BucketAlreadyExists
                            // error with respect
                            // to the usersBucket.
                            // TODO: move to `.is` once BKTCLT-9 is done and bumped in Cloudserver
                            if (err && !err.BucketAlreadyExists) {
                                log.error('error from metadata', {
                                    error: err,
                                });
                                return cb(err);
                            }
                            log.trace('Users bucket created');
                            // Finally put the key in the new format
                            // usersBucket
                            return metadata.putObjectMD(usersBucket,
                                key, omVal, {}, log, cb);
                        });
```

PR #2666, opened 2026-07-06
Add initializeBucketCapacity to seed empty-bucket metrics

delthas:
## Context

Bucket quota metrics live in the `__infostore` collection, keyed `bucket_<name>_<creationTimestamp>`, and are written today only by the s3utils `count-items` cron. A freshly created bucket — or a quota enabled on an empty bucket — has no metric document until the next cron run (up to 24h), so CloudServer's bucket-quota check finds nothing in Scuba and fails open (ARTESCA-17063).

`MongoClientInterface` already touches `__infostore` internally (`writeUUIDIfNotExists`), so this is a natural home for the write.

## Change

- Add `MongoClientInterface.initializeBucketCapacity(bucketName, creationDate, log, cb)` — idempotently inserts a zero-value bucket metric document via `updateOne({_id}, {$setOnInsert: zeroDoc}, {upsert: true})`. An existing document (e.g. one written by count-items with real values) is **never** overwritten.
  - `_id` = `bucket_<name>_<new Date(creationDate).getTime()>`.
  - Document shape matches what count-items produces / Scuba reads: `measuredOn` plus all-zero `usedCapacity` and `objectCount`, built from the module-level `emptyBucketCapacityMetrics` constant.
- Expose it through `MetadataWrapper` with a capability guard (no-op for non-mongodb backends, matching the existing `getUUID`/`countItems` pattern).

The consumer is CloudServer (CLDSRV-949, https://github.com/scality/cloudserver/pull/6221), which calls this on `createBucket` and on `bucketUpdateQuota` of a verifiably-empty bucket.

## Tests

Unit tests (`MongoClientInterface.spec.js`, real `mongodb-memory-server`): inserts a zero document and reads it back; confirms an existing document is not overwritten (idempotent); returns `InternalError` on update failure.

Issue: ARSN-610

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -151,62 +151,74 @@ export type InternalListObjectParams = {
     gt?: undefined;
 };
 
+export interface UsedCapacityMetrics {
+    current: Long | number;
+    nonCurrent: Long | number;
+    _currentCold: Long | number;
+    _nonCurrentCold: Long | number;
+    _currentRestored: Long | number;
+    _currentRestoring: Long | number;
+    _nonCurrentRestored: Long | number;
+    _nonCurrentRestoring: Long | number;
+    _incompleteMPUParts: Long | number;
+    // present on per-location and inflight-tagged bucket entries
+    _inflightsPreScan?: Long | number;
+}
+
+export interface ObjectCountMetrics {
+    current: Long | number;
+    nonCurrent: Long | number;
+    deleteMarker: Long | number;
+    _currentCold: Long | number;
+    _nonCurrentCold: Long | number;
+    _currentRestored: Long | number;
+    _currentRestoring: Long | number;
+    _nonCurrentRestored: Long | number;
+    _nonCurrentRestoring: Long | number;
+    _incompleteMPUUploads: Long | number;
+}
+
 export interface InfostoreDocument extends Document {
     _id: string | 'uuid';
     value?: string | ObjectMDStats;
     measuredOn?: string;
-    objectCount?: {
-        current: Long | number;
-        _currentCold: Long | number;
-        deleteMarker: Long | number;
-        nonCurrent: Long | number;
-        _nonCurrentCold: Long | number;
-        _currentRestored: Long | number;
-        _currentRestoring: Long | number;
-        _nonCurrentRestored: Long | number;
-        _nonCurrentRestoring: Long | number;
-        _incompleteMPUUploads: Long | number;
-    };
-    usedCapacity?: {
-        current: Long | number;
-        _currentCold: Long | number;
-        nonCurrent: Long | number;
-        _nonCurrentCold: Long | number;
-        _currentRestored: Long | number;
-        _currentRestoring: Long | number;
-        _nonCurrentRestored: Long | number;
-        _nonCurrentRestoring: Long | number;
-        _incompleteMPUParts: Long | number;
-    };
+    objectCount?: ObjectCountMetrics;
+    usedCapacity?: UsedCapacityMetrics;
     locations: {
         [key: string]: {
-            usedCapacity: {
-                current: Long | number;
-                nonCurrent: Long | number;
-                _currentCold: Long | number;
-                _nonCurrentCold: Long | number;
-                _currentRestored: Long | number;
-                _currentRestoring: Long | number;
-                _nonCurrentRestored: Long | number;
-                _nonCurrentRestoring: Long | number;
-                _inflightsPreScan: Long | number;
-                _incompleteMPUParts: Long | number;
-            };
-            objectCount: {
-                current: Long | number;
-                nonCurrent: Long | number;
-                _currentCold: Long | number;
-                _nonCurrentCold: Long | number;
-                _currentRestored: Long | number;
-                _currentRestoring: Long | number;
-                _nonCurrentRestored: Long | number;
-                _nonCurrentRestoring: Long | number;
-                _incompleteMPUUploads: Long | number;
-                deleteMarker: Long | number;
-            };
+            usedCapacity: UsedCapacityMetrics;
+            objectCount: ObjectCountMetrics;
         };
     };
 }
+
+// Canonical zero-value capacity/count metrics for an empty bucket, matching the
+// structure downstream metric readers expect.
+const emptyBucketCapacityMetrics: { usedCapacity: UsedCapacityMetrics; objectCount: ObjectCountMetrics } = {
+    usedCapacity: {
+        current: 0,
+        nonCurrent: 0,
+        _currentCold: 0,
+        _nonCurrentCold: 0,
+        _currentRestored: 0,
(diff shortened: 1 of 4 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2678,6 +2706,53 @@ class MongoClientInterface {
...
+        log: werelogs.Logger,
+        cb: ArsenalCallback<void>,
+    ) {
+        const i = this.getCollection<InfostoreDocument>(INFOSTORE);
+        if (!i) {
+            log.error('initializeBucketCapacity: error getting infostore collection');
+            return cb(errors.InternalError);
+        }
+        const _id = `bucket_${bucketName}_${new Date(creationDate).getTime()}`;
+        const doc = <InfostoreDocument>{
```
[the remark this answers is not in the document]
delthas: Good catch — the description was stale; it's now updated. `ownerCanonicalId`/`accountOwnerID` were intentionally dropped.

Nothing consumes the seed doc's `accountOwnerID`: Scuba's read path only touches `usedCapacity`/`objectCount`, and count-items fully replaces `__infostore` (atomic collection rename) before `updateInflightDeltas` runs, so the seed's copy would never be read. count-items itself only sets `accountOwnerID` conditionally (when inflights > 0), so omitting it here is consistent with the common case.

tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
maeldonn, line 1: Could we use async/await instead of callbacks ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2711,6 +2739,53 @@ class MongoClientInterface {
...
+        creationDate: string,
+        log: werelogs.Logger,
+        cb: ArsenalCallback<void>,
+    ) {
+        const i = this.getCollection<InfostoreDocument>(INFOSTORE);
+        if (!i) {
+            log.error('initializeBucketCapacity: error getting infostore collection');
+            return cb(errors.InternalError);
+        }
+        const _id = `bucket_${bucketName}_${new Date(creationDate).getTime()}`;
```
maeldonn: Should we verify that creationDate is valid ?
```ts
        const ts = new Date(creationDate).getTime();
        if (Number.isNaN(ts)) {
            log.error('initializeBucketCapacity: invalid creationDate', { bucketName, creationDate });
        }
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -208,6 +208,34 @@ export interface InfostoreDocument extends Document {
     };
 }
 
+// Canonical zero-value capacity/count metrics for an empty bucket, matching the
+// shape count-items produces so downstream readers see a consistent structure.
+const emptyBucketCapacityMetrics = {
```
SylvainSenechal: Btw, haven't looked at S3UTILS-224 but :

Should s3utils import this structure from Arsenal, instead of having database model defined twice in Arsenal/S3utils with a risk of desync ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2711,6 +2739,53 @@ class MongoClientInterface {
             });
     }
 
+    /**
+     * Initialize a zero-value bucket capacity metric document in the
+     * `__infostore` collection, unless one already exists. This allows quota
+     * checks to be served for a freshly created (or newly quota-enabled empty)
+     * bucket before the periodic count-items job has produced its metrics.
```
SylvainSenechal: Minor risk : Maybe the documentation shouldn't focus too much on what count-items is doing.
Count-items might later be changed, removed in favor of another more efficient system, in that case we would end up with out of date documentation

lib/storage/metadata/mongoclient/MongoClientInterface.ts
SylvainSenechal: Possibly.
Also can't we pass creationDate: Date ? Although it probably wont stop js codebase from sending trash on that field

lib/storage/metadata/mongoclient/MongoClientInterface.ts
SylvainSenechal: Nit because probably too annoying to do here, maybe a low priority ticket : 

I see we have this defined line 154 :
export interface InfostoreDocument extends Document {

I think this variable emptyBucketCapacityMetrics should be typed.
But InfostoreDocument fields (objectCount, usedCapacity) aren't defined.

Ideally, we should defined (objectCount, usedCapacity), and InfostoreDocument would 
use the proper types instead of inlining the definition, then we could define emptyBucketCapacityMetrics with a proper type

SylvainSenechal (approved):

tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
delthas: Done — converted the block to async/await (`promisify`-ing the callback method). Heads-up that the rest of this spec file is callback + `done`, so it's slightly inconsistent within the file now — happy to revert to callbacks if you'd rather keep it uniform.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
delthas: Added the NaN guard → `InternalError` (plus a test). I kept the parameter typed as `string`: the only callers pass `bucket.getCreationDate()` (which returns a string), so a `Date` type wouldn't buy anything and would fight the JS call sites.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
delthas: D2 done — extracted `UsedCapacityMetrics`/`ObjectCountMetrics`, and both `InfostoreDocument` and `emptyBucketCapacityMetrics` now use them.

On importing the shape into s3utils (D1): s3utils already depends on arsenal, but there's no literal shape to share. s3utils seeds its internal accumulator (`baseMetricsObject`, a different 30-field shape) through `_handleResults` and never builds this output object; and its values are BigInt (`0n`) whereas this constant is number (`0`), because the metrics persist as BSON `Long` (computed in BigInt, converted to `Long` on write). So a share would only touch a test helper, not the real runtime contract — leaving it as-is.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
delthas: Reworded the method's JSDoc and the constant comment to refer to "the periodic metrics job" / "downstream readers" instead of count-items specifically.

tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
maeldonn: Migrations usually cause temporary file inconsistencies, but I hope it doesn't last long.

maeldonn (approved):

PR #2670, opened 2026-07-07
impr(ARSN-608): Add support for passing targetAccount to Vault in request context

tmacro:


package.json
```diff
@@ -3,7 +3,7 @@
     "engines": {
         "node": ">=20"
     },
-    "version": "8.4.18",
+    "version": "8.4.19",
```
tcarmet, line 6: Depending on where this is supposed to land, I would redirect you to use the 8.5 version of arsenal since this is an addition (a small and harmless addition but addition nonetheless) 

But ultimately it's your call cause it might cause you some trouble (for example if you're aiming cloudserver 9.3, but on 9.4 you will be fine).

tcarmet (approved):
LGTM but I think the commits deserves a re-work between what's formatting and what new additional change I got a bit confused when reviewing

anurag4DSB (approved):
There are a lot of formatting changes, was hard to find the actual change(which is small). Would be nice to have the lint changes in a separate commit next time :)

tmacro:
@anurag4DSB 
> There are a lot of formatting changes, was hard to find the actual change(which is small). Would be nice to have the lint changes in a separate commit next time :)

I'll split them out. I didn't realize prettier had made that many changes.

PR #2672, opened 2026-07-09
ARSN-613: Properly return already exists error

benzekrimaha:
## Context

Concurrent MongoDB `createBucket` calls could silently behave as successful
idempotent writes. This prevented Cloudserver from distinguishing the winning
request from requests that lost the race and could allow a loser to overwrite
the winner's metadata.

This change restores an explicit `BucketAlreadyExists` contract matching the
other metadata backends. It must be deployed with
[Cloudserver #6222](https://github.com/scality/cloudserver/pull/6222), which
handles this result in normal and MPU shadow-bucket creation paths.

## Changes

- Use `$setOnInsert` for the metastore upsert so duplicate requests never
  overwrite existing bucket metadata.
- Return `BucketAlreadyExists` when the atomic upsert matches an existing
  metastore entry.
- Treat `BucketAlreadyExists` as success while initializing `usersBucket`.
- Treat `NamespaceExists` as success after a new metastore insert because an
  existing backing collection does not imply an existing bucket.
- Continue sharding after `NamespaceExists` and treat `AlreadyInitialized` as
  success.
- Keep callbacks outside Promise rejection handling so a callback exception
  cannot trigger a second callback.
- Return the Promise chain so callback exceptions propagate to the caller.

Issue: ARSN-613

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -361,6 +361,7 @@ class MongoClientInterface {
            usersBucket to have attributes, we pre-create the
            usersBucket attributes here (see bucketCreation.js line
            36)*/
+        const log = this.logger;
         const usersBucketAttr = new BucketInfo(
             constants.usersBucket,
             'admin',
@@ -368,9 +369,12 @@ class MongoClientInterface {
             new Date().toJSON(),
             BucketInfo.currentModelVersion(),
         );
-        return this.createBucket(constants.usersBucket, usersBucketAttr, this.logger, err => {
-            if (err) {
-                this.logger.fatal('error writing usersBucket ' + 'attributes to metastore', { error: err });
+        return this.createBucket(constants.usersBucket, usersBucketAttr, log, createErr => {
+            if (createErr?.is?.BucketAlreadyExists) {
+                return cb();
+            }
+            if (createErr) {
+                log.fatal('error writing usersBucket attributes to metastore', { error: createErr });
                 throw errors.InternalError;
             }
             return cb();
@@ -412,8 +416,17 @@ class MongoClientInterface {
         ).makeSerializable();
         const m = this.getCollection<BucketMetastoreDocument>(METASTORE);
 
+        let vFormat = this.defaultBucketKeyFormat;
+        if (
+            bucketName === constants.usersBucket ||
+            bucketName === PENSIEVE ||
+            bucketName.startsWith(constants.mpuBucketPrefix)
+        ) {
+            vFormat = BUCKET_VERSIONS.v0;
+        }
+
         const payload = {
-            $set: {
+            $setOnInsert: {
                 _id: bucketName,
                 value: {
                     ...newBucketMD,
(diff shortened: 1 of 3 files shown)


tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
```diff
@@ -396,12 +396,28 @@ describe('MongoClientInterface::createBucket', () => {
         client.client = {};
     }
 
-    it('should succeed when the collection already exists (NamespaceExists)', done => {
+    it('should return BucketAlreadyExists when metastore entry already exists', done => {
```
maeldonn: nit: Could you use async await on new tests instead of callbacks ?

maeldonn (approved):

tests/unit/storage/metadata/mongoclient/MongoClientInterface.spec.js
benzekrimaha: Done in b2c70d29 ,the stubbed createBucket tests now use async/await with a `createBucketPromised` helper and `assert.rejects` for error cases.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -361,19 +361,33 @@ class MongoClientInterface {
...
-            'admin',
-            'admin',
-            new Date().toJSON(),
-            BucketInfo.currentModelVersion(),
-        );
-        return this.createBucket(constants.usersBucket, usersBucketAttr, this.logger, err => {
-            if (err) {
-                this.logger.fatal('error writing usersBucket ' + 'attributes to metastore', { error: err });
+        const log = this.logger;
+        return this.getBucketAttributes(constants.usersBucket, log, err => {
```
francoisferrand: we should not make an extra API call !
- it is costly : extra IO → more load on mongo + more latency on the operation
- it does not protect against race condition → still need to handle actual conflict detection

francoisferrand (dismissed):

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -460,54 +465,46 @@ class MongoClientInterface {
                     log.debug('createBucket: failed to create bucket', { bucketName, result });
                     return cb(errors.InternalError);
                 }
+                if (result.matchedCount > 0 && result.upsertedCount === 0) {
```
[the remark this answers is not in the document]
benzekrimaha: Switched the metastore upsert from `$set` to `$setOnInsert` in `9654b07e` so a duplicate `createBucket` no longer overwrites existing bucket metadata before returning `BucketAlreadyExists`. Stub test updated accordingly (`modifiedCount: 0`, asserts `$setOnInsert`).

lib/storage/metadata/mongoclient/MongoClientInterface.ts
benzekrimaha: `usersBucketHack` now calls `createBucket` directly and treats `BucketAlreadyExists` as success when concurrent client setup wins the race. No `getBucketAttributes` pre-check only one write path, no extra Mongo read, and the race is handled by actual create conflict detection.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -460,54 +465,46 @@ class MongoClientInterface {
...
+                                .then(() => cb(null))
+                                .catch(err => {
+                                    log.error('createBucket: enabling sharding', { error: err });
+                                    return cb(errors.InternalError);
+                                });
                         })
                         .catch(err => {
+                            if (err.codeName === 'NamespaceExists') {
+                                log.debug('createBucket: collection already exists', { bucketName });
+                                return cb(errors.BucketAlreadyExists);
```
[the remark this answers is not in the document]
benzekrimaha: When the metastore upsert inserts a new entry, `NamespaceExists` on `createCollection` means an orphaned backing collection, not a bucket-level duplicate, returning `BucketAlreadyExists` there was inconsistent and broke the existing regression *"should create a bucket whose backing collection already exists"*. Fixed in `07aa045b`: treat `NamespaceExists` as success after a new metastore insert and continue with sharding if enabled. `BucketAlreadyExists` remains reserved for metastore duplicates.

delthas (approved):
LGTM code-wise, but didnt go far into the whole issue. Let's wait for Francois approve before merging.

maeldonn (approved):

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -361,16 +361,21 @@ class MongoClientInterface {
...
             'admin',
             new Date().toJSON(),
             BucketInfo.currentModelVersion(),
         );
-        return this.createBucket(constants.usersBucket, usersBucketAttr, this.logger, err => {
-            if (err) {
-                this.logger.fatal('error writing usersBucket ' + 'attributes to metastore', { error: err });
+        return this.createBucket(constants.usersBucket, usersBucketAttr, log, createErr => {
+            if (createErr?.is?.BucketAlreadyExists) {
+                // Concurrent setup won the race to create usersBucket.
```
francoisferrand: nit: comment is not really useless, don't know what race or which context could race... (and documenting that here would be weird as well)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -460,54 +465,50 @@ class MongoClientInterface {
...
+                                // Concurrent createBucket calls may race on
+                                // shardCollection after a successful
+                                // createCollection.
+                                if (err.codeName !== 'AlreadyInitialized') {
+                                    throw err;
+                                }
+                                log.debug('createBucket: collection already sharded', { bucketName });
+                            })
+                            .then(() => cb(null))
+                            .catch(err => {
```
francoisferrand: not correct: this would incorrectly catch an exception thrown by `cb`

(not clear if this was prevent before, with the refactoring...)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -460,54 +465,50 @@ class MongoClientInterface {
...
-                                    .then(() => cb(null))
-                                    .catch(err => {
-                                        log.error('createBucket: enabling sharding', { error: err });
-                                        return cb(errors.InternalError);
-                                    });
+                            if (err.codeName === 'NamespaceExists') {
+                                // Metastore insert succeeded: an orphaned backing
+                                // collection is not a bucket-level duplicate.
+                                log.debug('createBucket: collection already exists', { bucketName });
+                                return finishAfterCollectionReady();
```
francoisferrand: not clear what this changes: seems to do the same thing, less readable because of the callback preventing from reading sequentially...

lib/storage/metadata/mongoclient/MongoClientInterface.ts
benzekrimaha: Removed

lib/storage/metadata/mongoclient/MongoClientInterface.ts
benzekrimaha: The callback is now invoked from the terminal two-argument .then(success, failure), so an exception thrown by cb is not caught and converted into a second callback. createBucket also returns the Promise chain. I also added a regression test asserting the callback is called once when it throws.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
benzekrimaha: Removed `finishAfterCollectionReady` and restored the sequential `createCollection` → optional shardCollection flow. NamespaceExists and AlreadyInitialized are handled directly at their respective operation boundaries.

PR #2673, opened 2026-07-15
Add a shared newRequestLoggerFromRequest helper for x-scal-request-uids seeding

delthas:
`routes.ts` built a request logger from the inbound `x-scal-request-uids` header inline — read the header, guard its length (a bare `128` literal), then seed the werelogs UID chain or start a fresh one. Consumers that need the same behaviour (e.g. vault2's SSO/SAML request-logger seeding in scality/vault2#217) had to reimplement the whole pattern, and the copies had already drifted on edge cases.

This centralizes it:
- Export `maxRequestUidsLength = 128` from `lib/constants.ts`.
- Add `routesUtils.newRequestLoggerFromRequest(logger, req)` that owns the header name, the length cap, and the seed-or-fresh decision; `routes.ts` now uses it (dropping the private `isValidReqUids` and the inline block, plus two `@ts-ignore`s).

Consumers (scality/vault2#217) collapse their copy to a one-line delegate.

The shared helper also tightens two edge cases the old inline code handled loosely, so the S3 path now matches the intended policy:
- an empty header value starts a fresh chain instead of seeding from `''`;
- a header sent multiple times (array value) starts a fresh chain instead of passing an array to `newRequestLoggerFromSerializedUids`.

Tests: direct coverage of the helper (valid / absent / empty / array / length boundary).

The touched files were prettier-dirty on the base branch, so the first commit reformats them and the functional change lands prettier-clean on top.

Issue: ARSN-615

### DIFF
--- a/lib/s3routes/routes.ts
+++ b/lib/s3routes/routes.ts
@@ -53,12 +53,11 @@ function checkBucketAndKey(
     // if empty name and request not a List Buckets
     if (!bucketName && !(method === 'GET' && !objectKey)) {
         log.debug('empty bucket name', { method: 'routes' });
-        return (method !== 'OPTIONS') ?
-            errors.MethodNotAllowed : errorInstances.AccessForbidden
-                .customizeDescription('CORSResponse: Bucket not found');
+        return method !== 'OPTIONS'
+            ? errors.MethodNotAllowed
+            : errorInstances.AccessForbidden.customizeDescription('CORSResponse: Bucket not found');
     }
-    if (bucketName !== undefined && routesUtils.isValidBucketName(bucketName,
-        blacklistedPrefixes.bucket) === false) {
+    if (bucketName !== undefined && routesUtils.isValidBucketName(bucketName, blacklistedPrefixes.bucket) === false) {
         log.debug('invalid bucket name', { bucketName });
         if (method === 'DELETE') {
             return errors.NoSuchBucket;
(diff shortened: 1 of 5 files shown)


lib/s3routes/routes.ts
```diff
@@ -223,18 +217,9 @@ export default function routes(
...
-    }
-    const log =
-        reqUids !== undefined
-            ? // @ts-ignore
-              logger.newRequestLoggerFromSerializedUids(reqUids)
-            : // @ts-ignore
-              logger.newRequestLogger();
+    // logger is typed RequestLogger in this signature but is actually the
+    // parent Logger at runtime (it exposes newRequestLogger*).
+    const log = routesUtils.newRequestLoggerFromRequest(logger as unknown as Logger, req);
```
maeldonn: nit: Could you type `logger` as a `Logger` to avoid this ugly type casting ? Is it breaking something ?

maeldonn (approved):

lib/s3routes/routesUtils.ts
```diff
@@ -21,6 +21,18 @@ export function setServerHeader(value: string) {
     serverHeaderValue = value;
 }
 
+/**
+ * Build a request logger for an incoming request, seeding the werelogs UID
+ * chain from the `x-scal-request-uids` header when it carries a usable value.
+ */
+export function newRequestLoggerFromRequest(logger: Logger, req: http.IncomingMessage): RequestLogger {
+    const reqUids = req.headers['x-scal-request-uids'];
+    if (typeof reqUids !== 'string' || reqUids.length === 0 || reqUids.length >= constants.maxRequestUidsLength) {
```
SylvainSenechal, line 30: are you sure its >= 

If a constant "max text length = 10", it means my text can have 10 characters, here it would allow up to 9 🤔

SylvainSenechal (approved):

lib/s3routes/routes.ts
SylvainSenechal: ye its a bit ugly but I guess maybe its also annoying to fix properly

delthas: Good catch — done in 9b7cc388. The param really was just mistyped: `routes()`/`checkTypes` declared `logger` as `RequestLogger`, but at runtime it's the parent werelogs `Logger` (only ever used to build the per-request logger), which is why the old code needed the cast + `@ts-ignore`s. Retyped it to `Logger` (already imported) and dropped the cast — not breaking, since `logger` is never used as a `RequestLogger`. Build + tests green.

lib/s3routes/routesUtils.ts
delthas: Intentional — `>=` preserves arsenal's existing guard exactly. The old `isValidReqUids` was `return reqUids.length < 128` (reject 128+, so max accepted is 127); `reqUids.length >= maxRequestUidsLength` is the same threshold. So the constant is really an exclusive upper bound / anti-abuse cap rather than a precise "max allowed length" — you're right that the max accepted is 127, but that's unchanged from before this PR. Kept it byte-for-byte to avoid changing behaviour.

package.json
```diff
@@ -3,7 +3,7 @@
     "engines": {
         "node": ">=20"
     },
-    "version": "8.4.19",
+    "version": "8.4.20",
```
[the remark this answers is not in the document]
delthas: Not going to be 8.5.0, but I'll check if this impacts other users.

PR #2678, opened 2026-07-23
Support conditional writes on versioned puts in putObjectVerCase3

maeldonn:
Adds params.conditions support to putObjectVerCase3 so a version-specific put
becomes an atomic compare-and-write, failing with PreconditionFailed when the
stored version does not satisfy the condition. 
Needed by CLDSRV-952.

Issue: ARSN-619

### DIFF
--- a/lib/storage/metadata/conditions.js
+++ b/lib/storage/metadata/conditions.js
@@ -12,6 +12,7 @@ const supportedOperators = {
 
 const supportedStructuralOperators = {
     $or: true,
+    $and: true,
 };
 
 // supports strings and numbers
--- a/lib/storage/metadata/conditions.js
+++ b/lib/storage/metadata/conditions.js
@@ -16,45 +16,27 @@ const supportedStructuralOperators = {
 };
 
 // supports strings and numbers
-const _operatorType1 = joi.string().valid(
-    '$gt',
-    '$gte',
-    '$lt',
-    '$lte',
-);
+const _operatorType1 = joi.string().valid('$gt', '$gte', '$lt', '$lte');
 
 // supports strings, numbers, and boolean
-const _operatorType2 = joi.string().valid(
-    '$eq',
-    '$ne',
-);
+const _operatorType2 = joi.string().valid('$eq', '$ne');
 
 const _operatorType3 = joi.string().valid('$exists');
 
-const _valueType1 = joi.alternatives([
-    joi.string(),
-    joi.number(),
-]);
+const _valueType1 = joi.alternatives([joi.string(), joi.number()]);
 
-const _valueType2 = joi.alternatives([
-    joi.string(),
-    joi.number(),
-    joi.boolean(),
-]);
+const _valueType2 = joi.alternatives([joi.string(), joi.number(), joi.boolean()]);
 
 const _valueType3 = joi.boolean();
 
-const queryObject = joi.object({})
+const queryObject = joi
+    .object({})
     .pattern(_operatorType1, _valueType1)
     .pattern(_operatorType2, _valueType2)
     .pattern(_operatorType3, _valueType3)
     .xor(...Object.keys(supportedOperators));
 
-const metadataCondObject = joi.alternatives([
-    _valueType1,
-    _valueType2,
-    queryObject,
-]);
+const metadataCondObject = joi.alternatives([_valueType1, _valueType2, queryObject]);
 
 function validateConditionsObject(obj) {
     if (obj === undefined) {
--- a/lib/storage/metadata/conditions.js
+++ b/lib/storage/metadata/conditions.js
@@ -7,6 +7,11 @@ const supportedOperators = {
     $gte: true,
     $lt: true,
     $lte: true,
+    $exists: true,
+};
+
+const supportedStructuralOperators = {
+    $or: true,
 };
 
 // supports strings and numbers
@@ -23,6 +28,8 @@ const _operatorType2 = joi.string().valid(
     '$ne',
 );
 
+const _operatorType3 = joi.string().valid('$exists');
+
 const _valueType1 = joi.alternatives([
     joi.string(),
     joi.number(),
@@ -34,9 +41,12 @@ const _valueType2 = joi.alternatives([
     joi.boolean(),
 ]);
 
+const _valueType3 = joi.boolean();
+
 const queryObject = joi.object({})
     .pattern(_operatorType1, _valueType1)
     .pattern(_operatorType2, _valueType2)
+    .pattern(_operatorType3, _valueType3)
     .xor(...Object.keys(supportedOperators));
 
 const metadataCondObject = joi.alternatives([
@@ -58,5 +68,6 @@ function validateConditionsObject(obj) {
 
 module.exports = {
     supportedOperators,
+    supportedStructuralOperators,
     validateConditionsObject,
 };
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -62,6 +62,7 @@ const __UUID = 'uuid';
 const PENSIEVE = 'PENSIEVE';
 const __COUNT_ITEMS = 'countitems';
 const ASYNC_REPAIR_TIMEOUT = 15000;
+const MONGODB_DUPLICATE_KEY_ERROR = 11000;
 
 const MONGO_CONNECT_TIMEOUT_MS = process.env.MONGO_CONNECT_TIMEOUT_MS;
 const MONGO_SOCKET_TIMEOUT_MS = process.env.MONGO_SOCKET_TIMEOUT_MS;
@@ -952,7 +953,7 @@ class MongoClientInterface {
                  * are the same (affecting the first operation), or if the master
                  * version is updated concurrently.
                  */
-                if (err.code === 11000) {
+                if (err.code === MONGODB_DUPLICATE_KEY_ERROR) {
                     log.debug('putObjectVerCase1: error putting object version', {
                         code: err.code,
                         error: err.errmsg,
@@ -1072,7 +1073,7 @@ class MongoClientInterface {
             })
                 .then(() => callback(null, `{"versionId": "${objVal.versionId}"}`))
                 .catch(err => {
-                    if (err.code === 11000) {
+                    if (err.code === MONGODB_DUPLICATE_KEY_ERROR) {
                         // Failed condition turns the version upsert into an insert of an existing
                         // _id (dup-key). index 0 = version op in the ordered bulkWrite, so it's
                         // the version write that failed the condition.
@@ -1270,7 +1271,7 @@ class MongoClientInterface {
                             // we accept that the update fails if
                             // condition is not met, meaning that a more
                             // recent master was already in place
-                            if (err.code === 11000) {
+                            if (err.code === MONGODB_DUPLICATE_KEY_ERROR) {
                                 return cb(null, `{"versionId": "${objVal.versionId}"}`);
                             }
                             log.error('putObjectVerCase4: error upserting master', { error: err.message });
@@ -2731,7 +2732,7 @@ class MongoClientInterface {
                 return cb(null);
             })
             .catch(err => {
-                if (err.code === 11000) {
+                if (err.code === MONGODB_DUPLICATE_KEY_ERROR) {
                     // duplicate key error
                     return cb(errors.KeyAlreadyExists);
                 }
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -1107,7 +1107,8 @@ class MongoClientInterface {
             }
         }
 
-        return c.findOne({ _id: masterKey })
+        return c
+            .findOne({ _id: masterKey })
             .then(checkObj => {
                 const objUpsert = !checkObj;
                 // initiating array of operations with version creation/update
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -1072,28 +1072,48 @@ class MongoClientInterface {
             })
                 .then(() => callback(null, `{"versionId": "${objVal.versionId}"}`))
                 .catch(err => {
-                    log.error('putObjectVerCase3: error putting object version', { error: err.message });
                     if (err.code === 11000) {
+                        // Failed condition turns the version upsert into an insert of an existing
+                        // _id (dup-key). index 0 = version op in the ordered bulkWrite, so it's
+                        // the version write that failed the condition.
+                        if (params.conditions && err.writeErrors?.[0]?.index === 0) {
+                            log.info('putObjectVerCase3: version condition not met, rejecting write', {
+                                versionKey,
+                            });
+                            return callback(errors.PreconditionFailed);
+                        }
                         // We want duplicate key error logged however in
                         // case of the race condition mentioned above, the
                         // InternalError will allow for automatic retries
                         log.error('putObjectVerCase3:', errors.KeyAlreadyExists);
                         return callback(errors.InternalError);
                     }
+                    log.error('putObjectVerCase3: error putting object version', { error: err.message });
                     return callback(errors.NoSuchVersion);
                 });
         };
 
-        c.findOne({ _id: masterKey })
+        // a version failing the condition degenerates the upsert into an insert, raising a duplicate key error
+        const versionFilter: Record<string, any> = { _id: versionKey };
+        if (params.conditions && Object.keys(params.conditions).length > 0) {
+            try {
+                MongoUtils.translateConditions(0, 'value', versionFilter, params.conditions);
+            } catch (err) {
+                log.error('putObjectVerCase3: error creating mongodb filter', {
+                    error: reshapeExceptionError(err as ErrorLike),
+                });
+                return cb(errors.InternalError);
+            }
+        }
+
+        return c.findOne({ _id: masterKey })
             .then(checkObj => {
                 const objUpsert = !checkObj;
                 // initiating array of operations with version creation/update
                 const ops: AnyBulkWriteOperation<ObjectMetastoreDocument>[] = [
                     {
                         updateOne: {
-                            filter: {
-                                _id: versionKey,
-                            },
+                            filter: versionFilter,
                             update: {
                                 $set: {
                                     _id: versionKey,
--- a/lib/storage/metadata/mongoclient/utils.ts
+++ b/lib/storage/metadata/mongoclient/utils.ts
@@ -79,8 +79,13 @@ function _assignCondition(prefix: string, object: Record<string, any>, cond: Con
  * Translates a structural operator into a mongodb filter.
  * Each item is translated on its own, sharing the parent prefix.
  */
-function _assignStructuralOperator(depth: number, prefix: string, object: Record<string, any>, op: string,
-    items: Condition): void {
+function _assignStructuralOperator(
+    depth: number,
+    prefix: string,
+    object: Record<string, any>,
+    op: string,
+    items: Condition,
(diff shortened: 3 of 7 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1074,6 +1077,13 @@ class MongoClientInterface {
                 .catch(err => {
                     log.error('putObjectVerCase3: error putting object version', { error: err.message });
                     if (err.code === 11000) {
+                        // only the version op (index 0) carries params.conditions
+                        if (params.conditions && err.writeErrors?.[0]?.index === 0) {
```
SylvainSenechal, line 1080: Nit: Never really looked at this part of the code before, it was hard to read before your changes, it is still hard to read after your changes.

Can use temporary variables : 

`const versionConditionFailed = params.conditions && err.writeErrors?.[0]?.index === 0;`
Even go further 

```
const isUpdateVersionOperation = err.writeErrors?.[0]?.index === 0;
const isConditionalWrite = params.conditions

if(isUpdateVersionOperation && isConditionalWrite) {..}

```

Saying this because it took me some time to understand what the above new comment meant with "version op", and the fact that this putObjectEntry could do multiple operation wasn't directly expected from me

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1074,6 +1077,13 @@ class MongoClientInterface {
                 .catch(err => {
                     log.error('putObjectVerCase3: error putting object version', { error: err.message });
                     if (err.code === 11000) {
```
SylvainSenechal: nit as you didn't modify it : should declare a const, i had to google whats a 11000 error and it's important to understand what's happening inside this condition

tests/functional/metadata/mongodb/putObject.spec.js
```diff
@@ -525,6 +527,72 @@ describe('MongoClientInterface:metadata.putObjectMD', () => {
...
+                };
+
+                beforeEach(() => {
+                    putObjectMD = promisify(metadata.putObjectMD.bind(metadata));
+                    getObjectP = promisify(getObject);
+                    versionKey = formatVersionKey(OBJECT_NAME, VERSION_ID, variation.vFormat);
+                });
+
+                [
+                    { desc: 'a comparison', conditions: { number: { $gt: 42 } } },
```
SylvainSenechal, line 612: Ok for these tests, but I'm thinking : Should we have a test where like for example, the object metadata doesn't have the field "number: 24" at all, and we still have the conditions number gte 42 to see what happens ? 

Or what's the behavior when the condition is {number : {$gt: null}}.

Asking because I think we may encounter similar situation with our microVersionId

SylvainSenechal: Also discussion with claude about this, his response : 
(so yeah some of what I said is tested line 658, but not sure everything 🤔 )

---

"There's already a test at line 658 for when the version doesn't exist at all (insert path — condition is bypassed). But there's no test for when the version exists but is missing the field the condition references.

In MongoDB, querying { 'value.number': { $gt: 42 } } on a doc without a number field treats it as null, which fails the comparison — so the write would be rejected with PreconditionFailed. That's probably the right behavior, but it's untested and worth making explicit.

Worth adding two cases to the forEach block at line 611:


{ desc: 'a field missing from the stored object', conditions: { number: { $gt: 42 } } },
// stored object has no `number` field at all
And a separate test where the stored object has no number but the condition is { number: null } — to verify the null/missing distinction if that matters to callers.
"

lib/storage/metadata/bucketclient/BucketClientInterface.js
```diff
@@ -77,6 +78,14 @@ class BucketClientInterface {
     }
 
     putObject(bucketName, objName, objVal, params, log, cb) {
+        // reject rather than silently perform an unconditional write
```
SylvainSenechal: Not fully sure about the rejection. 
As I understand, we are gonna start using conditional upsert with your associated cloudserver ongoing PR, but the ticket about handling these conditional upsert for metadata is not done and we don't really have an ETA. 

So if I understand correctly, with this error, we are now forced to implement the conditional upsert or else we can't ship the feature at all ? As the conditional upsert is just an improvement for potential race condition, I think it's reasonable here to just log a warning and continue, but I don't have a strong opinion on this.

Also you already created the ticket for metadata, so I think you could comment it as a todo

lib/storage/metadata/MetadataWrapper.js
```diff
@@ -274,6 +274,14 @@ class MetadataWrapper {
         });
     }
 
+    /**
+     * Whether the backend atomically enforces params.conditions on putObjectMD
+     * @returns {boolean} true when supported
+     */
+    supportsConditionalPutObjectMD() {
```
SylvainSenechal: Wait this is supposed to be called by Cloudserver ? 
I'm not sure about this design we will talk about it tomorrow. 

More generally, I wonder if we need to gatekeep the conditional put : 

After all, we only implemented it for putObjectVer3, so technically, even for the mongodb backend we still don't fully support it. 

And the params.condition field, it existed before this pr, so we were already able to call these function with params.conditions not null and it would just be ignored (which is probably fine)

SylvainSenechal (changes requested):
The logic in putObjectVer3 is fine for me

Some discussion to have directly about the rejection of params.conditions

lib/storage/metadata/MetadataWrapper.js
SylvainSenechal: Even when discussing with ai, realized that it means there is now a risk that for the other function (putObjectVerX where X is not 3), if we currently pass conditions, it will now fail 🤔 (which it probably should i agree, but we probably don't wanna change the behavior now without thorough testing and review of each function, which I don't think we wanna do now)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
maeldonn: Added a comment explaining the index-0 check instead of a temp var. Let me know if you'd still prefer the variable.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
maeldonn: Done.

tests/functional/metadata/mongodb/putObject.spec.js
maeldonn: Added a test for a condition referencing a field missing from the stored version.

lib/storage/metadata/bucketclient/BucketClientInterface.js
maeldonn: This doesn't block shipping: bucketd reports `supportsConditionalPutObjectMD() === false`, so callers gating on it never send conditions here. The reject only catches a caller ignoring that, and warn-and-continue there = a silent unconditional write, the race we're trying to prevent. Can add a TODO to the metadata ticket if useful.

lib/storage/metadata/MetadataWrapper.js
maeldonn: No regression: only putObjectVerCase3 reads `params.conditions`. Case1/2/4 still ignore it exactly as before, so passing conditions there doesn't newly fail. Case3's enforcement is the only new behavior and it's covered by tests.

lib/storage/metadata/bucketclient/BucketClientInterface.js
```diff
@@ -77,6 +78,14 @@ class BucketClientInterface {
...
+        if (params && params.conditions) {
+            log.error('bucketd does not support conditional puts', {
+                bucketName,
+                objName,
+            });
+            return cb(errors.NotImplemented);
+        }
```
maeldonn:
```suggestion
```

lib/storage/metadata/bucketclient/BucketClientInterface.js
```diff
@@ -77,19 +77,31 @@ class BucketClientInterface {
     }
 
     putObject(bucketName, objName, objVal, params, log, cb) {
-        this.client.putObject(bucketName, objName, JSON.stringify(objVal),
-            log.getSerializedUids(), cb, params);
+        // reject rather than silently perform an unconditional write
+        if (params && params.conditions) {
```
benzekrimaha: no point guarding per backend: check `params.conditions && !this.supportsConditionalPutObjectMD()` once in `MetadataWrapper.putObjectMD`,  that also covers the memory/file backends, which currently ignore `conditions` silently.

benzekrimaha (commented):
One design comment, rest looks solid (good test coverage on the condition paths).

lib/storage/metadata/bucketclient/BucketClientInterface.js
SylvainSenechal: I talked with Maël, we are just not gonna do this at all : no need to return an error, params.conditions existed before this pr and was ignored. 
This pr is more of a best effort to have better database consistency, adding it to mongo for now without touching metadata is fine

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1072,28 +1073,49 @@ class MongoClientInterface {
...
+        // a version failing the condition degenerates the upsert into an insert, raising a duplicate key error
+        const versionFilter: Record<string, any> = { _id: versionKey };
+        if (params.conditions) {
+            try {
+                MongoUtils.translateConditions(0, 'value', versionFilter, params.conditions);
+            } catch (err) {
+                log.error('putObjectVerCase3: error creating mongodb filter', {
+                    error: reshapeExceptionError(err as ErrorLike),
+                });
+                return cb(errors.InternalError);
```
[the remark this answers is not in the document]
SylvainSenechal: not true anymore, pr description updated

benzekrimaha (dismissed):

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1072,28 +1073,49 @@ class MongoClientInterface {
...
                     }
+                    log.error('putObjectVerCase3: error putting object version', { error: err.message });
                     return callback(errors.NoSuchVersion);
                 });
         };
 
-        c.findOne({ _id: masterKey })
+        // a version failing the condition degenerates the upsert into an insert, raising a duplicate key error
+        const versionFilter: Record<string, any> = { _id: versionKey };
+        if (params.conditions) {
```
delthas: nit: `{}` is truthy, so an empty conditions object goes into `translateConditions`, which throws `InternalError` — the joi schema requires exactly one operator, so `{}` fails validation. Before this PR the same call succeeded with the conditions ignored.

`conditions: {}` already shows up in the putObjectVerCase1 unit tests, so it's a shape that gets passed around.

```suggestion
        if (params.conditions && Object.keys(params.conditions).length > 0) {
```

delthas (dismissed):
LGTM (see nit above). Hard to understand but I managed to grasp what we're doing here :)

SylvainSenechal:
@francoisferrand Added, you can take a look at the new exists/or logic

SylvainSenechal:
Will run prettier later

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1072,28 +1073,48 @@ class MongoClientInterface {
...
                         return callback(errors.InternalError);
                     }
+                    log.error('putObjectVerCase3: error putting object version', { error: err.message });
                     return callback(errors.NoSuchVersion);
                 });
         };
 
-        c.findOne({ _id: masterKey })
+        // a version failing the condition degenerates the upsert into an insert, raising a duplicate key error
+        const versionFilter: Record<string, any> = { _id: versionKey };
```
francoisferrand, line 1098: this is added in `putObjectVerCase3` ; which handles the case of updating an object with an (existing) versionId.

however, the same putObject may go through
- `putObjectVerCase4` when `repairMaster=true` - which is set by Cloudserver in some case → most likely needed to cover all cases
- `putObjectNoVer` when there is no versioning and `putObjectVerCase2` when versioning is suspended → no needed for CRR [which requires & enforces versioning], but should be added (in followup possibly) for consistency, to ensure future developers will not rely on this conditional write feature only to find it work only in some cases...

`putObjectVerCase1` is used for version creation (it allocates a new versionID !), so nothing to do there

lib/storage/metadata/mongoclient/utils.ts
```diff
@@ -75,6 +75,30 @@ function _assignCondition(prefix: string, object: Record<string, any>, cond: Con
     object[prefix] = cond;
 }
 
+/*
+* Translates a structural operator into a mongodb filter.
```
francoisferrand:
```suggestion
 * Translates a structural operator into a mongodb filter.
```

lib/storage/metadata/mongoclient/utils.ts
```diff
@@ -75,6 +75,30 @@ function _assignCondition(prefix: string, object: Record<string, any>, cond: Con
...
+    }
+    if (!Array.isArray(items) || items.length === 0) {
+        throw errors.InternalError;
+    }
+    object[op] = items.map((item: Condition) => {
+        const subFilter: Record<string, any> = {};
+        translateConditions(depth + 1, prefix, subFilter, item);
+        if (Object.keys(subFilter).some(k => supportedStructuralOperators[k])) {
+            throw errors.InternalError;
+        }
```
francoisferrand: `depth` may be used address this without parsing the "list" a second time : just add a guard `if (depth > 0)` to fail at the beginning?

```suggestion
function _assignStructuralOperator(depth: number, prefix: string, object: Record<string, any>, op: string, items: Condition): void {
      if (depth > 0) {
           throw errors.InternalError;
      }
```

Or do we want to support those at any level, just not too one inside the other? Also not sure the "limit" is correct:
- seems to work trivially: the code is actually simpler if we allow nested ops, and I don't see the risk we have in supporting these?
- the current check works only if two "structured" ops are nested directly: `$or: { $or: { "foo", "bar" }, "toto" }` ; but it still allows more complex nested: `$or: { $exists: $or: { "foo", "bar" } }`

lib/storage/metadata/mongoclient/utils.ts
```diff
@@ -75,6 +75,30 @@ function _assignCondition(prefix: string, object: Record<string, any>, cond: Con
...
+* Translates a structural operator into a mongodb filter.
+ * Each item is translated on its own, sharing the parent prefix.
+ * Nesting is not supported now: a translated item that still carries a
+ * structural operator (at any depth) is rejected.
+ */
+function _assignStructuralOperator(depth: number, prefix: string, object: Record<string, any>, op: string,
+    items: Condition): void {
+    if (op !== '$or') {
+        throw errors.InternalError;
+    }
```
francoisferrand: cannot real happen, since this is only called on `supportedStructuralOperators[]` condition ; and listing ops here explicitely is the exact thing we trying to avoid by having the `supportedStructuralOperators` map...

```suggestion
```

lib/storage/metadata/conditions.js
```diff
@@ -7,6 +7,11 @@ const supportedOperators = {
     $gte: true,
     $lt: true,
     $lte: true,
+    $exists: true,
+};
+
+const supportedStructuralOperators = {
+    $or: true,
```
francoisferrand, line 14: should we not add `$and` as well, as it works exactly the same?

lib/storage/metadata/mongoclient/utils.ts
SylvainSenechal: Let's remove the depth restriction its easier

lib/storage/metadata/conditions.js
SylvainSenechal: actually yeah I'm adding it, there was nothing to change to the code to add it beside here in this struct

lib/storage/metadata/mongoclient/MongoClientInterface.ts
SylvainSenechal: I want to delay all 3 for a next pr, I think modifications are moderate, but with new tests it will add ~200/500 lines of code. 

As you said : putObjectNoVer and VerCase2 are not used on non versioned objects

verCase4 is also not used : repairMaster=true is only set by cloudserver when objMd is undefined, and we only set the condition when objMd is defined. 

Tbh we should also probably support conditions for all put, even verCase1 that creates a version, could technically need conditions, I mean if its passed as a parameter of the function it should work...

I think it's tricky to work with these suite of functions because they take this very generic params.conditions parameter without using it, yet these function seems quite specialized so its debatable if we wanna support for conditions

francoisferrand: > verCase1 that creates a version, could technically need conditions

it cannot AFAIU:
- it is used exclusively for _creating_ a new ObjMD, and allocates a new versionID
- if there is a conflict on versionID (i.e. some race condition where 2 path try to create the an ObjMD with the same versionID - or if for some reason an ObjMD with the same versionID already exists), it will fail loudly
- the condition is a check on the _pre-existing_ content of ObjMD, already in DB

> this very generic params.conditions parameter without using it, yet these function seems quite specialized so its debatable if we wanna support for conditions

I think it is the other-way:
- these functions are not "specialized APIs". They are the different 'cases' that putObject needs to handle. `putObject()` is the actual API.
- `params.conditions` was added there to support conditional updates, so it should be implemented in every _relevant_ code path of putObject. Failing to do that changes the semantics of the flag: it is not "conditional updates", but something very hard to use like "conditional update when not repairing master in a versioned bucket where bucket is not suspended", and which will most likely cause issues if adopted to be used in other cases...
- in the implementation, some code path may not need any change: like when creating a new revision i.e. case1, where the semantics of "conditional update" does not mean anything. But this is an implementation detail (if we implement this with a single mongodb "upsert" call -not saying that we could, since we need to manage master- it would probably be set all the time, and mongo would do the job of ignoring the condition when there is no pre-existing data)

I fully agree there is lot of complexity though, and possibly not needed: the whole code kind of tries to rebuild the intent of the code through an reduced interface, where perhaps it would have been much simpler to have separate API for each intent ; and similarly many flags were added to cover different needs, which may be reduced to a limited number of simpler APIs...

But changing that is a very large endeavor, to try and understand all the cases and ways flags interact; so in the mean what we need to do is make our changes clean : so if we add a flag, make it behave consistently in every case...

> verCase4 is also not used

as you confirmed the new flag is not used, indeed that gives us some time to finish implementation and we can skip.

tests/unit/storage/metadata/mongoclient/utils.spec.js
```diff
@@ -215,6 +215,147 @@ describe('translate query object', () => {
...
+                        { $or: [
+                            { 'value.fieldB': { $exists: false } },
+                            { 'value.fieldC': { $lte: 'xyz' } },
+                        ] },
+                    ],
+                },
+            },
+        ],
+        [
+            'should throw when two sibling fields both contain a structural operator of the same type',
```
francoisferrand, line 301: why does this throw : it looks like a valid expression to me, should it not work?

we support (or do we actually?) a query on multiple fields: `{ fieldA: 4, fieldB: "foo" }`, so we should support when the value on each field is an expression?

(maybe not blocking, but weird enough that we should look at it, at least in a followup...)

francoisferrand (approved):

tests/unit/storage/metadata/mongoclient/utils.spec.js
SylvainSenechal: I'm gonna document it in the code, its a limitation of the current algo used, but we can still express the same thing by using explicit $and

delthas (approved):
LGTM. Not a fan of the pseudo abstraction we're extending here (that wants Mongo operators, so not super abstract in the end), but that's the code we have 😛 

Looks like prettier is complaining. Maybe add a "run prettier" commit before all your commits first then ensure each commit is prettier-passing?

SylvainSenechal (approved):

PR #2685, opened 2026-08-12
ARSN-620 Fix lifecycle listings stuck on PHD master keys

nicolas2bert:
**Problem**

Lifecycle listings can get stuck forever on PHD master keys.

A PHD (placeholder) master is written when the current version of a key is deleted. A repair job later promotes the newest surviving version back to master. Until the repair runs, the master key holds a PHD value. If the repair never runs, or if no version survives, the PHD stays.

DelimiterVersions accepted a PHD master and skipped it without touching the resume marker. That is fine for a plain version listing, which has no scan budget. It is not fine for the two lifecycle listings, DelimiterNonCurrent and DelimiterOrphanDeleteMarker. Both count every scanned entry against maxScannedLifecycleListingEntries and truncate when the budget runs out.

So a run of PHD masters longer than the scan limit burns the whole budget without moving the marker. The listing returns IsTruncated: true with no NextKeyMarker. The next listing starts at the same place and returns the same thing. Lifecycle never gets past that point, and nothing beyond it is ever expired.

**Example**

Keyspace (v0), scan limit 3:

phd-1 .. phd-6   dangling PHD masters
zebra\0v1        current
zebra\0v2        noncurrent -> must be expired

_Before:_

page 1: scans phd-1, phd-2, phd-3 -> truncated, no NextKeyMarker
page 2: identical request -> identical result
...forever. zebra\0v2 is never expired.

_After:_

page 1: phd-1, phd-2, phd-3 -> truncated, NextKeyMarker=phd-2
page 2: phd-3, phd-4, phd-5 -> truncated, NextKeyMarker=phd-4
page 3: phd-5, phd-6, zebra\0v1, zebra\0v2 -> zebra\0v2 listed

**What changed**

DelimiterVersions — new handlePHDMaster hook

Both PHD branches (keyHandler_NotSkippingV0 and keyHandler_NotSkippingV1) now call handlePHDMaster(key, versionId, value) instead of returning FILTER_ACCEPT directly.

The default implementation returns FILTER_ACCEPT and changes no state. Plain version listings behave exactly as before.

DelimiterOrphanDeleteMarker — treat a PHD as a key transition

The override runs the same new-key branch as addVersion():

- It advances prevKeyName / keyName, so truncation has a resume position.
- It emits the held delete-marker candidate first.
- It sets value = null, so a PHD is never listed as an orphan delete marker.

The second point matters. A delete marker is held in memory until an entry of another key proves it is an orphan. A PHD master is such an entry. If the marker moved past the candidate without emitting it, the candidate would never be scanned again, and the orphan delete marker would never expire.

scan limit 3, keyspace: banana\0v1 (delete marker), phd-1, phd-2, ...
banana\0v1 -> held as candidate
phd-1      -> new key: banana is proven orphan -> emitted, then marker moves

DelimiterNonCurrent — keep the marker one PHD key behind

The override sets nextKeyMarker to the previous PHD key, never to the key being scanned, and clears nextVersionIdMarker. The marker only moves forward.

A marker on the key being scanned would give the next listing a key-marker with no version-id-marker. S3 reads that as "start after every version of this key". The versions of a PHD master that still has some would be skipped, and they would never be expired. One key behind costs one re-scanned entry per truncation, and skips nothing.

The override leaves prevKey and staleDate untouched. This is deliberate. The next version key scanned is the newest surviving version under the PHD — the one the repair promotes back to master. It must stay classified as current:

apple (PHD)  -> marker left behind apple, prevKey untouched
apple\0v1    -> first version seen for apple -> current, protected
apple\0v2    -> noncurrent -> expirable, staleDate = v1's date

Setting prevKey = 'apple' would make apple\0v1 look noncurrent. Noncurrent version expiration would then delete the very version the repair needs to promote. That is data loss.

One fallback remains: if the scan limit lands on the first PHD of the listing, there is no previous PHD and no marker yet. The marker then points at that key, and that one key loses its versions for this pass. Progress matters more than one missed pass, and the alternative is the infinite loop described above.

**Tests**

- delimiterVersions.spec.js — guard: a PHD master is still not listed and still does not advance the marker in a plain version listing.
- delimiterNonCurrent.spec.js — marker set inside a PHD run; the newest surviving version under a PHD stays protected; crawl across a PHD run longer than the scan limit with marker feedback; scan limit landing exactly on a PHD.
- delimiterOrphanDeleteMarker.spec.js — NextMarker set inside a PHD run; a dangling PHD is never listed as an orphan; crawl across a PHD run; a held candidate is emitted when the PHD run begins.

All cases run against both v0 and v1 bucket formats.


lib/algos/list/delimiterNonCurrent.js
```diff
@@ -144,6 +147,70 @@ class DelimiterNonCurrent extends DelimiterVersions {
...
+     * prevKey='apple' here would classify apple\0v1 as noncurrent. NCVE would
+     * then expire the very version the repair needs to promote: data loss.
+     *
+     *  @param {String} key   - The PHD master key
+     *  @param {String} versionId - always undefined for a master key
+     *  @param {String} value - The PHD placeholder value
+     *  @return {number} - filter return value
+     */
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    handlePHDMaster(key, versionId, value) {
```
jonathan-gramain, line 198: I don't understand the reason to pass `versionId`,  I believe PHDs always has an undefined version ID here as their key doesn't have a version ID embedded (there's one within the metadata for internal purposes but it's not relevant for listing purposes).

Also, for a similar reason I think we don't need to pass `value` which is only internal to metadata, so we should be good with passing just the key.

lib/algos/list/delimiterNonCurrent.js
```diff
@@ -144,6 +147,70 @@ class DelimiterNonCurrent extends DelimiterVersions {
...
+            // Move the marker forward only. prevPHDKey can hold a key from an
+            // earlier run of PHDs that the listing already passed.
+            this.nextKeyMarker = this.prevPHDKey;
+            this.nextVersionIdMarker = undefined;
+        } else if (!this.nextKeyMarker) {
+            // No previous PHD, and no marker yet. Skip this key's versions
+            // rather than leave the listing unable to move forward.
+            this.nextKeyMarker = key;
+            this.nextVersionIdMarker = undefined;
+        }
```
jonathan-gramain, line 209: I'm not sure we should keep this second condition block, which seems to add the possibility to skip legit noncurrent versions (i.e. keeps a buggy behavior). My reasoning is, if we're there it means:
- `addVersion` wasn't called yet since it would have set `nextKeyMarker`
- hence either it's the first PHD
  - then no risk of having a stuck listing at this point
- or all we have seen so far are PHDs
  - then `prevPHDKey` has already been set earlier, and likely `nextKeyMarker` as well to the previous PHD key, starting from the 2nd PHD seen, so no risk of stuck listing either

So removing this block may remove the remaining buggy situation as `nextKeyMarker` will correctly be set already to the previous PHD.

Not 100% sure, please double check my reasoning.

lib/algos/list/delimiterNonCurrent.js
```diff
@@ -144,6 +147,70 @@ class DelimiterNonCurrent extends DelimiterVersions {
...
+     *
+     * THE BUG IT FIXES: a run of dangling PHD masters longer than
+     * maxScannedLifecycleListingEntries truncated the listing with no
+     * NextKeyMarker. The next page then repeated the first page, and the
+     * listing never moved forward.
+     *
+     * THE RULE: the marker moves to the PREVIOUS PHD key. It never points at
+     * the key being scanned. A marker on the scanned key gives the next listing
+     * a key-marker with no version-id-marker. S3 reads that as "start after
+     * every version of this key". The listing would then skip the versions of a
```
jonathan-gramain, line 161: This comment looks suspicious to me: "S3 reads that as ...":
- on one hand, "S3" is too generic and doesn't help the understanding, we could say "the listing algorithm" for example. If S3 refers to Cloudserver, I believe it just passes the received marker values as is to the next call, without further interpretation.
- on the other hand, I believe (and hope) it's not true: `DelimiterVersions` will scan all versions starting at `NextKeyMarker` if `NextVersionIdMarker` is not present (but it will skip the master).

I believe a more correct comment should say that the listing would skip the "master version of" and leave it unable to recognize that the upcoming version is the new current version. But it should still see it.

lib/algos/list/delimiterNonCurrent.js
```diff
@@ -144,6 +147,70 @@ class DelimiterNonCurrent extends DelimiterVersions {
...
+     * PHD master that still has some, and NCVE would never see them. One key
+     * behind costs one re-scanned entry per truncation, and skips nothing.
+     *
+     * Example, scan limit 3, dangling PHD masters phd-1 ... phd-6:
+     *   page 1: phd-1, phd-2, phd-3 -> truncated, NextKeyMarker=phd-2
+     *   page 2: phd-3, phd-4, phd-5 -> truncated, NextKeyMarker=phd-4
+     *   page 3: phd-5, phd-6        -> done
+     *
+     * THE FALLBACK: on the first PHD of a listing there is no previous PHD key,
```
jonathan-gramain, line 170: cf my comment, I believe the fallback should be removed.

lib/algos/list/delimiterNonCurrent.js
```diff
@@ -144,6 +147,70 @@ class DelimiterNonCurrent extends DelimiterVersions {
...
+     * surviving version under the PHD, and the repair promotes it back to
+     * master. Untouched state keeps that version classified as current, so the
+     * listing never returns it as an expirable noncurrent version.
+     *
+     * Example, a PHD master with two surviving versions:
+     *   apple (PHD)  -> marker stays behind apple, prevKey untouched
+     *   apple\0v1    -> first version seen for apple -> current, protected
+     *   apple\0v2    -> noncurrent -> expirable, staleDate = v1's date
+     *
+     * apple\0v1 is not deduplicated as the master copy: a PHD gets its
```
jonathan-gramain, line 187:
```suggestion
     * apple\0v1 is not duplicated as the master copy: a PHD gets its
```

lib/algos/list/delimiterOrphanDeleteMarker.js
```diff
@@ -171,6 +171,57 @@ class DelimiterOrphanDeleteMarker extends DelimiterVersions {
...
+    handlePHDMaster(key, versionId, value) {
+        if (key !== this.keyName) {
+            if (this.value) {
+                this._addOrphan();
+            }
+            this.prevKeyName = this.keyName;
+            this.keyName = key;
+            this.value = null;
+        }
+        return FILTER_ACCEPT;
```
jonathan-gramain, line 222: maybe simpler:

```suggestion
        this.addVersion(key, null, null);
        return FILTER_ACCEPT;
```

Should be equivalent since PHDs always have a different key than the previous entry.

lib/algos/list/delimiterNonCurrent.js
nicolas2bert: - bien vu pour "S3". c'est pas CloudServer, lui il passe juste le marker tel quel. celui qui l'interprete vraiment c'est DelimiterVersions.genMDParamsV0(). renommé, le commentaire dit ça maintenant.
- par contre sur le comportement, un key-marker sans version-id-marker skip toute la key, versions incluses:


```
if (this.versionIdMarker) {
    params.gte = `${this.keyMarker}${VID_SEP}`;        // key\x00 -> demarre SUR la premiere version
} else {
    params.gt  = `${this.keyMarker}${inc(VID_SEP)}`;   // key\x01 -> demarre APRES toutes les versions
}
```

lib/algos/list/delimiterNonCurrent.js
nicolas2bert: you are right, removing them.

lib/algos/list/delimiterNonCurrent.js
nicolas2bert: - you're right, I removed it.
- on thing but I don't think it will be an issue since it requires: maxScannedLifecycleListingEntries = 1. 
On the very first PHD we can still truncate with no marker, if the budget dies on the next entry. that needs maxScannedLifecycleListingEntries = 1. and at 1 nothing advances anyway: the resumed listing re-reads the marker version to get staleDate, that's the whole budget. 
The old code loops there too, just one key later, so the fallback was covering a config that is already broken.

lib/algos/list/delimiterOrphanDeleteMarker.js
nicolas2bert: thanks

lib/algos/list/delimiterNonCurrent.js
```diff
@@ -144,6 +147,94 @@ class DelimiterNonCurrent extends DelimiterVersions {
...
+     *
+     *   apple (PHD)  -> marker moves, prevKey untouched
+     *   apple\0v1    -> first entry seen for apple -> current, not listed
+     *   apple\0v2    -> same key again -> noncurrent -> listed, staleDate = v1
+     *
+     * (`apple\0v1` is not dropped as the master copy the way a normal master
+     * would be: a PHD gets its versionId at delete time, and that id matches no
+     * version key.)
+     *
+     * If this method set prevKey = 'apple', addVersion() would see `apple\0v1`
```
jonathan-gramain, line 214:
```suggestion
     * If this method sets prevKey = 'apple', addVersion() would see `apple\0v1`
```

jonathan-gramain (approved):

lib/algos/list/delimiterNonCurrent.js
```diff
@@ -144,6 +147,94 @@ class DelimiterNonCurrent extends DelimiterVersions {
...
+     *     call 3: phd-5, phd-6        -> done
+     *
+     * WHY THE MARKER POINTS AT THE PREVIOUS PHD, NOT THE ONE JUST SCANNED
+     * A marker that carries a key but no version id means "I am done with this
+     * key, skip it completely". genMDParamsV0() turns it into the range
+     * `gt: <key>\x01`. Version keys look like `<key>\x00<versionId>`, and 0x00
+     * sorts below 0x01, so that range starts past every version of the key.
+     *
+     * The listing is not done with the PHD it just scanned. That PHD may still
+     * have surviving versions below it, and NCVE has to see them. If the marker
```
anurag4DSB, line 188: what is NCVE?

anurag4DSB (approved):

lib/algos/list/delimiterNonCurrent.js
anurag4DSB: got it its the NoncurrentVersionExpiration

nicolas2bert:
@bert-e approve

nicolas2bert:
ping

PR #2688, opened 2026-08-19
ARSN-617: clean-read filter hiding non-localized versions

DarkIsDude:
## Context

https://scality.atlassian.net/browse/ARSN-617

In a clean room, metadata is replicated before the object data is copied locally: until then the version's `dataStoreName` still points at the remote production site. Such a version is **non-localized** and must not be visible to clients.

## What this adds

A per-call `hideNonLocalizedVersions` flag on the metadata read and listing APIs (`getObject`, `getObjects`, `getBucketAndObject`, `listObject`, `listMultipartUploads`).

## Notes for ARSN-618

ARSN-618 answers the same "which version is current" question at write time, unconditionally, and can reuse the condition helpers and the `getLatestVersion` filter parameter added here. Read-time resolution stays necessary for entries that never traverse the write path (mongodump/mongorestore bootstrap and re-bootstrap loads); since the design rejected read-time master filtering, the design and ARSN-618 should record that it now exists. We can also decide to drop it as this use case should not exist in a DR architecture.


lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -301,6 +308,7 @@ class MongoClientInterface {
         this.replicationGroupId = replicationGroupId;
         this.database = database;
         this.isLocationTransient = isLocationTransient;
+        this.getLocationConstraints = getLocationConstraints;
```
francoisferrand: this does not look the a list of "location constraints" : but really the actual list/map of locations.

("Location constraint" is a parameter of a bucket, which indicates in AWS in which region the bucket is; and which we use instead to track the _default_ data location for the bucket)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1512,6 +1520,37 @@ class MongoClientInterface {
...
+     * markers and PHD keys carry no dataStoreName, are matched by $nin, and
+     * are therefore never hidden.
+     * @param {Boolean} [hide] whether the versions must be hidden
+     * @return {Object | null} query fragment, or null when nothing to filter
+     */
+    private nonLocalizedFilter(hide?: boolean): object | null {
+        if (!hide) {
+            return null;
+        }
+        const excluded = this.nonLocalizedLocations();
```
francoisferrand: list of locations do not change, best to compute this list just once in constructor

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2434,6 +2485,27 @@ class MongoClientInterface {
     ) {
         const c = this.getCollection<ObjectMetastoreDocument>(bucketName);
         const getLatestVersion = this.getLatestVersion;
+        // when the master keys are resolved, the stream must keep the
+        // non-localized masters so that they can be replaced by the newest
+        // localized version of their object
```
francoisferrand: not needed : c.f. design and next ticket ARSN-618, the master **_is not updated_** when inserting a non-localized version... so only versions need filtering.

or is there something which was missed in the design?

lib/storage/metadata/mongoclient/Mongoclient.md
```diff
@@ -330,3 +330,51 @@ In non versioned buckets, master object events are the ones to be processed.
...
+(`hideNonLocalizedVersions`) on the read and listing APIs — `getObject`, `getObjects`,
+`getBucketAndObject`, `listObject`, `listMultipartUploads` — so any metadata
+backend can implement the same contract. `MetadataWrapper` sets the flag on
+every such call when the deployment runs with clean read enabled (set on the
+user-facing Cloudserver only: Backbeat's internal Cloudserver must see all the
+entries), and refuses to start when the backend does not implement it.
+
+The location configuration is provided by the embedder through the
+`getLocationConstraints` parameter, evaluated on each call so that
+configuration updates are picked up.
```
francoisferrand: configuration cannot be updated dynamically. The locations are defined in a config file, and service is restarted when the list changes.

lib/storage/metadata/mongoclient/Mongoclient.md
```diff
@@ -330,3 +330,51 @@ In non versioned buckets, master object events are the ones to be processed.
...
+  overwrites it nor is overwritten by it;
+- delete markers and PHD keys carry no `dataStoreName`, are matched by `$nin`,
+  and are therefore never hidden;
+- a master listing (`DelimiterMaster`) keeps the objects whose current version
+  is not localized, and lists them with their newest localized version — the
+  master key is resolved through `getLatestVersion` on the way out of the
+  stream, the same way a PHD key is resolved. An object with no localized
+  version at all is dropped from the listing;
+- `getObject` resolves a hidden master the same way, so a listing and a `GET`
+  agree on which version is current;
```
francoisferrand: The current version **_IS_** the localized version (the non-localized version in an "internal" version, so to say):
- `getObject` should only point to latest localized version
- `GetObjectVersion` returns the version, whether it is localized or not

Backbeat needs (internally) be able to see all the versions, but I don't think it needs the master: thinking of the use cases, maybe it is actually safer if even backbeat does not _list_ these objets:
- Lifecycle must not delete these objects (or delete the next "non-current" versions), 
- Bucket notification acts on versions, no listing or master access - so not affected
- Replication cannot (yet) replicate the objects, so should ignore them
- Pull-replication does not perform listing or access master, but **must** be able to both read and update the metadata of any version
- Listing versions may be needed for tooling (e.g. finding non-localized objects), but this should not affect listing of masters

Some filtering will be done in backbeat indeed, and we probably need to "strict" when answering to external client (i.e. reject reading object/version if not localized) ; but we should avoid complexfying the code for path which do not need it (e.g. listing master or resolving non-localized master): are there situations where do you think this is needed?

lib/storage/metadata/mongoclient/Mongoclient.md
```diff
@@ -330,3 +330,51 @@ In non versioned buckets, master object events are the ones to be processed.
...
+  documents after the limit would under-fill pages and report a complete
+  listing while entries remain;
+- in a metadata search, the filter is added as an `$and` element after the
+  search query has been merged, so that a search on `dataStoreName` neither
+  overwrites it nor is overwritten by it;
+- delete markers and PHD keys carry no `dataStoreName`, are matched by `$nin`,
+  and are therefore never hidden;
+- a master listing (`DelimiterMaster`) keeps the objects whose current version
+  is not localized, and lists them with their newest localized version — the
+  master key is resolved through `getLatestVersion` on the way out of the
```
francoisferrand: the master should always point to a localized version, per the design : next (arsenal) ticket is about making sure putObject handles this

maeldonn:
Please remove the footer `Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>` from your commits 🙏

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2434,6 +2485,27 @@ class MongoClientInterface {
...
+            if (nonLocalizedLocations.length === 0) {
+                return false;
+            }
+            // in v0 the master and version keys share the same range: only the
+            // master keys, which carry no version id, are resolved
+            if (vFormat === BUCKET_VERSIONS.v0 && obj.key.includes(VID_SEP)) {
+                return false;
+            }
+            try {
+                return nonLocalizedLocations.includes(JSON.parse(obj.value).dataStoreName);
```
francoisferrand: this can be quite costly: JSON parse() call (object MD can be large, esp. for bug MPUs) + linear search (though not sure how many CRR locations we would have)

lib/storage/metadata/mongoclient/Mongoclient.md
francoisferrand: > Read-time resolution stays necessary for entries that never traverse the write path (mongodump/mongorestore bootstrap and re-bootstrap loads)

Bootstrap should no copy master either (since they are not localized), and thus master resolution is not needed: the master is only created when the ObjMD is finally "localized" - which happens regular arsenal calls and would be handled by ARSN-618

DarkIsDude: Not at all, as explained in the description. I was not really sure if I should be strict or not, you seems to have the answer. I'll then remove that and have a less defensive code 🙏

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: Removed, only versions are filtered now — `DelimiterMaster` is back to its original state.

lib/storage/metadata/mongoclient/Mongoclient.md
DarkIsDude: Done. One case where it is still needed: when an object has no master at all, `getObject` falls through to `getLatestVersion`, and unfiltered that returns the non-localized version. So the fallback takes the filter and reports `NoSuchKey` — one argument on an existing call, no master resolution.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -301,6 +306,12 @@ class MongoClientInterface {
         this.replicationGroupId = replicationGroupId;
         this.database = database;
         this.isLocationTransient = isLocationTransient;
+        // a version is non-localized when its data location still refers to a remote site; $nin also matches the
+        // empty or absent dataStoreName of delete markers and PHD keys, which must stay visible
+        const nonLocalized = Object.entries(locations ?? {})
```
francoisferrand:
```suggestion
        const nonLocalizedLocations = Object.entries(locations ?? {})
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -301,6 +306,12 @@ class MongoClientInterface {
         this.replicationGroupId = replicationGroupId;
         this.database = database;
         this.isLocationTransient = isLocationTransient;
+        // a version is non-localized when its data location still refers to a remote site; $nin also matches the
+        // empty or absent dataStoreName of delete markers and PHD keys, which must stay visible
```
francoisferrand: comment is hard to relate to the following lines, unless you already know what it means : nothing here creates to "version", "dateStoreName"... So not sure it really helps?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1710,6 +1728,7 @@ class MongoClientInterface {
      * @param {Object} c collection
      * @param {String} objName object name
      * @param {String} vFormat bucket version format
+     * @param {Object | null} nonLocalizedFilter query fragment hiding the non-localized versions
```
francoisferrand: should we really pass `null` ?
Passing `{}` does the same when unpacked, but (a bit) easier to read at the call site (we see an empty object instead of a completely non-descript `null`)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1565,7 +1578,7 @@ class MongoClientInterface {
                     // If no master found then object is either non existent
                     // or last version is delete marker
                     if (!doc || doc.value.isPHD) {
-                        this.getLatestVersion(c, objName, vFormat, log, (err, value?) => {
+                        this.getLatestVersion(c, objName, vFormat, nonLocalizedFilter, log, (err, value?) => {
```
francoisferrand: should we always pass `this.nonLocalizedQuery` ?

seems weird to try to get the master (necessarily non-localized) when no versionID is specified; and falling back to searching for the latest version:
- if we find the master document, we would return it (even it not actually the latest one -this is the expectation of the _clean view_-, i.e. master is the latest localized)
- but if we don't find it, then we look through _all_ versions, and may return a non-localized one

Practically, not sure it really matters (the only non-versioned calls would be from external, i.e. non-localized) ; but does not seem consistent and may thus bite as some point...

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1838,7 +1859,7 @@ class MongoClientInterface {
         vFormat: string,
         log: werelogs.Logger,
     ) {
-        this.getLatestVersion(c, objName, vFormat, log, (err, value?) => {
+        this.getLatestVersion(c, objName, vFormat, null, log, (err, value?) => {
```
francoisferrand: not correct: repair will typically recreate the master (in case of conflict).
since the master is the latest localized version, repair must keep the flag...

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1878,7 +1899,7 @@ class MongoClientInterface {
     ) {
         const masterKey = formatMasterKey(objName, vFormat);
         // Check if there are other versions available
-        this.getLatestVersion(c, objName, vFormat, log, (err, version?) => {
+        this.getLatestVersion(c, objName, vFormat, null, log, (err, version?) => {
```
francoisferrand: same : deleting a PHD should not promote a non-localized master

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2476,7 +2498,7 @@ class MongoClientInterface {
                     transform(obj, encoding, callback) {
                         if (Version.isPHD(obj.value)) {
                             const key = obj.key.slice(DB_PREFIXES.Master.length);
-                            getLatestVersion(c, key, BUCKET_VERSIONS.v1, log, (err, version?) => {
+                            getLatestVersion(c, key, BUCKET_VERSIONS.v1, null, log, (err, version?) => {
```
francoisferrand:
```suggestion
                            getLatestVersion(c, key, BUCKET_VERSIONS.v1, nonLocalizedFilter, log, (err, version?) => {
```

lib/storage/metadata/MetadataWrapper.js
```diff
@@ -160,6 +161,18 @@ class MetadataWrapper {
             this.implName = 'cdmi';
         }
         this._listingParser = params.customListingParser || _parseListEntries;
+        // silently ignoring the flag would expose the versions it must hide
+        if (params.hideNonLocalizedVersions && this.implName !== 'mongoclient') {
+            throw new Error(`hiding the non-localized versions is not supported by the ${this.implName} backend`);
```
francoisferrand: I wonder if we can we really throw here?

it means we must make sure to pass the flag only when the config is correct, and would thus require more logic in any component that use it to consider both if it is needed (depending on the request/...) AND if it is supported (depending on the deployment/backend).

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1620,6 +1633,10 @@ class MongoClientInterface {
         if (objects.length > 1000) {
             return callback(errorInstances.InternalError.customizeDescription('cannot get more than 1000 objects'));
         }
+        // the flag is set per call, hence identical on every entry of the batch
+        const nonLocalizedFilter = objects.some(({ params }) => params?.hideNonLocalizedVersions)
```
francoisferrand: does not very efficient... esp. since it is actually set on _all_ objects (and I see no reason it would be passed differently for each object)

anyway the semantic is non correct (we apply the param to all if present on a single object), so best to keep it single: let's just look at the first object only?
(or find another)

lib/storage/metadata/MetadataWrapper.js
```diff
@@ -160,6 +161,18 @@ class MetadataWrapper {
...
+        }
+        this._hideNonLocalizedVersions = Boolean(params.hideNonLocalizedVersions);
+    }
+
+    _readParams(params) {
+        if (!this._hideNonLocalizedVersions) {
+            return params;
+        }
+        return Object.assign({}, params, { hideNonLocalizedVersions: true });
```
francoisferrand: Should allow explicit override in `params`:
```suggestion
        return Object.assign({ hideNonLocalizedVersions: true }, params);
```

(was written more clearly in the design "_the per-call flag stays available underneath for finer-grained control later (e.g. an administrative route bypassing the filter)._"  than in the ticket: "_Define the clean-read filter at the metadata-interface level, as a **per-call** flag on the read and listing APIs_")

francoisferrand (changes requested):

lib/storage/metadata/MetadataWrapper.js
DarkIsDude: This runs once in the `MetadataWrapper` constructor, and both inputs come from the deployment config: no component has to check backend support at request time. Not sure to follow you on it, as the request is not considered here. Not a risk to check that on the config as we check it only if we have the flag.

lib/storage/metadata/MetadataWrapper.js
DarkIsDude: 👏 thanks

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: 😂 if you start to ask for comment removal, we are done 😂

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: The fallback is filtered too, `nonLocalizedFilter` is passed to `getLatestVersion`. After 418, this is a no-op. But that's a good question I'm open minded on it, even if I think this is the good approach as more accurate and no cost ? The master-maintenance path writes the master, which is by definition the latest localized version, whereas read paths must honour the caller's view.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: Thanks 🙏

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: I think it's better and more accurate to keep null. This is easier for deployment where we don't have CRR location. Let me know 🙏. Also null is already the "no filter" value produced by the constructor (no CRR location) and by the `params?.hideNonLocalizedVersions ? … : null` ternaries

lib/storage/metadata/mongoclient/MongoClientInterface.ts
francoisferrand: > read paths must honour the caller's view

the "master" is computed only from "localized" entries, so:
- it does not depend on the caller's view
- should never return a non-localized ObjMD as master

It particular, it means that if the caller request no filter (i.e. all entries) it would have a different behavior if the master is found (→ ignore newer non-localized entries, since master is the latest localized entry) vs when we hit the fallback (→ find latest entry, including non-localized).

So this is really about consistency in the API : passing the filter was consistent with your initial proposal where we _always_ computed the master; now that we don't, the code should ensure we never bypass that "definition" of master?

lib/storage/metadata/MetadataWrapper.js
francoisferrand: `hideNonLocalizedVersions` is both a deployment-level configuration -so set by a component, based on its configuration- and a per-call override:
- with this assert, this requires some other component (like Cloudserver or ZKOP) to check if the flags are combined: e.g. if we deploy zenko with metadata backend, we introduce some flag where we combine both orthogonal concerns (what kind of MD backend / which instances require clean view) and thus break the MD abstraction.
- similarly (though not asserted in this PR), each call which needs such overrides (in backbeat...) would need to adjust the behavior depending on backend

You are totally correct that _clean view_ does not work if not implemented in the backend; but I fear this is not the right place, and that it would create more work/code in other places to work around the check (for exemple when we try to run Zenko with metadata backend - though that work starts with simpler features). Practically, there are anyway many other things to do if we want to support DR/pull-replication on metadata backend and/or on Zenko: so I don't think there is much risk either if we do not flag this error.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
francoisferrand: my concern was more about the places where the function is called, where `null` is kind of obscure:
```
this.getLatestVersion(c, objName, params.vFormat, null, log, (err, mstObjVal?) => {
```

But maybe the better approach is not using `{}`, but a constant whose name would express what it means: something like:

```
const hideNonLocalizedVersionsFilter = { ... };
const includeNonLocalizedVersionsFilter = null;
this.getLatestVersion(c, objName, params.vFormat, includeNonLocalizedVersionsFilter, log, (err, mstObjVal?) => {
```

(not a blocker though)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1217,7 +1228,7 @@ class MongoClientInterface {
             },
         )
             .then(() =>
-                this.getLatestVersion(c, objName, params.vFormat, log, (err, mstObjVal?) => {
+                this.getLatestVersion(c, objName, params.vFormat, null, log, (err, mstObjVal?) => {
```
francoisferrand, line 1233: nit: not sure about this one (e.g. fine if we write a localized object; but if we write a non-localized one not sure we should on reach in putObjectVerCase4?), but should be handle in next PR, covering the write path and update of master.

tests/unit/storage/metadata/mongoclient/hideNonLocalizedVersions.spec.js
francoisferrand, line 1: I don't think these test cover all the fallback/corner cases : i.e. the cases where we fail to get master, and end up calling GetLatestVersion

(and I wonder if it would not be better to move these tests into each "method"'s test, so the behavior of each API is documented in a single place...)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
francoisferrand: (this applies to all `this.getLatestVersion()` calls, i.e. the corner cases where we dynamically recompute the master)

DarkIsDude: Agreed and fixed, thanks. Thanks 🙏

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: Good point. Not sure we should define a new const, but improve the doc ? It's what I did.

lib/storage/metadata/MetadataWrapper.js
DarkIsDude: Right, and the throw only covered the deployment flag — a per-call one reached an unsupported backend anyway: now it logs an error and leaves the flag off, and asks the backend for the support instead of its name. Make sense for you ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: Agreed, that's the next one: writing a non-localized version shouldn't move the master at all. The JSDoc now records why it passes null until then.

maeldonn:
This PR is very difficult to review.
- Prettier formatting: Please group all formatting into a single initial commit (or ideally, a separate stacked PR).
- History cleanup: The first commit modifies a file that doesn't actually show changes in the final diff. We should clean up the git history whenever there are large, unnecessary changes like this.

DarkIsDude:
> This PR is very difficult to review.
> 
>     * Prettier formatting: Please group all formatting into a single initial commit (or ideally, a separate stacked PR).
> 
>     * History cleanup: The first commit modifies a file that doesn't actually show changes in the final diff. We should clean up the git history whenever there are large, unnecessary changes like this.

I'm not sure it's in our guidelines to have one dedicated commit for all prettier. The guidelines is put prettier in dedicated commit. Several if needed as you can work from one file to a new file during dev time. Anyway, Claude can rework the history easily so it's done. Do you think we could have split this PR in several part ?

lib/storage/metadata/mongoclient/Mongoclient.md
```diff
@@ -330,3 +330,33 @@ In non versioned buckets, master object events are the ones to be processed.
...
+
+Clean read hides those versions from the clients. It is a **per-call flag**
+(`hideNonLocalizedVersions`) on the read and listing APIs so any metadata
+backend can implement the same contract. `MetadataWrapper` sets the flag on
+every such call when the deployment runs with clean read enabled (set on the
+user-facing Cloudserver only: Backbeat's internal Cloudserver must see all the
+entries), and logs an error leaving it off when the backend does not
+implement it.
+
+Only the versions are filtered: the master key always points at a localized
```
maeldonn: `copyVersionOverMaster` in the functional tests models the opposite case: a master restored from the source site, still non-localized. There, `getObject` resolves to the localized version but `DelimiterMaster` returns the master as-is, so ListObjects and GetObject disagree on the same key.

Filter the master listing too, or document the gap here.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2669,6 +2711,7 @@ class MongoClientInterface {
         const internalParams = {
             mainStreamParams: extensionParams,
             mongifiedSearch: params.mongifiedSearch,
+            hideNonLocalizedVersions: params.hideNonLocalizedVersions,
```
maeldonn: Are you sure the flag is meant to reach the MPU listing? The shadow bucket holds no versions, and the doc says only versions are filtered.

As far as I can tell an overview key takes its `dataStoreName` from the controllingLC, which `controllingBackendParam` rejects when the location is `isCRR`, so I don't see what this would filter out. Is there a case I'm missing?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -259,6 +262,7 @@ class MongoClientInterface {
     private replicationGroupId: string;
     private database: string;
     private isLocationTransient: Function;
+    private nonLocalizedQuery: object | null;
```
maeldonn, line 265:
```suggestion
    private nonLocalizedQuery: Filter<ObjectMetastoreDocument> | null;
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1718,6 +1743,7 @@ class MongoClientInterface {
         c: Collection<ObjectMetastoreDocument>,
         objName: string,
         vFormat: string,
+        nonLocalizedFilter: object | null,
```
maeldonn, line 1746: nit: could be refactored with an option object

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1710,6 +1730,11 @@ class MongoClientInterface {
      * @param {Object} c collection
      * @param {String} objName object name
      * @param {String} vFormat bucket version format
+     * @param {Object | null} nonLocalizedFilter query fragment hiding the
+     * non-localized versions. Always the deployment filter, never the caller's
+     * view: the version returned stands in for the master, the newest localized
+     * version. `null` filters nothing: no `isCRR` location, or a write path
+     * waiting on ARSN-618
```
maeldonn, line 1737:
```suggestion
     * @param {Object | null} nonLocalizedFilter query fragment hiding the non-localized
     * versions, `null` to filter nothing.
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2434,6 +2461,9 @@ class MongoClientInterface {
     ) {
         const c = this.getCollection<ObjectMetastoreDocument>(bucketName);
         const getLatestVersion = this.getLatestVersion;
+        const nonLocalizedFilter = params.hideNonLocalizedVersions ? this.nonLocalizedQuery : null;
+        // the master is the newest localized version, whatever the caller's view
+        const phdResolutionFilter = this.nonLocalizedQuery;
```
maeldonn, line 2466: What is PHD ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1565,7 +1580,8 @@ class MongoClientInterface {
                     // If no master found then object is either non existent
                     // or last version is delete marker
                     if (!doc || doc.value.isPHD) {
-                        this.getLatestVersion(c, objName, vFormat, log, (err, value?) => {
+                        // the master is the newest localized version, whatever the caller's view
```
maeldonn, line 1583: this comment is duplicated, does it mean we should do something ?

maeldonn (approved):
LGTM, left a few minor nits. Logic looks fine, but since I'm not an expert in this area, please wait for a third review from the team.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
delthas: Master PlaceHolDer (IIRC when we delete the latest version and must repair the master to point to the previous latest version)

delthas (commented):
LGTM code-style wise but this is quite subtle so needs a review from Francois.

(Leaving the approves to Mael and Francois)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: These three call sites are the only ones that pass `this.nonLocalizedQuery` instead of the caller-dependent `nonLocalizedFilter`, and the reason has to be visible at each of them.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: I thought about it but the method is a public one that will ask a global refacto. Happy to do it, but should we ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1565,7 +1581,8 @@ class MongoClientInterface {
...
```
francoisferrand: nit: for consistency, should be "latest"
```suggestion
                        // the master is the latest localized version, whatever the caller's view
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
francoisferrand: so should we not pass `this.nonLocalizedQuery` here as well, to consistently return a non-localized master?
(and thus better not to add a new parameter from `getLatestVersion` at all, so we don't unexpectedly return a non-localized latest version - and have much less impact on the existing `xxxObjects.spec.js`)

```suggestion
                this.getLatestVersion(c, objName, params.vFormat, this.nonLocalizedQuery, log, (err, mstObjVal?) => {
```

(write will come indeed in next PR, and probably affect which case of putObjectVer... needs to be used when creating a new non-localized ObjMD or updating an existing version with localized ObjMD)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2121,7 +2146,7 @@ class MongoClientInterface {
                     // getting the last version if master not found
                     // (either object non existent or last version is a delete marker)
                     if (!mst) {
-                        return this.getLatestVersion(c, objName, params.vFormat, log, (err, version?) => {
+                        return this.getLatestVersion(c, objName, params.vFormat, null, log, (err, version?) => {
```
francoisferrand: should we not pass this.nonLocalizedQuery here as well?

```suggestion
                        return this.getLatestVersion(c, objName, params.vFormat, this.nonLocalizedQuery, log, (err, version?) => {
```

tests/functional/metadata/mongodb/hideNonLocalizedVersions.spec.js
```diff
@@ -0,0 +1,527 @@
...
+
+    /**
+     * Writes a localized version of an object, through the regular write path:
+     * the master key follows the version.
+     * @param {String} objName - object key
+     * @param {Object} [extraMD] - additional object metadata fields
+     * @param {Function} cb - callback(err, versionId)
+     * @return {undefined}
+     */
+    function putLocalizedVersion(objName, extraMD, cb) {
```
francoisferrand: nit: should we not write these function (in the test file) with async/promises instead of callback?

```suggestion
async function putLocalizedVersion(objName, extraMD) {
    [...]
    return await promisify(cb => metadata.putObjectMD(BUCKET_NAME, objName, objVal, versionParams, logger, cb)();
}
```

tests/unit/storage/metadata/MetadataWrapperHideNonLocalized.spec.js
```diff
@@ -0,0 +1,119 @@
...
+    it('should ignore the flag on a backend not implementing it', () => {
+        const error = sinon.stub(logger, 'error');
+        const metadata = new MetadataWrapper('mem', { hideNonLocalizedVersions: true }, null, logger);
+        assert.strictEqual(metadata._hideNonLocalizedVersions, false);
+        assert.strictEqual(error.calledOnce, true);
+    });
+
+    it('should set the flag on the read and listing calls', done => {
+        const metadata = buildWrapper(true);
+        metadata.getObjectMD(bucketName, objName, {}, logger, () => {
```
francoisferrand: `async.series` or `async` would be better, to allow "linear" code

lib/storage/metadata/mongoclient/MongoClientInterface.ts
SylvainSenechal: nit: I understood because i spent some time on this thread, more direct comment version :

"ignore hideNonLocalizedVersions parameter and always query the newest localized version which is the master"

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1710,6 +1730,11 @@ class MongoClientInterface {
      * @param {Object} c collection
      * @param {String} objName object name
      * @param {String} vFormat bucket version format
+     * @param {Object | null} nonLocalizedFilter query fragment hiding the
```
SylvainSenechal: nit : 
almost feels like this could be a boolean "fetchNonLocalizedVersions". 
then the function would use this.nonLocalizedQuery ?

SylvainSenechal: Ok I think thats what you do for other functions (and you dnot even pass the parameter), but here, ai said its more complicated because of the way this function is used, doesnt have access to this, whatever its fine

lib/storage/metadata/mongoclient/readStream.js
SylvainSenechal, line 85: Non related but found by ai while discussing about this file : 

When searchOptions countains a top level $or, it will overwrite this query.$or and deleted elements will show up in the results. 
Will create a ticket to double confirm it and fix if its confirmed

SylvainSenechal (approved):
barely any comments from me, I think it's ok

tests/unit/storage/metadata/mongoclient/hideNonLocalizedVersions.spec.js
```diff
@@ -0,0 +1,260 @@
...
+        // the fallback stands in for the master, so it filters whatever the caller asked for
+        function stubFallback(master) {
+            const collection = { findOne: () => Promise.resolve(master) };
+            sinon.stub(client, 'getCollection').callsFake(() => collection);
+            return sinon.stub(client, 'getLatestVersion').callsFake((...args) => args[5](null, {}));
+        }
+
+        it('should exclude the non-localized locations when the flag is set', done => {
+            captureFilter(filter => {
+                assert.deepStrictEqual(filter['value.dataStoreName'], { $nin: ['dr-source'] });
```
SylvainSenechal, line 54: Nit: 
I think the current structure assert that if findOne is called, we have to verify whats in the deepStrictEqual.
But if findOne is never called, the test also passes. 
I think there are some if statement in this clientInterface that can trigger early return so ideally we should assert findOne is called, but I think you can also leave as it is

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: Done — and merged with Sylvain's wording on the same comment: `// ignore hideNonLocalizedVersions: the master is the latest localized version`, applied to the three call sites.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: Good point, made it direct: `// ignore hideNonLocalizedVersions: the master is the latest localized version` (kept "latest" for consistency, as asked in the other thread).

tests/unit/storage/metadata/mongoclient/hideNonLocalizedVersions.spec.js
DarkIsDude: Fixed: `findOne` is now a spy and the helper asserts `callCount === 1`, so a test that never reaches the query fails instead of passing silently.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: 🤝 agree, resolving the master against the localized versions only.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: On this one I should say no. We are in a delete here, if we have no localized, what should be the behaviour ?
- If the answer is refuse, then yes, we should use it.
- If the answer is accept, then no, we should not use it. Thanks to that we can delete.

Maybe we have a following JIRA for that and the topic must be moved after ?

tests/functional/metadata/mongodb/hideNonLocalizedVersions.spec.js
DarkIsDude: I'm always in favor of refactoring. I had that in mind but other reviewers tell us that it was already hard to review, so I tried to keep the diff as small as possible 🙏. Done

lib/storage/metadata/mongoclient/MongoClientInterface.ts
francoisferrand: ah, this function is not very clear: I would need a table I think...

- if the intent is to "delete the object", i.e. no versionId specified
  → should only delete only the master, and try to lookup only localized objects
  → if no master key and no localized master is found, then it is indeed an error ("not found")
- however it seems we can pass a versionId, so this is an actual version deletion: it tries to get the master to see if it needs to go through this.deleteObjectVerMaster or this.deleteObjectVerNotMaster - likely so that the master document can be updated when the version we delete _is_ the master
   → if the _version_ is not localized, we should go in the "not master" path: it can't be the master, so no need to even try. So should lookup non-localized.
   → I guess we could even optimized it, and skip the whole master lookup if we see the object being detected is non-localized
   → need to review the `deleteObjectVerNotMaster` path though, to ensure nothing breaks when deleting a version "newer" than the master

That said, there is some contention with the write path here (i.e. next PR) : a non-localized object cannot be master -so should pass `this.nonLocalizedQuery`, and we can drop that parameter I think since this is the last place it is used-, but _today_ the write path has not been updated so it could happen...

➡  What do you think?
➡ I this reasoning is confirmed, I would prefer keeping the code simple/naive, i.e. a simple invariant "master is non localized" (so getLatestVersion only returns non-localized versions) ; and we close the gap (if any) ideally in the write-path PR, or in another followup if it seems bigger or isolated?

SylvainSenechal: Joining as im working on the next pr ARSN 618 for writes

It seems the problem we have may be simplified if we consider this invariant "any non localized object version will never be a master". It should be possible to implement (and its the design actually), but I think it there is one place where we could end up with non localized master : During the mongo dump / bootstrap, so not sure the invariant can hold, unless we explicitly develop the bootstrap to not create non localized masters 🤔

francoisferrand: > During the mongo dump / bootstrap, so not sure the invariant can hold, unless we explicitly develop the bootstrap to not create non localized masters 🤔

The mongodump pipeline needs to be updated to mark the object non-localized: so that is part of the job indeed. Should be fine in most case (i.e. just drop Master documents if data is not empty), but there may be corner cases to address (e.g. null-version master....) indeed

→ However, since we make sure we 'skip' non-localized in both listing and GetObjMD, even having a "stored" non-localized master should be fine : it may be stored, but will not be visibile
→ May need to think further on it, but it seems to me this would only happen on the "oldest" revision, so skipping the master like this cannot "hide" another version
→ The discussion here is on the _fallbacks_, i.e. when the code tries to "recompute" a missing master : we should not be affected I think

Anyway this is a good point, and worth adding mention in the tickets about "triggering" the localization (@delthas) and in test plan ticket.

DarkIsDude: Agreed and done. Thanks. `getLatestVersion` always filters now, so the parameter is gone and the existing specs go back to their original signature.

For the mongodump, I think we can indeed to the stuff on our side.

We can keep this code indeed simple, as we seen during the review the logic is already complex. And do the needed things in the write path.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -2123,15 +2146,17 @@ class MongoClientInterface {
                     if (!mst) {
                         return this.getLatestVersion(c, objName, params.vFormat, log, (err, version?) => {
                             if (err) {
-                                return next(err);
+                                // without a localized version the deleted one cannot be
+                                // the master: let the non-master path report the miss
+                                return next(err.is.NoSuchKey ? null : err, null);
```
francoisferrand: likely possible to optimize, by completely _skipping_ the whole master delete for non-localized objects: either early in the function (and hope we don't have a non-localized master) or at least skip the call to `getLatestVersion` : if we non-localized, we know already getLatestVersion cannot return an object with the same versionId....
→ let's create a followup?

francoisferrand (approved):

francoisferrand:
git history as many iterative commits, which change the same thing multiple times, and makes it hard to read/understand the git history _afterwards_ ; worth cleaning up before merging? 🙏

PR #2691, opened 2026-08-31
support params.conditions in all putObject cases

maeldonn:
Issue: ARSN-624

### DIFF
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -1232,20 +1232,23 @@ class MongoClientInterface {
     ) {
         const versionKey = formatVersionKey(objName, params.versionId, params.vFormat);
         const masterKey = formatMasterKey(objName, params.vFormat);
-        c.updateOne(
-            {
-                _id: versionKey,
-            },
-            {
-                $set: {
-                    _id: versionKey,
-                    value: objVal,
+        const versionFilter = buildConditionsFilter('putObjectVerCase4', { _id: versionKey }, params.conditions, log);
+        if (!versionFilter) {
+            return cb(errors.InternalError);
+        }
+        return c
+            .updateOne(
+                versionFilter,
+                {
+                    $set: {
+                        _id: versionKey,
+                        value: objVal,
+                    },
                 },
-            },
-            {
-                upsert: true,
-            },
-        )
+                {
+                    upsert: true,
+                },
+            )
             .then(() =>
                 this.getLatestVersion(c, objName, params.vFormat, log, (err, mstObjVal?) => {
                     if (err?.is.NoSuchKey) {
@@ -1311,6 +1314,11 @@ class MongoClientInterface {
                 }),
             )
             .catch(err => {
+                // a version failing the condition degenerates the upsert into an insert, raising a duplicate key error
+                if (hasConditions(params.conditions) && err.code === MONGODB_DUPLICATE_KEY_ERROR) {
+                    log.info('putObjectVerCase4: version condition not met, rejecting write', { versionKey });
+                    return cb(errors.PreconditionFailed);
+                }
                 log.error('putObjectVerCase4: error upserting object version', { error: err.message });
                 return cb(errors.InternalError);
             });
--- a/lib/storage/metadata/mongoclient/MongoClientInterface.ts
+++ b/lib/storage/metadata/mongoclient/MongoClientInterface.ts
@@ -1054,9 +1054,19 @@ class MongoClientInterface {
         const versionId = generateVersionId(this.instanceId, this.replicationGroupId);
         objVal.versionId = versionId;
         const masterKey = formatMasterKey(objName, params.vFormat);
-        c.updateOne({ _id: masterKey }, { $set: { value: objVal }, $setOnInsert: { _id: masterKey } }, { upsert: true })
+        const filter = buildConditionsFilter('putObjectVerCase2', { _id: masterKey }, params.conditions, log);
+        if (!filter) {
+            return cb(errors.InternalError);
+        }
+        return c
+            .updateOne(filter, { $set: { value: objVal }, $setOnInsert: { _id: masterKey } }, { upsert: true })
             .then(() => cb(null, `{"versionId": "${objVal.versionId}"}`))
             .catch(err => {
+                // a master failing the condition degenerates the upsert into an insert, raising a duplicate key error
+                if (hasConditions(params.conditions) && err.code === MONGODB_DUPLICATE_KEY_ERROR) {
+                    log.info('putObjectVerCase2: master condition not met, rejecting write', { masterKey });
+                    return cb(errors.PreconditionFailed);
+                }
                 log.error('putObjectVerCase2: error putting object version', { error: err.message });
                 return cb(errors.InternalError);
             });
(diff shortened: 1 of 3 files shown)


lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1392,6 +1451,18 @@ class MongoClientInterface {
                         )
                         .then(doc => {
                             if (!doc?.value) {
+                                if (hasConditions(params.conditions)) {
+                                    return collection.findOne(putFilter).then(existing => {
```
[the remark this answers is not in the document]
maeldonn: Fixed.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1026,9 +1054,19 @@ class MongoClientInterface {
         const versionId = generateVersionId(this.instanceId, this.replicationGroupId);
         objVal.versionId = versionId;
         const masterKey = formatMasterKey(objName, params.vFormat);
-        c.updateOne({ _id: masterKey }, { $set: { value: objVal }, $setOnInsert: { _id: masterKey } }, { upsert: true })
+        const filter = buildConditionsFilter('putObjectVerCase2', { _id: masterKey }, params.conditions, log);
```
francoisferrand, line 1057: is this case (putObjectVerCase2) used for metadata _updates_ as well, or just for creating new versions in the suspended-versioning case?
(conditional _updates_ should only be used for, well, updates 😇... Not sure if I see an issue here, but I think it's better to double-think this, and make sure we have predictable outcome...)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1317,7 +1355,10 @@ class MongoClientInterface {
             return this.putObjectNoVerWithOplogUpdate(collection, bucketName, objName, value, params, log, cb);
         }
         const key = formatMasterKey(objName, params.vFormat);
-        const putFilter = { _id: key };
+        const putFilter = buildConditionsFilter('putObjectNoVer', { _id: key }, params?.conditions, log);
```
francoisferrand, line 1358: same: can we distinguish object "replacement" from actual "metadata updates" ? do we need to, or can it be part of the "conditional update semantics" ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1366,11 +1412,22 @@ class MongoClientInterface {
...
-            $or: [{ 'value.deleted': { $exists: false } }, { 'value.deleted': { $eq: false } }],
+            ...notDeletedFilter,
         };
+        if (hasConditions(params.conditions)) {
+            const conditionsFilter = buildConditionsFilter('putObjectNoVerWithOplogUpdate', {}, params.conditions, log);
+            if (!conditionsFilter) {
+                return cb(errors.InternalError);
+            }
+            // conditions go under $and to leave the deleted flag $or above untouched
+            findFilter.$and = [conditionsFilter];
```
francoisferrand, line 1429: why not like the others:

```suggestion
const findFilters = buildConditionsFilter('putObjectNoVerWithOplogUpdate', {
    ...putFilter,
    ...notDeletedFilter,
}, params.conditions, log);
if (!putFilter) {
    return cb(errors.InternalError);
}
```

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1392,6 +1449,19 @@ class MongoClientInterface {
                         )
                         .then(doc => {
                             if (!doc?.value) {
+                                if (hasConditions(params.conditions)) {
+                                    // a transiently deleted document is absent, not a condition mismatch
+                                    return collection.findOne({ ...putFilter, ...notDeletedFilter }).then(existing => {
```
francoisferrand, line 1454: that looks fishy (if we can't find it, we can't).....and I think this is actually not correct : conditional updates is for _updates_ (i.e. modifying metadata: transition, replication status...), which should _never_ go through this code (which "builds" a fake oplog update entry, to simulate the ObjMD was deleted then created) ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
maeldonn: Only suspended-versioning PUTs land here: `getPutObjectVerStrategy` picks case2 when `params.versionId === ''`. Metadata updates go to case3/case4 (explicit versionId) or `putObjectNoVer`. Case2 always mints a new versionId and replaces the master, so the condition is checked against the master being overwritten, which is the If-Match semantic for a PUT.

Same caveat as case3/case4/putObjectNoVer: if no master exists, the upsert inserts and the write succeeds instead of failing the precondition. The filter is a race guard on an existing doc, not a replacement for the existence check upstream.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
maeldonn: We can't tell them apart here: the same function serves both (`putObjectNoVer` does a non-versioned PUT and a tagging update, case3/case4 do a versioned PUT and a metadata update on that version). Nothing in the params distinguishes them.

But it doesn't matter, since conditions are opt-in per call: only the conditional-write path passes `params.conditions`, internal metadata updates pass none. So the semantic stays uniform: apply this write only if the stored document still matches.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
maeldonn: This base filter already has the not-deleted `$or`, and `translateConditions` throws `InternalError` if the conditions also use `$or` (it won't overwrite an existing structural operator). So they're built separately and combined under `$and`.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
maeldonn: This isn't the metadata-update path. CloudServer only sets needOplogUpdate on PUT/copy/MPU complete, as s3:ReplaceArchivedObject, when replacing an archived object in a non-versioned bucket. So a PUT with If-Match on an archived object lands here and does need the condition.

The findOne is needed because findOneAndUpdate returns null both when the doc is missing and when it fails the conditions, and we want NoSuchKey vs PreconditionFailed. Failure path only.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
francoisferrand: > which is the If-Match semantic for a PUT

This is a good point, but seems like a bad news : there may be 2 semantics for `If-Match`, and I don't think we can differentiate them

- When we upload new data, and thus _create a new ObjMD_ (i.e. version-suspended, replace a non-versionned object, or create a new version) : we may want indeed something similar to If-Match, looking at the current master/...
- When we _update an existing ObjMD_ (i.e. lifecycle or replication changes some internal metadata fields), what we want is actually to compare with the existing metadata of the same object

The second semantic should not consider a different revision of the object : only the metadata already stored, and it would thus ignore previous master........except that in version-suspended or non-versionned bucket, everything is just replacing the ObjMD : so a put can be either a metadata update (not from CRR which requires versioning, but from lifeycle for exemple...) or a new object upload (which replaces data and metadata) 😱

francoisferrand: - To implement the second semantic, maybe we need to make the condition apply only when the eTag _does not_ match. This way we have the same semantics everywhere.
- To implement the first semantics, we would need to implement it as well for versioned put: and not create a new version if the previous one (doesn't) match the condition

PR #2692, opened 2026-08-31
Support object versioning on GCP locations

maeldonn:
### Not handled yet

| Gap | Impact | What it would take |
|---|---|---|
| Object versioning on the GCS bucket is a hard requirement: without it, the delete issued behind a delete marker destroys the data instead of archiving it | Silent data loss | Block writes on a non-versioned bucket, instead of only reporting it in the healthcheck |
| GCS lifecycle rules on noncurrent versions (`isLive: false`, `daysSinceNoncurrentTime`, `numNewerVersions`) can reap generations that S3 still serves as current | Deferred and silent data loss, with no error at write time | Document the precondition, and consider detecting such rules at configuration time |
| GCS never promotes a noncurrent generation, so the backend state diverges from the S3 view after an undelete or a version delete | Confusing during support, and the root cause of the lifecycle exposure above | Optional: promote by copy, accepting the new generation, the non-atomicity and the storage class reset |


lib/storage/data/external/GcpClient.js
```diff
@@ -38,10 +39,11 @@ class GcpClient extends AwsClient {
         this._gcpBucketName = config.bucketName;
         this._mpuBucketName = config.mpuBucket;
         this._createGcpKey = this._createAwsKey;
-        this._gcpParams = Object.assign(this._s3Params, {
-            mainBucket: this._gcpBucketName,
+        this._gcpParams = {
```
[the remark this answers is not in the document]
maeldonn: `this.config` is the SDK v3 resolved client config (arsenal config lives in `this._config`), so that fallback already resolved to `undefined` before the rename and it's unreachable anyway, since `GcpManagedUpload` is only used by `GcpService.upload()`, which has no caller left.

lib/constants.ts
```diff
@@ -128,7 +128,7 @@ export const externalBackendHealthCheckTimeout = 5000;
 // as a string as first parameter of the get/delete methods.
 export const clientsRequireStringKey = { sproxyd: true, cdmi: true };
 export const hasCopyPartBackends = { aws_s3: true, gcp: true };
-export const versioningNotImplBackends = { azure: true, gcp: true };
+export const versioningNotImplBackends = { azure: true };
```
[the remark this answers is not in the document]
maeldonn: Updated.

benzekrimaha, line 131: Removing `gcp` here lets CloudServer put versioned objects on a GCP location from what I understand. The delete issued behind a delete marker then destroys the live object if the GCP bucket is not versioned. I think we can potentially block writes (or at least the delete-marker put) unless versioning is enabled. Do we have  a followup ticket for the other gaps (suspended state for example) as they stay out of this PR scope?

lib/storage/data/external/GcpClient.js
```diff
@@ -55,6 +57,57 @@ class GcpClient extends AwsClient {
...
+    // best-effort: NoSuchKey means a previous marker already removed it.
+    // GCS returns no generation on delete, so the marker stays
+    // metadata-only with no backend version
+    _putDeleteMarker(keyContext, reqUids, callback) {
+        const gcpKey = this._createGcpKey(keyContext.bucketName, keyContext.objectKey, this._bucketMatch);
+        const log = createLogger(reqUids);
+
+        return this._client
+            .send(
+                new DeleteObjectCommand({
```
benzekrimaha: On a versioned GCP bucket that archives live objects; without versioning it just deletes them which results in silent data loss.
`AwsClient.put` already handles `isDeleteMarker`. We shouldn't duplicate it here if GCP returns no generation, don't fail on a missing VersionId for that case.
Also we should not call `callback` from inside `.then`: if it throws you re-enter the error handler... We should `await` the send, then call the callback. Also match `NoSuchKey` the same way `AwsClient.delete` does (`err instanceof NoSuchKey`), not only `err.name`.

lib/constants.ts
maeldonn: You're right, nothing detects a non-versioned GCS bucket today: GCS returns a generation whether or not versioning is on, so the missing-version check never fires here. Blocking writes and reporting it in the healthcheck are tracked in the followup, with the other items from the "Not handled yet" table.

lib/storage/data/external/GcpClient.js
maeldonn: Updated

lib/storage/data/external/GcpClient.js
```diff
@@ -282,9 +335,12 @@ class GcpClient extends AwsClient {
             );
         }
 
+        // the versionId suffix is translated into the GCS source
+        // generation header by the GcpService middleware
+        const copySourceVersion = sourceVersionId ? `?versionId=${encodeURIComponent(sourceVersionId)}` : '';
```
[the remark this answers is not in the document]
maeldonn: Updated

PR #2694, opened 2026-09-01
Copy the source version instead of the latest on external backends

maeldonn:
CopyObject and UploadPartCopy dropped the source object's dataStoreVersionId, so the backend server-side copy always read the latest version of the source key while the destination metadata still described the requested version.

Thread the source version down the external-backend copy path (DataWrapper -> MultipleBackendGateway -> external clients) and address the copy source as `key?versionId=...` on AWS locations. AzureClient and PfsClient take the new argument without using it.

Issue: ARSN-627


lib/storage/data/external/GcpClient.js
```diff
@@ -256,7 +256,7 @@ class GcpClient extends AwsClient {
         });
     }
 
-    uploadPartCopy(request, gcpSourceKey, sourceLocationConstraintName, config, log, callback) {
+    uploadPartCopy(request, gcpSourceKey, sourceVersionId, sourceLocationConstraintName, config, log, callback) {
```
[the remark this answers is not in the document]
maeldonn: Addressed in a later commit of #2692

lib/storage/data/MultipleBackendGateway.js
```diff
@@ -383,6 +384,7 @@ class MultipleBackendGateway {
                     request,
                     destLocationConstraintName,
                     externalSourceKey,
+                    externalSourceVersionId,
```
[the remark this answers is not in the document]
maeldonn: Fixed.

lib/storage/data/external/GcpClient.js
benzekrimaha, line 266: GCP isn't in the same boat as azure/pfs, and it goes both ways:
- here the argument is accepted and dropped, so UploadPartCopy on GCP still copies the latest version => the bug this PR fixes isn't fixed for GCP...
- `copyObject` isn't overridden, so GCP inherits the AwsClient one and does get `?versionId=` appended. GcpClient sets `dataStoreVersionId`, so that suffix will be there. GCP selects a version with `x-goog-copy-source-generation`, not a `?versionId=` on the copy source , so it ends up part of the object name. Thus CopyObject on GCP goes from "copies the wrong version" to "NoSuchKey".
Can we make the suffix a small overridable method on AwsClient, so GcpClient returns '' (or the
generation) instead of inheriting the S3 form? and have a followup ticket for the GCP version. I think that the description should say AWS-only too.

tests/unit/storage/data/DataWrapper.spec.js
```diff
@@ -521,6 +521,7 @@ describe('DataWrapper', () => {
                             request,
                             'destBackend',
                             'sourceKey',
+                            undefined,
```
benzekrimaha, line 547: The bug  from what I get is that DataWrapper dropped `dataStoreVersionId`, and the only assertion here is that it passes `undefined`. can we add the case where the dataLocator actually carries a version, and assert it reaches the client? the AwsClient tests are good but they're given the version by hand.

lib/storage/data/external/AwsClient.js
```diff
@@ -755,10 +756,11 @@ class AwsClient {
         const sourceAwsBucketName = config.getAwsBucketName(sourceLocationConstraintName);
         const metadataDirective = request.headers['x-amz-metadata-directive'];
         const metaHeaders = trimXMetaPrefix(getMetaHeaders(request.headers));
+        const copySourceVersion = sourceVersionId ? `?versionId=${encodeURIComponent(sourceVersionId)}` : '';
```
benzekrimaha: nit: the version is encoded but the key isn't, and we're the ones introducing `?` as the delimiter , a source key containing a `?` now resolves to the wrong version... probably
already broken for other reasons, but worth a line IMO

lib/storage/data/external/GcpClient.js
maeldonn: Done

lib/storage/data/external/AwsClient.js
maeldonn: Right, and it's pre-existing: without the suffix, a key like `photo?versionId=999` was already parsed by S3 as key `photo` + version `999`. Encoding the key would change `CopySource` for every copy, versioned or not, so I'd rather keep that out of this fix.

tests/unit/storage/data/DataWrapper.spec.js
maeldonn: Added the positive case for both paths

PR #2701, opened 2026-09-11
ARSN-640: end version listing when max keys reached on a null key

jonathan-gramain:
## Context

`DelimiterVersions.filter()` caches a null key (a key ending with the version separator) so it can be output later at its correct position among the versions of that key. The cached key is flushed by calling `handleKey()` when a subsequent key is reached, but the return value of that call was discarded.

When the flush hits the max keys limit, `handleKey()` returns `FILTER_END`, meaning the listing is complete. Ignoring it made `filter()` return `FILTER_ACCEPT` or `FILTER_SKIP` instead, so the listing kept scanning. With a large number of consecutive null keys — where each new key flushes the previously cached one, hence produces a `FILTER_END` that is dropped — the whole remaining range was scanned even though no further entry could ever be returned, causing listing timeouts.

## Changes

- `lib/algos/list/delimiterVersions.ts`: return `FILTER_END` from `filter()` when flushing the cached null key returns it.

  Other return values only qualify the cached null key, not the current one, so they are still discarded: the current key is evaluated on its own afterwards, taking into account any state change caused by the flush (e.g. entering `SkippingPrefix` after a common prefix was added).

- `tests/unit/algos/list/delimiterVersions.spec.js`: new unit test feeding 100 consecutive null keys with `maxKeys: 2`, checking the filter results are exactly `[ACCEPT, ACCEPT, ACCEPT, END]` — the 4th entry flushes the 3rd, which no longer fits — and that the truncated result and its markers are unchanged. The test fails before the fix (all 100 keys scanned, every result `FILTER_ACCEPT`) in both v0 and v1 formats.

- `package.json`: version bump to 8.2.52.

## Testing

`tests/unit/algos/list/` passes (461 tests), in both v0 and v1 formats.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jonathan-gramain:
ping

BourgoisMickael (approved):

tcarmet (approved):

PR #2707, opened 2026-09-15
fix bug with mongo search option removing deletion filter when top le…

SylvainSenechal:
ISSUE: ARSN-641

Found this while reviewing the pr of the target branch https://github.com/scality/Arsenal/pull/2688#discussion_r4016162587

- mongo query is supposed to filter out deleted elements
- The query can contains extra search option, added with an object.assign that can erase the first deletion filter with top level $or statement

Added a functional test to verify the bug
Added the fix and rerun the functional test => its fixed

Side note : I think impact is minimal, I don't believe we use top level $or filters


SylvainSenechal:
Will apply linter after parent pr is merged

benzekrimaha (approved):
LGTM code wise , on the ci side  prettier is failing along with codecov...

DarkIsDude (approved):
Thanks for the spot and the PR 🙏

PR #2708, opened 2026-09-16
dr clean view : do not set non localized versions as master versions

SylvainSenechal:
ISSUE: ARSN-618

Implementing Clean View for object whose metadata are still sitting in the source location.

The previous PR ARSN-617 dealt with Clean View : Reads
This one is for Clean View : Writes

When mongo processor sends a put request, the mongo interface will now check if the object's location is localized or not. Depending on that it will (or not) update the master : A non localized object version data cannot be promotoed to be a master


lib/storage/metadata/mongoclient/MongoClientInterface.ts
SylvainSenechal, line 1211: Mongo processor will only use VerCase4 to put non localized objects : 
https://github.com/scality/backbeat/blob/57765e8f67e61fd7fb2ba36facbda6cc70054a31/extensions/mongoProcessor/MongoQueueProcessor.js#L562

And bootstrap writes directly in mongo without using the Arsenal functions.

So this PR ends up only change put ver case 4.

I could add something like this in the other put functions : 
```
if (this.nonLocalizedLocations.has(objVal.dataStoreName)) {
  return err / log err / log warn
}
```

because in theory, these put function could later be used to put non localized objects (like if we were to start supporting non versioning objects, we would not use ver case 4 but another), but I don't think we wanna do that

SylvainSenechal: Update, I did change the other functions, but for some of them the change can be discussed

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -887,6 +889,17 @@ class MongoClientInterface {
...
     }
 
+    /**
+     * A version is non-localized when its location points to data living on
+     * a different site: its metadata is here, but the data is
+     * not copied yet. Such a version must never be exposed as master.
+     * @param {Object} objVal object metadata
+     * @return {Boolean} true if the version is non-localized
+     */
+    isNonLocalized(objVal: ObjectMDData): boolean {
```
SylvainSenechal, line 899: Nit: We can discuss the naming of this, because in the functions later we end up with stuff like 

if not not
if (!this.isNonLocalized(objVal)) {

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1040,6 +1064,15 @@ class MongoClientInterface {
         const versionId = generateVersionId(this.instanceId, this.replicationGroupId);
         objVal.versionId = versionId;
         const masterKey = formatMasterKey(objName, params.vFormat);
+        // the master is the only document this path writes : a non-localized version
```
SylvainSenechal, line 1067: I did this for putObjectNoVer, putObjectWithCond and putObjectVerCase2, but it's **really** debatable

Logging an error but continuing...
These 3 functions only write masters. But also from what I found, they won't be called by the mongo processor, so it's more of a security, but as it is it's just not really useful.

Could return an error instead, but currently these errors won't be handled either, or just not do anything and consider that these functions will never be called with a non localized object

lib/storage/metadata/mongoclient/MongoClientInterface.ts
SylvainSenechal: And second update, technically, the mongo processor can also call putObjectNoVer, but only for null version, but it could happen in theory : User creates a non versioned bucket and writes objects, then set bucket versioning, then setup disaster recovery on the bucket.

SylvainSenechal, line 1372: As I wrote in another comment, this function may be a bit more tricky :

- it can be called by mongo processor for null versions (the scenario would be : User creates non versioned bucket and writes objects, then turn versioning bucket on, then setup disaster recovery)
- In that scenario, we need to write the master version even if its non localized otherwise we just dont store anything for the object..
- It's probaly still fine because regardless of what we do, non localized entries will still be hidding by clean read

SylvainSenechal:
I'll lint later

package.json
```diff
@@ -3,7 +3,7 @@
     "engines": {
         "node": ">=20"
     },
-    "version": "8.5.17",
+    "version": "8.5.18",
```
DarkIsDude: 8.6.0-preview I guess ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: You just describe the feature, we have citadel for that. `A version is non-localized when its location points to data living on a different site` is enough ?

DarkIsDude: `isLocalized` ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -947,8 +960,18 @@ class MongoClientInterface {
         // updating or deleting master depending on the last version put
         // in v0 the master gets updated, in v1 the master gets deleted if version is
         // a delete marker or updated otherwise.
-        const masterOp = this.updateDeleteMaster(objVal.isDeleteMarker || false, params.vFormat, filter, update, true);
-        ops.push(masterOp);
+        // a non-localized version must never be exposed as master: write the version
```
DarkIsDude: we just describe the behaviour ?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: For me it's fine. A little bit defensive but I'm in. Instead of a comment, you can put that directly in the error message (it's the goal).

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1217,20 +1254,33 @@ class MongoClientInterface {
...
+                    $set: {
+                        _id: versionKey,
+                        value: objVal,
+                    },
+                },
+                {
+                    upsert: true,
+                },
+            );
+        // A version whose data still lives on a different site (happens with Disaster Recovery)
```
DarkIsDude: To avoid too many comments in the code that are duplicated, we should (not sure it's needed) the .md related to it ?

DarkIsDude (commented):
Code look goods to me 🙏. I'll monitor the Francois review to check if some new stuff come to my mind but nothing notice on my side 🙏

package.json
SylvainSenechal: ah yeah

lib/storage/metadata/mongoclient/MongoClientInterface.ts
SylvainSenechal: I think im gonna use this unless François has another suggestion

francoisferrand: non-localized is clear enough when discussing, but hard to read in code indeed... I actually ended removing lots of it, and sticking more simpler terms (copy data, replicate...)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
francoisferrand: This is safety in depth : we implement a new behavior/invariant (master can never be non-localized), so we should indeed cover all path. Some may not happen in this specific case, but we may later use the same feature in a different way or have a bug somewhere, which would cause it to break the invariant...

lib/storage/metadata/mongoclient/MongoClientInterface.ts
francoisferrand: `isFullyReplicated` maybe?
nice-enough in the context of pull replication, but I fear it could be harder to understand here...

`isFullyPullReplicated` clarifies it, but is quite ugly...

`isDataCopied` may be ok?

or the opposite, `isPendingPullReplication` - a bit long, but has a nice simetry with the PENDING replication status. Maybe that would me my preferred name, what do you think?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
francoisferrand: `putObjectVerCase4` is the "repair master" path ; but if the object is non localized, the master shoudl not be updated: so getPutObjectVerStrategy shoudl route non-localized objects to `putObjectVerCase3` instead?

lib/storage/metadata/mongoclient/MongoClientInterface.ts
francoisferrand: (btw, this could be managed in getPutObjectVerStrategy, instead of doing it 3 times? i.e. let it return `null` if the "case" is not valid, and putObject can return an error)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1162,14 +1193,17 @@ class MongoClientInterface {
...
-                        const masterOp = this.updateDeleteMaster(
-                            objVal.isDeleteMarker || false,
-                            params.vFormat,
-                            filter,
-                            update,
-                            objUpsert,
-                        );
-                        ops.push(masterOp);
+                        // a master never holds a non-localized version, see Mongoclient.md
+                        if (!this.isNonLocalized(objVal)) {
```
francoisferrand: this function could be optimized : this function starts by looking up the master, which is not needed here (this should always be an upsert, for non-localized objects)

lib/storage/metadata/mongoclient/MongoClientInterface.ts
```diff
@@ -1328,6 +1378,15 @@ class MongoClientInterface {
         log: werelogs.Logger,
         cb: ArsenalCallback<void>,
     ) {
+        // the master is the only document this path writes : a non-localized version
+        // reaching this function is unexpected
+        if (this.isNonLocalized(value)) {
+            log.error('putObjectNoVer: writing a non-localized version as master', {
```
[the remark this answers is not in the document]
francoisferrand: this is an error. We should never reach here : if we do, it means the contract is broken : log error and fail.

tests/functional/metadata/mongodb/nonLocalizedMaster.spec.js
francoisferrand: file name is weird : there is no non-localized master...

→ should jsut go in the same file as the other "clean view" tests
→ I don't remember how this file was called (nonLocalizedRead.ts maybe?), we could match (nonLocalizedWrite.ts) or just name the whole thing cleanview.ts ?

tests/functional/metadata/mongodb/nonLocalizedMaster.spec.js
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * Replicates a version the way the clean-room mongo-processor does: the
+     * version id is preserved and the master is repaired, i.e. putObjectVerCase4.
+     * @param {String} objName - object key
+     * @param {String} dataStoreName - location the version data lives on
+     * @param {String} [versionId] - version id, generated when omitted
+     * @return {Promise<String>} the version id written
+     */
+    async function replicateVersion(objName, dataStoreName, versionId) {
```
francoisferrand: you mean this function simulates the metadata update once data has been copied?

so should not be "replicate version", but `putLocalizedObjectMD` or something like this.

nit: maybe easier to read if you have multiple "objects md" (constants) for each state, e.g. non-localized, localized... so the test reads cleanly `putObjectMD(v1NonLocalized); putObjectMD(v2NonLocalized); putObjectMD(v1Localized)`.... and you don't need to find a name which matches both the action (putObjectMD) and what it puts (an update to non-localized obj md)

tests/functional/metadata/mongodb/nonLocalizedMaster.spec.js
```diff
@@ -0,0 +1,184 @@
...
+                const localizedVersionId = await replicateVersion('pfx-mixed', LOCAL_LOCATION);
+                // a newer version sorts first, and would take over the master if localized
+                await replicateVersion('pfx-mixed', SOURCE_LOCATION);
+                const master = await getMaster('pfx-mixed', variation.vFormat);
+                assert(master, 'the master key should still be there');
+                assert.strictEqual(master.value.versionId, localizedVersionId);
+                assert.strictEqual(master.value.dataStoreName, LOCAL_LOCATION);
+            });
+
+            it('should repair the master when the version gets localized', async () => {
```
francoisferrand: nit: out of curiosity, which "case" does this go through?

(good that it works directly, I feared we would need some tweaks here...)

tests/functional/metadata/mongodb/nonLocalizedMaster.spec.js
```diff
@@ -0,0 +1,184 @@
...
+                assert.strictEqual(version.value.dataStoreName, SOURCE_LOCATION);
+                assert.strictEqual(await getMaster('pfx-nonlocalized', variation.vFormat), null);
+            });
+
+            it('should create a master for a localized version', async () => {
+                const versionId = await replicateVersion('pfx-localized', LOCAL_LOCATION);
+                const master = await getMaster('pfx-localized', variation.vFormat);
+                assert(master, 'the master key should have been written');
+                assert.strictEqual(master.value.versionId, versionId);
+            });
```
francoisferrand: nothing new here, this is all the other tests, right?

tests/functional/metadata/mongodb/nonLocalizedMaster.spec.js
```diff
@@ -0,0 +1,184 @@
...
+            it('should keep the master on the localized version when a newer one is not localized', async () => {
+                const localizedVersionId = await replicateVersion('pfx-mixed', LOCAL_LOCATION);
+                // a newer version sorts first, and would take over the master if localized
+                await replicateVersion('pfx-mixed', SOURCE_LOCATION);
+                const master = await getMaster('pfx-mixed', variation.vFormat);
+                assert(master, 'the master key should still be there');
+                assert.strictEqual(master.value.versionId, localizedVersionId);
+                assert.strictEqual(master.value.dataStoreName, LOCAL_LOCATION);
+            });
+
```
francoisferrand: should also have tests
- when we _update_ the metadata of a localized object
- for the other cases (non versioned...) → fail

francoisferrand (dismissed):

lib/storage/metadata/mongoclient/MongoClientInterface.ts
DarkIsDude: If we do that (I'm fine with `isPendingPullReplication`), we must update the design and already code done. Should be pretty easy but it's a must have to avoid 2 naming for the same concept. This is the kind of tech debt what we must really not insert in the codebase 🙏.

lib/storage/metadata/mongoclient/MongoClientInterface.ts
SylvainSenechal: I'm gonna keep it then and return an error. 
I don't think we can easily change the strategy dispatcher, some functions are actually not called from that dispatcher (dispatcher calls case1/2/3/4/noVer), but we have functions like putObjectWithCond that needs the error too

tests/functional/metadata/mongodb/nonLocalizedMaster.spec.js
SylvainSenechal: Yeah let's move it to a single clean view file

tests/functional/metadata/mongodb/nonLocalizedMaster.spec.js
SylvainSenechal: It's case 4 (replicateVersion is called with repair master = true)

tests/functional/metadata/mongodb/nonLocalizedMaster.spec.js
SylvainSenechal: yeah its redundant

tests/functional/metadata/mongodb/nonLocalizedMaster.spec.js
SylvainSenechal: Yeah I'm rewriting it so that we have 

await putReplicated(v1Localized);
await putReplicated(v2NonLocalized); 

which should be more readable

lib/storage/metadata/mongoclient/MongoClientInterface.ts
SylvainSenechal: Ok i'm changing it to isLocalized for now because I agree with Edouard about mismatch naming

But I like isPendingPullReplication, so if the citadel design is already not referencing "localized" aymore, I think i can update arsenal to also use this new isPendingPullReplication term

lib/storage/metadata/mongoclient/MongoClientInterface.ts
SylvainSenechal: I don't know if its a good idea to do that. If I understand your point, when object is non localized, case4 kinda becomes case3 so we could reroute to case3 🤔 But I'm thinking regardless, we have to keep the logic in case4 as we said we need to handle all non localized write case regardless of if its used or not. I'm a bit afraid of breaking something by rerouting request now

PR #2709, opened 2026-09-21
ARSN-642: prepare Arsenal to run on Node 24

francoisferrand:
Node 22 enters Maintenance in Oct 2026 and reaches end-of-life 30 April 2027. Node 24 is Active LTS until April 2028, so it's time to get Arsenal running there.

## CI

`tests.yaml` now runs a Node 22/24 matrix (compile/lint/KMIP jobs move fully to 24; Codecov upload stays on the 24 leg). Keeping 22 in the mix a while longer since cloudserver, backbeat and scuba are still on it during this transition.

## engines.node -> >=22

Was `>=20`, which went EOL in April 2026 and wasn't being tested anymore. Deliberately not going to `>=24` yet — cloudserver, backbeat and scuba need to move first, or Node 22 needs to go EOL (April 2027), whichever comes first.

## url.parse() removal

`url.parse()` is runtime-deprecated (DEP0169). Six call sites replaced.

Three are straight swaps to `new URL()`. Three needed a purpose-built parser (`lib/utils/requestUrl.ts`) instead, because `new URL()` normalizes in ways that change an S3 request path: it collapses `.`/`..` segments, reads a leading `//` as an authority, and percent-encodes non-ASCII. `/bucket/a/../b` becomes `/bucket/b`; `//bucket/key` loses its bucket. On `getCanonicalizedResource.ts` (v2 signature path) that would change what gets signed — a request could end up verified against a different resource than it addresses. Supplying a base doesn't help either: absolute URLs ignore it entirely.

`parseRequestTarget()` reproduces the legacy parser's behavior for these shapes (differentially tested against real `url.parse()` output, hundreds of thousands of cases, zero mismatches) without ever resolving a scheme or host. Exported it from Arsenal's index so cloudserver/backbeat can use the same helper for their own `url.parse()` removals, since they have the identical hazard (copy-source header, routeBackbeat).

## Scality dependency refresh

Bumped once each of them had a Node 24-ready release: `fcntl` 0.3.1, `httpagent` 1.1.1, `sproxydclient` 8.2.3 (now tagged, previously pinned by commit), `werelogs` 8.2.5, `@scality/eslint-config-scality` 8.3.3. Lockfile refreshed accordingly.

## Verification

- Unit suite green on Node 22 and Node 24 (167 suites, 4006 passed, 9 skipped, run sequentially post-rebase to avoid MongoMemoryServer port contention).
- Build and lint pass on Node 24.
- `yarn audit` before/after: prod advisories unchanged (65, none of the bumped Scality packages carried any); dev advisories dropped from 52 to 48 — 4 DoS advisories fixed via the Guidelines 8.3.3 refresh dropping its old markdown-it/linkify-it chain.

## Out of scope

Deprecated dependency cleanup (`level`, `ioctl`, `prom-client`, `ajv`, `@types/ioredis`) and publishing — tracked separately.

Issue: ARSN-642


.github/workflows/tests.yaml
```diff
@@ -9,6 +9,12 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # 24 is the target runtime; 22 is what engines.node still allows and
+        # what cloudserver/backbeat/scuba run until they move over.
+        node-version: ['22', '24']
```
SylvainSenechal, line 17: Btw you used this in a lot of repo recently but it means every tests will run twice 🤔  You could've put 24 directly 🤔 , I guess the few remaining repo on node 22 will soon be updated considering you do like one repo per day 🤔

lib/utils/requestUrl.ts
```diff
@@ -0,0 +1,96 @@
+/**
+ * Splitting of request targets (`/bucket/key?acl`) without `url.parse()`, which
```
SylvainSenechal, line 2: A bit heavy to read, while parseRequestTarget could use one or two input->output examples

tests/unit/auth/v2/canonicalization.spec.js
```diff
@@ -224,4 +224,38 @@ describe('canonicalization', () => {
...
+            assert.strictEqual(resourceFor('/bucket/a/../b'), '/bucket/a/../b');
+            assert.strictEqual(resourceFor('/bucket/a/./b'), '/bucket/a/./b');
+        });
+
+        it('should keep a key that itself starts with a slash', () => {
+            assert.strictEqual(resourceFor('//bucket/key'), '//bucket/key');
+        });
+
+        it('should leave a non-ASCII key unencoded', () => {
+            assert.strictEqual(resourceFor('/bucket/ключ'), '/bucket/ключ');
```
SylvainSenechal, line 246: 😆

tests/unit/utils/requestUrl.spec.js
```diff
@@ -0,0 +1,92 @@
+'use strict';
+
+const assert = require('assert');
+
+const { parseRequestTarget } = require('../../../lib/utils/requestUrl');
+
+/*
+ * Expected values below were captured from the `url.parse()` this replaces, so
```
SylvainSenechal, line 8: Its nit and more of a remark, connected to the other new file requestUrl.ts

I notice claude likes to make comment referencing stuff that it just removed, it's okish in the context of the pr, but for the codebase it's weird,
It feels like its outputing its thought context into the comments. I'm talking about the "were captured from the url.parse() this repllaces", not just for this pr but something i notice in general, I don't think this is super good to keep written this way.

SylvainSenechal (approved):

benzekrimaha (approved):

lib/utils/requestUrl.ts
benzekrimaha: Indeed the comment here is stating legacy vs new not really helpful I think , could be kept ( and less verbose ) as the commit body just in case

delthas (approved):

.github/workflows/tests.yaml
francoisferrand: 2 reasons:
- even if for a short time, no reason to drop compatibiltiy. Esp. in many repos we don't even need a new branch, this is really just "marking" that node24 is supported engine version...
- in a few weeks/months we sould add node26, and need to support both 24 and 26 for migration; so good to keep the matrix job

